Issue #217, opened 2024-05-08
[Feature Request] More basic settings for Android (and iOS) emulator like Biometrics or Rotation


Martinocom:
**Description**
My personal and our company app are using Fingerprint to authenticate. Using emulators inside the RN-IDE has not the option to setup nor trigger the Fingerprint/FaceID features, effectively blocking us to test and develop those part of the app. This is only an example, because I think there are other devs that uses maybe some other stuff like rotation, shake, tilt or battery/cellular options. All of those are missing from settings in the RN-IDE.

Managing them from the emulator settings directly from Android Studio causes the "Changes detected, please rebuild" popup that won't disappear until rebuild has been made. Besides, having an Android Studio window opened kills the purpose of having the IDE integrated with VS Code.

If I would set a priority, the Fingerprint/FaceID and rotation features are probably the most used ones and I would like to have them.

**Environment**
- RN-IDE: 0.0.9
- RN: 0.73.7

filip131311:
Already on main Will be included in Radon v1.10.0

PR #236, opened 2024-05-12
Add support for simulating biometric authorization on ios

filip131311:
Description: 
This PR introduces an option to simulate biometric authorization. 
You can simulate: 
- matching face/touch id 
- not matching face/touch id
- biometric authorization enrollment.

Additionally convenient commands were added to vs code to allow for binding a shortcut to the action of matching or non-matching. The default  keybindings are the same as in xcode.

Relates to : #217

https://github.com/user-attachments/assets/88cc5f2b-fd71-4abd-b302-853578b721de

How was it tested: 

a simple example app is added as a part of this PR that allows testing introduced functionality .

Future improvments: 
- [ ]  android support


packages/vscode-extension/package.json
```diff
@@ -61,6 +61,18 @@
         "title": "Diagnostics",
         "category": "React Native IDE",
         "enablement": "!RNIDE.extensionIsActive"
+      },
+      {
+        "command": "RNIDE.matchBiometricAuthorization",
+        "title": "Matching Biometric Authorization ",
```
jakub-gonet, line 67: Maybe "Perform biometric auth"? It's not clear what it does currently.

packages/vscode-extension/package.json
```diff
@@ -61,6 +61,18 @@
...
+        "category": "React Native IDE",
+        "enablement": "RNIDE.extensionIsActive"
+      },
+      {
+        "command": "RNIDE.nonMatchBiometricAuthorization",
+        "title": "Non Matching Biometric Authorization",
```
jakub-gonet, line 73: Similar here, maybe "Perform failed biometric auth"?

packages/vscode-extension/src/common/Project.ts
```diff
@@ -111,6 +112,7 @@ export interface ProjectInterface {
 
   getDeviceSettings(): Promise<DeviceSettings>;
   updateDeviceSettings(deviceSettings: DeviceSettings): Promise<void>;
+  sendBiometricAuthorization(match: boolean): Promise<void>;
```
jakub-gonet, line 115: I'd name the parameter `(is)successful`, match doesn't convey much meaning iMO.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -125,6 +125,42 @@ export class IosSimulatorDevice extends DeviceBase {
...
+    await exec("xcrun", [
+      "simctl",
+      "--set",
+      deviceSetLocation,
+      "spawn",
+      this.deviceUDID,
+      "notifyutil",
+      "-p",
+      "com.apple.BiometricKit.enrollmentChanged",
+    ]);
```
jakub-gonet, line 148: Why we do it twice and only once we send an argument?

packages/vscode-extension/src/common/Project.ts
```diff
@@ -8,6 +8,7 @@ export type DeviceSettings = {
     longitude: number;
     isDisabled: boolean;
   };
+  biometricEnrollment: boolean;
```
jakub-gonet, line 11: Maybe `biometricsRegistered`, `hasEnrolledBiometrics`, etc?

jakub-gonet: BTW, how do we sync device settings across workspaces? I didn't see code reading things using xcrun.

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -117,7 +119,57 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
...
               <span className="device-settings-large-text-indicator" />
             </div>
+            <div className="device-settings-margin" />
+            {projectState.selectedDevice?.platform === DevicePlatform.IOS && (
+              <DropdownMenu.Sub>
+                <DropdownMenu.SubTrigger className="dropdown-menu-item">
+                  <span className="codicon codicon-layout" />
+                  Biometrics
+                  <span className="codicon codicon-chevron-right right-slot" />
+                </DropdownMenu.SubTrigger>
```
jakub-gonet, line 129: Can we please extract it to the different component?

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -117,7 +119,57 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
...
+                    <DropdownMenu.Item
+                      className="dropdown-menu-item"
+                      onSelect={() => {
+                        project.updateDeviceSettings({
+                          ...deviceSettings,
+                          biometricEnrollment: !deviceSettings.biometricEnrollment,
+                        });
+                      }}>
+                      <span className="codicon codicon-layout-sidebar-left" />
+                      Enrolment
```
jakub-gonet, line 142: This doesn't mean much to me, I'd preferably see here "Registered", "Configured" or similar.

jakub-gonet (commented):
I also wonder if we can detect if app is requesting biometrics and show the UI automatically instead of hiding it in two step submenu.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
filip131311: We don't do exactly the same thing, once we set the argument and the second one informs the system about the change.  https://keith.github.io/xcode-man-pages/notifyutil.1.html

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
filip131311: enrolled is used by ios simulator if there is a word that is easiest to understand for people how are using this feature it is this one

packages/vscode-extension/src/common/Project.ts
filip131311: why? matching (face/finger) is exactly what you do. so it is a match or it is not. In fact when you fail to match you are not necessary unsuccessful because not matching strangers face is expected behaior

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
filip131311: Screenshot 2024-08-28 at 11 43 31

packages/vscode-extension/src/common/Project.ts
filip131311: we don't sync across workspaces, we remember workspace settings tho, by calling changeSettings on boot

kmagiera (approved):

PR #275, opened 2024-05-20
Fix Android crashing on reload

filip131311:
This PR solves an issue, with android reloads, that was causing new emulator  device to crash, because the old one was still active. 

Solution: 
check if the process with the same avd is running before booting the device.  

Before: 

https://github.com/software-mansion/react-native-ide/assets/159789821/3b9a5111-97ce-4b02-b8b0-51125f83bc52

After: 

https://github.com/software-mansion/react-native-ide/assets/159789821/427ae727-e5aa-44ef-bf30-0bb2e9fbbdc5

### DIFF
--- a/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
@@ -84,7 +84,7 @@ export class AndroidEmulatorDevice extends DeviceBase {
     });
     await subprocess;
     if (runningPid) {
-      await exec("kill", ["-9", `${runningPid}`]);
+      process.kill(Number(runningPid), 9);
     }
   }
 
--- a/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
@@ -72,7 +72,7 @@ export class AndroidEmulatorDevice extends DeviceBase {
     ]);
   }
 
-  private async checkIfOldDeviceIsRunning() {
+  private async ensureOldEmulatorProcessExited() {
     return new Promise<void>((resolve, reject) => {
       let runningPid: string | undefined;
       const subprocess = exec("ps", ["-A"]);
@@ -95,7 +95,7 @@ export class AndroidEmulatorDevice extends DeviceBase {
 
   async bootDevice() {
     // this prevents booting device with the same AVD twice
-    await this.checkIfOldDeviceIsRunning();
+    await this.ensureOldEmulatorProcessExited();
 
     const avdDirectory = getOrCreateAvdDirectory();
     const subprocess = exec(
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -66,10 +72,30 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+        }
+      });
+
+      subprocess.on("close", async () => {
+        if (runningPid) {
+          await exec("kill", ["-9", `${runningPid}`]);
+        }
+        resolve();
+      });
+    });
```
kmagiera, line 93: 1) use `-o` to format ps output (it may be different depending on system settings)
2) Don't use promise, just await on subprocess and then do `if (runingPid) await exec(...)`

kmagiera (changes requested):

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
filip131311: done

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -17,6 +17,7 @@ import { EXPO_GO_PACKAGE_NAME, fetchExpoLaunchDeeplink } from "../builders/expoG
 
 export const EMULATOR_BINARY = path.join(ANDROID_HOME, "emulator", "emulator");
 const ADB_PATH = path.join(ANDROID_HOME, "platform-tools", "adb");
+const DISPOSE_TIMEOUT = 3000;
```
kmagiera, line 20:
```suggestion
const DISPOSE_TIMEOUT = 9000;
```

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -66,10 +72,25 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+    const regexpPattern = new RegExp(`(\\d+)\\s.*qemu.*-avd ${this.avdId}`);
+    lineReader(subprocess).onLineRead(async (line) => {
+      const regExpResult = regexpPattern.exec(line);
+      if (regExpResult) {
+        runningPid = regExpResult[1];
+      }
+    });
+    await subprocess;
+    if (runningPid) {
+      await exec("kill", ["-9", `${runningPid}`]);
```
kmagiera, line 87: process.kill(runningPid, 9)

watch out as this method may throw when PID doesn't exist. Also PID needs to be passed as number

kmagiera (approved):

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
kmagiera: https://nodejs.org/api/process.html#processkillpid-signal

PR #304, opened 2024-05-26
fix: add `bundle` command as a fallback to CocoaPods check

### LINKED ISSUE #303: CocoaPods installation check not working with Bundler
**Description*
Within my project, I use [Bundler](https://bundler.io/) so CocoaPods commands run through `bundle exec pod` as opposed to only `pod`.
Therefore, [the `pod --version` check fails](https://github.com/software-mansion/react-native-ide/blob/main/packages/vscode-extension/src/dependency/DependencyChecker.ts#L126-L128) with an error that can't be dismissed (which has been already reported in #297).

**Environment**
- Version of currently used React Native IDE: `v0.0.14`
- Version of React Native/Expo used in the project: `0.72.13`
- Any other versions of tools related to the problem:
```
pod: 1.13.0
bundler: 2.1.4
```

**Screenshots**
|Warning|Doctor|
|---|---|
|Screenshot 2024-05-26 at 13 40 28|Screenshot 2024-05-26 at 13 40 42|

lucasbento:
Fixes #303.

This PR adds `bundle exec pod --version` as a "second attempt" to verify that CocoaPods is installed, the current implementation that's solely relying in `pod --version` to check for CocoaPods installation will fail in projects using [Bundler](https://bundler.io/).

Since `bundle` commands are tied to the specific project installation, it's necessary to execute the command from within the project directory path.

### DIFF
--- a/packages/vscode-extension/src/dependency/DependencyChecker.ts
+++ b/packages/vscode-extension/src/dependency/DependencyChecker.ts
@@ -1,4 +1,4 @@
-import { Webview, Disposable } from "vscode";
+import { Webview, Disposable, workspace } from "vscode";
 import { coerce, gte } from "semver";
 import { Logger } from "../Logger";
 import fs from "fs";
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/dependency/DependencyChecker.ts
```diff
@@ -123,21 +123,32 @@ export class DependencyChecker implements Disposable {
...
+    let isCocoaPodsInstalled = await checkIfCLIInstalled("pod --version", {
+      env,
     });
+
+    // If `pod --version` fails, try `bundle exec pod --version` as a second attempt
+    if (!isCocoaPodsInstalled && workspace.workspaceFolders?.length) {
+      isCocoaPodsInstalled = await checkIfCLIInstalled("bundle exec pod --version", {
+        env,
+        // Run the command in the current project directory
+        cwd: workspace.workspaceFolders?.[0].uri.fsPath,
```
jakub-gonet, line 136: Can we safely assume that first workspace directory is always an app directory?

packages/vscode-extension/src/dependency/DependencyChecker.ts
```diff
@@ -123,21 +123,32 @@ export class DependencyChecker implements Disposable {
   }
 
   public async checkCocoaPodsInstalled() {
-    const installed = await checkIfCLIInstalled("pod --version", {
-      env: { ...process.env, LANG: "en_US.UTF-8" },
+    const env = { ...process.env, LANG: "en_US.UTF-8" };
+    let isCocoaPodsInstalled = await checkIfCLIInstalled("pod --version", {
```
jakub-gonet, line 127: Nit: Can we fit it in the one line please?

packages/vscode-extension/src/dependency/DependencyChecker.ts
```diff
@@ -123,21 +123,32 @@ export class DependencyChecker implements Disposable {
...
+      env,
     });
+
+    // If `pod --version` fails, try `bundle exec pod --version` as a second attempt
+    if (!isCocoaPodsInstalled && workspace.workspaceFolders?.length) {
```
jakub-gonet, line 132: Nit: can we please make it more explicit? `workspace.workspaceFolders?.length > 0`

packages/vscode-extension/src/dependency/DependencyChecker.ts
lucasbento: Not always, it largely depends on the users' setup but this is a fair try IMHO.
We could attempt to look for the correct folder within the `fsPath`, though.

jakub-gonet:
This wasn't active for quite a while now so I'm closing it. If need arises, we will reimplement it in the future. Thanks!

PR #308, opened 2024-05-27
Fix clickability of wrapped items in button-group-top

uguraktas:
The `button-group-top` div contains various buttons, dropdowns, and other interactive elements. When the width of the container becomes small, these items wrap to the next line. However, the wrapped items were not clickable, causing usability issues. This PR addresses the clickability problem of the wrapped items in the `button-group-top `div.

### DIFF
--- a/packages/vscode-extension/src/webview/views/PreviewView.css
+++ b/packages/vscode-extension/src/webview/views/PreviewView.css
@@ -7,10 +7,13 @@
 
   margin: 10px;
   gap: 4px;
+  position: relative;
 }
 
 .button-group-top {
   margin-bottom: 8px;
+  position: relative;
+  z-index: 1;
 }
 
 .feedback-button {


packages/vscode-extension/src/webview/views/PreviewView.css
```diff
@@ -7,10 +7,13 @@
 
   margin: 10px;
   gap: 4px;
+  position: relative;
 }
 
 .button-group-top {
   margin-bottom: 8px;
+  position: relative;
```
jakub-gonet, line 15: Isn't `position: relative` a default?

uguraktas: I have updated

uguraktas:
> Can you please include screenshots?

Settings icons was not clickable 

Screenshot 2024-05-28 at 00 50 25

kmagiera:
I believe this isn't the right fix for this issue. The problem seems to be in the fact that when settings button is moved to second line, it gets covered by the simulator view and hence is not clickable. What we really should do is adjust the layout such that on narrow screens we still keep it in the top-right corner and for example can make the url bar a bit shorter to compensate for that. Would you be willing to update your PR with such a change?

kmagiera (changes requested):
Should've requested changes with my previous comment. As mentioned there, we should adjust the position of that item such that it never wraps to the second row

kmagiera:
Closing as this has been eventually addressed in #363

PR #317, opened 2024-05-28
Add static geolocation

filip131311:
This PR adds an ability to set static device location. 

added functionality:
- setting location
- enable/disable location services
- coordinates validation

https://github.com/software-mansion/react-native-ide/assets/159789821/a22754ba-a832-4fab-a13a-23545da14b79

https://github.com/software-mansion/react-native-ide/assets/159789821/cf25e213-2af1-41d9-9ddd-6baef5c3444d

Fixes: #330

### DIFF
--- a/packages/vscode-extension/src/common/utils.ts
+++ b/packages/vscode-extension/src/common/utils.ts
@@ -27,25 +27,3 @@ export function throttle<T extends (...args: any[]) => any>(
     recentArgs = args;
   } as T;
 }
-
-export function throttleWithTrailing<T extends (...args: any[]) => any>(
-  func: T,
-  limitMs: number
-): (...args: Parameters<T>) => void {
-  let timeout: NodeJS.Timeout | null = null;
-  let recentArgs: any;
-
-  return function (...args: Parameters<T>) {
-    recentArgs = args;
-
-    if (!timeout) {
-      timeout = setTimeout(() => {
-        timeout = null;
-        if (recentArgs) {
-          func(...recentArgs);
-          recentArgs = null;
-        }
-      }, limitMs);
-    }
-  };
-}
--- a/packages/vscode-extension/src/common/utils.ts
+++ b/packages/vscode-extension/src/common/utils.ts
@@ -27,3 +27,25 @@ export function throttle<T extends (...args: any[]) => any>(
     recentArgs = args;
   } as T;
 }
+
+export function throttleWithTrailing<T extends (...args: any[]) => any>(
+  func: T,
+  limitMs: number
+): (...args: Parameters<T>) => void {
+  let timeout: NodeJS.Timeout | null = null;
+  let recentArgs: any;
+
+  return function (...args: Parameters<T>) {
+    recentArgs = args;
+
+    if (!timeout) {
+      timeout = setTimeout(() => {
+        timeout = null;
+        if (recentArgs) {
+          func(...recentArgs);
+          recentArgs = null;
+        }
+      }, limitMs);
+    }
+  };
+}
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -64,6 +64,10 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
   ) ?? {
     appearance: "dark",
     contentSize: "normal",
+    location: {
+      latitude: 0,
+      longitude: 0,
+    },
   };
 
   constructor(private readonly deviceManager: DeviceManager) {
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -65,8 +65,9 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
     appearance: "dark",
     contentSize: "normal",
     location: {
-      latitude: 50.04865344199642,
-      longitude: 19.965474780486165,
+      latitude: 50.048653,
+      longitude: 19.965474,
+      isDisabled: false,
     },
   };
 
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -65,8 +65,8 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
     appearance: "dark",
     contentSize: "normal",
     location: {
-      latitude: 0,
-      longitude: 0,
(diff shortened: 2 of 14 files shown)


packages/vscode-extension/src/webview/views/DeviceLocationView.tsx
```diff
@@ -0,0 +1,91 @@
...
+import React, { useState, useEffect } from "react";
+
+import "./DeviceLocationView.css";
+import Button from "../components/shared/Button";
+
+export function DeviceLocationView() {
+  const { project, deviceSettings } = useProject();
+
+  const [latitude, setLatitude] = useState("");
+  const [longitude, setLongitude] = useState("");
```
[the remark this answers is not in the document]
jakub-gonet: To be honest, I'd remove all useStates here. The main source of data is coming from `deviceSettings` and you can compute all of transformed data using it. useState + useEffect is an antipattern in React.

jakub-gonet (changes requested):

packages/vscode-extension/src/project/project.ts
```diff
@@ -63,6 +63,10 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
   ) ?? {
     appearance: "dark",
     contentSize: "normal",
+    location: {
+      latitude: 0,
+      longitude: 0,
```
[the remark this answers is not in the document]
filip131311: I added the ability to disable location, but decided to stick to location being set be default, as a little easter egg our office is set as first location.

packages/vscode-extension/src/webview/views/DeviceLocationView.tsx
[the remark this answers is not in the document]
filip131311: Thank you done!

packages/vscode-extension/src/webview/views/DeviceLocationView.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <Button type="submit">Set Current Location</Button>
+      </form>
+      <Label>Disable Location</Label>
+      <select
+        className="select-disable-location"
+        defaultValue={deviceSettings.location.isDisabled ? "true" : "false"}
+        onChange={handleDisableLocation}>
+        <option value="true">YES</option>
+        <option value="false">NO</option>
+      </select>
```
kacperkapusciak, line 105: this should be a switch https://www.radix-ui.com/primitives/docs/components/switch

packages/vscode-extension/src/webview/views/DeviceLocationView.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    // @ts-ignore
+    const latitude = event.target[0].value;
+    // @ts-ignore
+    const longitude = event.target[2].value;
+    // @ts-ignore
+    const latDirection = event.target[1].value;
+    // @ts-ignore
+    const lonDirection = event.target[3].value;
```
kacperkapusciak, line 20:
```suggestion
const [latitude, latDirection, longitude, lonDirection] = event.target.map(item => item.value);
```

the `.target[0]`, `.target[1]`, `.target[2]`... screams array destructuring.

here it's not a trivial example but I believe it is a bit readable? we could split it to 2 lines as well something like that
```js
  const locationValues = event.target.map(item => item.value);
  const [latitude, latDirection, longitude, lonDirection] = locationValues;
```

packages/vscode-extension/src/webview/views/DeviceLocationView.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const handleDisableLocation = (e: any) => {
+    e.preventDefault();
+
+    let isDisabled;
+    if (e.target.value === "true") {
+      isDisabled = true;
+    } else {
+      isDisabled = false;
+    }
```
kacperkapusciak, line 40:
```suggestion
    const isDisabled = e.target.value === "true";
```
why

packages/vscode-extension/src/webview/views/DeviceLocationView.tsx
```diff
@@ -0,0 +1,108 @@
...
+      ...deviceSettings,
+      location: {
+        ...deviceSettings.location,
+        latitude: latDirection === "S" ? -parseFloat(latitude) : parseFloat(latitude),
+        longitude: lonDirection === "W" ? -parseFloat(longitude) : parseFloat(longitude),
+      },
+    });
+  };
+
+  const handleDisableLocation = (e: any) => {
```
kacperkapusciak, line 32: let's not use `any`

packages/vscode-extension/src/webview/views/DeviceLocationView.tsx
```diff
@@ -0,0 +1,108 @@
...
+              step="0.000001"
+              defaultValue={
+                deviceSettings.location.latitude < 0
+                  ? -deviceSettings.location.latitude
+                  : deviceSettings.location.latitude
+              }
+            />
+            <select
+              className="direction"
+              defaultValue={deviceSettings.location.latitude < 0 ? "S" : "N"}>
```
kacperkapusciak, line 68: i'm not a fan of magic strings like "S" or "N". They either should be constants on top of the file or grouped as an enum as they never change

kacperkapusciak (changes requested):

kmagiera (changes requested):
We discussed this over call but posting high level comments such that they don't get lost:
 1) let's accept single line input for coords and allow different formatting (use some js libfor parsing). It is convinient to be able to paste the coords from somewhere instead of interacting with four separate controls
 2) don't put form controls below "save" button, the enable/disable switch should be first actually
 3) we don't need save, let's just save on change assuming the provided coords are valid

packages/vscode-extension/src/common/utils.ts
```diff
@@ -27,3 +27,25 @@ export function throttle<T extends (...args: any[]) => any>(
     recentArgs = args;
   } as T;
 }
+
+export function throttleWithTrailing<T extends (...args: any[]) => any>(
```
kmagiera, line 31: how is this different than "throttle"? Maybe add some comment here? Also from reading the code it appears like you can actually use the existing method, no?

kmagiera (approved):
Leaving some minor comments in code to fix before merging

PR #320, opened 2024-05-28
Add support for multi-root workspaces

### LINKED ISSUE #172: Support for multi-root workspaces
I work on many applications at once, and I use multi-root workspace to achieve so. However, when I open a multi-root workspace, even if one of the roots contains `launch.json`, I'm met with the following message:

> React Native IDE couldn't find root application folder in this workspace. Please make sure that the opened workspace contains a valid React Native or Expo project. The way extension verifies the project is by looking for either: app.json, metro.config.js, or node_modules/react-native folder. If your project structure is different, you can set the app root using launch configuration.

It'd be nice to be able to use the extension without the need of opening each project in a separate window.

thejoaov:
This adds support for multi-root workspaces by adding a new appRoot setting. This will allow the definition of the root folder of the app relative to the workspace, which is beneficial for monorepos. Additionally, this changes adjusts the handling of launch configurations to accommodate these multi-root setups.

This also adds a monorepo for testing purposes on /test-apps with this structure:
- apps
	- mobile: the primary application using expo
- packages
	- eslint-config: shared eslint config
	- feature-home: an example of a package containing a feature, with storybook
	- ui: an example of ui library, with storybook

The monorepo was based on [byCedric/expo-monorepo-example](https://github.com/byCedric/expo-monorepo-example)

Closes #172 

https://github.com/user-attachments/assets/ceae5d3a-afdb-45cb-aba9-52ee65eb023e


kmagiera:
Thanks for sending this. Before reviewing, I'd like if you could expand the PR description a bit. I'd like it to include a list of updates you're making and when practical, also explain why it's been decided this way.

As for the placement of the new menu, I don't think the place you'd put it in is the best possible place. However, for the time being I'd be ok with keeping it there. The only think I'm wondering is whether the menu is going to be present for projects with only a single workspace?

kmagiera:
It feels like appRoot shouldn't really be persistent with settings. VSCode settings have workspace and global visibilities. There is no way to limit a setting for workspace only making them not the right choice for us when configuring something that's project specific. For this purpose we either use launch configuration or workspace-level persistence (check how we store simulator zoom or dark/mode settings).

I'm not completely familiar with the setup you'd like to cover here, but do you expect this setting to be something you'd check into the version control? As that's typically what people do with workspace-based settings. I feel like it's something that each individual on the team may want to change at any point and seems to be more like a transient state rather than a setting.

kmagiera (changes requested):
Only now I realized I haven't click "request changes" after adding my last comment

thejoaov:
@kmagiera Makes sense, I agree, it's really not a good idea to use the extension settings to manage the appRoot. In that case, I'll make the modifications to use the workspace settings or a local state as soon as I have time.

kmagiera:
Hi, sorry for the delay here. I missed the updates as the PR was not marked as ready for review. I'm assuming it now is ready and we should look into the code

thejoaov:
hey @kmagiera, applied the changes you requested, can you review it again please?

TimBroddin:
This is a really valuable addition to an already great extension. Is there any reason why this was closed?

PR #323, opened 2024-05-29
Check fingerprint on save when editing node_modules

jakub-gonet:
Tested by editing is-number package in the node_modules directory.

### DIFF
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -48,6 +48,7 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
 
   private nativeFilesChangedSinceLastBuild: boolean;
   private workspaceWatcher!: FileSystemWatcher;
+  private fileSaveWatcherDisposable!: Disposable;
 
   private deviceSession: DeviceSession | undefined;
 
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/project/project.ts
```diff
@@ -96,6 +97,11 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
     this.workspaceWatcher.onDidChange(() => this.checkIfNativeChanged());
     this.workspaceWatcher.onDidCreate(() => this.checkIfNativeChanged());
     this.workspaceWatcher.onDidDelete(() => this.checkIfNativeChanged());
+    this.fileSaveWatcherDisposable = workspace.onDidSaveTextDocument(({ fileName }) => {
+      if (fileName.includes("node_modules")) {
```
kmagiera, line 101: Is it necessary to do the filename check here? It seems like it should matter as otherwise we'd get the watcher triggered. If that's the case I'd just skip it given we have throttling in place and that it'd be more error prone to some other possible updates people are making

jakub-gonet: I initially wanted to await running fingerprinting all the time when people edit their files but we already have fs watchers in place. Will change.

kmagiera (approved):

PR #324, opened 2024-05-29
Prevent infinite loading when creating first device

jakub-gonet:
In some code paths we didn't unset the isLoading state which sometimes caused the loading spinner to last indefinitely. This popped out mostly when, for some reason, dialog didn't show up.

Tested by creating new ios device on empty devices view.


kmagiera (commented):
Can you expand the description a bit. What was the problem before? How does this change fixes it?

kmagiera (approved):

PR #349, opened 2024-06-05
docs: Remove Hire Us Container from subpages

patrycjakalinska:
Hire Us Container is redundant in `Privacy Policy`, `Terms Of Use` and `Refund Policy` subpages, so we decided to hide it there.

Before:
image
image
image

After:
image
image
image


kacperkapusciak (approved):
thank u!

PR #371, opened 2024-06-12
Add tests

jakub-gonet:
Adds [vscode-test](https://github.com/microsoft/vscode-test) package that allows for testing extension.

We configure it's user data directory because it creates a socket and socket's path is used as a key. Unfortunately, it only supports 103 characters so long paths get trimmed and it doesn't clean the socket before next run.

For tests, it uses mocha. We additionally use [sinon](https://sinonjs.org/releases/v18/) for creating mocks.

Mocha only supports JS files, we transpile them using TS in the dist directory, under `test/`. We ignore this directory when creating a VSIX package.

Tests are structured as follows:
- `integration-tests` directory for more end-to-end testing.
- `XXX.test.ts` files next to implementation to unit test it.

To run tests execute `npm run test` command which rebuilds webview and extension, then running tests.
It expects simulator executable in the dist directory, you can regenerate it with `npm run build:sim-server-debug`.

Few sources that can be used as a reference for writing tests:
- [Basic mocking](https://github.com/microsoft/vscode/issues/45531)
- [Sample project for vscode-test](https://github.com/microsoft/vscode-test/tree/main/sample)
- [VSCode-js-debug extension tests](https://github.com/microsoft/vscode-js-debug/tree/main/src/test)

### DIFF
--- a/packages/vscode-extension/.gitignore
+++ b/packages/vscode-extension/.gitignore
@@ -2,3 +2,4 @@ out/
 dist/
 node_modules/
 *.vsix
+.vscode-test
--- a/packages/vscode-extension/.vscode-test.js
+++ b/packages/vscode-extension/.vscode-test.js
@@ -1,0 +1,11 @@
+const { defineConfig } = require("@vscode/test-cli");
+const { tmpdir } = require("os");
+const { sep } = require("path");
+const { mkdtempSync } = require("fs");
+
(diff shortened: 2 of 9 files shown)


jakub-gonet:
It's possible to run tests without running VSC instance but we'd need to mock vscode imports somehow. I left it for the future as it's good enough in the current shape.

kmagiera (approved):
Looks good 👍

PR #384, opened 2024-06-14
Fix displaying component names in the preview URL

jakub-gonet:
When `preview(<View />)` was used, URL bar shown "preview:undefined".

Turns out that forwarded ref components have different structure for `.type` which doesn't include meaningful name in it.

### DIFF
--- a/packages/vscode-extension/lib/preview.js
+++ b/packages/vscode-extension/lib/preview.js
@@ -14,6 +14,21 @@ function stringifyProps(obj) {
   return keyValuePairs.join(" ");
 }
 
+function getComponentName({ type }) {
+  const name = type.name;
+  const isClassOrFunctionComponent = name !== undefined;
+  if (isClassOrFunctionComponent) {
+    return name;
+  }
+
+  const isForwardedRef = type["$$typeof"] === Symbol.for("react.forward_ref");
+  if (isForwardedRef) {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/lib/preview.js
```diff
@@ -14,6 +14,21 @@ function stringifyProps(obj) {
...
+function getComponentName({ type }) {
+  const name = type.name;
+  const isClassOrFunctionComponent = name !== undefined;
+  if (isClassOrFunctionComponent) {
+    return name;
+  }
+
+  const isForwardedRef = type["$$typeof"] === Symbol.for("react.forward_ref");
+  if (isForwardedRef) {
+    return "(forwarded ref)";
```
filip131311, line 26: I think it would be more usefull for the user to see a generic name like "Component ${index}" here, because it could help them navigate the menu between multiple components.

filip131311 (changes requested):

kmagiera (approved):
Wonder if we could use the version of this method that react devtools rely on: https://github.com/facebook/react/blob/main/packages/shared/getComponentNameFromType.js#L56

Otherwise it looks good so accepting.

jakub-gonet:
We could but it doesn't give us much. We can revisit this is need arises.

PR #390, opened 2024-06-17
Avoid closing IDE accidentally

kewinzaq1:
## Problem
Currently, there is a chance of accidentally closing the IDE panel and it can be an unpleasant experience
## Solution
Simple confirmation popup to avoid headaches.  The following popup only appears if users click `cmd+w/ctr+w` 
https://github.com/software-mansion/react-native-ide/assets/77052270/70c45f35-9d89-4b99-ad97-9f69e0cbc96e

https://github.com/software-mansion/react-native-ide/issues/387

### DIFF
--- a/packages/vscode-extension/src/panels/Tabpanel.ts
+++ b/packages/vscode-extension/src/panels/Tabpanel.ts
@@ -20,13 +20,23 @@ export class TabPanel implements Disposable {
   public static currentPanel: TabPanel | undefined;
   private readonly _panel: WebviewPanel;
   private webviewController: WebviewController;
+  private changeViewStateDisposable: Disposable;
 
   private constructor(panel: WebviewPanel, context: ExtensionContext) {
     this._panel = panel;
     this._panel.iconPath = Uri.joinPath(context.extensionUri, "assets", "logo.svg");
     // Set an event listener to listen for when the panel is disposed (i.e. when the user closes
     // the panel or when the panel is closed programmatically)
-    this._panel.onDidDispose(() => this.dispose());
+
+    commands.executeCommand("setContext", "RNIDE.isTabPanelFocused", this._panel.active);
+
+    this.changeViewStateDisposable = this._panel.onDidChangeViewState((e) => {
+      commands.executeCommand("setContext", "RNIDE.isTabPanelFocused", this._panel.active);
+    });
+
+    this._panel.onDidDispose(() => {
+      this.dispose();
+    });
 
     // Set the HTML content for the webview panel
     this._panel.webview.html = generateWebviewContent(
@@ -88,6 +98,9 @@ export class TabPanel implements Disposable {
     // this is triggered when the user closes the webview panel by hand, we want to reset open_panel_on_activation
     // key in this case to prevent extension from automatically opening the panel next time they open the editor
     extensionContext.workspaceState.update(OPEN_PANEL_ON_ACTIVATION, undefined);
+
+    this.changeViewStateDisposable.dispose();
+    commands.executeCommand("setContext", "RNIDE.isTabPanelFocused", false);
 
     TabPanel.currentPanel = undefined;
 
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -68,6 +68,12 @@
         "command": "RNIDE.openDevMenu",
         "key": "ctrl+alt+z",
         "mac": "ctrl+cmd+z"
+      },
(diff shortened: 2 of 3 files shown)


kmagiera (changes requested):
Thanks for this. I like the idea but would suggest the following changes:
1) use onDidChangeActiveTextEditor instread of the tab group change event
2) register the event in Tabpanel instead of extension.ts and make sure we call dispose on the result of registration when dispose is triggered on the panel
3) Use panel.active field to tell whether the IDE panel is focused (https://code.visualstudio.com/api/references/vscode-api#WebviewPanel)
4) Rename `isTabActive` to `isTabPanelFocused` since word "active" is ambiguous in this context
5) Rename "doubleClickClose" to "closeWithConfirmation"

kewinzaq1:
@k

> Thanks for this. I like the idea but would suggest the following changes:
> 
> 1. use onDidChangeActiveTextEditor instread of the tab group change event
> 2. register the event in Tabpanel instead of extension.ts and make sure we call dispose on the result of registration when dispose is triggered on the panel
> 3. Use panel.active field to tell whether the IDE panel is focused (https://code.visualstudio.com/api/references/vscode-api#WebviewPanel)
> 4. Rename `isTabActive` to `isTabPanelFocused` since word "active" is ambiguous in this context
> 5. Rename "doubleClickClose" to "closeWithConfirmation"



> Thanks for this. I like the idea but would suggest the following changes:
> 
> 1. use onDidChangeActiveTextEditor instread of the tab group change event
> 2. register the event in Tabpanel instead of extension.ts and make sure we call dispose on the result of registration when dispose is triggered on the panel
> 3. Use panel.active field to tell whether the IDE panel is focused (https://code.visualstudio.com/api/references/vscode-api#WebviewPanel)
> 4. Rename `isTabActive` to `isTabPanelFocused` since word "active" is ambiguous in this context
> 5. Rename "doubleClickClose" to "closeWithConfirmation"

Thanks for the review.
I tried to go with` onDidChangeActiveTextEditor` however I ended up checking the active state 
 `this._panel.active`  before update had happened, so I moved to the direct method on the panel; `onDidChangeViewState` it allows avoiding this problem in a pretty straightforward way. What do you think about this approach?

kewinzaq1:
@kmagiera Thanks for the review.
I tried to go with` onDidChangeActiveTextEditor` however I ended up checking the active state 
 `this._panel.active`  before the update had happened, so I moved to the direct method on the panel; `onDidChangeViewState` allows for avoiding this problem pretty straightforwardly. What do you think about this approach?

packages/vscode-extension/src/panels/Tabpanel.ts
```diff
@@ -26,7 +26,18 @@ export class TabPanel implements Disposable {
...
+
+    commands.executeCommand("setContext", "RNIDE.isTabPanelFocused", this._panel.active);
+
+    const changeViewEvent = this._panel.onDidChangeViewState((e) => {
+      commands.executeCommand("setContext", "RNIDE.isTabPanelFocused", this._panel.active);
+    });
+
+    this._panel.onDidDispose(() => {
+      changeViewEvent.dispose();
+      commands.executeCommand("setContext", "RNIDE.isTabPanelFocused", false);
```
kmagiera, line 38: to keep the schema similar to what we have in other classes, I'd prefer to keep all the logic responsible for diposing inside of the `dispose` method from line 97. Can you make it so that `onDidDispose` just calls `this.dispose()` and move the rest of the added code to that method. This will require that we store event listener as class memeber but that's ok as we do the same with other disposable things.

For the cases when there are more than a couple of disposable items we have this `subscriptions` concept that's used in extension.ts but I'd rather not use it here as long as we have just a few things to dispose.

kmagiera (changes requested):
This is much better. I'm ok with using `onDidChangeViewState` if it works and gets triggered at the right moment. Only added one more comment inline and this should be good to go.

packages/vscode-extension/src/panels/Tabpanel.ts
kewinzaq1: @kmagiera thanks for the suggestions I applied them. `This will require that we store event listener as class memeber but that's ok as we do the same with other disposable things.` - yeah.. at first I thought about this and that was the reason I avoided that... however separating things and keeping a consistent approach across code-base makes more sense.

kmagiera (approved):
This is great. Thanks for working on the requested changes!

PR #399, opened 2024-06-21
Fix JSON reporter for Expo

jakub-gonet:
Previously, we overrode metro-config to inject our config in Expo apps because Expo doesn't allow for customizing metro config directly.

Expo generates config based on metro-config but then customizes it by overriding reporter setting (see [here](https://github.com/expo/expo/blob/a208802b51c73a2799048d30f1d4b9240bffc6c0/packages/%40expo/cli/src/start/server/metro/instantiateMetro.ts#L94-L108)

To work around that, we replace the base class for TerminalReporter that expo CLI uses with out metro_reporter that we want to set as the reporter in configuration.

*Test plan:*
1. Run expo devclient project
2. Get loading percentage on "waiting app to load" step (which is the percentage from metro that is read using json reporter)


kmagiera (changes requested):
As discussed offline. Would be good to avoid importing internals from expo cli especially that we already had issues with some changes related to bundling in the past. There is `setEventReporter` method that's exposed from the config loader in cli codebase. Perhaps we could use that one?

kmagiera (approved):

PR #401, opened 2024-06-21
Refresh preview on fast refresh

### LINKED ISSUE #288: Preview is not working well when component changes.
**Description*
Preview is not working well when the component changes. 
I have a component which has title props. I changed the title props value but ui is not updated. After clicked open preview, UI is updated.

**Environment**
- RN IDE: v0.0.14
- EXPO: ~51.0.1 RN: 0.74.1

Video 
https://github.com/software-mansion/react-native-ide/assets/37181742/efb354e4-4295-4306-bc7b-b42c3fc9eca7

jakub-gonet:
Fast refresh is triggered when props are changed in previewed component. We didn't accommodate for that previously, as we only rendered preview once, after clicking a button from code lenses.

Additionally:
- cleared up logic in code lenses and added a check for preview import to prevent false positives.
- added check preventing crash on calling `preview()` without any component.
- fixed URL being out of sync when switching between preview/non-preview

Fixes #288.

### DIFF
--- a/packages/vscode-extension/lib/expo_router_plugin.js
+++ b/packages/vscode-extension/lib/expo_router_plugin.js
@@ -1,7 +1,6 @@
-import { store } from "expo-router/build/global-state/router-store.js";
-
-import { useSyncExternalStore, useEffect, useState } from "react";
+import { useSyncExternalStore, useEffect } from "react";
 import { useRouter } from "expo-router";
+import { store } from "expo-router/build/global-state/router-store.js";
 
 function computeRouteIdentifier(pathname, params) {
   return pathname + JSON.stringify(params);
@@ -9,13 +8,11 @@ function computeRouteIdentifier(pathname, params) {
 
 function useRouterPluginMainHook({ onNavigationChange }) {
   const router = useRouter();
-  const [initCallbacks, setInitCallbacks] = useState([]);
   const routeInfo = useSyncExternalStore(
     store.subscribeToRootState,
     store.routeInfoSnapshot,
     store.routeInfoSnapshot
   );
-
   const pathname = routeInfo?.pathname;
   const params = routeInfo?.params;
   useEffect(() => {
@@ -26,16 +23,6 @@ function useRouterPluginMainHook({ onNavigationChange }) {
       id: computeRouteIdentifier(pathname, params),
     });
   }, [pathname, params]);
-
-  useEffect(() => {
-    if (router.navigationRef) {
-      for (const callback of initCallbacks) {
-        callback();
-      }
-      setInitCallbacks([]);
-    }
-  }, [router.navigationRef]);
-
   return {
     getCurrentNavigationDescriptor: () => {
       const snapshot = store.routeInfoSnapshot();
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/lib/wrapper.js
```diff
@@ -209,6 +216,10 @@ export function PreviewAppWrapper({ children, ...rest }) {
       };
       LoadingView.hide = () => {
         devtoolsAgent._bridge.send("RNIDE_fastRefreshComplete");
+        if (isRunningPreview()) {
+          // refresh preview component
+          openPreview(getCurrentSceneName());
```
kmagiera, line 221: I like all the other changes in this PR except for this one. It doesn't seem like the right choice to manually fix fast-refresh behavior. I wonder why does it in the main app but not in preview,  and would expect it to "just work" here as well? Maybe it's because preview is the root element and fast refresh doesn't support it? Maybe it'd be enough to just wrap it in some other root component?

packages/vscode-extension/src/webview/components/UrlBar.tsx
```diff
@@ -57,7 +56,10 @@ function UrlBar({ project, disabled }: UrlBarProps) {
         <span className="codicon codicon-refresh" />
       </IconButton>
       <IconButton
-        onClick={() => project.goHome()}
+        onClick={() => {
+          project.goHome();
+          setUrlList([]);
```
jakub-gonet, line 61: This isn't ideal but we relied on metro reload to do it for us. In the future, we need to keep `/` always in the URL list. I'll create a ticket for that.

packages/vscode-extension/lib/expo_router_plugin.js
```diff
@@ -1,18 +1,21 @@
-import { useSyncExternalStore, useEffect } from "react";
-import { useRouter } from "expo-router";
 import { store } from "expo-router/build/global-state/router-store.js";
```
jakub-gonet, line 1: Moved the store import to the top so it's more evident that those two files are different only there.

packages/vscode-extension/lib/wrapper.js
```diff
@@ -47,35 +51,34 @@ export function PreviewAppWrapper({ children, ...rest }) {
...
     },
-    [rootTag]
+    [rootTag, devtoolsAgent]
   );
 
   const closePreview = useCallback(() => {
-    const SceneTracker = require("react-native/Libraries/Utilities/SceneTracker");
-    const isRunningPreview = isPreviewUrl(SceneTracker.getActiveScene().name);
-    if (isRunningPreview) {
+    if (currentPreviewKey) {
```
kmagiera, line 81: maybe check active scene name, we don't need the exact preview name here as we only check if its running or not

kmagiera (changes requested):
Added some comments inline. Also please remove changes to navigation integration interface. If there are issues in expo router integration due to early navChange calls, this should be handled by that implementation alone. Also such changes shouldn't be a part of this PR as they seem to be fixing a different problem

packages/vscode-extension/lib/preview.js
```diff
@@ -1,36 +1,31 @@
...
+  return (
+    <View style={{ flex: 1, justifyContent: "center", alignItems: "center" }}>
+      {previewData.component}
+    </View>
+  );
 }
 
 export function preview(component) {
-  if (component._source == null) {
+  if (!component || component._source === null) {
```
kmagiera, line 20: I think we may want to have `==` check here. I think `_source` could be `undefined` here too, and since it's internal field I wouldn't rely on it being set to `null`

kmagiera (approved):
Looks good, please check the inline comment I added tho

PR #409, opened 2024-06-25
Add possibility to hide zoom controls

michalzuk:
In case of a tight panel width, it’s easy for the simulator to overlap with the zoom controls. This PR adds an option in the settings to toggle the visibility of these controls. For me it's extremally useful with the “Fit” zoom setting.

Before:
image

After:
image

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -10,6 +10,10 @@ export type DeviceSettings = {
   };
 };
 
+export type WindowState = {
+  hoveredEdge: "top" | "bottom" | "left" | "right" | null;
+};
+
 export type ProjectState = {
   status:
     | "starting"
@@ -83,6 +87,7 @@ export interface ProjectEventMap {
   log: { type: string };
   projectStateChanged: ProjectState;
   deviceSettingsChanged: DeviceSettings;
+  windowStateChanged: WindowState;
   navigationChanged: { displayName: string; id: string };
   needsNativeRebuild: void;
 }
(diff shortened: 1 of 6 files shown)


kmagiera:
This is a good point and we actually discussed this internally. Our preference for this case would be to handle that without additional options. I'd like to keep the number of possible settings as small as possible such that it is easy to navigate. In this scenario I'd like the controls to hide automatically and reveal automatically when you hover over the left edge of the panel

kmagiera (changes requested):
Forgot to request changes on this one with my last comment

michalzuk:
@kmagiera Thanks for providing more details. I've made changes to reflect desired behavior in the app.

packages/vscode-extension/src/common/Project.ts
```diff
@@ -101,6 +106,9 @@ export interface ProjectInterface {
   getDeviceSettings(): Promise<DeviceSettings>;
   updateDeviceSettings(deviceSettings: DeviceSettings): Promise<void>;
 
+  getWindowState(): Promise<WindowState>;
+  updateWindowState(newState: Partial<WindowState>): Promise<void>;
```
kmagiera, line 110: I don't think we want to add this here. This is an interface for communicating between the frontend of the extension (what gets displayed in the panel) and the main extension process (what runs in NodeJS environment inside VSCode). It doesn't seem like it's needed by the main process to understand whether someone is hovering edges of the device.

The change you're making belongs fully into the frontend code. It's sole purpose is to present zoom controls in a way they don't cover device window and because of that there should be no changes made outside of the front-end part which lives under `packages/vscode-extension/src/webview/`

PR #416, opened 2024-07-01
Make clicking progress message more discoverable

jakub-gonet:
The ability to click the startup message to see more was pretty obscure. Now, if some startup step takes more than 30s, we underline the message so users understand clickability. Most of the time it's not needed, this is meant for situations where RNIDE gets stuck on some step.

https://github.com/software-mansion/react-native-ide/assets/12465392/80ef8bff-9509-46cb-9650-fc8813fe695f

### DIFF
--- a/packages/vscode-extension/src/webview/components/PreviewLoader.tsx
+++ b/packages/vscode-extension/src/webview/components/PreviewLoader.tsx
@@ -1,4 +1,5 @@
-import { useEffect, useState } from "react";
+import classNames from "classnames";
+import { useEffect, useRef, useState } from "react";
 
 import "./PreviewLoader.css";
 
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/webview/components/PreviewLoader.tsx
```diff
@@ -9,14 +10,23 @@ import { StartupMessage, StartupStageWeight } from "../../common/Project";
...
 function PreviewLoader({ onRequestShowPreview }: { onRequestShowPreview: () => void }) {
   const { projectState, project } = useProject();
   const [progress, setProgress] = useState(0);
 
+  const [previousProgress, setPreviousProgress] = useState({
+    status: projectState.status,
+    startupMessage: projectState.startupMessage,
+  });
+  const [isLoadingSlowly, setIsLoadingSlowly] = useState(false);
+  console.log({ isLoadingSlowly });
```
kmagiera, line 27: nope

packages/vscode-extension/src/webview/components/PreviewLoader.tsx
```diff
@@ -25,23 +35,44 @@ function PreviewLoader({ onRequestShowPreview }: { onRequestShowPreview: () => v
...
+        isLoadingSlowlyTimeout.current = setTimeout(() => {
+          setIsLoadingSlowly(true);
+        }, slowLoadingThresholdMs);
+      }
+    }
+    setPreviousProgress({
+      status: projectState.status,
+      startupMessage: projectState.startupMessage,
+    });
+  }, [projectState]);
```
kmagiera, line 74: shouldn't we just have projectState.status on deps list and then we don't need to add previousStatus as new state because useEffect will handle this for us

kmagiera (changes requested):

kmagiera (approved):

PR #419, opened 2024-07-03
Add information about keybindings to dropdown menu. 

filip131311:
### Description: 

This PR is addressing a problem with discoverability of a keybinding that allows users to open dev-menu (cmd+shift+z by default). The new UI hopefully, will clear up the confusion users may had, caused by  their experience with expo shortcuts. It also provides a function that allows to check current keybinding for commands contributed by RNIDE, even if user has set custom ones.  

### Potential improvements: 
This functionality could be further improved by adding a listener that would inform the UI about changes in user keybindings (aka. keybindings.json file), but as it seems very unlikely that somebody is going to have a dropdown menu open and at the same time changing keybindings in vscode settings, it does not seem to be a priority right now. 

### Note 
keybinding.json is the only place we should expect keybinding to be changed, according to [this article](https://code.visualstudio.com/docs/getstarted/keybindings)
also alexdima explains here why they do not plan on introducing per workspace keybindings [here](https://github.com/Microsoft/vscode/issues/4504)

### before 
Screenshot 2024-07-03 at 13 55 29

### after 
Screenshot 2024-07-08 at 15 56 28

### DIFF
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -33,6 +33,10 @@ import stripAnsi from "strip-ansi";
 import { minimatch } from "minimatch";
 import { IosSimulatorDevice } from "../devices/IosSimulatorDevice";
 import { AndroidEmulatorDevice } from "../devices/AndroidEmulatorDevice";
+import path from "path";
+import { homedir } from "node:os";
+import fs from "fs";
+import JSON5 from "json5";
 
 const DEVICE_SETTINGS_KEY = "device_settings_v2";
 const LAST_SELECTED_DEVICE_KEY = "last_selected_device";
@@ -370,6 +374,52 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
 
   public onActiveFileChange(filename: string, followEnabled: boolean) {
     this.deviceSession?.onActiveFileChange(filename, followEnabled);
+  }
+
+  public async getCommandsCurrentKeyBinding(commandName: string) {
+    const packageJsonPath = path.join(extensionContext.extensionPath, "package.json");
+    const extensionPackageJson = require(packageJsonPath);
+    let keybindingsJsonPath;
+    let keybindingsJson;
+    try {
+      keybindingsJsonPath = path.join(
+        homedir(),
+        "Library/Application Support/Code/User/keybindings.json"
+      );
+      // can not use require because the file may contain comments
+      keybindingsJson = JSON5.parse(fs.readFileSync(keybindingsJsonPath).toString());
+    } catch (e) {
+      Logger.error("error while parsing keybindings.json", e);
+      return undefined;
+    }
+
+    const isRNIDECommand =
+      extensionPackageJson.contributes.commands &&
+      extensionPackageJson.contributes.commands.find((command: any) => {
+        return command.command === commandName;
+      });
+    if (!isRNIDECommand) {
+      Logger.warn("Trying to access a keybinding for a command that is not part of an extension.");
+      return undefined;
+    }
+
+    const userKeybinding = keybindingsJson.find((command: any) => {
+      return command.command === commandName;
+    });
+    if (userKeybinding) {
+      return userKeybinding.key;
+    }
+
+    const defaultKeybinding = extensionPackageJson.contributes.keybindings.find(
+      (keybinding: any) => {
+        return keybinding.command === commandName;
+      }
+    );
+    if (defaultKeybinding) {
+      return defaultKeybinding.mac;
+    }
+
+    return undefined;
   }
 
   public async getDeviceSettings() {
--- a/packages/vscode-extension/src/webview/components/SettingsDropdown.tsx
+++ b/packages/vscode-extension/src/webview/components/SettingsDropdown.tsx
@@ -7,6 +7,7 @@ import ManageDevicesView from "../views/ManageDevicesView";
 import { ProjectInterface } from "../../common/Project";
 import DoctorIcon from "./icons/DoctorIcon";
 import { useWorkspaceConfig } from "../providers/WorkspaceConfigProvider";
+import { KeybindingInfo } from "./shared/KyebindingInfo";
 
 interface SettingsDropdownProps {
   children: React.ReactNode;
@@ -50,7 +51,16 @@ function SettingsDropdown({ project, isDeviceRunning, children, disabled }: Sett
               project.openDevMenu();
             }}>
             <span className="codicon codicon-code" />
-            Open dev menu
+            <div
+              style={{
+                display: "flex",
+                width: "100%",
+                flexDirection: "row",
+                justifyContent: "space-between",
+              }}>
+              Open dev menu
+              <KeybindingInfo commandName="RNIDE.openDevMenu" />
+            </div>
           </DropdownMenu.Item>
 
           <DropdownMenu.Sub>
--- a/packages/vscode-extension/src/webview/components/shared/KyebindingInfo.tsx
+++ b/packages/vscode-extension/src/webview/components/shared/KyebindingInfo.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { ReactNode, useEffect, useState } from "react";
 import "./KeybindingInfo.css";
 import { useProject } from "../../providers/ProjectProvider";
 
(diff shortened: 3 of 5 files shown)


packages/vscode-extension/src/project/project.ts
```diff
@@ -372,6 +376,52 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
     this.deviceSession?.onActiveFileChange(filename, followEnabled);
   }
 
+  public async getCommandsCurrentKeyBinding(commandName: string) {
```
kmagiera, line 379: can we extract this method from project.ts ? It seems to be very loosely related and we already have a ton of code in this file

packages/vscode-extension/src/project/project.ts
```diff
@@ -372,6 +376,52 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
+    const packageJsonPath = path.join(extensionContext.extensionPath, "package.json");
+    const extensionPackageJson = require(packageJsonPath);
```
kmagiera, line 381: note that require will load the file only once and then subsequent calls will just load the cached version.

packages/vscode-extension/src/project/project.ts
```diff
@@ -372,6 +376,52 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
+      return undefined;
+    }
+
+    const isRNIDECommand =
+      extensionPackageJson.contributes.commands &&
+      extensionPackageJson.contributes.commands.find((command: any) => {
+        return command.command === commandName;
+      });
+    if (!isRNIDECommand) {
+      Logger.warn("Trying to access a keybinding for a command that is not part of an extension.");
```
kmagiera, line 402: what's the reason we're adding this restriction? I saw some other PR that was overriding the behavior of the hotkey for closing the tab, so it may sometimes be useful to read hotkeys for other commands too, no?

packages/vscode-extension/src/webview/components/shared/KyebindingInfo.tsx
```diff
@@ -0,0 +1,59 @@
...
+      cmd: "\u2318", // ⌘
+      ctrl: "\u2303", // ⌃
+      alt: "\u2325", // ⌥
+      option: "\u2325", // ⌥
+      shift: "\u21E7", // ⇧
+      tab: "\u21E5", // ⇥
+      up: "\u2191", // ↑
+      down: "\u2193", // ↓
+      left: "\u2190", // ←
+      right: "\u2192", // →
```
kmagiera, line 32: can't we just have the symbols in the strings here instead of unicode codes and symbols in the comments?

packages/vscode-extension/src/webview/components/shared/KyebindingInfo.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const Symbol = ({ children }: SymbolProps) => {
+  return <span className="symbol">{children}</span>;
+};
+
+export const KeybindingInfo = ({ commandName }: KeybindingInfoProps) => {
+  const { project } = useProject();
+  const [keybinding, setKeybinding] = useState<string[]>([]);
+
+  function translateToUnicode(symbol: string) {
```
kmagiera, line 21: this method doesn't need to be defined inside of the component or am I missing something?

packages/vscode-extension/src/webview/components/shared/KyebindingInfo.tsx
```diff
@@ -0,0 +1,59 @@
...
+  commandName: string;
+}
+
+interface SymbolProps {
+  children: ReactNode;
+}
+
+const Symbol = ({ children }: SymbolProps) => {
+  return <span className="symbol">{children}</span>;
+};
```
kmagiera, line 15: This doesn't seem like a particularily interesting component, you can just inline <span/> in the place where you use it

kmagiera (approved):
Looks good, please address inline comments before merging

packages/vscode-extension/src/project/project.ts
filip131311: that's okay since keybindings set in package.json are static.

packages/vscode-extension/src/project/project.ts
filip131311: I am not entirely  sure how to write a universal method, since the commands default  keybinding might be defined in multiple places (depending on if it is native for vscode or contributed by another extension), I feel like It is a significant enough challange to do it, when it is needed, a specially that I can't think of a use case for informing our users about keybindings that are not contributed by us. Right now the only use case I see is to increase discoverability of our on functionality.

packages/vscode-extension/src/project/project.ts
filip131311: @kmagiera I agree but It seems that the best solution would be to create a new interface for utilities like that for example: 
- reportIssue(): Promise<void>;
- openDevMenu(): Promise<void>;
- openFileAt(filePath: string, line0Based: number, column0Based: number): Promise<void>;
- movePanelToNewWindow(): void;

I'll create another PR for that at it is out of scope for this one. What do you think?

packages/vscode-extension/src/project/project.ts
```diff
@@ -372,6 +376,52 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
+    let keybindingsJson;
+    try {
+      keybindingsJsonPath = path.join(
+        homedir(),
+        "Library/Application Support/Code/User/keybindings.json"
+      );
+      // can not use require because the file may contain comments
+      keybindingsJson = JSON5.parse(fs.readFileSync(keybindingsJsonPath).toString());
+    } catch (e) {
+      Logger.error("error while parsing keybindings.json", e);
```
jakub-gonet, line 392: Nit: capitalize the error.

packages/vscode-extension/src/project/project.ts
jakub-gonet: Seconded, it'd be nice to have it in utils/ or common/. Refactor can be done later.

packages/vscode-extension/src/project/project.ts
```diff
@@ -372,6 +376,52 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
 
+  public async getCommandsCurrentKeyBinding(commandName: string) {
+    const packageJsonPath = path.join(extensionContext.extensionPath, "package.json");
+    const extensionPackageJson = require(packageJsonPath);
+    let keybindingsJsonPath;
+    let keybindingsJson;
+    try {
+      keybindingsJsonPath = path.join(
+        homedir(),
+        "Library/Application Support/Code/User/keybindings.json"
```
jakub-gonet, line 387: Wonder if we want allow user to override this for non-standard VSCode installation.

packages/vscode-extension/src/project/project.ts
```diff
@@ -372,6 +376,52 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
+      );
+      // can not use require because the file may contain comments
```
jakub-gonet, line 389: Typo: "Cannot" or even remove the comment.

packages/vscode-extension/src/project/project.ts
```diff
@@ -372,6 +376,52 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
+    } catch (e) {
+      Logger.error("error while parsing keybindings.json", e);
+      return undefined;
+    }
+
+    const isRNIDECommand =
+      extensionPackageJson.contributes.commands &&
+      extensionPackageJson.contributes.commands.find((command: any) => {
+        return command.command === commandName;
+      });
```
jakub-gonet, line 400: Nit: this doesn't return boolean, maybe you want something like: 
```
const isRNIDECommand =  extensionPackageJson.contributes.commands?.includes({command}) => command === commandName);
```

packages/vscode-extension/src/project/project.ts
```diff
@@ -372,6 +376,52 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
+
+    const userKeybinding = keybindingsJson.find((command: any) => {
+      return command.command === commandName;
+    });
+    if (userKeybinding) {
+      return userKeybinding.key;
+    }
+
+    const defaultKeybinding = extensionPackageJson.contributes.keybindings.find(
+      (keybinding: any) => {
```
jakub-gonet, line 414: This could be better typed.

packages/vscode-extension/src/webview/components/SettingsDropdown.tsx
```diff
@@ -50,7 +51,16 @@ function SettingsDropdown({ project, isDeviceRunning, children, disabled }: Sett
...
             }}>
             <span className="codicon codicon-code" />
-            Open dev menu
+            <div
+              style={{
+                display: "flex",
+                width: "100%",
+                flexDirection: "row",
+                justifyContent: "space-between",
+              }}>
```
jakub-gonet, line 60: Please use CSS.

packages/vscode-extension/src/webview/components/shared/KyebindingInfo.tsx
```diff
@@ -0,0 +1,55 @@
...
+    alt: "⌥",
+    option: "⌥",
+    shift: "⇧",
+    tab: "⇥",
+    up: "↑",
+    down: "↓",
+    left: "←",
+    right: "→",
+  };
+  return icons[symbol.toLowerCase() as keyof typeof icons] || symbol.toUpperCase();
```
jakub-gonet, line 22: Nit: use `??` instead of `||`.

packages/vscode-extension/src/webview/components/shared/KyebindingInfo.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const KeybindingInfo = ({ commandName }: KeybindingInfoProps) => {
+  const { project } = useProject();
+  const [keybinding, setKeybinding] = useState<string[]>([]);
+
+  useEffect(() => {
+    project.getCommandsCurrentKeyBinding(commandName).then((res) => {
```
jakub-gonet, line 30: Nit: `async`? Also would be awesome if we got better name than `res`.

PR #421, opened 2024-07-04
Fix expo-router navigation

jakub-gonet:
When going in and out of preview, we remount main app. Previously, we'd navigate to given path immediately which wasn't working as navigator was still unitialized.

This PR changes the following things:
1. We no longer call navigation change request immediately after closePreview. This was wrong as the new app haven't had a chance to render and mount. Instead we're changing closePreview to return promise that resolves when mounting the app is complete and only then take necessary actions
2. We updated navigation change request handling on the expo router plugin side, such that it checks whether the navigator is in ready state. If not, we wait for onReady event.

|Before|After|
|-|-|
|<video src="https://github.com/software-mansion/react-native-ide/assets/12465392/f5327e03-1316-4159-92d5-d9a550b48439" />|<video src="https://github.com/software-mansion/react-native-ide/assets/12465392/f1e046e4-ad84-4f91-80a5-bc5f761372f1" />| 

## Testing
1. Open expo-router example, navigate to details
3. Open preview (i.e. from UglyButton)
4. Use URL bar to navigate back to details (w/o this change it'd navigate to home screen instead of details)


packages/vscode-extension/lib/expo_router_v2_plugin.js
```diff
@@ -33,8 +54,12 @@ function useRouterPluginMainHook({ onNavigationChange }) {
         id: computeRouteIdentifier(snapshot.pathname, snapshot.params),
       };
     },
-    requestNavigationChange: ({ pathname, params }) => {
-      router.push(pathname, params);
+    requestNavigationChange: (navigationDescriptor) => {
```
kewinzaq1, line 57: @jakub-gonet Maybe we could simplify this approach and use while loop instead? What do you think?
```tsx
while (store.navigationRef?.isReady() === false) {
        await new Promise((resolve) => setTimeout(resolve, 100));
      }
      requestNavigationChange(navigationDescriptor);
```

kmagiera (approved):

PR #422, opened 2024-07-04
Allow disabling device frame

kewinzaq1:
### The improvement
This Pull Request introduces an option to disable the device frame. It can be handy for people working with one screen. The motivation to implement this gives me: https://github.com/software-mansion/react-native-ide/issues/418

### Result
allow-disabling-device-frame

https://github.com/software-mansion/react-native-ide/assets/77052270/c99d69ca-f44a-41de-a177-e18ed5a68bff


kmagiera:
I'm fine with adding this as an option, just not sure how I feel about placing it under the device settings as those were meant to be options of the device while frame is purely a panel presentation setting (more like zoom levels). Also, when you add option like that to project, it will default to undefined and so the new users and people who upgrade the extension will get no frame by default until they filp the switch.

Wanted to call in @jakub-gonet for an opinion on this one

kewinzaq1:
@kmagiera 
Thanks for pointing out the possibility with unexpectedly disabling frame - the fallback should prevent this issue [packages/vscode-extension/src/project/project.ts](https://github.com/software-mansion/react-native-ide/pull/422/files#diff-8d5143a1f3455de6dbfaa820c4b22f0f3af065a16dd2cf0a27a5e4f3b59d52d6)

packages/vscode-extension/src/project/project.ts
```diff
@@ -60,17 +60,28 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
+    if (!savedSettings) {
+      return {
+        appearance: "dark",
+        contentSize: "normal",
+        location: {
+          latitude: 50.048653,
+          longitude: 19.965474,
+          isDisabled: true,
+        },
+        showFrame: true,
```
filip131311, line 75: I think settings like that should be located tin WorkspaceConfiguration, similarly to how "Panel Location" is implemented. We could possibly  even just relay on vscode internal ui for "User Settings".

filip131311 (commented):
I left inline comments.

jakub-gonet (changes requested):
We don't have a better place for that, we may think about moving it to IDE configuration since it typically isn't changed on regular basis. For now, I feel we can leave it in the device settings.

Before merging, we need to fix weird layout issue that happens after switching + change the color of switch to match the rest of the theme.

https://github.com/software-mansion/react-native-ide/assets/12465392/a3d58f14-efcd-485f-b45b-659e51183155

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -112,6 +113,26 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
             <span className="codicon codicon-location" />
             Set Device Location
           </DropdownMenu.Item>
+          <Label>Device Frame</Label>
+          <form>
+            <div style={{ display: "flex", alignItems: "center" }}>
```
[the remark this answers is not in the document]
kewinzaq1: @jakub-gonet https://github.com/software-mansion/react-native-ide/pull/422#pullrequestreview-2168429727 
`change the color of switch to match the rest of the theme.` - What do you exactly mean? Do you mean color the checked state color?

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -112,6 +113,26 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
...
+              <label className="Label" htmlFor="device-frame" style={{ paddingRight: 15 }}>
```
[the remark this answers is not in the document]
kewinzaq1: @jakub-gonet Is there a sense to use this component? This component currently is a p tag what does not play well with accessibility, for instance cannot toggle clicking the label. The two ideas I have is to remove the `Label` and just use the className since this component does not do anything, zero custom logic, or to fix the label component tag, change it to label and allow passing label props to it. What do you think make more sense?

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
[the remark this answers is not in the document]
jakub-gonet: Yes, exactly that.

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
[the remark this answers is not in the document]
jakub-gonet: `<Label />` is shared which makes it more difficult. We can try to change it to use `label` and pass optional `for` prop. If that breaks some of the uses then I think we should continue to use `<Label />` as is for consistency, even if broken.

We can always refactor later instead of overcomplicating this PR.

packages/vscode-extension/src/webview/views/DeviceLocationView.css
```diff
@@ -51,30 +51,3 @@ input.coordinate:disabled {
 .supported-formats li {
   margin-bottom: 5px;
 }
-
```
kewinzaq1, line 54: moved to https://github.com/software-mansion/react-native-ide/pull/422/files#diff-e73cae9b9a44401cdd88d9db6605d882d057713ccd90af8482aa2b5dcef581b7

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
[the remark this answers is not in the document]
kewinzaq1: @jakub-gonet `--navy-light-60` is fine?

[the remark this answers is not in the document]
kewinzaq1: @jakub-gonet `we need to fix weird layout issue that happens after switching` - if you think about the notch area, it happens because of the way how the mask looks, we can disable mask or change it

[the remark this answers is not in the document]
kewinzaq1: @jakub-gonet now it looks a bit better and the unexpected width increase does not happen anymore

https://github.com/user-attachments/assets/2a3cb9d1-84ca-40a0-b6d1-97f986b05e5d

jakub-gonet:
Looks good, I'm thinking about adding some border around the preview. Right now, it blends with the background.

Nonblocking though, please rebase and we can get it merged.

jakub-gonet (approved):

jakub-gonet:
@kewinzaq1 will merge in few days; I need to finish what I'm doing right now first. I thought it over and I think we should store it in the vscode settings (`extensionContext.workspaceState`) but I can do that change for you if you want.

jakub-gonet:
I don't think so, just settings menu is sufficient. It's more of a question if we should have it in settings or in JSON and I concluded that JSON isn't that discoverable.

We have already some options that we store there and aren't accessible to user.

kewinzaq1:
@jakub-gonet Thanks for the answer, I asked a bit too early, it's my first experience implementing such thing and I asked only about the JSON pointing out `ReactNativeIDE.showDeviceFrame`, but as I saw later it's connected with the native vscode ui settings of extension, and maybe we should keep it because panel location setting is accessible here and would be nice to keep consistence? What do you think?
native-vscode-setting
custom-ide-settings

packages/vscode-extension/src/webview/components/SettingsDropdown.tsx
```diff
@@ -58,7 +60,22 @@ function SettingsDropdown({ project, isDeviceRunning, children, disabled }: Sett
               <KeybindingInfo commandName="RNIDE.openDevMenu" />
             </div>
           </DropdownMenu.Item>
-
+          <div className="dropdown-menu-item">
+            <span className="codicon codicon-device-mobile" />
+            <label className="switch-group-center" htmlFor="device-frame" style={{ width: "100%" }}>
+              <span>Show Device Frame</span>
+              <Switch.Root
+                style={{ transform: "scale(0.75)", marginLeft: "auto" }}
```
kewinzaq1, line 68: @jakub-gonet The default size is too big for settings dropdown, and I added inline transform to reduce its size. Maybe I should create `switch-root--small` or just keep it?

packages/vscode-extension/src/project/project.ts
```diff
@@ -265,9 +265,15 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
-          if (this.projectState.status === "runtimeError") return;
+          if (this.projectState.status === "starting") {
+            return;
+          }
+          if (this.projectState.status === "incrementalBundleError") {
+            return;
+          }
+          if (this.projectState.status === "runtimeError") {
+            return;
+          }
```
jakub-gonet, line 276: Please revert changes in project.ts, it's not strictly needed in this PR and will conflict with my other changes.

packages/vscode-extension/src/webview/components/Preview.css
```diff
@@ -48,7 +48,6 @@
 
 .phone-screen {
   position: absolute;
-  width: calc(100% - 14px);
```
jakub-gonet, line 51: AFAIK, this was needed for touch area margins. Does it break something when frame isn't visible?

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -327,11 +330,30 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
     device: device!,
   });
 
+  const workspace = useWorkspaceConfig();
+
+  const isFrameDisabled = workspace.showDeviceFrame === false;
+
+  function MaybeFrame() {
```
jakub-gonet, line 337: Please just name it `DeviceFrame` and move `workspace` & `isFrameDisabled` inside.

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -327,11 +330,30 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
...
+    if (!device) {
+      return null;
+    }
+    return (
+      <img
+        src={device.frameImage}
+        className="phone-frame"
+        style={{
+          opacity: isFrameDisabled ? 0 : 1,
```
jakub-gonet, line 346: `opacity` because it still needs to be included in layout calculation?

jakub-gonet (commented):
Actually settings may be nicer as you suggested. People don't necessarily change frame visibility that often.

test-apps/react-native-75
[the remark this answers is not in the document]
jakub-gonet: Nevermind, this shouldn't be in repository in the first place.

packages/vscode-extension/src/webview/components/Preview.tsx
kewinzaq1: @jakub-gonet `cssPropertiesForDevice` takes `isFrameDisabled`  because styles depend on it. So I cannot move these to `DeviceFrame` component

packages/vscode-extension/src/webview/components/Preview.css
kewinzaq1: @jakub-gonet Right it's needed will be restored

packages/vscode-extension/src/webview/components/shared/SwitchGroup.css
```diff
@@ -0,0 +1,32 @@
+.switch-group-center {
+  display: flex;
+  align-items: center;
+  gap: var(--spacing-sm);
+}
+
+.switch-root {
```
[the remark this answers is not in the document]
kewinzaq1: @jakub-gonet I did and looks fine. 
light
dark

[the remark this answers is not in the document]
kewinzaq1: @jakub-gonet I did and looks fine. 
light
dark

packages/vscode-extension/src/webview/components/Preview.tsx
kewinzaq1: @jakub-gonet yes

packages/vscode-extension/src/project/project.ts
kewinzaq1: @jakub-gonet changes are only caused by formatting, previously someone did not format it and the rule `Expected { after 'if' condition.eslint[curly](https://eslint.org/docs/rules/curly)` was not satisfied

jakub-gonet:
Will do some changes in the evening and try to merge it. Thanks for being patient.

jakub-gonet (approved):

jakub-gonet:
💚 💙 💜 💛 ❤️

Followup: #490

PR #423, opened 2024-07-05
Add device setting to control privacy permissions

kmagiera:
This PR adds a device setting that allows for resetting privacy permissions.

The new setting helps with scenarios where you want to test flows that rely on asking users for different types of permissions.

The current implementations supports resetting the following permissions on iOS:
 - all
 - location
 - photos
 - contatcs
 - calendar

On Android due to the limitations of the method we currently use, only resetting all permissions is supported.

On top of the above, the app may behave differently depending on the system. On Android, the app process is always terminated when we reset permission if at least one permission was granted prior to reset. On iOS the app never restarts.

An important consideration here is that we may want to restart the app regardless of the operating system, as the behavior where the app keeps running may lead to some unpredictable behavior. This may also end up getting the app in a state that is not expected by the developers. Typically for end-users on Android and iOS it is not possible to reset the privacy settings  – you can change the setting to disallow or grant the permission later on, but you can never get it back to the state when it's unknown unless you uninstall the app. So this new option we are adding here may get the app into a state that will never be possible to replicate in real-life and therefore there is no need to handle such state for the developer.

The biggest downside of the current implementation is that we use the restart flow for the case when the permission is reset on Android. This leads to a longer than necessary time to interact with the app, as the current restart flow will also reboot the emulator, potentially trigger build and also restart metro. I plan to address this in a follow up PR, where we will add a new quicker flow for handling restarts that doesn't reset the emulator and metro.

Here's how the new dropdown menu look for iOS:
image

Test plan:
We're adding new location screen to expo-router example and extending that example to ask for location permission.
This has been tested on this new example with two scenarios: 1) get location info, allow, reset, get again 2) get location info, disallow, reset permission, get again. In both scenarios it is expected to get the permission popup. On Android the app will always restart when resetting the permission.


packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -395,3 +412,18 @@ function convertToSimctlSize(size: DeviceSettings["contentSize"]): string {
       return "extra-extra-extra-large";
   }
 }
+
+function convertToSimctlPermission(permissionType: AppPermissionType): string {
```
jakub-gonet, line 416: Do we expect more to come? Right this function is doing nothing as far as I can see.

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -32,10 +32,25 @@ interface DeviceSettingsDropdownProps {
...
+  { label: "Reset All Permissions", value: "all", icon: "check-all" },
+];
+
 function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownProps) {
-  const { project, deviceSettings } = useProject();
+  const { project, deviceSettings, projectState } = useProject();
   const { openModal } = useModal();
 
+  const resetOptions =
+    projectState.selectedDevice?.platform === "iOS" ? resetOptionsIOS : resetOptionsAndroid;
```
jakub-gonet, line 52: If we expect device to be selected here, you can use `projectState.selectedDevice!.platform` to remove possibly undefined result.

test-apps/expo-router/app/location.js
```diff
@@ -0,0 +1,33 @@
+import { useState } from 'react';
+import { View, Text, Button } from 'react-native';
+import * as Location from 'expo-location';
+
+export default function Another() {
+  const [location, setLocation] = useState(null);
+
+  async function readLocation() {
+    let { status } = await Location.requestForegroundPermissionsAsync();
+    console.log('Permission status', status);
```
jakub-gonet, line 10: Remove.

test-apps/expo-router/app/location.js
```diff
@@ -0,0 +1,33 @@
...
+    if (status !== 'granted') {
+      setLocation('Permission to access location was denied');
+      return;
+    }
+
+    let location = await Location.getCurrentPositionAsync({});
+    console.log('Location', location);
```
jakub-gonet, line 17: Remove.

test-apps/expo-router/app/location.js
```diff
@@ -0,0 +1,33 @@
...
+
+  async function readLocation() {
+    let { status } = await Location.requestForegroundPermissionsAsync();
+    console.log('Permission status', status);
+    if (status !== 'granted') {
+      setLocation('Permission to access location was denied');
+      return;
+    }
+
+    let location = await Location.getCurrentPositionAsync({});
```
jakub-gonet, line 16: `const`.

test-apps/expo-router/app/location.js
```diff
@@ -0,0 +1,33 @@
...
+    console.log('Location', location);
+    setLocation(
+      `${location.coords.latitude.toFixed(
```
jakub-gonet, line 19: Nit: Maybe destructuring would save some space here.

test-apps/expo-router/app/location.js
```diff
@@ -0,0 +1,33 @@
+import { useState } from 'react';
+import { View, Text, Button } from 'react-native';
+import * as Location from 'expo-location';
+
+export default function Another() {
```
jakub-gonet, line 5: Nit: Maybe LocationReader or something?

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -250,6 +250,23 @@ export class IosSimulatorDevice extends DeviceBase {
     ]);
   }
 
+  async resetAppPermissions(appPermission: AppPermissionType, build: BuildResult) {
+    if (build.platform !== Platform.IOS) {
+      throw new Error("Invalid platform");
+    }
```
jakub-gonet, line 256: Why we need this check?

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
kmagiera: we don't. the menu is active regardless of whether device is selected. This can only happen for a very short time and in that case I don't think anyone will expect the reset options to do anything. I don't feel like adding extra code here to handle such case

test-apps/expo-router/app/location.js
kmagiera: this is in the test app, I'd prefer to keep it there especially that we want to test logging too

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
kmagiera: to narrow down the type of build from BuildResult to IosBuildResult which has some fields that we need

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
kmagiera: I updated it such that I have a separate list of accepted service names. The idea was that we levarage typescript to catch the case when we add new item in the future and it isn't mapping directly to the name that we can use with simctl

jakub-gonet (approved):

PR #424, opened 2024-07-05
Avoid rebooting and rebuilding for certain type of restarts

kmagiera:
This PR changes the way we handle restarts / reloads.

Before, when fast-refresh wasn't possible, we'd use `selectDevice` flow, which always kills the simulator process, restarts it, starts new preview etc. This tend to take long and definitely much longer than needed when we only want the app process to be restarted.

An example when this scenario occurs frequently is when updating privacy permissions. In that case, the app process is killed by the system, and we need to restart it. Rebooting emulator in such a case is unnecessary.

The new restart flow now looks as follows:
 - unless force rebuild is requested (in which case we restart everything and also rebuild the app) we follow the procedure below:
 -  we check if fast-refresh can be performed – for that we test if devtools connection is active which means that the app process is connected to devtools server
 - if not, we now try restarting the app process, we reuse the existing device and build if possible. If any of that doesn't work or is stale, this will trigger an exception and we will move to the next phase
 - finally, if the above fails, we go back to the select-device flow, which resets the simulator and build

Test plan:
this has been tested on top of permissions PR where I check that the app process restart is performed after resetting permissions.


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -54,26 +49,40 @@ export class DeviceSession implements Disposable {
...
     await this.device.changeSettings(deviceSettings);
     progressCallback(StartupMessage.Building);
     this.buildResult = await this.disposableBuild.build;
     progressCallback(StartupMessage.Installing);
     await this.device.installApp(this.buildResult, false);
-    progressCallback(StartupMessage.Launching);
-    await this.device.launchApp(this.buildResult, this.metro.port, this.devtools.port);
 
-    const waitForPreview = this.device.startPreview().then(() => {
+    this.device.startPreview().then(() => {
```
jakub-gonet, line 81: NIt: `await` here.

jakub-gonet (approved):
Looks good but I must admit that launching/restarting/soft-restarting logic is getting gnarly.

packages/vscode-extension/src/project/deviceSession.ts
kmagiera: awaiting here would block calling this.launch thats below. So this should stay as is

PR #425, opened 2024-07-08
Mitigate the negative effects of double x button in secondary, sidebar. 

filip131311:
### Description 
Because, secondary sidebar in vscode has a "close" button in top right corner, the user is faced with two "X" buttons that do different things; one closes the secondary sidebar and in effect hides our extension and the second closes our extension completely. 

### Solution 
To mitigate  the negative effects this PR overrides the command used to hide secondary sidebar and closes React native IDE completely before hiding it, that makes two buttons perform similar actions instead of doing two different things, which is an improvement.  

### Hiding the second button
The natural question to ask is why can't we disable, our "close RNIDE" button when IDE is in secondary sidebar. Unfortunately there is now reliable way to check in which sideBar the viewContainer contributed by an extension is located at any given time. There is a closed issue on that topic here. One workaround for that would be to check if context key "activeAuxiliary" has value matching our extension contributed viewContainer, because since it is an activeAuxiliary it has to be in Auxiliary SideBar, unfortunately the id of Extension contributed view Containers is randomly generated, you can read the code [here](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/services/views/browser/viewDescriptorService.ts) the first relevant line is 502. 

There is hope of discovering the generated viewContainer id by checking the activeAxiliary and activeViewlet context fields while we know about its location, but it would require a getContext() API that is not currently implemented, an active issue on that can be found [here](https://github.com/microsoft/vscode/issues/10471).


packages/vscode-extension/src/extension.ts
```diff
@@ -136,6 +137,35 @@ export async function activate(context: ExtensionContext) {
...
+    registeredCommandDisposable.dispose(); // must dispose to avoid endless loops
+
+    const wasIDEPanelVisible = SidePanelViewProvider.currentProvider?.view?.visible;
+
+    // run the built-in closeAuxiliaryBar command
+    await commands.executeCommand("workbench.action.closeAuxiliaryBar");
+
+    const isIDEPanelVisible = SidePanelViewProvider.currentProvider?.view?.visible;
+
+    // if closing of Auxiliary bar affected the visibility of SidePanelView, we assume that it means that it was pined to the secondary sidebar.
```
jakub-gonet, line 150: Typo: pinned.

packages/vscode-extension/src/extension.ts
```diff
@@ -136,6 +137,35 @@ export async function activate(context: ExtensionContext) {
     commands.registerCommand("RNIDE.diagnose", diagnoseWorkspaceStructure)
   );
 
+  async function closeAuxiliaryBar(registeredCommandDisposable: Disposable) {
+    registeredCommandDisposable.dispose(); // must dispose to avoid endless loops
```
jakub-gonet, line 141: Nit: may be more readable with helper like `interceptCommand(name: string, fn: () -> Disposable): Disposable`.

jakub-gonet (approved):

PR #433, opened 2024-07-10
Support React Native 0.75

kmagiera:
This PR adds test app for react-native 0.75 (Rc 4 for now as that's the most recent version).

Supporting 0.75 required one change in wrapper.js where we'd import LoadingView component from React Native internals which got moved to DevLoadingView. This PR updates the necessary code to load the right file depending on its availability.

Test plan:
run RN 75 example on Android and iOS
run also older examples to see that the wrapper change didn't affect them


jakub-gonet (commented):
Inspect breaks when used. [This branch](https://github.com/software-mansion/react-native-ide/compare/kmagiera/rn75...jgonet/kmagiera/rn75) has a fix – seems like       `getInspectorDataForViewAtPoint` `require` is loading a wrong module (NativeDevMenu) but I don't know why. Module IDs match.

Those two modules have a weird coupling – DevMenu doesn't work at all and when commenting out its `require`, `getInspectorDataForViewAtPoint` gets different error (expected function, got undefined) instead of current "expected function, got object" one.

packages/vscode-extension/lib/wrapper.js
```diff
@@ -17,9 +17,33 @@ export function registerNavigationPlugin(name, plugin) {
 
 let navigationHistory = new Map();
 
+const NativeModules = {
```
kmagiera, line 20: Those are not native modules. Rename to sth like "RNInternals"

kmagiera:
Other than that this looks ok. Can't accept this apparently as I'm the original author of this PR

jakub-gonet (approved):

PR #435, opened 2024-07-10
Create a new Utility Class callable from webview

filip131311:
### Description 

This PR addresses the problem with `ProjectInterface` holding common utilities that are not connected with project class on domain level. To combat that we create a new `UtilityInterface` that can be called from webview and execute code in on extension side. 

#### Note: 
This also introduces a minor cleans up the `getCommandsCurrentKeyBinding` method.


packages/vscode-extension/src/common/utils.ts
```diff
@@ -1,29 +1,8 @@
-export function throttle<T extends (...args: any[]) => any>(
-  func: T,
-  limitMs: number
-): (...args: [...Parameters<T>, force?: boolean]) => ReturnType<T> {
-  let timeout: NodeJS.Timeout | null = null;
-  let recentArgs: any;
+export interface UtilsInterface {
```
jakub-gonet, line 1: Avoid adding `Interface` suffix, it's pure visual noise.

packages/vscode-extension/src/webview/providers/UtilsProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { UtilsInterface } from "../../common/Utils";
+
+const utils = makeProxy<Utils>("Utils");
+
+type UtilsContextType = {
+  utils: UtilsInterface;
+};
+
+const UtilsContext = createContext<UtilsContextType>({
+  utils,
```
jakub-gonet, line 13: It's better to flatten this IMO and put utils directly.

jakub-gonet (approved):

Issue #439, opened 2024-07-11
IDE panel gets into weird state when last used device got removed


kmagiera:
**Description**

Here are steps to reproduce:
1. Open project A and use device A
2. Open project B, switch to a different device and remove device A
3. Open again project A
4. It will now try to use device A which has been removed in the meantime and result in displaying a device loading state

This issue isn't critical as you can simply use device picker to select a new device, but I think the correct behavior should be that we select some other, available device instead of trying to boot the one that does no longer exist.

**Environment**
- Version of currently used React Native IDE: 0.16
- Version of React Native/Expo used in the project: any

PR #441, opened 2024-07-11
Fix opening a different workspace after device was removed. 

filip131311:
### Description
In this [commit](https://github.com/software-mansion/react-native-ide/commit/24703d99ed2657fa0996ac987633edec08e7b45c) we introduced a cacheing strategy for listing all devices, because the cache was workspace specific it caused problems when switching workspaces. This PR fixes those issues by opting for globalState, as a caching mechanism. 

#### How to test:
- Open project A and use device A
- Open project B, switch to a different device and remove device A
- Open again project A
- It will now select the first available device.

fixes: #439

### DIFF
--- a/packages/vscode-extension/src/devices/DeviceManager.ts
+++ b/packages/vscode-extension/src/devices/DeviceManager.ts
@@ -85,12 +85,12 @@ export class DeviceManager implements Disposable, DeviceManagerInterface {
   private async loadDevices(forceReload = false) {
     if (forceReload) {
       // Clear the cache when force reload is requested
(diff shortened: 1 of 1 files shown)


jakub-gonet (approved):
LGTM, I'd suggest some more descriptive name for commit like "Share devices cache to avoid stale entries" with appropriate commit description.

jakub-gonet:
Also I'm not really sure it fixes https://github.com/software-mansion/react-native-ide/issues/439 as it is concerned about removing last device from device list.

PR #442, opened 2024-07-11
Fix - display url params 

### LINKED ISSUE #443: URL Parameters don't display in the search
Currently, URL parameters don't display in the search panel
- Version of currently used React Native IDE 0.16.0

https://github.com/software-mansion/react-native-ide/assets/77052270/9f2b4b59-0b6d-4444-be48-ce927234eb49

kewinzaq1:
### Problem
Currently, URL parameters don't display in the search panel
Fixes #443 
https://github.com/software-mansion/react-native-ide/assets/77052270/fc3f2489-5e7d-46cd-865b-473aa54a0eaa

### How Has This Been Tested: 
Go to a page with parameters, and then you can see this problem, after this fix this issue no longer happens
https://github.com/software-mansion/react-native-ide/assets/77052270/f1c20113-86a1-4ea3-816c-9fc5fb7e4d4e


packages/vscode-extension/lib/wrapper.js
```diff
@@ -82,9 +82,12 @@ export function PreviewAppWrapper({ children, initialProps, ..._rest }) {
 
   const handleNavigationChange = useCallback(
     (navigationDescriptor) => {
+      const queryParams = new URLSearchParams(navigationDescriptor.params).toString();
       navigationHistory.set(navigationDescriptor.id, navigationDescriptor);
       devtoolsAgent?._bridge.send("RNIDE_navigationChanged", {
-        displayName: navigationDescriptor.name,
+        displayName: `${navigationDescriptor.name}${
```
kmagiera, line 88: I'd much rather prefer for the name to already contain params for routes with parameters. We support expo router here only for the time being but the API is designed to be routing agnostic and for other solutions params may be formatted diferently.

The semantics of the "name" prop here is the "display name", so it is ok for the expo router plugin to format the params when appropriate

kmagiera (changes requested):
Thanks for this. I pointed out in an inline comment that we'd want the expo router plugin to append params when passing "name" prop for the route. Would be happy to merge this when you made that change

packages/vscode-extension/lib/wrapper.js
kewinzaq1: @kmagiera got it sounds reasonable

packages/vscode-extension/lib/expo_router_plugin.js
```diff
@@ -17,9 +17,12 @@ function useRouterPluginMainHook({ onNavigationChange }) {
   const pathname = routeInfo?.pathname;
   const params = routeInfo?.params;
 
+  const displayParams = new URLSearchParams(params).toString();
```
kewinzaq1, line 20: info: copy pasted since as I know the v2 plugin will be removed and similar approach was recommended in previous implementation

kmagiera (approved):
This looks good, thank you!

PR #445, opened 2024-07-15
Extract just preview address from sim-server logs

filip131311:
This PR adreses the problem that introduced by a change in sim-server that logs: 
`Stream live on ${adres}` instead of `${adres}` when the preview is ready. 

Additionally a new approach allows us to make sure that the url is the required  stream and frees up the sim-server to log other links in the future.

### DIFF
--- a/packages/vscode-extension/src/devices/preview.ts
+++ b/packages/vscode-extension/src/devices/preview.ts
@@ -33,10 +33,15 @@ export class Preview implements Disposable {
         reject(new Error("Preview server exited without URL"));
       });
 
+      const streamURLRegex = /(http:\/\/[^ ]*stream\.mjpeg)/;
+
       lineReader(subprocess).onLineRead((line) => {
-        if (line.includes("http://")) {
-          Logger.debug(`Preview server ready ${line}`);
-          this.streamURL = line;
+        const match = line.match(streamURLRegex);
+
+        if (match) {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/devices/preview.ts
```diff
@@ -33,10 +33,15 @@ export class Preview implements Disposable {
         reject(new Error("Preview server exited without URL"));
       });
 
+      const streamURLRegex = /(http:\/\/[^ ]*stream\.mjpeg)/;
+
       lineReader(subprocess).onLineRead((line) => {
-        if (line.includes("http://")) {
-          Logger.debug(`Preview server ready ${line}`);
-          this.streamURL = line;
+        const match = line.match(streamURLRegex);
```
jakub-gonet, line 39: Suggestion: you can do: `const [_full, url] = line.match(streamURLRegex);`

jakub-gonet (approved):

packages/vscode-extension/src/devices/preview.ts
jakub-gonet: Also, you can simplify the regex to `/http:\/\/.+stream\.mjpeg/;` and just use full match.

filip131311: I can't do that because match maybe null, I could do deconstruction inside the if(match) scope, but it seems unnecessary.

Issue #397, opened 2024-06-19
Metro Unable to resolve "scheduler"


renanmav:
**Description**

After updating to v0.0.15, I'm getting this error:

```
2024-06-19 19:31:11.714 [info] Metro Waiting on http://localhost:8081
2024-06-19 19:31:11.714 [info] Metro Logs for your project will appear below.
2024-06-19 19:31:17.736 [info] Launch preview /Users/renanmav/.vscode/extensions/swmansion.react-native-ide-0.0.15-darwin-arm64/dist/sim-server-executable ios,85BCDCE8-A1A8-4900-B49B-B9E3D8AB8073,/Users/renanmav/Library/Caches/com.swmansion.react-native-ide/Devices/iOS
2024-06-19 19:31:17.736 [info] Will wait for app ready and for preview
2024-06-19 19:31:19.413 [info] Preview server: Server running...
2024-06-19 19:31:19.414 [info] Preview server ready http://localhost:60708/stream.mjpeg
2024-06-19 19:31:19.414 [info] Preview server: http://localhost:60708/stream.mjpeg
2024-06-19 19:31:25.241 [info] Metro iOS Bundling failed 4121ms node_modules/expo-router/entry.js (1620 modules)
2024-06-19 19:31:25.242 [info] Metro Unable to resolve "scheduler" from "../../.vscode/extensions/swmansion.react-native-ide-0.0.15-darwin-arm64/lib/rn74/ReactNativeRenderer-dev.js"
```

**Environment**
- Version of currently used React Native IDE (Extensions tab -> click on React Native IDE extension -> version is located next to the extension name)
- Version of React Native/Expo used in the project.
- Any other versions of tools related to the problem.

jakub-gonet:
What React Native version does your project use?

Issue #396, opened 2024-06-19
metro error: Error: Unable to resolve module __RNIDE_lib__/rn74/ReactFabric-dev.js (pnpm workspace)


imWildCat:
**Description*
Metro cannot be started for unknown reason

**Environment**
- Version of currently used React Native IDE: 0.0.15
- Version of React Native/Expo used in the project: 0.74.2
- Any other versions of tools related to the problem.

Project is like:

```
apps/browser-extension/package.json
apps/content-renderer/package.json
apps/desktop/package.json
apps/mobile/package.json
package.json // project root with pnpm-workspace
```

```.npmrc
node-linker=hoisted 
```

renanmav:
Same as #397

PR #446, opened 2024-07-16
Add a mechanism for detecting if node_modules are installed and installing them

filip131311:
### Description: 
This PR adds a way for our extension to detect if node_modules are installed and inserts the check into start up flow. This is the first step towards, React Native IDE detecting and installing every dependency needed for a project with minimal user input. 

##### Limitations: 
as per #397 and #396 react-native-IDE currently does not work with pnpm projects, while we are actively working on solving those issues, this PR supports only setups using `yarn` or `npm`.


packages/vscode-extension/src/dependency/DependencyInstaller.ts
```diff
@@ -1,77 +1,81 @@
-import { Webview, Disposable } from "vscode";
-import { Logger } from "../Logger";
-import { DependencyChecker } from "./DependencyChecker";
-import { command } from "../utilities/subprocess";
-import { getIosSourceDir } from "../builders/buildIOS";
-import { getAppRootFolder } from "../utilities/extensionContext";
-import { getLaunchConfiguration } from "../utilities/launchConfiguration";
+// This code is deprecated in favour of DependencyManager.ts
```
jakub-gonet, line 1: I'd just remove it, you can checkout single files while being on other branch.

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -85,6 +96,67 @@ export class DependencyChecker implements Disposable {
...
+        installationCommand = "bun install";
+        break;
+    }
+
+    const subprocess = command(installationCommand, {
+      cwd: workspacePath,
+      silentErrorsOnExit: true,
+    });
+
+    await subprocess;
```
jakub-gonet, line 148: Nit: `await command(...)`.

packages/vscode-extension/src/panels/WebviewController.ts
```diff
@@ -51,22 +49,18 @@ export class WebviewController implements Disposable {
     this.setWebviewMessageListener(webview);
 
     // Set the manager to listen and change the persisting storage for the extension.
-    this.dependencyChecker = new DependencyChecker(webview);
-    this.dependencyChecker.setWebviewMessageListener();
-
-    this.dependencyInstaller = new DependencyInstaller(webview);
-    this.dependencyInstaller.setWebviewMessageListener();
+    this.dependencyManager = new DependencyManager(webview);
+    this.dependencyManager.setWebviewMessageListener();
```
jakub-gonet, line 53: Why we don't do it in the c-tor?

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -32,6 +33,17 @@ export async function resolvePackageManager(): Promise<PackageManagerName> {
     } else if (isNpm) {
       return "npm";
     }
+    try {
+      const packageManager = require(path.join(workspacePath, "package.json")).packageManager;
```
jakub-gonet, line 37: How often is this specified? Why we're doing regex matching on it? Don't we want to use lock-files for detection?

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -44,3 +56,68 @@ export function isPackageManagerAvailable(manager: PackageManagerName): boolean
...
+  } catch (e) {
+    return false;
+  }
+}
+
+async function isYarnInstalled(): Promise<boolean> {
+  const workspacePath = getAppRootFolder();
+  try {
+    // because "yarn check" was removed from yarnv2 we use npm's method for checking dependencies
+    // npm is takeing into consideration yarn.lock since version 7, which means
```
jakub-gonet, line 78: Typo: "taking".

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -44,3 +56,68 @@ export function isPackageManagerAvailable(manager: PackageManagerName): boolean
...
+  // TODO: add pnpm support
+  return false;
+}
+
+async function isBunInstalled(): Promise<boolean> {
+  // TODO: add bun support
+  return false;
+}
+
+export async function isNodeModulesInstalled(manager: PackageManagerName): Promise<boolean> {
```
jakub-gonet, line 112: This function names are confusing, I'd expect it checks for installation status of packages in node_modules but it uses `isXInstalled` inside which suggests only checking if some packager is available.

packages/vscode-extension/src/utilities/subprocess.ts
```diff
@@ -84,7 +84,10 @@ export function execSync(name: string, args?: string[], options?: execa.SyncOpti
   return result;
 }
 
-export function command(commandWithArgs: string, options?: execa.Options) {
+export function command(
+  commandWithArgs: string,
+  options?: execa.Options & { silentErrorsOnExit?: boolean }
```
jakub-gonet, line 89: This option name was confusing – "silent" is used as a verb and we don't ignore exit messages, we ignore all messages. Maybe typical `quiet` option would be more fitting?

packages/vscode-extension/src/project/metro.ts
```diff
@@ -126,11 +130,12 @@ export class Metro implements Disposable {
 
   public async startInternal(
     resetCache: boolean,
-    progressListener: (newStageProgress: number) => void
+    progressListener: (newStageProgress: number) => void,
+    dependencies: Promise<any>[]
```
jakub-gonet, line 134: `any` + why we're adding additional parameter for that here? `dependencies` is typically used for react hooks + we could just await up in the calling code, we don't use and need this argument at all.

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -59,6 +60,7 @@ export class DeviceSession implements Disposable {
 
     progressCallback(StartupMessage.AttachingDebugger);
     await this.startDebugger();
+    isNodeModulesInstalled("npm");
```
jakub-gonet, line 63: Why we specify npm here? Also, this does nothing except slowing down run procedure.

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -85,6 +96,67 @@ export class DependencyChecker implements Disposable {
     return installed;
   }
 
+  public async checkNodeModulesInstalled() {
+    const packageManager = await resolvePackageManager();
+
+    // if(!isPackageManagerAvailable(packageManager)){
+    //   // TODO some error handling
+    // }
```
jakub-gonet, line 104: Remove or implement.

packages/vscode-extension/src/utilities/packageManager.ts
filip131311: We do use lock files to detect it but in rare case that they do not exist, we try checking the package.json instead of immediately fallbacking to npm. 
We perform regex to get rid of package managers version in the name e.g. `yarn@1.2.2` 

It is specified in most projects started with RN CLI

packages/vscode-extension/src/panels/WebviewController.ts
filip131311: do you think somethnig like `isNpmModulesInstalled` makes more sense?

packages/vscode-extension/src/utilities/packageManager.ts
filip131311: isNpmModulesInstalled

filip131311: do you think something like that makes more senes?

packages/vscode-extension/src/project/deviceSession.ts
filip131311: remanence of debbuging process

packages/vscode-extension/src/project/metro.ts
filip131311: this is constructed like that to avoid awaiting inside Project.start() to quickly move on to selecting device.

packages/vscode-extension/src/utilities/packageManager.ts
jakub-gonet: It's definitely better.

packages/vscode-extension/src/project/project.ts
```diff
@@ -497,6 +506,18 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
+  private async installNodeModules(): Promise<void> {
+    const nodeModulesStatus = await this.dependencyManager.checkNodeModulesInstalled();
+
+    if (nodeModulesStatus.installed) {
+      Logger.debug("Node Modules installed");
+      return;
+    }
+
+    await this.dependencyManager.installNodeModules(nodeModulesStatus.packageManager);
+    Logger.debug("Node Modules installed");
```
jakub-gonet, line 518: Nit: 
```
if (!nodeModulesStatus.installed) {
   await this.dependencyManager.installNodeModules(nodeModulesStatus.packageManager);
}
Logger.debug("Node Modules installed");
```

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -32,15 +31,91 @@ export async function resolvePackageManager(): Promise<PackageManagerName> {
...
+
+    // because npm marks packages installed with yarn as "extraneous" we need to check if there are any other problems.
+    if (parsedJson.problems) {
+      parsedJson.problems.forEach((element: string) => {
+        if (!element.startsWith("extraneous")) {
+          return false;
+        }
+      });
+    }
+    return true;
```
jakub-gonet, line 94: If we do inverse, we can exit early. Also, `return` in `forEach` doesn't work as intended.


```ts
for(const problem of parsedJson.problems ?? []) {
  if(element.startsWith("extraneous")) { return true; }
}
```

jakub-gonet (commented):
Except for one bug in `installNodeModules` function , it's ready to go.

PR #462, opened 2024-07-22
Integration with Windows

p-malecki:
This PR adds integration with Windows OS and implements some minor fixes:
- hides the option to add an iOS device from Windows user 
- when removing an Android device, the emulator process is first killed, followed by removal from the file system and a reload of the devices list
- adds API level 35 (android 15) to sdkmanager
- fixes an issue with the invisible device type placeholder in CreateDeviceView

There still is an issue with removing AVD files when removing device on Windows, but remains unsolved for now.

resolves: #447


packages/vscode-extension/scripts/build-sim-server.sh
```diff
@@ -14,7 +14,11 @@ submodule_status=$(git submodule status ../simulator-server)
 if [[ $submodule_status == -* ]]; then # submodule is not initialized
 
 submodule_hash=$(git ls-tree HEAD ../simulator-server | awk '{print $3}')
-product_path="$output_dir/sim-server-Release-${submodule_hash}"
+if [[ "$OSTYPE" == "msys" ]]; then
```
filip131311, line 17: can `"$OSTYPE" == "win32"` ever happen?

packages/vscode-extension/scripts/build-sim-server.sh
```diff
@@ -58,6 +62,11 @@ fi # submodule check
 mkdir -p "$output_dir"
 target_location="$output_dir/sim-server"
 
+if [[ "$OSTYPE" == "msys" ]]; then # rename to *.exe on Windows
```
filip131311, line 65: can `"$OSTYPE" == "win32"` ever happen?

packages/vscode-extension/src/dependency/DependencyChecker.ts
```diff
@@ -57,10 +49,23 @@ export class DependencyChecker implements Disposable {
             Logger.debug("Received checkExpoInstalled command.");
             this.checkExpoInstalled();
             return;
-          case "checkPodsInstalled":
-            Logger.debug("Received checkPodsInstalled command.");
-            this.checkPodsInstalled();
-            return;
+        }
+        if (process.platform !== "win32") {
```
filip131311, line 53: `DependencyChecker.ts` does not exist anymore please rebase

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -15,8 +16,12 @@ import { AppPermissionType, DeviceSettings } from "../common/Project";
 import { getAndroidSystemImages } from "../utilities/sdkmanager";
 import { EXPO_GO_PACKAGE_NAME, fetchExpoLaunchDeeplink } from "../builders/expoGo";
 
-export const EMULATOR_BINARY = path.join(ANDROID_HOME, "emulator", "emulator");
-const ADB_PATH = path.join(ANDROID_HOME, "platform-tools", "adb");
+ export const EMULATOR_BINARY = (process.platform === "win32")
+  ? path.join(ANDROID_HOME, "emulator", "emulator.exe")
+  : path.join(ANDROID_HOME, "emulator", "emulator");
+const ADB_PATH = (process.platform === "win32")
```
filip131311, line 22: can `msys` ever happen?

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -32,11 +32,15 @@ export class Preview implements Disposable {
         // the promise to prevent the caller from waiting indefinitely
         reject(new Error("Preview server exited without URL"));
       });
+      
+      const streamURLRegex = /(http:\/\/[^ ]*stream\.mjpeg)/;
```
filip131311, line 36: Why is it a change here? I believe it is already on main?

packages/vscode-extension/src/project/project.ts
```diff
@@ -584,7 +599,7 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
 
   // used in callbacks, needs to be an arrow function
   private removeDeviceListener = async (_devices: DeviceInfo) => {
```
filip131311, line 601: I know it was like that but I think `_devices` should be called `device` or even better `removedDevice` since its just one.

packages/vscode-extension/src/utilities/android.ts
```diff
@@ -2,21 +2,33 @@ import path from "path";
 import os from "os";
 import fs from "fs";
 
-export const ANDROID_HOME =
-  process.env.ANDROID_HOME || path.join(os.homedir(), "Library/Android/sdk");
 
-const ANDROID_STUDIO_PATH = "/Applications/Android Studio.app";
+export const ANDROID_HOME = process.env.ANDROID_HOME || (process.platform === "win32")
```
filip131311, line 6: please use `??` instead of `||`

packages/vscode-extension/src/utilities/subprocess.ts
```diff
@@ -13,10 +13,14 @@ function overridePWD<T extends execa.Options>(options?: T) {
...
   // therefore removing the risk of the subprocess using the wrong working directory.
+  if (process.platform === 'win32') {
+    return options
+  }
+  
   if (options?.cwd) {
     return { ...options, env: { ...options.env, PWD: options.cwd } };
   }
-  return undefined;
+  return options;
```
filip131311, line 23: again this change is already on main

packages/vscode-extension/src/utilities/android.ts
```diff
@@ -2,21 +2,33 @@ import path from "path";
...
+  : "/Applications/Android Studio.app";
 
 function findJavaHome() {
   // we try to use java bundled with Android Studio if exists
   // in newer distributions java is placed under Contents/jbr/Contents/Home and with some
   // older (but still recent ones) it is under Contents/jre/Contents/Home
   // we check if the first path exists and if not we try the second one
-  const jbrPath = path.join(ANDROID_STUDIO_PATH, "Contents/jbr/Contents/Home");
+
+  const jbrPath = (process.platform === "win32")
```
filip131311, line 20: might be nice to create a similar utill `getPlatform() ` as you did for webview.

filip131311: or maybe isWindows() and isMacOS()

filip131311 (changes requested):
Left some inline coments

packages/vscode-extension/src/webview/utilities/consts.ts
```diff
@@ -75,3 +75,19 @@ export const AndroidSupportedDevices: DeviceProperties[] = [
     maskImage: pixel6amask,
   },
 ] as const;
+
+
+function getPlatform() {
+  // https://stackoverflow.com/a/73619128
+  if (typeof navigator.userAgentData !== 'undefined' && navigator.userAgentData != null) {
```
jakub-gonet, line 82: This function has a lot of redundant logic.

```ts
function getPlatform() {
  return navigator.userAgentData?.platform ?? navigator.platform ?? 'unknown';
}
```

should have the same semantics.

packages/vscode-extension/src/webview/views/CreateDeviceView.tsx
```diff
@@ -37,12 +38,14 @@ function useSupportedDevices() {
...
-          items: iOSSupportedDevices.map((device) => buildSelections(device, Platform.IOS)),
-        },
+    (isOSX)
+      ? ((iosSimulatorError !== undefined)
+        ? { label: "iOS – error, check diagnostics", items: [] }
+        : {
+            label: "iOS",
+            items: iOSSupportedDevices.map((device) => buildSelections(device, Platform.IOS)),
+          })
+      : { label: "", items: [] },
```
jakub-gonet, line 48: Please don't nest ternary operators.

packages/vscode-extension/src/utilities/sdkmanager.ts
```diff
@@ -79,7 +81,9 @@ export async function getAndroidSystemImages(): Promise<AndroidSystemImageInfo[]
 
 // example input: 'android-34/default/arm64-v8a/data'
 function mapToSystemImageInfo(systemImagePath: string) {
-  const [imageName, systemImageType, arch] = systemImagePath.split("/");
+  const [imageName, systemImageType, arch] = (process.platform === "win32") 
+    ? systemImagePath.replace(SYSTEM_IMAGES_PATH + path.sep, "").split(path.sep)
```
jakub-gonet, line 85: How's the input look on windows?

packages/vscode-extension/src/project/project.ts
```diff
@@ -117,33 +117,48 @@ export class Project implements Disposable, MetroDelegate, ProjectInterface {
...
+        }
+      };
+      this.deviceManager.addListener("devicesChanged", listener);
+    }
+  }
 
-    const devices = await this.deviceManager.listAllDevices();
-    if (!selectInitialDevice(devices)) {
+   // Try to select device from devices list different than removedDevice in single attempt.
+  private async trySelectingAnotherDevice(removedDevice: DeviceInfo) {
```
jakub-gonet, line 156: If I'm not mistaken, this is a copy of `trySelectingInitialDevice`. Why we're getting remove device ID from `listAllDevices()` in the first place?

packages/vscode-extension/src/webview/views/DevicesNotFoundView.tsx
```diff
@@ -97,6 +97,10 @@ function DevicesNotFoundView() {
   }
 
   async function createIOSDevice() {
+    if (!isOSX) {
+      return
+    }
```
jakub-gonet, line 102: Just don't call it on unavailable platforms, there's no need to guard against it.

jakub-gonet (changes requested):

packages/vscode-extension/src/project/project.ts
jakub-gonet: The underscore is signifying that the argument is not used.

packages/vscode-extension/src/dependency/DependencyChecker.ts
p-malecki: I've merged main into this branch, so this file doesn't exist now.

packages/vscode-extension/src/devices/preview.ts
p-malecki: As above I've merged main into this branch.

packages/vscode-extension/src/utilities/android.ts
p-malecki: Changed.

packages/vscode-extension/src/utilities/subprocess.ts
p-malecki: Also solved.

packages/vscode-extension/src/utilities/android.ts
p-malecki: I've implemented `Platform` class that takes care of selecting the right paths depending on the platform.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -15,8 +16,12 @@ import { AppPermissionType, DeviceSettings } from "../common/Project";
 import { getAndroidSystemImages } from "../utilities/sdkmanager";
 import { EXPO_GO_PACKAGE_NAME, fetchExpoLaunchDeeplink } from "../builders/expoGo";
 
-export const EMULATOR_BINARY = path.join(ANDROID_HOME, "emulator", "emulator");
-const ADB_PATH = path.join(ANDROID_HOME, "platform-tools", "adb");
+ export const EMULATOR_BINARY = (process.platform === "win32")
+  ? path.join(ANDROID_HOME, "emulator", "emulator.exe")
+  : path.join(ANDROID_HOME, "emulator", "emulator");
```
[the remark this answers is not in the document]
p-malecki: I've made something similar to your proposal. Although it's in two separate versions in src/utilities and src/webview utilities.  I'm not sure if it is correct approach.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -475,6 +466,24 @@ export async function listEmulators() {
   );
 }
 
+export async function ensureOldEmulatorProcessExited(avdId: string) {
+  let runningPid: string | undefined;
+  const command = (process.platform === "darwin") ? "ps" : 'powershell.exe "Get-WmiObject Win32_Process | Select-Object ProcessId, CommandLine | Out-String -Width 10000"';
```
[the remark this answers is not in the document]
p-malecki: Done, and also made a pattern variable for suitable regex depending on os.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -443,12 +432,14 @@ export async function createEmulator(displayName: string, systemImage: AndroidSy
 const UUID_REGEX = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
 async function getAvdIds(avdDirectory: string) {
   const { stdout } = await exec(EMULATOR_BINARY, ["-list-avds"], {
-    env: { ANDROID_AVD_HOME: avdDirectory },
+    env: process.platform === 'darwin'
+  ? { ANDROID_AVD_HOME: avdDirectory }
+  : { ANDROID_SDK_HOME: avdDirectory + "\\.." }
```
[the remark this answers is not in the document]
p-malecki: Changed. On Windows to find AVDs we have to look for folder that contains avdDirectory. Strange but works

packages/vscode-extension/src/webview/utilities/consts.ts
p-malecki: In new `Platform` class your simpler version is used.

packages/vscode-extension/src/utilities/sdkmanager.ts
p-malecki: sample input looks like 'C:\\Users\\USER\\AppData\\Local\\Android\\Sdk\\system-images\\android-35\\google_apis_playstore\\x86_64\\data' so we have to truncate the beginning.

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -150,14 +157,28 @@ export class DeviceManager implements Disposable, DeviceManagerInterface {
...
+
+
   public async removeDevice(device: DeviceInfo) {
     if (device.platform === Platform.IOS) {
-      await removeIosSimulator(device.UDID, SimulatorDeviceSet.RN_IDE);
+      // kill simulator
     }
     if (device.platform === Platform.Android) {
-      await removeEmulator(device.avdId);
+      await ensureOldEmulatorProcessExited(device.avdId);
```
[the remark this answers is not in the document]
p-malecki: Made ensureOldEmulatorProcessExited static function in AndroidEmulatorDevice.

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -150,14 +157,28 @@ export class DeviceManager implements Disposable, DeviceManagerInterface {
...
+          if (device.platform === Platform.IOS) {
+            await removeIosSimulator(device.UDID, SimulatorDeviceSet.RN_IDE);
+          }
+          if (device.platform === Platform.Android) {
+            await removeEmulator(device.avdId);
+          }
+          resolve();
+      })
+    ]);
+    await this.loadDevices(true);
```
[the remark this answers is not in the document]
p-malecki: I've restored that.

packages/vscode-extension/src/project/project.ts
p-malecki: I thought that we should select different device after removing, but I guess it's not necessary and we ignore those errors in debug console when removed device is selected.

packages/vscode-extension/tsconfig.json
```diff
@@ -10,7 +10,10 @@
     "esModuleInterop": true,
     "skipLibCheck": true,
     "allowJs": true,
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "types": [
+      "./node_modules/user-agent-data-types"
```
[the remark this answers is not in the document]
p-malecki: Agree. I've removed that package and replace with type any instead of userAgentData type.

packages/vscode-extension/src/utilities/platform.ts
```diff
@@ -0,0 +1,21 @@
+import * as os from "os";
+
+export class Platform {
```
jakub-gonet, line 3: I kinda don't like the fact that we renamed Platform to DevicePlatform since it's used way more frequently. I'd see this named `NativePlatform` or `HostPlatform` and potentially shared with webview in global variable instead of having two classes for it.

packages/vscode-extension/src/utilities/platform.ts
```diff
@@ -0,0 +1,21 @@
...
+  static OS: "macos" | "windows" | "unknown" = (() => {
```
jakub-gonet, line 4: It can't get to "unknown" union member.

packages/vscode-extension/src/utilities/platform.ts
```diff
@@ -0,0 +1,21 @@
...
+      default:
+        return "macos";
+    }
+  })();
+
+  static select(obj: { macos: any; windows: any }) {
+    if (Platform.OS !== "unknown" && Platform.OS in obj) {
+      return obj[Platform.OS];
+    }
+  }
```
jakub-gonet, line 20: If possible, use `unknown` instead of `any`. We don't have any fallback here so following implementation is equivalent:

```js
static select(options: { macos: unknown; windows: unknown }) {
  return options[Platform.OS];
}
```

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -106,10 +114,20 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+    const command = Platform.select({
+      macos: "ps",
+      windows:
+        'powershell.exe "Get-WmiObject Win32_Process | Select-Object ProcessId, CommandLine | ConvertTo-Csv -NoTypeInformation"',
+    });
+    const args = Platform.select({ macos: ["-Ao", "pid,command"], windows: [] });
+    const subprocess = exec(command, args);
+    const pattern = Platform.select({
+      macos: `(\\d+)\\s.*qemu.*-avd ${avdId}`,
+      windows: `"(\\d+)".*qemu.*-avd ${avdId}`,
```
jakub-gonet, line 128: Why pattern is different between platforms? Shouldn't `.*` catch `\s` as well?

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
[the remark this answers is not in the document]
jakub-gonet: This is unexpected since we control the directory where they're saved.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -106,10 +114,20 @@ export class AndroidEmulatorDevice extends DeviceBase {
     }
   }
 
-  private async ensureOldEmulatorProcessExited() {
+  public static async ensureOldEmulatorProcessExited(avdId: string) {
```
jakub-gonet, line 117: Make it private please. Static is fine but I'm personally not a fan of it.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -475,7 +496,12 @@ export async function listEmulators() {
   );
 }
 
-export function removeEmulator(avdId: string) {
+export async function removeEmulator(avdId: string) {
+  // ensure to kill emulator process before removing avd files used by that process
+  if (Platform.OS == "windows") {
+    await AndroidEmulatorDevice.ensureOldEmulatorProcessExited(avdId);
```
jakub-gonet, line 502: Can we use `this` instead of `AndroidEmulatorDevice`? Statics should be available there IIRC.

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -51,10 +51,14 @@ export class DeviceManager implements Disposable, DeviceManagerInterface {
   }
 
   public async acquireDevice(deviceInfo: DeviceInfo) {
-    if (deviceInfo.platform === Platform.IOS) {
+    if (deviceInfo.platform === DevicePlatform.IOS) {
+      if (process.platform !== "darwin") {
```
jakub-gonet, line 55: Replace with helper. I'd actually throw here, no one should ask for iOS device on non-macOS.

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -102,10 +106,13 @@ export class DeviceManager implements Disposable, DeviceManagerInterface {
       Logger.error("Error fetching emulators", e);
       return [];
     });
-    const simulators = listSimulators().catch((e) => {
-      Logger.error("Error fetching simulators", e);
-      return [];
-    });
+    const simulators =
+      process.platform === "darwin"
```
jakub-gonet, line 110: Replace with the helper.

packages/vscode-extension/src/utilities/sdkmanager.ts
jakub-gonet: It shouldn't look like that. Can you please change the truncation code in the `recursiveSystemImagePathsSearch`? It hardcodes unix separator.

packages/vscode-extension/src/project/project.ts
jakub-gonet: We _should_ select different device. My point is that: at the time you call `listAllDevices()` the device that was deleted shouldn't be there. Then you don't need to add unnecessary workarounds for "device that was deleted but state wasn't updated so we need to avoid selecting it again".

packages/vscode-extension/src/utilities/sdkmanager.ts
p-malecki: You're right. That's much simpler.

packages/vscode-extension/src/project/project.ts
p-malecki: I've noticed that after removing device debug console logs `"Couldn't start device session Error: Command failed with exit code 204: xcrun simctl --set /Users/USER/Library/Caches/com.swmansion.react-native-ide/Devices/iOS install ED79C082-FDEF-46E3-94A0-088E0AE337AD /Users/USER/Library/Developer/Xcode/DerivedData/exporouterexample-fdhmhyxcaikoqicwmzpzgpagpcxx/Build/Products/Debug-iphonesimulator/exporouterexample.app..."` so my guess was that it selected the removed device. What is the reason that causes that log?

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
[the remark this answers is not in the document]
p-malecki: It was unnecessary so I've restored previous version of `getAvdIds`.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
p-malecki: No we can't. Function `removeEmulator` is not a member of `AndroidEmulatorDevice` and in `bootDevice` case  I found: "Static members are not directly accessible using the [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) keyword from non-static methods."

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
p-malecki: It's static because we use it in `removeEmulator ` function which is not a member of `AndroidEmulatorDevice `.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
p-malecki: Yes, a pattern without \s is equivalent to the current.

packages/vscode-extension/src/webview/providers/UtilsProvider.tsx
```diff
@@ -1,7 +1,7 @@
 import { PropsWithChildren, useContext, createContext } from "react";
 import { makeProxy } from "../utilities/rpc";
 import { Utils } from "../../utilities/utils";
-import { UtilsInterface } from "../../common/Utils";
+import { UtilsInterface } from "../../common/utils";
```
jakub-gonet, line 4: Please revert before changing.

PR #473, opened 2024-07-29
Move app event handling to device session

jakub-gonet:
⚠️ Depends on https://github.com/software-mansion/react-native-ide/pull/472.

Second PR in refactoring effort, see https://github.com/software-mansion/react-native-ide/pull/472 for overall notes.

Extracts events coming from app to `DeviceSession` in order to move devtools there in the future. Also extract device selection step to its own function in order to cut code in future refactorings.

### DIFF
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -1,7 +1,7 @@
 import { Disposable, commands, workspace, window, DebugSessionCustomEvent } from "vscode";
 import { Metro, MetroDelegate } from "./metro";
 import { Devtools } from "./devtools";
-import { DeviceSession } from "./deviceSession";
+import { AppEvent, DeviceSession, EventDelegate } from "./deviceSession";
 import { Logger } from "../Logger";
 import { BuildManager, didFingerprintChange } from "../builders/BuildManager";
 import { DeviceAlreadyUsedError, DeviceManager } from "../devices/DeviceManager";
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/project/project.ts
```diff
@@ -81,6 +83,39 @@ export class Project implements Disposable, MetroDelegate, DebugSessionDelegate,
...
+  //#endregion
+
+  //#region App events
+  onAppEvent<E extends keyof AppEvent, P = AppEvent[E]>(event: E, payload: P): void {
+    switch (event) {
+      case "appReady":
+        Logger.debug("App ready");
+        if (this.reloadingMetro) {
+          this.reloadingMetro = false;
+          this.updateProjectState({ status: "running" });
```
kmagiera, line 99: not sure where this whole refactor is going. It seems like you're moving the logic responsible for handling selected device into device session. In that case shouldn't we also move state to live there or maybe that's what some of the follow up PRs is doing?

packages/vscode-extension/src/project/project.ts
```diff
@@ -81,6 +83,39 @@ export class Project implements Disposable, MetroDelegate, DebugSessionDelegate,
...
   }
 
+  //#region Build progress
+  onStateChange(state: StartupMessage): void {
+    this.updateProjectStateForDevice(this.projectState.selectedDevice!, { startupMessage: state });
+  }
+  //#endregion
+
+  //#region App events
+  onAppEvent<E extends keyof AppEvent, P = AppEvent[E]>(event: E, payload: P): void {
```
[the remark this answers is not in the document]
jakub-gonet: We have to, [contextual typing](https://github.com/microsoft/TypeScript/pull/6118) aren't a thing for methods implementing interfaces.

packages/vscode-extension/src/project/project.ts
jakub-gonet: Discussed offline, to sum up:
we want to move as much state from the project as we can, including current `projectState`. This can't be done in one go so I opted to do it in smaller bites.
Right now, we're establishing what events change the state so it's clearer what's happening.

As for shared state – metro, devtools – we need to think of good way of keeping them in DeviceSession that allows to create multiple DeviceSession for multiple devices. I haven't got an answer to that, yet.

PR #475, opened 2024-07-29
Implement restartProcess reload option

jakub-gonet:
⚠️ Depends on https://github.com/software-mansion/react-native-ide/pull/474.

Fourth PR in refactoring effort, see https://github.com/software-mansion/react-native-ide/pull/472 for overall notes.

- Adds `onPreviewReady` callback to `Project`.
- Adds `DeviceSession.perform("restartProcess")`

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -10,8 +10,6 @@ import { AndroidEmulatorDevice } from "../devices/AndroidEmulatorDevice";
 import { getLaunchConfiguration } from "../utilities/launchConfiguration";
 import { DebugSession, DebugSessionDelegate } from "../debugging/DebugSession";
 
-type PreviewReadyCallback = (previewURL: string) => void;
-
 type PerformAction =
   | "rebuild"
   | "reboot"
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -102,30 +103,26 @@ export class DeviceSession implements Disposable {
...
     this.buildResult = await this.disposableBuild.build;
     this.eventDelegate.onStateChange(StartupMessage.Installing);
     await this.device.installApp(this.buildResult, false);
-
-    this.device.startPreview().then(() => {
-      previewReadyCallback(this.device.previewURL!);
-    });
-
-    await this.launch();
+    await this.perform("restartProcess");
```
kmagiera, line 125: not a huge fan of the perform method, it seems like it takes string and just calls back to the right method, so why can't we just call that method directly (in this case its launchApp)

kmagiera (approved):

packages/vscode-extension/src/project/deviceSession.ts
jakub-gonet: It's a good idea. I did it because it better resembled current structure for me while refactoring.

PR #477, opened 2024-07-29
Move buildManager into device session

jakub-gonet:
⚠️ Depends on https://github.com/software-mansion/react-native-ide/pull/476.

Sixth PR in refactoring effort, see https://github.com/software-mansion/react-native-ide/pull/472 for overall notes.

This change was mostly focused on moving `BuildManager` into `DeviceSession` but I also did a lot of small improvements in project's and device session's code.

### DIFF
--- a/packages/vscode-extension/src/project/metro.ts
+++ b/packages/vscode-extension/src/project/metro.ts
@@ -211,8 +211,9 @@ export class Metro implements Disposable {
   }
 
   public async reload() {
+    const appReady = this.devtools.appReady();
     await fetch(`http://localhost:${this._port}/reload`);
-    await this.devtools.appReady();
+    await appReady;
   }
 
   public async getDebuggerURL() {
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/project/metro.ts
```diff
@@ -211,18 +211,8 @@ export class Metro implements Disposable {
...
-          this.devtools.removeListener(waitForAppReady);
-          resolve();
-        }
-      };
-      this.devtools.addListener(waitForAppReady);
-    });
-
     await fetch(`http://localhost:${this._port}/reload`);
-    await appReady;
+    await this.devtools.appReady();
```
kmagiera, line 215: this changes the behavior. Before we'd start listening before the fetch and now this is reversed. The difference is that fetch is also async, but because of that it may already be too late to start listening for event after fetch is complete, becuase the event might've already be sent and we may miss it

kmagiera (changes requested):

kmagiera (approved):

PR #478, opened 2024-07-29
Per device settings

jakub-gonet:
⚠️ Depends on https://github.com/software-mansion/react-native-ide/pull/477.

Moves device settings from project to device. The added value is that we now support keeping different settings for different devices.

https://github.com/user-attachments/assets/6705121b-7ccb-440f-affc-93a9b014d3fe


jakub-gonet:
I actually started wondering if splitting it between devices is a good idea. It may be that you want it synchronized most of the time. Maybe it's better to keep it centralized in the DeviceManager instead. It certainly doesn't belong to Project IMO.

kmagiera:
yes, I agree. I'd rather keep it in sync. I don't think it is a common scenario for people to have multiple devices and when they do I don't think they keep track of different settings that they changed on different devices

jakub-gonet:
Will move it to other class later, without changing behavior.

PR #479, opened 2024-07-29
Add more fine-grained reload controls

jakub-gonet:
⚠️ Depends on https://github.com/software-mansion/react-native-ide/pull/477.

Exposes already implemented `DeviceSession.perform(type)` to extension UI.

https://github.com/user-attachments/assets/bcb1341b-0061-45d6-9d16-5d9e4f9bd38e

## Test plan

Tested by repeatedly switching between devices (every one of four combinations) and resetting permissions. Other features look good, too.


packages/vscode-extension/src/common/Project.ts
```diff
@@ -55,6 +55,14 @@ export const StartupStageWeight = [
   { StartupMessage: StartupMessage.AttachingDebugger, weight: 1 },
 ];
 
+export type ReloadAction =
+  | "rebuild"
+  | "reboot"
+  | "reinstall"
+  | "restartProcess"
+  | "reloadJs"
+  | "hotReload";
```
kmagiera, line 64: Can we have some small docs for the different actions in comments?
For example:
"rebuild" – performs a clean build of the app and clears metro cache
"reboot" – ???

packages/vscode-extension/src/project/project.ts
```diff
@@ -266,7 +267,13 @@ export class Project
       }
     }
   }
-  //#endregion
+
+  public async reload(type: ReloadAction): Promise<boolean> {
+    this.updateProjectState({ status: "starting" });
+    const success = (await this.deviceSession?.perform(type)) ?? false;
+    this.updateProjectState({ status: "running" });
```
kmagiera, line 274: is this ok that we always move to running state after reload even if it doesn't succeed?

kmagiera (changes requested):
Looks mostly good, for the code, I left a few comments inline.

My other comment is about the UI. I'd prefer to make it a bit more subtle. Maybe we can make the arrow appear only when you hover the reload button? For example xcode has similar options for the run button: 
image


On top of that I feel like this PR is the last from the series of other changes and it deserves some more evolved test plan to make sure nothing broke here or as a part of refactoring.

Specifically I'd be worried about the restart functionality that we perform when changing permissions on Android. As well as whether we handle cases when you switch device while build is still running on the previous device.

jakub-gonet:
Updated dropdown interactions:


https://github.com/user-attachments/assets/bcb1341b-0061-45d6-9d16-5d9e4f9bd38e

jakub-gonet:
Tested by repeatedly switching between devices (every one of four combinations) and resetting permissions. Other features look good, too.

packages/vscode-extension/src/project/project.ts
jakub-gonet: For now, we assume it can't fail. We'll revisit after we adapt the lower API to return booleans on success instead of throwing exceptions. I'll add a comment here.

kmagiera (approved):
Accepting, but it'd be nice to move the arrow such that it's within the hiver radius and also make it so that it shows and hides with the button overlay.

PR #481, opened 2024-07-30
Add multitouch support

p-malecki:
This PR adds implementation of multitouch support for iOS and Android devices.
To use multitouch feature press and hold `Option` button, then two markers indicating touch points show up. While holding `Option` button and pressing `left mouse` button movements and button clicks ("up", "down") are sent to the sim-server subprocess. To pan touch points press `Option` and `Shift` buttons at the same time.  When the mouse leaves device's screen area, multitouch ends with "up" command.
Instead of `Option` button use `Alt` button on Windows.

⚠️After implementing the backend part in sim-server (#470) ensure that a proper multitouch command is set in `sendMultiTouch` function in `src/devices/preview.ts` file.

resolves: #449

### DIFF
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -250,12 +250,8 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
   function onMouseMove(e: MouseEvent<HTMLDivElement>) {
     e.preventDefault();
     if (isMultiTouching) {
-      if (isPanning) {
-        moveAnchorPoint(e);
-      }
-      if (isPressing) {
-        sendMultiTouch(e, "Move");
-      }
+      isPanning && moveAnchorPoint(e);
+      isPressing && sendMultiTouch(e, "Move");
       setMirroredTouchPoint(getMirroredTouchPosition(anchorPoint));
     } else if (isPressing) {
       sendTouch(e, "Move");
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -401,6 +401,25 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
     device: device!,
   });
 
+  const TouchPointMarker: React.FC<TouchPoint> = ({ x, y }) => {
+    const styles: React.CSSProperties = {
+      position: "absolute",
+      top: `${y * 100}%`,
+      left: `${x * 100}%`,
+      width: "33px",
+      height: "33px",
+      backgroundColor: "rgba(175, 175, 175, 0.75)",
+      borderRadius: "50%",
+      borderColor: "rgba(135, 135, 135, 0.6)",
+      borderWidth: "1px",
+      borderStyle: "solid",
+      transform: "translate(-50%, -50%)",
+      boxShadow: isPressing ? "none" : "2px 2px 6px 1px rgba(0, 0, 0, 0.2)",
+      display: isMultiTouchVisible ? "block" : "none",
+    };
+    return <div style={styles} />;
+  };
+
   return (
     <>
       <div
@@ -420,6 +439,9 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
                   }}
                   className="phone-screen"
                 />
+
+                <TouchPointMarker x={touchPoint.x} y={touchPoint.y} />
+                <TouchPointMarker x={mirroredTouchPoint.x} y={mirroredTouchPoint.y} />
 
                 {inspectFrame && (
                   <div className="phone-screen phone-inspect-overlay">
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -302,6 +302,12 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
       sendTouch(e, "Up");
       setIsPressing(false);
     }
+    if (isMultiTouching) {
+      sendMultiTouch(e, "Up");
+      setIsMultiTouching(false);
+      setIsPanning(false);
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -116,8 +116,19 @@ type Props = {
...
+  }
+
   const wrapperDivRef = useRef<HTMLDivElement>(null);
   const [isPressing, setIsPressing] = useState(false);
+  const [isMultiTouching, setIsMultiTouching] = useState(false);
+  const [isPanning, setIsPanning] = useState(false);
+  const [touchPoint, setTouchPoint] = useState<TouchPoint>({ x: 0.5, y: 0.5 });
+  const [mirroredTouchPoint, setMirroredTouchPoint] = useState<TouchPoint>({ x: 0.5, y: 0.5 });
+  const [anchorPoint, setAnchorPoint] = useState<TouchPoint>({ x: 0.5, y: 0.5 });
+  const [isMultiTouchVisible, setIsMultiTouchVisible] = useState(false);
```
jakub-gonet, line 131: We should be able to remove this state and use `isMultiTouching`.

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -116,8 +116,19 @@ type Props = {
...
 function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Props) {
+  interface TouchPoint {
+    x: number;
+    y: number;
+  }
+
   const wrapperDivRef = useRef<HTMLDivElement>(null);
   const [isPressing, setIsPressing] = useState(false);
+  const [isMultiTouching, setIsMultiTouching] = useState(false);
+  const [isPanning, setIsPanning] = useState(false);
```
jakub-gonet, line 127: At this point I kinda want to make it an enum (string based instead of direct TS enum) instead of using three independent states. If you want, you can experiment with it.

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -327,6 +400,25 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
...
+      width: "33px",
+      height: "33px",
+      backgroundColor: "rgba(175, 175, 175, 0.75)",
+      borderRadius: "50%",
+      borderColor: "rgba(135, 135, 135, 0.6)",
+      borderWidth: "1px",
+      borderStyle: "solid",
+      transform: "translate(-50%, -50%)",
+      boxShadow: isPressing ? "none" : "2px 2px 6px 1px rgba(0, 0, 0, 0.2)",
+      display: isMultiTouchVisible ? "block" : "none",
```
jakub-gonet, line 417: Can we move it to conditional rendering?

```
{isMultiTouchVisible && <TouchPointMarker x={x} y={y} />}
```

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -327,6 +400,25 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
     device: device!,
   });
 
+  const TouchPointMarker: React.FC<TouchPoint> = ({ x, y }) => {
+    const styles: React.CSSProperties = {
```
jakub-gonet, line 404: Please move it outside of Preview so it doesn't get redefined every render. You also [don't need](https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/) `React.FC` here. If you move `styles` inline, you don't need to type it, we don't get much from that variable anyway.

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -116,8 +116,19 @@ type Props = {
...
+    x: number;
+    y: number;
+  }
+
   const wrapperDivRef = useRef<HTMLDivElement>(null);
   const [isPressing, setIsPressing] = useState(false);
+  const [isMultiTouching, setIsMultiTouching] = useState(false);
+  const [isPanning, setIsPanning] = useState(false);
+  const [touchPoint, setTouchPoint] = useState<TouchPoint>({ x: 0.5, y: 0.5 });
+  const [mirroredTouchPoint, setMirroredTouchPoint] = useState<TouchPoint>({ x: 0.5, y: 0.5 });
```
jakub-gonet, line 129: Can we always compute it instead of keeping it in the state?

jakub-gonet (changes requested):

packages/vscode-extension/src/webview/components/Preview.tsx
p-malecki: `mirroredTouchPoint` can be replaced with computing. I've changed it.
However, `moveAnchorPoint` needs to track previous mouse position so `touchPoint` state is used there before computing a new one. Also `getMirroredTouchPosition` function is called in place where mouse event is not available (after change it's used in TouchPointMarker component) so we need to store mouse position in `touchPoint`.

packages/vscode-extension/src/webview/components/Preview.tsx
p-malecki: Each of those states is responsible for different aspect:
- `isPressing` indicates whether the left mouse button is currently pressed.
- `isMultiTouching` indicates whether the Option button is being held, changing the visibility of multitouch markers.
- `isPanning` indicates whether the Shift button is being held; when the user stops pressing it, the isMultiTouching variable is used to determine whether multitouch markers were previously visible.

I tried to simplify it but keeping previous state of an enum would be necessary, so I'll omit adding that enum.

packages/vscode-extension/src/webview/components/Preview.tsx
p-malecki: `mirroredTouchPoint` can be replaced with computing. I've changed it.
However, `moveAnchorPoint` needs to track previous mouse position so `touchPoint` state is used there before computing a new one. Also `getMirroredTouchPosition` function is called in place where mouse event is not available (after change it's used in TouchPointMarker component) so we need to store mouse position in `touchPoint`.

jakub-gonet (approved):

PR #492, opened 2024-08-09
Change multitouch UI markers

p-malecki:
This PR changes the style of multitouch UI markers to be more visible and fixes onMouseLeave event logic to avoid sending a touch signal twice when the mouse leaves the screen.


packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -474,13 +493,31 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
                 />
 
                 {isMultiTouching && (
-                  <TouchPointMarker x={touchPoint.x} y={touchPoint.y} isPressing={isPressing} />
+                  <TouchPointMarker
+                    x={touchPoint.x}
+                    y={touchPoint.y}
+                    isPressing={isPressing}
+                    color="175, 175, 175"
```
filip131311, line 500: Why do you add those values here /or why did you defined default ones in the component definition. 
Also please extract magic numbers like that to constants.

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -137,24 +137,37 @@ type TouchPointMarkerProps = {
...
+  color?: string;
+  pressedColor?: string;
 };
 
-function TouchPointMarker({ x, y, isPressing }: TouchPointMarkerProps) {
+function TouchPointMarker({
+  x,
+  y,
+  isPressing,
+  color = "175, 175, 175",
```
filip131311, line 149: I left another note in the place you use this component, but is it necessary to pass it here as a prop can't we achieve this with css?

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -253,11 +266,13 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
   type MouseMove = "Move" | "Down" | "Up";
   function sendTouch(event: MouseEvent<HTMLDivElement>, type: MouseMove) {
     const { x, y } = getTouchPosition(event);
+    console.log("FRYTKI SINGLE", x, y, type);
```
filip131311, line 269: please remove console.logs

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -253,11 +266,13 @@ function Preview({ isInspecting, setIsInspecting, zoomLevel, onZoomChanged }: Pr
...
     project.dispatchTouch(x, y, type);
   }
 
   function sendMultiTouch(event: MouseEvent<HTMLDivElement>, type: MouseMove) {
     const { x, y } = getTouchPosition(event);
+    console.log("FRYTKI SINGLE", x, y, anchorPoint.x, anchorPoint.y, type);
```
filip131311, line 275: please remove console.logs

filip131311 (changes requested):
Please adress inline coments

filip131311 (approved):
LGTM

PR #495, opened 2024-08-13
Add nonce for inline script

jakub-gonet:
Fixes "Refused to execute inline script because it violates the following Content Security Policy" error.


filip131311 (approved):
LGTM

PR #496, opened 2024-08-13
Fix diagnostics for windows

p-malecki:
This PR hides iOS dependencies (Xcode, CocoaPods, Pods) from Windows users and introduces an 'Optional' status in InstallationStatus with a gray close icon, differing from the red icon used for the 'NotInstalled' status.


jakub-gonet (approved):
Didn't test locally, code looks fine.

PR #504, opened 2024-08-19
Add buildScript option

### LINKED ISSUE #150: Ability to use already created builds
Building the app takes a lot of time. It'd be nice to have a way to point to a build file and completely skip the building part that happens in the beginning

jakub-gonet:
Adds new configuration options to launch config:
- `buildScript` – replacing default build process with any script that outputs built app's path as last line of standard output (e.g. fetching app binary from the internet or using built app from local fs). It's an object with `ios` and `android` string keys.
- `eas` – replacing default build process with fetching app from EAS build service. It's an object with `profile`, `useBuildType`, and `buildUUID` keys. `profile` is used to select correct build profile from `eas.json`, `useBuildType` is either `"latest"` or `"id"` used to select either latest suitable build or selecting build via build UUID. `buildUUID` is required when using `useBuildType: "id"`.

You can't specify both `eas` and `buildScript` for one platform.

## Test plan

- `buildScript` – simple `"echo SOME_PATH"` script for both platforms.
- `eas` – tested both platforms, for "latest" and "id" strategies.

## Followups

- Add "matchAppVersion" option to use builds that match current branch app version for platform.
- If using EAS, notify the user in the UI and provide a link to monitor build process (or periodically poll API that returns progress if it exists).
- Consider best location to store build artifacts (right now uses system's temporary directory).
- Consider allowing drag & drop for built artefacts.
- Consider building a screen to select builds.

Resolves #150.


packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -76,14 +76,38 @@ export async function buildAndroid(
...
+    );
+    let apkPath: string | undefined;
+    lineReader(buildProcess).onLineRead((line) => {
+      apkPath = line.trim();
+    });
+
+    await buildProcess;
+
+    if (!apkPath || fs.existsSync(apkPath)) {
+      throw new Error("Build script didn't output any app path");
```
filip131311, line 93: I believe that dis error will cause webview to display buildErrorAlert, but this error alert is redirecting the user to build logs, which is not usefull in this case, could you make a special case for external builds that displays appropriate message and redirects to launchConfiguration? 

Screenshot 2024-08-23 at 14 24 43

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -76,14 +76,38 @@ export async function buildAndroid(
...
 ): Promise<AndroidBuildResult> {
+  const { buildScript, env, android } = getLaunchConfiguration();
+
+  if (buildScript) {
+    const buildProcess = cancelToken.adapt(
+      exec(buildScript.name, ["android", ...(buildScript.args ?? [])])
+    );
+    let apkPath: string | undefined;
+    lineReader(buildProcess).onLineRead((line) => {
+      apkPath = line.trim();
```
filip131311, line 87: is it always just one line? Is is the output EAS returns?

packages/vscode-extension/package.json
```diff
@@ -186,6 +186,26 @@
                 "type": "string",
                 "description": "Location of Metro config relative to the workspace. This is used for using custom configs for e.g. development."
               },
+              "buildScript": {
+                "type": "object",
+                "description": "Script used to build app or fetch app from known location. Executed as part of building process. Receives 'ios' or 'android' as first argument and should return a path to built app.",
```
filip131311, line 191: Is it the output EAS returns?

filip131311 (changes requested):
I left inline comment and question.

packages/vscode-extension/src/builders/buildAndroid.ts
jakub-gonet: We don't want to assume that it's always EAS. We expect a script that outputs a single line in stdout. I forgot to do that but we should update docs to document this new option.

kmagiera: I think it'd be easier if we let the script output stuff to stdout, otherwise people may need to add stdout forwarding to the scripts they already use and that'd make the scripts less readable. Instead, we can expect that when successful, the script returns code 0 and the path is the last line from the output.

Given that, the original code is ok, as it remembers the last line that was printent. However, we can make it so that the rest of the output is forwarded to extension logs such that it may be possible to get some feedback from the build script or see it output much like we see the output of gradle runs.

packages/vscode-extension/package.json
kmagiera, line 191: Expand on this. The script will not "return path", it prints it to stdout. Add note that this is "advanced" options for setups where non-standard build instructions are used

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -76,14 +76,38 @@ export async function buildAndroid(
...
+    });
+
+    await buildProcess;
```
kmagiera, line 90: this may also throw in which case I don't know what  the expected result to the user will be? Maybe we should handle it separately and give some more descriptive error

kmagiera (changes requested):
Added some comments inline

packages/vscode-extension/src/builders/customBuild.ts
```diff
@@ -0,0 +1,262 @@
...
+};
+
+export async function runExternalBuild(
+  cancelToken: CancelToken,
+  platform: DevicePlatform,
+  externalCommand: string
+): Promise<string> {
+  const { stdout, lastLine: binaryPath } = await runExternalScript(cancelToken, externalCommand);
+
+  let easBinaryPath = await downloadAppFromEas(stdout, platform, cancelToken);
```
kmagiera, line 69: I understand this code is generic and not specific to EAS, why we use EAS in the method name then?

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -318,27 +323,37 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
         Logger.error("Error while uninstalling will be ignored", e);
       }
+    };
+
+    // adb install sometimes fails because we call it too early after the device is initialized.
+    // we haven't found a better way to test if device is ready and already wait for boot_completed
+    // flag in waitForEmulatorOnline. But even after that even is delivered, adb install also sometimes
+    // fails claiming it is too early. The workaround therefore is to retry install command.
+    if (forceReinstall) {
+      uninstallApp(build.packageName);
```
kmagiera, line 333: the code before would await for uninstall, this one doesn't

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -301,15 +301,20 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
-    // we haven't found a better way to test if device is ready and already wait for boot_completed
-    // flag in waitForEmulatorOnline. But even after that even is delivered, adb install also sometimes
-    // fails claiming it is too early. The workaround therefore is to retry install command.
-    if (forceReinstall) {
+
+    // allowNonZeroExit is set to true to not print errors when INSTALL_FAILED_UPDATE_INCOMPATIBLE occurs.
+    const installApk = (allowDowngrade: boolean) =>
+      exec(
+        ADB_PATH,
+        ["-s", this.serial!, "install", ...(allowDowngrade ? ["-d"] : []), "-r", build.apkPath],
```
kmagiera, line 309: any reason why we can't just always "allow downgrade" ??

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -76,14 +78,30 @@ export async function buildAndroid(
   outputChannel: OutputChannel,
   progressListener: (newProgress: number) => void
 ): Promise<AndroidBuildResult> {
+  const { buildScript, env, android } = getLaunchConfiguration();
+
+  if (buildScript?.android) {
+    const apkPath = await runExternalBuild(
```
kmagiera, line 84: it'd make sense to also pass `env` from launch configuration to external build scripts, no?

packages/vscode-extension/src/builders/customBuild.ts
jakub-gonet: It is specific to EAS, we assume certain JSON structure.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
jakub-gonet: `adb` only allows downgrading debuggable packages. If we assume that every app we install is debuggable then we can always do it.

kmagiera: that's a good point. I think it's ok to assume process is debuggable unless it's expo go

kmagiera (approved):
I think we can merge it as is. Would be great to do a quick follow up to add better caching, specifically for the eas builds. Now every fingerprint change will result in us downloading the same EAS build while as long as build ID stays the same, there's no reason to download the file again

PR #505, opened 2024-08-20
Add authentication

szewczukk:
In this PR we ask the user for an API token, store it in secure storage and pass it as a param in to the sim-server.

Related PRs:
https://github.com/software-mansion-labs/customer-portal/pull/15
https://github.com/software-mansion-labs/simulator-server/pull/167


packages/vscode-extension/src/devices/preview.ts
```diff
@@ -22,8 +22,14 @@ export class Preview implements Disposable {
       Platform.select({ macos: "sim-server-executable", windows: "sim-server-executable.exe" })
     );
 
+    const token = await extensionContext.secrets.get("API_KEY");
+    if (!token) {
+      window.showErrorMessage("No token provided!");
```
filip131311, line 27:
```suggestion
      window.showErrorMessage("No license token provided!");
```

packages/vscode-extension/src/extension.ts
```diff
@@ -66,6 +67,11 @@ export function deactivate(context: ExtensionContext): undefined {
 export async function activate(context: ExtensionContext) {
   handleUncaughtErrors();
 
+  let apiKey = await context.secrets.get("API_KEY");
```
filip131311, line 70: It could be best to remove this constant to a const and import it when needed.

packages/vscode-extension/src/extension.ts
```diff
@@ -363,3 +375,16 @@ async function fixMacosBinary(context: ExtensionContext) {
...
+
+async function storeAPIKey(secrets: SecretStorage) {
+  const key = await window.showInputBox({ password: true, prompt: "Your API key" });
+
+  if (!key) {
+    window.showErrorMessage("No API key provided.", "Dismiss");
+    return;
+  }
+
+  await secrets.store("API_KEY", key);
```
filip131311, line 387: Again it would be best to extract magic constants to a const.

filip131311 (commented):
I left some inline comments, this PR is also blocked by BETA, so I don't aprove to avoid accidental merging.

jakub-gonet (commented):
IMO would be nice to have some feature flag defined somewhere (e.g. `CHECK_LICENSE_KEY`, along with description and expected toggle date) and all calls wrapped by it so we can merge it right away so we don't need to deal with conflicts.

kmagiera:
Closing this one as we don't need it right now and will likely need to redo some stuff around this concept once we do need it.

PR #506, opened 2024-08-21
Skip waiting to load on windows

jakub-gonet:
There's currently an error which prevents sending `RNIDE_appReady` events. This results in IDE being stuck at "Waiting for an app to load" state. We workaround it on Windows to never wait.

This is caused by wrapper.js not being loaded as app's wrapper. This also means we can't use inspector functionality and other things that rely on communication with the app.


filip131311 (approved):
LGTM

p-malecki (approved):

PR #507, opened 2024-08-21
Add UI for launch configuration

filip131311:
This PR adds a UI for lunch Configuration and links to it in case of a build error. 
The configuration panel consist mostly of input fields, but build schemas for iOS are auto detected and composed into a selection list. 

Proposed future improvements: 
- adding auto detection for build configurations, product flavors and build types. 
- adding a configuration for build source
- adding tooltips for each configuration

Screenshot 2024-08-21 at 12 53 17
Screenshot 2024-08-21 at 12 53 24
Screenshot 2024-08-21 at 12 58 14


packages/vscode-extension/src/webview/views/LaunchConfigurationView.css
```diff
@@ -0,0 +1,39 @@
...
+  justify-content: space-between;
+  border-radius: 4px;
+  padding: 0 8px;
+  font-size: 13px;
+  line-height: 1;
+  height: 26px;
+  gap: 5px;
+  color: var(--swm-default-text);
+  background-color: var(--swm-select-background);
+  cursor: pointer;
```
jakub-gonet, line 14: Why we need that?

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -0,0 +1,267 @@
...
+      return;
+    }
+    update("ios", { scheme, configuration: newConfiguration });
+  };
+
+  const availableXcodeSchemes = xcodeSchemes.map((xcodeScheme) => {
+    return { value: xcodeScheme, label: xcodeScheme };
+  });
+
+  availableXcodeSchemes.push({ value: "Auto", label: "Auto" });
```
jakub-gonet, line 85: Can we use undefined as a value and not introduce special case for "Auto"?

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+function IosConfiguration({ scheme, configuration, update, xcodeSchemes }: iosConfigurationProps) {
+  const configurationInputRef = useRef<HTMLInputElement>(null);
+
+  const onSchemeChange = (newScheme: string) => {
+    if (newScheme === "Auto") {
+      update("ios", { scheme: undefined, configuration });
+      return;
+    }
+    update("ios", { scheme: newScheme, configuration });
```
jakub-gonet, line 69: Nit: My opinion but I dislike using returns like that. I'd write this code as:

```
let scheme: string | undefined = newScheme;
if(newScheme === "Auto") {
  scheme = undefined;
}
update("ios", { scheme: newScheme, configuration });
```

so it doesn't have multiple points of doing the same side-effect and we only manipulate the input.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -0,0 +1,267 @@
...
+  ) => void;
+}
+
+function AndroidConfiguration({ buildType, productFlavor, update }: androidConfigurationProps) {
+  const buildTypeInputRef = useRef<HTMLInputElement>(null);
+  const productFlavorInputRef = useRef<HTMLInputElement>(null);
+
+  const onBuildTypeBlur = () => {
+    const newBuildType = buildTypeInputRef.current?.value;
+    if (newBuildType === "Auto" || !newBuildType) {
```
jakub-gonet, line 122: First condition already takes care of undefined case.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -0,0 +1,267 @@
...
+  };
+
+  return (
+    <div className="container">
+      <div className="setting-description">App Root:</div>
+      <input
+        ref={appRootInputRef}
+        className="input-configuration"
+        type="string"
+        defaultValue={!!appRoot ? appRoot : "Auto"}
```
jakub-gonet, line 187: `appRoot ?? "Auto"`.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -0,0 +1,267 @@
...
+      />
+    </div>
+  );
+}
+
+interface appRootConfigurationProps {
+  appRoot?: string;
+  update: <K extends keyof LaunchConfigurationOptions>(
+    key: K,
+    value: LaunchConfigurationOptions[K]
```
jakub-gonet, line 164: Can we extract it to own type? It's repeated multiple times in this file.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+  const onIsExpoChange = (newIsExpo: string) => {
+    if (newIsExpo === "true") {
+      update("isExpo", true);
+      return;
+    } else if (newIsExpo === "false") {
+      update("isExpo", false);
+      return;
+    }
+    // @ts-ignore
```
jakub-gonet, line 251: Why ts-ignore?

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -0,0 +1,267 @@
...
+    update("isExpo", "Auto");
+  };
+
+  return (
+    <div className="container">
+      <div className="setting-description">Is Expo:</div>
+      <Select
+        value={isExpo !== undefined ? isExpo.toString() : "Auto"}
```
jakub-gonet, line 259: `isExpo?.toString() ?? "Auto"`

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -0,0 +1,267 @@
+import "./View.css";
+import "./LaunchConfigurationView.css";
+import Label from "../components/shared/Label";
+import { useLaunchConfig } from "../providers/LaunchConfigProvider";
+import { LaunchConfigurationOptions } from "../../common/LaunchConfig";
+import Select from "../components/shared/Select";
+import { useRef } from "react";
+import { valid } from "semver";
```
jakub-gonet, line 8: Is it used?

packages/vscode-extension/src/panels/LaunchConfigController.ts
```diff
@@ -0,0 +1,112 @@
...
+      const launchConfiguration = workspace.getConfiguration(
+        "launch",
+        workspace.workspaceFolders![0].uri
+      );
+
+      const configurations = launchConfiguration.get<Array<any>>("configurations")!;
+
+      const RNIDEConfiguration = configurations.find((config) => {
+        return config.type === "react-native-ide";
+      });
```
jakub-gonet, line 30: A bit shorter:
```suggestion
      const RNIDEConfiguration = configurations.find(({type}) => type === "react-native-ide");
```

jakub-gonet (commented):
Seems okay from architectural point of view, mostly highlighted a lot of small things.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.css
```diff
@@ -0,0 +1,39 @@
...
+.setting-description {
+  font-size: 10px;
+  margin-bottom: 5px;
+}
+
+.container {
+  flex: 1;
+
+  padding-right: 30px;
+  padding-left: 10px;
```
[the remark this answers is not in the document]
filip131311: Why?

packages/vscode-extension/src/utilities/xcode.ts
```diff
@@ -0,0 +1,92 @@
...
+  } else if (xcodeprojFiles.length >= 1) {
+    xcodeprojLocation = getParentDirectory(xcodeprojFiles[0]);
+  }
+
+  if (xcodeprojLocation) {
+    return {
+      workspaceLocation,
+      xcodeprojLocation,
+      isWorkspace: workspaceLocation !== undefined,
+    } as IOSProjectInfo;
```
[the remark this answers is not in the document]
filip131311: This is correctly typed, but ts does not understand that  workspaceLocation !== undefined means isWorkspace is always true when workspaceLocation is not undefined and vice versa always false when it's undefined, TS thinks it's a `boolean`. alternatives would include: 
- changing the type of IOSProjectInfo to look something like: 
```
type IOSProjectInfo =
  {
      workspaceLocation?: string;
      xcodeprojLocation: string;
      isWorkspace: boolean;
    }
```
but that would make typing less useful 
- writing the return as: 
```
return workspaceLocation !== undefined ? 
  {
     workspaceLocation,
      xcodeprojLocation,
      isWorkspace: true
  }  : {
     workspaceLocation,
     xcodeprojLocation,
     isWorkspace: false
  }
```
but that would hurt runtime performance.

packages/vscode-extension/src/panels/LaunchConfigController.ts
```diff
@@ -0,0 +1,112 @@
...
+  private configListener: Disposable | undefined;
+
+  constructor() {
+    const getCurrentConfig = () => {
+      const launchConfiguration = workspace.getConfiguration(
+        "launch",
+        workspace.workspaceFolders![0].uri
+      );
+
+      const configurations = launchConfiguration.get<Array<any>>("configurations")!;
```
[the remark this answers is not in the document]
filip131311: Record is not an Array and the result is an Array.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
filip131311: unfortunately no the problem lies in our constrction of makeProxy function: 
```
export function makeProxy<T extends object>(objectName: string) {
  return new Proxy<T>({} as T, {
    get(_, methodName) {
      return (...args: any[]) => {
        const currentCallId = `${instanceToken}:${globalCallCounter++}`;
        let argsWithCallbacks = args.map((arg) => {
          if (typeof arg === "function") {
            maybeInitializeCallbackMessageListener();
            const callbackId =
              callbackToID.get(arg) || `${instanceToken}:${globalCallbackCounter++}`;
            callbackToID.set(arg, callbackId);
            idToCallback.set(callbackId, arg);
            return {
              __callbackId: callbackId,
            };
          } else {
            return arg;
          }
        });

        vscode.postMessage({
          command: "call",
          callId: currentCallId,
          object: objectName,
          method: methodName,
          args: argsWithCallbacks,
        });
        return new Promise((resolve, reject) => {
          registerCallResultPromise(currentCallId, resolve, reject);
        });
      };
    },
  });
}
```
specifically that if the last argument is undefined the `...args: any[]` length it shorter and it does not send the appropriate message.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
filip131311: I am not concerned with undefined but with empty string here

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
filip131311: unfortunately no because isExpo can be false and false is falsy

packages/vscode-extension/src/panels/LaunchConfigController.ts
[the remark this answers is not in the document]
jakub-gonet: I'm talking about `any` here.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
jakub-gonet: Can you please make it more explicit then?

packages/vscode-extension/src/common/LaunchConfig.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export interface LaunchConfigEventMap {
+  launchConfigChange: LaunchConfigurationOptions;
+}
+
+export interface LaunchConfigEventListener<T> {
+  (event: T): void;
+}
+
+export type LaunchConfigUpdateType = <K extends keyof LaunchConfigurationOptions>(
```
jakub-gonet, line 27: `LaunchConfigUpdater` maybe?

packages/vscode-extension/src/webview/views/LaunchConfigurationView.css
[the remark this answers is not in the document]
jakub-gonet: Doesn't really scale when window is resized.

jakub-gonet (approved):

PR #508, opened 2024-08-21
Fix lists without key prop errors

p-malecki:
This PR adds keys to lists in DeviceSettingsDropdown and IconButtonWithOptions to prevent error messages.


filip131311 (approved):
LGTM

PR #510, opened 2024-08-22
Replace rc version of React native 75 with the stable release

filip131311:
This PR add a new React native 75 test app to the repository to replace the old one that was running rc version of react native.


filip131311:
@jakub-gonet done.

jakub-gonet (approved):

PR #513, opened 2024-08-23
Hide Zoom controls

filip131311:
This PR hides zoom controls outside of the screen unless user hovers over the left border of our web view. 

https://github.com/user-attachments/assets/03a169b0-bc0a-4251-ac18-195a8d8a2e5c

### DIFF
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -1,4 +1,4 @@
-import { useState, useRef, useEffect, MouseEvent, forwardRef, RefObject } from "react";
+import { useState, useRef, useEffect, MouseEvent, forwardRef, RefObject, ReactNode } from "react";
 import clamp from "lodash/clamp";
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -137,6 +139,37 @@ function TouchPointMarker({ isPressing }: { isPressing: boolean }) {
...
 }
 
+type ButtonGroupLeftProps = {
+  children: ReactNode;
+};
+
+function ButtonGroupLeft({ children }: ButtonGroupLeftProps) {
+  const [isMouseOver, setIsMouseOver] = useState(false);
+
+  const hideButtonGroupTimeout = useRef<any | undefined>();
```
jakub-gonet, line 149: `any`, please type it better.

jakub-gonet (changes requested):
We need something signifying that zoom controls are on the left. If you don't know about it, it's hard to discover.

jakub-gonet (approved):

PR #515, opened 2024-08-23
Set environment on startup

jakub-gonet:
When launching VS Code from Applications, we sometimes get a `PATH` env var that doesn't have path to node, npm, or other tools. This happens because some tool managers like nvm, asdf, mise, etc update `PATH` only in rc files, which are only used when launching shells interactively.

To prevent errors, we start user's shell (or ZSH in case $SHELL isn't defined) interactively once when opening IDE panel for the first time and use its env with execa. In case of an error, we show a warning a proceed.

## Testing
To test this, run VS Code from applications, comment out `preLaunchTask` in `launch.json`, run `npm run build:extension-debug` and `npm run watch:extension` and launch extension development task (F5 or run launch.json).

You should not see any error in debug console and extension should work correctly.

### DIFF
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -23,9 +23,7 @@ import {
   setAppRootFolder,
   setExtensionContext,
 } from "./utilities/extensionContext";
-import { command } from "./utilities/subprocess";
-import path from "path";
-import os from "os";
+import { command, setupPathEnv } from "./utilities/subprocess";
 import fs from "fs";
 import { SidePanelViewProvider } from "./panels/SidepanelViewProvider";
 import { PanelLocation } from "./common/WorkspaceConfig";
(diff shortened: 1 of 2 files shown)


kmagiera (changes requested):
What are the command that fail due to lack of PATH? If there are commands that require shell, they should use `command` instead of `exec`

packages/vscode-extension/src/utilities/subprocess.ts
```diff
@@ -2,18 +2,59 @@ import { Logger } from "../Logger";
 import execa, { ExecaChildProcess } from "execa";
 import readline from "readline";
 import { Platform } from "./platform";
+import util from "util";
+import { exec as bareExec } from "child_process";
+import { getAppRootFolder } from "./extensionContext";
+const nodeExec = util.promisify(bareExec);
```
kmagiera, line 8: use `command` instead of this?

packages/vscode-extension/src/extension.ts
```diff
@@ -88,7 +86,18 @@ export async function activate(context: ExtensionContext) {
   commands.executeCommand("setContext", "RNIDE.sidePanelIsClosed", false);
 
   async function showIDEPanel(fileName?: string, lineNumber?: number) {
-    commands.executeCommand("setContext", "RNIDE.sidePanelIsClosed", false);
+    await commands.executeCommand("setContext", "RNIDE.sidePanelIsClosed", false);
+
+    if (Platform.OS === "macos") {
+      try {
+        await setupPathEnv();
```
kmagiera, line 93: this isn't the right place do do this, as when side panel mode is on, this will be called too late. We should move this to extension activation

packages/vscode-extension/src/utilities/subprocess.ts
```diff
@@ -2,18 +2,59 @@ import { Logger } from "../Logger";
...
+  // cwd option when it's set for the subprocess call therefore removing the
+  // risk of the subprocess using the wrong working directory.
+
+  // Additionally, we overwrite PATH env variable using env from interactive
+  // shell, ensuring that we have access to node and other tools, even when
+  // VSCode is launched as an application and not from the terminal.
+  if (pathEnv) {
+    options = {
+      ...options,
+      env: { ...options?.env, PATH: pathEnv ?? options?.env?.PATH },
```
kmagiera, line 55: expression after `??` will never be used. I also think that when `env.PATH` is specified, we shouldn't override it

kmagiera (approved):

Issue #454, opened 2024-07-17
Storyboook integration


darthez:
Feature similar to preview from react-native-ide but should work with storybook

PR #516, opened 2024-08-27
Integration with Storybook

p-malecki:
This PR introduces Storybook integration and adds the following features:

- Selecting a story via codeLens adds its preview to `__RNIDE_previews`, extracts it from the Storybook `View` class, and renders it in the IDE as preview.
- Enables switching between story previews using the UrlBar.
- Triggers the IDE to open upon selecting a story when the IDE is closed.
- Hides CodeLens when previews and Storybook stories are commented.
- Checks for the installation of the Storybook dependency.
- Adds Storybook as an optional dependency in the diagnostics view.
- For metro config setups that exclude the `.storybook` folder from the bundle when the Storybook flag is disabled, overrides the resolveRequest in the metro config to ensure it remains included.

Defined Storybook's minimum version as "5.2.0" due to its support for Component Story Format (CSF).
Added `expo-storybook` to test apps.

 resolves: #454


packages/vscode-extension/lib/storybook_helper.js
```diff
@@ -0,0 +1,43 @@
...
+    const story = view._preview.storyStore.getStoryContext(preparedStory);
+    view._setStory(story);
+
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    const { unboundStoryFn: StoryComponent } = story;
+
+    const key = `sb://${storyId}`;
+    global.__RNIDE_previews.set(key, {
+      component: (
+        <>{StoryComponent ? <StoryComponent {...story} /> : <Text>Failed to render story.</Text>}</>
```
jakub-gonet, line 34: Do we need a fragment here? Does it makes sense to register the story if component is undefined?

packages/vscode-extension/lib/wrapper.js
```diff
@@ -128,6 +130,14 @@ export function PreviewAppWrapper({ children, initialProps, ..._rest }) {
     return closePreviewPromise;
   }, [rootTag]);
 
+  const showStorybookStory = useCallback(
+    async (componentTitle, storyName) => {
+      const previewKey = await storybookPreview(componentTitle, storyName);
+      previewKey !== undefined && openPreview(previewKey);
+    },
+    [rootTag, handleNavigationChange]
```
jakub-gonet, line 138: Why we need those deps here if we don't use it?

packages/vscode-extension/src/project/project.ts
```diff
@@ -76,6 +78,7 @@ export class Project
     this.trySelectingInitialDevice();
     this.deviceManager.addListener("deviceRemoved", this.removeDeviceListener);
     this.detectedFingerprintChange = false;
+    this.storybookStatus = false;
```
jakub-gonet, line 81: `status` implies an enum. Can we use `storybookInstalled` or similar?

packages/vscode-extension/src/project/project.ts
```diff
@@ -390,6 +396,14 @@ export class Project
     this.deviceSession?.startPreview(appKey);
   }
 
+  public async showStorybookStory(componentTitle: string, storyName: string) {
+    if (this.storybookStatus) {
+      this.devtools.send("RNIDE_showStorybookStory", { componentTitle, storyName });
+    } else {
+      Logger.error("Storybook is not installed.");
```
jakub-gonet, line 403: This is a soft error, right? It won't show in the UI?

packages/vscode-extension/src/providers/PreviewCodeLensProvider.ts
```diff
@@ -13,26 +13,75 @@ export class PreviewCodeLensProvider implements CodeLensProvider {
     token: CancellationToken
   ): CodeLens[] | Thenable<CodeLens[]> {
     const text = document.getText();
+    const isStory = /.*stories.?(js|ts|jsx|tsx)$/.test(document.fileName);
```
jakub-gonet, line 16: You don't need `.*` part here if you don't capture it.

packages/vscode-extension/src/providers/PreviewCodeLensProvider.ts
```diff
@@ -13,26 +13,75 @@ export class PreviewCodeLensProvider implements CodeLensProvider {
...
+    return componentName;
+  }
+
+  addStorybookCodeLenses(
+    text: string,
+    document: TextDocument,
+    codeLenses: CodeLens[],
+    componentName: string
+  ) {
+    const storyRegex = /^(?:(?!\/\/).)*export const (\w+)(?::\s*\w+\s*)? =/gm;
```
jakub-gonet, line 57: I'd change the first dot to space so it's more obvious we want to avoid commented out exports.

packages/vscode-extension/lib/metro_helpers.js
```diff
@@ -64,13 +64,33 @@ function adaptMetroConfig(config) {
   // Handle the case when resolver is not defined in the config
   if (!config.resolver) {
     config.resolver = {};
+  } else {
+    const originalResolveRequest = config.resolver?.resolveRequest;
+    if (originalResolveRequest) {
+      // Override the resolveRequest method to include files in the ".storybook/"
+      // directory, which contains "storybook.requires.js".
```
kmagiera, line 71: Add more descriptive comment here to explain why its here:

> Standard storybook setup relies on `resolveRequest` being overriden in order to exclude storybook files from bein imported into the bundle. The files are only included when STORYBOOK_ENABLED environment variable is set. Apparently, we can't set that variable for the whole metro process as you'd normally do with storybook, because it also controlls swapping out the main app entrypoint which also accesses that env constant via expo-constants module. We here implement a workaround which only sets the env variable for the duration of resolveRequest call and reset it back afterwards such that it only impacts resolution process.

packages/vscode-extension/lib/storybook_helper.js
```diff
@@ -0,0 +1,45 @@
...
+
+    const key = `sb://${storyId}`;
+    global.__RNIDE_previews.set(key, {
+      component: <StoryComponent {...story} />,
+      name: storyId,
+    });
+
+    return key;
+  } catch (e) {
+    console.log("Failed to select story: ", e);
```
kmagiera, line 41: use console.error for errors

packages/vscode-extension/src/providers/PreviewCodeLensProvider.ts
```diff
@@ -13,26 +13,75 @@ export class PreviewCodeLensProvider implements CodeLensProvider {
     token: CancellationToken
   ): CodeLens[] | Thenable<CodeLens[]> {
     const text = document.getText();
+    const isStory = /stories.?(js|ts|jsx|tsx)$/.test(document.fileName);
```
kmagiera, line 16: add comment what this regex should do, for example:
> we detect whether a file is a storybook story based on name. If is ends with .stories.js / ts / etc we treat it as a story file

packages/vscode-extension/src/providers/PreviewCodeLensProvider.ts
```diff
@@ -13,26 +13,75 @@ export class PreviewCodeLensProvider implements CodeLensProvider {
...
+      if (!componentName) {
+        return [];
+      }
+      this.addStorybookCodeLenses(text, document, codeLenses, componentName);
+    }
+    this.addPreviewCodeLenses(text, document, codeLenses);
+    return codeLenses;
+  }
+
+  extractComponentName(text: string): string | null {
```
kmagiera, line 35: add comment how this logic works

test-apps/expo-router/metro.config.js
```diff
@@ -1,4 +1,35 @@
...
+//     process.env.STORYBOOK_ENABLED !== "true" &&
+//     defaultResolveResult?.filePath?.includes?.(".storybook/")
+//   ) {
+//     return {
+//       type: "empty",
+//     };
+//   }
+
+//   return defaultResolveResult;
+// };
```
kmagiera, line 33: remove?

kmagiera (approved):
👍 👍 👍 👍

Issue #514, opened 2024-08-23
Cannot run project


hf-yevhen:
**Description*
Cannot run project in React Native IDE - diagnostics cannot find Node Modules.
In my project I'm using monorepo (yarn workspaces)

<img width="469" alt="Screenshot 2024-08-23 at 13 00 57" src="https://github.com/user-attachments/assets/7ceb3be5-3f9f-4830-a4fa-93c266bd0cc5">

**Environment**
- Version of currently used React Native IDE: v0.0.17
- Version of React Native/Expo used in the project: Expo SDK 51 (expo 51.0.28, react-native 0.74.5)

node: 20.12.2
yarn: 1.22.22

dpecha777:
i have the same issue

filip131311:
Hello, thank you for your feedback, would you mind sharing the logs of React Native IDE? To find them you can go to the output section of your vscode and choose "React Native IDE" from the list. 

<img width="661" alt="Screenshot 2024-08-23 at 14 32 10" src="https://github.com/user-attachments/assets/093cf9a4-6534-46ea-87ea-13b869f967e4">

dpecha777:
Sure there it is :) 


`2024-08-23 14:37:02.091 [info] Found app root folder: /Users/dominik/Work/EDGY/EdgyTasks
2024-08-23 14:37:02.091 [info] Setup dependency checker listeners.
2024-08-23 14:37:02.091 [info] Installing Node Modules
2024-08-23 14:37:02.091 [info] Launching devtools
2024-08-23 14:37:02.091 [info] Launching metro
2024-08-23 14:37:02.102 [info] Devtools started on port 61723
2024-08-23 14:37:02.202 [info] Received checkNodejsInstalled command.
2024-08-23 14:37:02.202 [info] Received checkAndroidEmulatorInstalled command.
2024-08-23 14:37:02.202 [info] Android Emulator installed: true
2024-08-23 14:37:02.203 [info] Received checkXcodeInstalled command.
2024-08-23 14:37:02.204 [info] Received checkCocoaPodsInstalled command.
2024-08-23 14:37:02.205 [info] Received checkNodeModulesInstalled command.
2024-08-23 14:37:02.206 [info] Received checkReactNativeInstalled command.
2024-08-23 14:37:02.206 [info] Check react-native module version. Version found: 0.74.5. Minimum version: 0.71.0
2024-08-23 14:37:02.206 [info] Minimum React Native version installed: true
[...]

PR #517, opened 2024-08-27
Fix yarn workspaces

filip131311:
This PR fixes node_modules detection for yarn workspaces and adds a working example of a set up in which workspaces are used. 

relates to: #514


packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -14,49 +18,74 @@ async function pathExists(p: string) {
...
+  function findWorkspace(appRoot: string) {
+    let currentDir = appRoot;
+    let parentDir = path.resolve(currentDir, "..");
+    while (parentDir !== currentDir) {
+      currentDir = parentDir;
+      parentDir = path.resolve(currentDir, "..");
+      const packageJsonPath = path.join(currentDir, "package.json");
+      if (!existsSync(packageJsonPath)) {
+        continue;
+      }
```
jakub-gonet, line 31: I think we should just read the file and catch if doesn't exist, something like that:
```ts
function walkDirectoriesUntilRoot(
  start: string,
  shouldStop: (current: string) => boolean
): string | undefined {
  let currentDir = start;
  let parentDir = path.resolve(currentDir, "..");
  while (parentDir !== currentDir) {
    currentDir = parentDir;
    parentDir = path.resolve(currentDir, "..");

    if (shouldStop(currentDir)) {
      return currentDir;
    }
  }
  return undefined;
}

function findWorkspace(appRoot: string) {
  return walkDirectoriesUntilRoot(appRoot, (currentDir) => {
    try {
      const packageJsonPath = path.join(currentDir, "package.json");
      const hasWorkspaces = require(packageJsonPath).workspaces !== undefined;
      return hasWorkspaces;
    } catch (error) {
      // No package.json
      return false;
    }
  });
}
```

jakub-gonet (approved):

packages/vscode-extension/src/utilities/packageManager.ts
filip131311: i did that just reaarengedsome functions to reuse code

packages/vscode-extension/src/extension.ts
```diff
@@ -273,31 +321,24 @@ async function findAppRootFolder() {
...
-    "**/node_modules/react-native/package.json",
-    null
-  );
-  if (rnPackageLocation) {
-    return Uri.joinPath(rnPackageLocation, "../../..").fsPath;
+  if (appRootCandidates.length > 1) {
+    const openLaunchConfigButton = "Open Launch Configuration";
+    window
+      .showErrorMessage(
+        `multiple reat-native applications were detected in the workspace, ${appRootCandidates[0]} were chosen, but you can change the root of your application in lunch configuration.`,
```
kmagiera, line 330: add some text that says that in order to get rid of this error, people should specify app root in the launch configuration

kmagiera (approved):

packages/vscode-extension/src/extension.ts
jakub-gonet: Also, "lunch configuration" typo. 🥪

DorianMazur:
This PR probably broke react-native-ide. Now node_modules don't work. I had to downgrade to 0.17

filip131311:
@DorianMazur hello, thank you so much for your feedback would you mind sharing React Native IDE logs from 0.18 version with us? you can find them in the output section of your vscode. Why do you think it is this PR that broke it for you?

DorianMazur:
I'm using yarn with corepack in a project and it basically can't find node_modules. After downgrading to 0.17 everything works like a charm.

PR #522, opened 2024-09-02
Fix wrapper to work on Windows

jakub-gonet:
We're using filenames which are platform-specific to check what files should be modified. Windows uses backslashes so it wasn't ever matched.

## Testing plan

Run any app on windows and check if RN IDE functionality works (inspector, etc).

### DIFF
--- a/packages/vscode-extension/lib/babel_transformer.js
+++ b/packages/vscode-extension/lib/babel_transformer.js
@@ -52,11 +52,15 @@ overrideModuleFromAppDir("@babel/plugin-transform-react-jsx-self", {
   visitor: {},
 });
 
-function transformWrapper({ filename, src, ...rest }) {
+function transformWrapper({ filename, src, options, plugins }) {
+  function isTransforming(unixPath) {
+    return filename.endsWith(path.normalize(unixPath));
+  }
+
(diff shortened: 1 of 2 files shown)


p-malecki (approved):

packages/vscode-extension/lib/babel_transformer.js
```diff
@@ -97,7 +101,7 @@ function transformWrapper({ filename, src, ...rest }) {
     }
   }
 
-  return transform({ filename, src, ...rest });
+  return transform({ filename, src, options, plugins });
```
kmagiera, line 104: why this change here? I actually fixed one issue in another transformer a while ago that was assuming only three elements in that object exists while `plugins` was added afterwards to the specification. Given we don't use all of them it'd be more future proof to have the ellipsis here

kmagiera (approved):
Approving, but please reverse spread operator change

packages/vscode-extension/lib/babel_transformer.js
kmagiera: Here's the PR: https://github.com/kristerkari/react-native-svg-transformer/pull/354/files

kmagiera:
Also, can you provide the test plan here?

PR #526, opened 2024-09-04
URL bar improvements

p-malecki:
This PR introduces several enhancements to the URL bar:
- The `home` button now opens the root URL (`/{}`) with empty params. Initial params may be may be introduced in future PRs.
- The `go back` button navigates back through the URL history up to a limit of 20 URLs.
- A list of the 5 most recently visited URLs is added to the top of `Select.Content` in the URL bar, followed by an alphabetically sorted list of all visited URLs.
- Improved responsiveness of `Select.Content` and `Select.Trigger` components.
- Added a scroll buttons for long URL lists in the `Select.Content`.
- Long URL names now break at dashes and display in full upon hover in `Select.Content`.
- Added `Expo Router` as an optional dependency to the diagnostics view.

Screenshots:
Screenshot 2024-09-04 at 11 15 34
Screenshot 2024-09-04 at 11 15 59
Screenshot 2024-09-03 at 13 35 04


packages/vscode-extension/src/webview/components/UrlBar.tsx
```diff
@@ -35,16 +35,50 @@ function ReloadButton({ project, disabled }: ReloadButtonProps) {
...
+
+      function removeDynamicPostfix(url: string): string {
+        return url.split("?")[0];
+      }
+
+      const newRecord = {
+        name: navigationData.displayName, //removeDynamicPostfix(), TODO
+        id: navigationData.id,
+      };
+      const isNotInHistory = !urlHistory.length || urlHistory[0] !== newRecord.id;
```
jakub-gonet, line 59: Please be more explicit – `urlHistory.length === 0`

packages/vscode-extension/src/webview/components/UrlBar.tsx
```diff
@@ -35,16 +35,50 @@ function ReloadButton({ project, disabled }: ReloadButtonProps) {
...
+
+      setUrlList((urls) => [newRecord, ...urls.filter((record) => record.id !== newRecord.id)]);
+      setRecentUrlList((recentUrls) => {
+        const filteredRecentUrls = [
+          newRecord,
+          ...recentUrls.filter((record) => record.id !== newRecord.id),
+        ];
+        return filteredRecentUrls.length > MAX_RECENT_URL_SIZE
+          ? filteredRecentUrls.slice(0, MAX_RECENT_URL_SIZE)
+          : filteredRecentUrls;
```
jakub-gonet, line 69: I think always slice'ing isn't that bad. We lose some performance on copying but it's much more readable.

packages/vscode-extension/src/webview/components/UrlBar.tsx
```diff
@@ -35,16 +35,50 @@ function ReloadButton({ project, disabled }: ReloadButtonProps) {
...
+        return url.split("?")[0];
+      }
+
+      const newRecord = {
+        name: navigationData.displayName, //removeDynamicPostfix(), TODO
+        id: navigationData.id,
+      };
+      const isNotInHistory = !urlHistory.length || urlHistory[0] !== newRecord.id;
+
+      setUrlList((urls) => [newRecord, ...urls.filter((record) => record.id !== newRecord.id)]);
```
jakub-gonet, line 61: This could better show intent if you did:
```
function moveAsMostRecent(urls, url) {
  return [newUrl, ...urls.filter((record) => record.id !== newRecord.id)]
}
```

then you can replace this call and call to `setRecentUrlList` below.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -1,29 +1,93 @@
...
+  HTMLDivElement,
+  PropsWithChildren<Select.SelectItemProps & { urlWidth: number }>
+>(({ children, urlWidth, ...props }, forwardedRef) => {
+  const itemStyle = { maxWidth: urlWidth - 20 };
+  return (
     <Select.Item className="url-select-item" {...props} ref={forwardedRef}>
       <Select.ItemText>
-        <div className="url-select-item-text">{children}</div>
+        <div className="url-select-item-text" style={itemStyle}>
+          {children}{" "}
```
jakub-gonet, line 14: I think you don't want the space here.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -1,29 +1,93 @@
...
     </Select.Item>
-  )
-);
+  );
+});
 
 interface UrlSelectProps {
   value: string;
   onValueChange: (newValue: string) => void;
+  recentItems: { id: string; name: string }[];
```
jakub-gonet, line 24: Consider extracting it to it's own type, like `NavigationItem`.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -1,29 +1,93 @@
...
+  const [windowWidth, setWindowWidth] = useState(window.innerWidth);
+  const longestUrl = Math.max(...items.map((item) => item.name.length));
+  let urlWidth = Math.min(Math.max(longestUrl * 7.5, 180), windowWidth * 0.6);
+  const maxCharsInSingleLine = Math.floor(urlWidth / 7.5);
+
+  const handleValueChange = (newValue: string) => {
+    const dashIndex = newValue.indexOf("#");
+    const strippedValue = dashIndex === -1 ? newValue : newValue.substring(dashIndex + 1);
+    onValueChange(strippedValue);
+  };
```
jakub-gonet, line 44:
```suggestion
  const handleValueChange = (newSelection: string) => {
    const stripped = newSelection.replace(/^recent#/, '')
    onValueChange(stripped)
  };
```

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -1,29 +1,93 @@
...
+
+  useEffect(() => {
+    window.addEventListener("resize", handleResize);
+    return () => {
+      window.removeEventListener("resize", handleResize);
+    };
+  }, []);
+
+  function splitLines(text: string) {
+    // Reformats text to max line lenght characters per line, breaking at last special char if possible.
```
jakub-gonet, line 58:
```suggestion
    // Reformats text to max line length characters per line, breaking at last special char if possible.
```

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -1,29 +1,93 @@
...
+
+    let result = "";
+    let startIndex = 0;
+    while (startIndex < text.length) {
+      if (startIndex + maxCharsInSingleLine > text.length) {
+        result += text.substring(startIndex);
+        break;
+      }
+      let chunk = text.substring(startIndex, startIndex + maxCharsInSingleLine);
+      let nextBreakPoint = findBreakPoint(chunk, startIndex);
```
jakub-gonet, line 82: `const`.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -32,16 +96,37 @@ function UrlSelect({ onValueChange, items, value, disabled }: UrlSelectProps) {
...
-                  </SelectItem>
-                )
-            )}
+            <Select.Group>
+              <Select.Label className="url-select-label">Recently used:</Select.Label>
+              {recentItems.map(
+                (item) =>
+                  item.name && (
+                    <SelectItem value={`recent#${item.id}`} key={item.id} urlWidth={urlWidth}>
+                      {item.name.length > maxCharsInSingleLine ? splitLines(item.name) : item.name}
```
jakub-gonet, line 109: Why we can't achieve this with CSS?

p-malecki: I've managed to remove the computation and passing of urlWidth and use css to determine components' width.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -1,29 +1,93 @@
...
+function UrlSelect({ onValueChange, recentItems, items, value, disabled }: UrlSelectProps) {
+  // We use two lists for URL selection: one with recently used URLs and another
+  // with all available URLs. Since recentItems is a subset of items, each recentItems's
+  // value is prefixed to differentiate their origins when presented in the Select
+  // component. This prefix is stripped off when the selected value is passed back
+  // through onValueChange.
+  const [windowWidth, setWindowWidth] = useState(window.innerWidth);
+  const longestUrl = Math.max(...items.map((item) => item.name.length));
+  let urlWidth = Math.min(Math.max(longestUrl * 7.5, 180), windowWidth * 0.6);
+  const maxCharsInSingleLine = Math.floor(urlWidth / 7.5);
```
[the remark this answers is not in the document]
p-malecki: The char size is approximately 8px, but urlWidth was mostly determined experimentally. It has now been replaced with a better solution.

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -103,7 +104,7 @@ function PreviewView() {
   return (
     <div className="panel-view">
       <div className="button-group-top">
-        <UrlBar project={project} disabled={devicesNotFound} />
+        <UrlBar project={project} disabled={devicesNotFound} resetUrlHistory={!isRunning} />
```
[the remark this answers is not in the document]
p-malecki: It's now resetting at key prop change.

packages/vscode-extension/src/project/project.ts
```diff
@@ -236,8 +238,11 @@ export class Project
     }
   }
 
-  public async goHome() {
-    await this.reloadMetro();
+  public async goHome(homeUrl: string) {
+    console.log("FRYTKI ", homeUrl);
```
filip131311: please remove

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -103,10 +103,8 @@ function PreviewView() {
   return (
     <div className="panel-view">
       <div className="button-group-top">
-        <UrlBar project={project} disabled={devicesNotFound} />
-
+        <UrlBar key={resetKey} project={project} disabled={devicesNotFound} />
```
filip131311, line 106: why do we pas project as a prop here? Is UrlBar Ever used outside of the ProjectContextProvider?

packages/vscode-extension/src/webview/providers/DependenciesProvider.tsx
```diff
@@ -202,6 +207,9 @@ export default function DependenciesProvider({ children }: PropsWithChildren) {
         case "installingPods":
           updateDependency("Pods", { error: undefined, installed: InstallationStatus.InProgress });
           break;
+        case "isExpoRouterInstalled":
+          updateDependency("ExpoRouter", data, { isOptional: true });
```
filip131311, line 211: It is not always optional, if the set up uses expo router it is required.

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -315,6 +320,26 @@ export class DependencyManager implements Disposable {
     return installed;
   }
 
+  public async checkExpoRouterInstalled() {
+    const status = checkMinDependencyVersionInstalled(
```
filip131311, line 324: this function alone does not give as information if project uses expo router or not

filip131311 (changes requested):
I left some inline coments

packages/vscode-extension/src/webview/views/PreviewView.tsx
p-malecki: I've changed it to use useProject instead of props.

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -460,3 +462,18 @@ export function checkMinDependencyVersionInstalled(dependency: string, minVersio
...
+
+export function isExpoRouterProject() {
+  try {
+    const appRoot = getAppRootFolder();
+    const packageJson = require(path.join(appRoot, "package.json"));
+    const hasExpoRouterDependency = Object.values<string>(packageJson.dependencies).some(
+      (dependencies: string) => {
+        return dependencies.includes("expo-router");
+      }
+    );
```
jakub-gonet, line 474:
```suggestion
    const hasExpoRouter = Object.values<string>(packageJson.dependencies).some((dependency) => dependency === "expo-router");
```

In the future, we can think about creating `maybeRequire` that returns undefined instead of throwing.

packages/vscode-extension/src/project/project.ts
```diff
@@ -236,8 +238,10 @@ export class Project
     }
   }
 
-  public async goHome() {
-    await this.reloadMetro();
+  public async goHome(homeUrl: string) {
+    if (this.expoRouterInstalled) {
+      await this.openNavigation(homeUrl);
+    }
```
jakub-gonet, line 244: What if it's not installed?

packages/vscode-extension/src/webview/components/UrlBar.tsx
```diff
@@ -65,32 +98,34 @@ function UrlBar({ project, disabled }: UrlBarProps) {
...
-        disabled={disabled || urlList.length == 0}>
+        disabled={disabled || urlList.length < 1}>
         <span className="codicon codicon-home" />
       </IconButton>
       <UrlSelect
         onValueChange={(value: string) => {
           project.openNavigation(value);
         }}
-        items={urlList}
+        recentItems={recentUrlList}
```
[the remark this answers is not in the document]
p-malecki: When changing from a preview to another URL, two route changes happen: {displayName: '', id: '{}'} and {displayName: '/', id: '/{}'}. We skip storing the first route with an empty displayName to prevent saving home URL entries twice.

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -435,3 +462,16 @@ export function checkMinDependencyVersionInstalled(dependency: string, minVersio
 export async function checkAndroidEmulatorExists() {
   return fs.existsSync(EMULATOR_BINARY);
 }
+
+export function isExpoRouterProject() {
+  try {
+    const appRoot = getAppRootFolder();
+    const packageJson = require(path.join(appRoot, "package.json"));
```
filip131311, line 469: requireNoCashe

packages/vscode-extension/src/project/project.ts
p-malecki: It that case we reload metro. I've changed it.

packages/vscode-extension/src/webview/components/UrlBar.tsx
```diff
@@ -35,64 +25,93 @@ function ReloadButton({ project, disabled }: ReloadButtonProps) {
...
+  const MAX_URL_HISTORY_SIZE = 20;
+  const MAX_RECENT_URL_SIZE = 5;
+
+  const [urlList, setUrlList] = useState<UrlItem[]>([]);
+  const [recentUrlList, setRecentUrlList] = useState<UrlItem[]>([]);
+  const [urlHistory, setUrlHistory] = useState<string[]>([]);
+
   useEffect(() => {
+    function moveAsMostRecent(urls: UrlItem[], newUrl: UrlItem) {
+      return [newUrl, ...urls.filter((record) => record.id !== newUrl.id)];
```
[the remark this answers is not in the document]
jakub-gonet: I don't agree, trading 2ms of performance for worse readability isn't worth it.

packages/vscode-extension/src/webview/components/UrlBar.tsx
```diff
@@ -35,64 +25,102 @@ function ReloadButton({ project, disabled }: ReloadButtonProps) {
...
+
+  const sortedUrlList = useMemo(() => {
+    return [...urlList].sort((a, b) => a.name.localeCompare(b.name));
+  }, [urlList]);
 
   return (
     <>
+      <IconButton
+        onClick={() => {
+          console.log("FRYTKI ", urlHistory, urlList);
```
jakub-gonet, line 84: Remove.

filip131311 (approved):
I aprove but left some coments

PR #530, opened 2024-09-09
New React Native debugger support

kmagiera:
This PR adds the new React Native debugger support to IDE. The new debugger will be enabled starting from RN 0.76, but it can be used in prior versions (0.72 even) after enabling build configuration flag.

Here's the gist of the changes:
1) We updated the way we lookup websocket debugger URL after reading the list of available pages from /list/json endpoint. The old debugger had a different format for page IDs that we used to detect the runtime that supported reloads. With the new debugger this is no longer necessary as c++ adapter handles that and gives us a stable device id hash. To make this compatible across different RN versions, we now use the new lookup mechanism, and if we don't find any matches we fallback to the old lookup method
2) The new debugger uses a project alias when compiling source maps. Instead of using absolute paths for filenames, it now uses a prefix "/[metro-project]". This needs to be taken into account when transforming file names the editor uses. We account for this in debug adapter implementation that takes a new configuration option. When launching the debugger, we detect whether it uses new or old debugger (based on the page naming) and sets up the alias processing when necessary.

## Caveat

The path resolution bit may not work on Windows yet as we use "/[metro-project" as prefix, and I'm not sure if metro uses backslash on Windows or not, and if it does I don't know if path.join will handle this. So this change needs to be tested and fixed on Windows.

## Test plan
This has been tested with RN 76 test app from #535 on Android and iOS by adding some breakpoints in the app code.
Also test old debugger flow, especially on Android, as we touched some lines of code that were there to translate debugger endpoints to local network addresses.

### DIFF
--- a/packages/vscode-extension/src/debugging/DebugAdapter.ts
+++ b/packages/vscode-extension/src/debugging/DebugAdapter.ts
@@ -71,6 +71,8 @@ class MyBreakpoint extends Breakpoint {
 export class DebugAdapter extends DebugSession {
   private variableStore: VariableStore = new VariableStore();
   private connection: WebSocket;
+  private absoluteProjectPath: string;
+  private projectPathAlias?: string;
   private configuration: DebugConfiguration;
   private threads: Array<Thread> = [];
   private sourceMaps: Array<[string, string, SourceMapConsumer]> = [];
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -368,7 +380,10 @@ export class DebugAdapter extends DebugSession {
   }
 
   private cdpMessageId = 0;
-  private cdpMessagePromises: Map<number, (result: any) => void> = new Map();
+  private cdpMessagePromises: Map<
+    number,
+    { resolve: (result: any) => void; reject: (error: any) => void }
```
jakub-gonet, line 385: Nit: extract to own type so it fits in one line and replace `any` with `unknown`.

packages/vscode-extension/src/project/metro.ts
```diff
@@ -228,12 +233,11 @@ export class Metro implements Disposable {
     return websocketAddress;
   }
 
-  private async fetchDebuggerURL() {
-    // query list from http://localhost:${metroPort}/json/list
-    const list = await fetch(`http://localhost:${this._port}/json/list`);
-    const listJson = await list.json();
-    // with metro, pages are identified as "deviceId-pageId", we search for the most
-    // recent device id and want want to use special -1 page identifier (reloadable page)
+  private lookupWsAddressForOldDebugger(listJson: any) {
```
jakub-gonet, line 236: Nit: improve types for documentation. Not sure if last two keys are always present and what's the type of `reactNative`.
```
type DebuggerPage = {
  id: string; // format: deviceId-pageId – pre 0.76 (number, -1 | 1), post 0.76 (hex string, number)
  webSocketDebuggerUrl: string;
  title: string;
  reactNative: ???;
}
```

We could also collapse two functions into one and return `[isNewDebugger, Url]` tuple to flatten those ifs below (at little alloc cost).

jakub-gonet (approved):

PR #532, opened 2024-09-10
Add device localization on ios

filip131311:
This PR introduces a functionality for changing device localization on ios. 

https://github.com/user-attachments/assets/2259b6c2-c490-466c-bdcf-3fc3f7ad9381


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -116,6 +122,10 @@ export class DeviceSession implements Disposable {
   private async bootDevice(deviceSettings: DeviceSettings) {
     this.eventDelegate.onStateChange(StartupMessage.BootingDevice);
     await this.device.bootDevice();
+    if (await this.device.checkLocale(deviceSettings.locale)) {
+      await this.device.changeLocale(deviceSettings.locale);
+      await this.device.resetDevice();
+    }
```
kmagiera, line 128: seems like this is implementation detail for a given device, shouldn't it be a part of changeSetting implementation that's specific to device rather than being handled in a generic way?

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -164,6 +197,38 @@ export class IosSimulatorDevice extends DeviceBase {
...
+  async changeLocale(newLocale: Locale): Promise<boolean> {
+    const deviceSetLocation = getOrCreateDeviceSet();
+    const languageCode = newLocale.match(/([^_-]*)/)![1];
+    await exec("xcrun", [
+      "simctl",
+      "--set",
+      deviceSetLocation,
+      "spawn",
+      this.deviceUDID,
+      "defaults",
```
kmagiera, line 209: I wonder if we could simplify the process and avoid restarts if we use the same method for updating defaults as we do for things like schemeapproval list (see below)

kmagiera (commented):
added a few comments inline. As they may impact parts of the implementation I'll be taking another look once you respond to the comments

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
filip131311: Unfortunately no, we could change the way we modify relavent, plist but we still have no way to modify what operating system keeps in memory at runtime.

kmagiera: ok, but we could still update that before the boot. Now we boot the device to only update this setting and the reboot.

filip131311: you were right I was able to avoid reset on on boot

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -119,7 +137,72 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+    const locale = newLocale.replace("_", "-");
+    const SETTINGS_SYSTEM_PATH = "/data/system/users/0/settings_system.xml";
+
+    await exec(ADB_PATH, ["-s", this.serial!, "root"]);
+
+    // most of the time persist.sys.locale will be enough to change phone localization,
+    // but if the user has more then languages set up then a setting in settings_system are created
+    // in such a case we need to change them too. It seems that if the setting exists in
+    // settings_system it is persistent even if the number of locales drops back to one.
+    await exec(ADB_PATH, ["-s", this.serial!, "shell", "setprop", "persist.sys.locale", locale]);
```
kmagiera, line 167: Have you tried using `-change-language` and `change-locale` options from `emulator`:
image

I tried and they seem to work, although when used they trigger cold boot of the device. So perhaps worth combining with some cached information about the current language

kmagiera (commented):
Leaving a couple more comments inline

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
filip131311: unfortunately, this approach won't work because it does the same thing as `adb shell setprop persist.sys.locale ${locale}` witch means it works only as long as locales are not set in `/data/system/users/0/settings_system.xml` (system settings) witch will override the persist.sys.locale prop. Moreover the approach with emulator cli tool does not provide significant benefits as it still turns on the device before reseting it agin to load changes.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -164,6 +247,24 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+    this.emulatorProcess?.kill(9);
+    await this.internalBootDevice();
+  }
+
+  async bootDevice(deviceSettings: DeviceSettings): Promise<void> {
+    await this.internalBootDevice();
+
+    if (await this.shouldUpdateLocale(deviceSettings.locale)) {
+      await this.changeLocale(deviceSettings.locale);
+      await this.resetDevice();
```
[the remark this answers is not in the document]
filip131311: It is neccesery to boot it twice i added a coment explaining that, in short we use adb shell to change locale so device needs to be booted. The solution you provided (`-change-locale`) unfortunately has the same short comings as just using `adb shell setprop persist.sys.locale` and moreover boots device twice anyway (performs cold boot after changing locale).

packages/vscode-extension/src/webview/views/DeviceLocalizationView.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const handleSearchChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const newSearchPhrase = event.target.value;
+    setSearchPhrase(newSearchPhrase);
+  };
+
+  return (
+    <div className="localization-container">
+      <div className="search-bar">
+        <input
```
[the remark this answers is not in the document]
filip131311: I already did that, there is a bug in the frontend that persist focus on deviceSettingsDropdown after it was closed, but it is outside of scope for this PR so I'll look into it in another one


here is an example of how the component works when deviceSettingsDropdown is not focused: 
https://github.com/user-attachments/assets/b31fd863-c4c6-4062-8913-b20004683d00

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
filip131311: The easiest way to reproduce this problem is to on device go to settings>system>languages and add more then one language that will ensure that `/data/system/users/0/settings_system.xml` has system_locales field

packages/vscode-extension/src/webview/views/DeviceLocalizationView.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const onCancel = () => {
+    openModal("Localization", <DeviceLocalizationView />);
+  };
+
+  return (
+    <div className="localization-change-wrapper">
+      <h2 className="localization-change-title">
+        Are you sure you want to change the localization to <i>{locale.Description}</i>?
+      </h2>
+      <p className="localization-change-subtitle">This action will reset the device.</p>
```
kmagiera, line 109: "reset" is a wrong wording here. This may indicate we want to clear the device settings or something while we only want to reboot the device. I'd rather state it
"Changing localization requires rebooting the simulator"

kmagiera (approved):
Added small comment about wording in the popup dialog. Otherwise looks good 👍

PR #535, opened 2024-09-11
Add test app for react native 76

filip131311:
This PR replaces render in `babel_transformer.js` in order for inspector to work and adds test app for react native 76 in version rc 0 as it is the newest one at the moment.


kmagiera (approved):

kmagiera:
We need a change in `babel_transformer.js` for the inspector to work and likely #530 to be merged in order for the debugger to work

PR #542, opened 2024-09-13
Fix alternative xcode paths

filip131311:
This PR allows non standard xcode setups.

##### How to test: 
- change the name of xcode.app in your application directory
- update xcode-select using `xcode-select --switch ${PathToYourXCode}/Contents/Developer`
- run ide

### DIFF
--- a/packages/docs/docs/troubleshooting.md
+++ b/packages/docs/docs/troubleshooting.md
@@ -83,4 +83,4 @@ If you need to install an older version of an IDE, you can do so by navigating t
 
 ### -sec-num- Configureing Alternative Xcode Versions
 
-If you are using alternative Xcode version ( e.g. xcode-beta, "xcodes" ) IDE will only work if xcode-select points to the correct directory to set it up run: `xcode-select --switch ${PathToYourXCode}/Contents/Developer`
+If you are using alternative Xcode version ( e.g. xcode-beta, ["xcodes"](https://www.xcodes.app/) IDE will only work if xcode-select points to the correct directory to set it up run: `xcode-select --switch ${PathToYourXCode}/Contents/Developer`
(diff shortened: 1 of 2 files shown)


packages/docs/docs/troubleshooting.md
```diff
@@ -80,3 +80,7 @@ Here is what you can try when the extension got stuck on some errors:
 If you need to install an older version of an IDE, you can do so by navigating to the cogwheel menu next to the "install" button in the market place. 
 
 <img width="698" alt="download-older-version" src="/img/docs/marketplace_install_older_version.png"/>
+
+### -sec-num- Configureing Alternative Xcode Versions
+
+If you are using alternative Xcode version ( e.g. xcode-beta, "xcodes" ) IDE will only work if xcode-select points to the correct directory to set it up run: `xcode-select --switch ${PathToYourXCode}/Contents/Developer`
```
kmagiera, line 86: can we add link to xcodes here such that it is clear for people to understand what we refer to here in case they don't know what they use.

Also: have you tried xcodes here? It seems like it has `select` command so I'd expect that it sets the value with xcode-select, otherwise building swift/xcode projects from command line wouldn't work? Regardless of that I think we can keep this in the instruction, but it'd be good to know whether we should expect this to work when someone is using xcodes or not

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -23,7 +23,19 @@ export class Preview implements Disposable {
...
 
     Logger.debug(`Launch preview ${simControllerBinary} ${this.args}`);
-    const subprocess = exec(simControllerBinary, this.args, { buffer: false });
+
+    let simControllerBinaryEnv: { DYLD_FRAMEWORK_PATH: string } | undefined;
+
+    if (Platform.OS === "macos") {
+      const { stdout } = await exec("xcode-select", ["-p"]);
+      const DYLD_FRAMEWORK_PATH = path.join(stdout, "Library", "PrivateFrameworks");
+      simControllerBinaryEnv = { DYLD_FRAMEWORK_PATH };
```
kmagiera, line 32: can we add some logging here, would be good for troubleshooting in case xcode-select prints out some trash data

kmagiera (approved):
Looks good. Please address inline comments prior to merging

PR #543, opened 2024-09-13
Use touchX command with extra args to dispatch touch events to sim-server

kmagiera:
This PR builds on top of changes from https://github.com/software-mansion-labs/simulator-server/pull/173 where we add support for multitouch events to be dispatched to device (Android for the time being).

This change restructures the multitouch handling code both on the front-end and in extension code such that we follow the new protocol. Now, instead of having separate command for multitouch, we allow for the dispatchTouch method to take an array of touches (typically one or two). Then, we send `touchDown` or similar command with additional coordinates after the x and y of the first pointer are added.

The main change on the frontend was that we don't send leading point and anchor points but instead send two concrete touch points,. This removes coupling because sim-server and frontend code as otherwise we'd need to have code that calculates touch points on both sides (including logic related to rounding etc)

This PR also updates some parts of the rpc code that I needed for one stage of the changes since we werent handling `null` value correctly (which apparently has typeof = 'object' in JS)

## Test plan

This has been tested on custom sim-server build that includes https://github.com/software-mansion-labs/simulator-server/pull/173 by opening an app with zoomable control on Android (I used photo gallery)

### DIFF
--- a/packages/vscode-extension/src/devices/preview.ts
+++ b/packages/vscode-extension/src/devices/preview.ts
@@ -52,8 +52,8 @@ export class Preview implements Disposable {
   }
 
   public sendTouches(touches: Array<TouchPoint>, type: "Up" | "Move" | "Down") {
-    const fff = `touch${type} ` + touches.map((pt) => `${pt.xRatio} ${pt.yRatio}`).join(" ") + "\n";
-    this.subprocess?.stdin?.write(fff);
+    const touchesCoords = touches.map((pt) => `${pt.xRatio} ${pt.yRatio}`).join(" ");
+    this.subprocess?.stdin?.write(`touch${type} ${touchesCoords}\n`);
   }
 
   public sendKey(keyCode: number, direction: "Up" | "Down") {
--- a/packages/vscode-extension/src/devices/preview.ts
+++ b/packages/vscode-extension/src/devices/preview.ts
@@ -51,18 +51,9 @@ export class Preview implements Disposable {
     });
   }
 
-  public sendTouch(
-    touchPoint: TouchPoint,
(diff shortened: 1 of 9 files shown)


packages/vscode-extension/src/devices/preview.ts
```diff
@@ -50,20 +51,9 @@ export class Preview implements Disposable {
...
-    xAnchorRatio: number,
-    yAnchorRatio: number,
-    type: "Up" | "Move" | "Down"
-  ) {
-    // this.subprocess?.stdin?.write(
-    //   `multitouch${type} ${xRatio} ${yRatio} ${xAnchorRatio} ${yAnchorRatio}\n` // TODO set proper multitouch simserver command
-    // );
+  public sendTouches(touches: Array<TouchPoint>, type: "Up" | "Move" | "Down") {
+    const fff = `touch${type} ` + touches.map((pt) => `${pt.xRatio} ${pt.yRatio}`).join(" ") + "\n";
+    this.subprocess?.stdin?.write(fff);
```
p-malecki, line 56: Maybe change `fff` to something more meaningful?
```suggestion
    const touchCommand = `touch${type} ` + touches.map((pt) => `${pt.xRatio} ${pt.yRatio}`).join(" ") + "\n";
    this.subprocess?.stdin?.write(touchCommand);
```

p-malecki (approved):

packages/vscode-extension/src/devices/preview.ts
kmagiera: sorry, this one got in by accident. Was debugging it and then forgot to inline it back again. Will change that

PR #560, opened 2024-09-18
Modify line reader to always consume stderr and mark stderr in callback

kmagiera:
This PR modifies the behavior of lineReader which would previously only tried on reading from stderr when requested.

Since typically stderr is used to pass on important information and given almost all uses in out codebase would set stderr as `true`, this method now connects to stderr by default. 

For the use-cases where we want to be able to tell whether line comes from stderr or stdout (like with the sim-server stdout that is used as communication and control channel), the line callback gets an additional argument that indicates whether line comes from stderr.

## Test plan

Run expo-router, try clean rebuild on Android and iOS

### DIFF
--- a/packages/vscode-extension/src/devices/preview.ts
+++ b/packages/vscode-extension/src/devices/preview.ts
@@ -37,7 +37,12 @@ export class Preview implements Disposable {
 
       const streamURLRegex = /(http:\/\/[^ ]*stream\.mjpeg)/;
 
-      lineReader(subprocess).onLineRead((line) => {
+      lineReader(subprocess).onLineRead((line, stderr) => {
+        if (stderr) {
+          Logger.warn("sim-server err:", line);
+          return;
+        }
+
         const match = line.match(streamURLRegex);
 
         if (match) {
@@ -46,7 +51,7 @@ export class Preview implements Disposable {
           this.streamURL = match[1];
           resolve(this.streamURL);
         }
-        Logger.debug("Preview server:", line);
+        Logger.debug("sim-server out:", line);
       });
     });
   }
--- a/packages/vscode-extension/src/devices/preview.ts
+++ b/packages/vscode-extension/src/devices/preview.ts
@@ -39,19 +39,19 @@ export class Preview implements Disposable {
 
       lineReader(subprocess).onLineRead((line, stderr) => {
         if (stderr) {
-          Logger.warn("sim-server err:", line);
+          Logger.info("sim-server:", line);
           return;
         }
 
         const match = line.match(streamURLRegex);
 
         if (match) {
-          Logger.debug(`Preview server ready ${match[1]}`);
+          Logger.info(`Stream ready ${match[1]}`);
 
           this.streamURL = match[1];
           resolve(this.streamURL);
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/utilities/subprocess.ts
```diff
@@ -62,26 +63,25 @@ function overrideEnv<T extends execa.Options>(options?: T): T | undefined {
  * When using this methid, the subprocess should be started with buffer: false option
  * as there's no need for allocating memory for the output that's going to be very long.
  */
-export function lineReader(childProcess: ExecaChildProcess<string>, includeStderr = false) {
-  const input = childProcess.stdout;
-  if (!input) {
-    throw new Error("Child process has no stdout");
+export function lineReader(childProcess: ExecaChildProcess<string>) {
+  if (!childProcess.stdout) {
+    throw new Error("Child process doesn't have stdout");
```
filip131311, line 68: is it possible for the process to not have stdout but have stderr? If so and if we decide that stderr should be passed on by default, maybe it would be useful to allow line reader to exist without stdout?

filip131311 (approved):
approved, left one question inline

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -37,7 +37,12 @@ export class Preview implements Disposable {
 
       const streamURLRegex = /(http:\/\/[^ ]*stream\.mjpeg)/;
 
-      lineReader(subprocess).onLineRead((line) => {
+      lineReader(subprocess).onLineRead((line, stderr) => {
+        if (stderr) {
+          Logger.warn("sim-server err:", line);
+          return;
+        }
```
balins, line 44: I'm a bit skeptical about this, simserver runs a few processes concurrently and one of them can log to `stderr` a thing that is unrelated to the `stdout` output that is awaited by the IDE in the meantime. Loggers by default output to `stderr`, even if a message is for informational/debugging purposes, so I wouldn't assume that something appearing on `stderr` means an error occured.

What I'd suggest is expecting the communication from simserver on `stdout` (responses to commands passed on `stdin` like `preview_ready`, `replay_ready`, `replay_error`) and ignore `stderr`, leaving it for simserver's logging purposes. It would require no changes in this repo, on simserver's side I'd bring back the default logging to stdout and replace `log::info!` calls with `println!` when a message is a command response.

packages/vscode-extension/src/utilities/subprocess.ts
kmagiera: This is only possible if the process is configured to pipe stdout. We typically use lineReader to read from stdout hence it crashes if stdout is not present. If we ever need to reac stderr and not stdout we may need to change this function to support such usecase.

packages/vscode-extension/src/devices/preview.ts
kmagiera: I understand, however it is still important that we read stderr and forward that to the logs. For example the recent crash that we were seeing due to unresolved shared library was something that got printed out by the system into stderr. We didn't see that in the extension logs at all because of that.

PR #561, opened 2024-09-18
Adjust for new simserver CLI, update simserver submodule

balins:
This PR changes the commands that are sent to simserver to comply with new format.


kmagiera (approved):
This PR needs to udpate sim-server submodule to the revision that includes the changes. Otherwise it'll use the old version with incompatible commands

balins:
Updated the submodule to the latest version including running Android emulator without Xcode on macOS and separated stderr and stdout output.

PR #562, opened 2024-09-18
Migrate cashes directories

filip131311:
This PR changes the location of caches directory for all new devices, old devices remain in old cache location and can still be used and removed by the user by the user.


kmagiera (approved):
Looks good 👍

PR #563, opened 2024-09-19
Improve a flow of loading devices and xcode detection

filip131311:
this PR addresses a few issues with the loading devices  flow: 
- `checkIfCLIInstalled` no longer logs errors when returning false instead just logs with debug level, this change makes the logs more alined with expected results for setups without xcode
- initial device load is performed with force, as xcode installation could change between session of the ide
- initial loading screen no longer "jumps" between states 
- removed `DYLD_FRAMEWORK_PATH` as it is no longer needed
- `trySelectingInitialDevice()` creates "devicesChanged" listeners before selectingInitial device is complete, so it catches cases in which `lastDeviceId` points to the device that no longer exists (e.g. because of xcode update)

### DIFF
--- a/packages/vscode-extension/src/dependency/DependencyManager.ts
+++ b/packages/vscode-extension/src/dependency/DependencyManager.ts
@@ -166,7 +166,7 @@ export class DependencyManager implements Disposable {
 
     await command(installationCommand, {
       cwd: workspacePath,
-      quiet: true,
+      quietErrorsOnExit: true,
     });
 
     this.webview.postMessage({
@@ -425,7 +425,11 @@ export class DependencyManager implements Disposable {
 export async function checkIfCLIInstalled(cmd: string, options: Record<string, unknown> = {}) {
   try {
     // We are not checking the stderr here, because some of the CLIs put the warnings there.
-    const { stdout } = await command(cmd, { encoding: "utf8", quiet: true, ...options });
+    const { stdout } = await command(cmd, {
+      encoding: "utf8",
+      quietErrorsOnExit: true,
+      ...options,
+    });
     const result = stdout.length > 0;
     Logger.debug(`CLI: ${cmd} ${result ? "" : "not"} installed `);
     return result;
--- a/packages/vscode-extension/src/dependency/DependencyManager.ts
+++ b/packages/vscode-extension/src/dependency/DependencyManager.ts
@@ -198,12 +198,9 @@ export class DependencyManager implements Disposable {
 
   /* iOS-related */
   public async checkXcodeInstalled() {
-    const isXcodebuildInstalled = await checkIfCLIInstalled("xcodebuild -version");
-    const isXcrunInstalled = await checkIfCLIInstalled("xcrun --version");
-    const isSimctlInstalled = await checkIfCLIInstalled("xcrun simctl help");
-    const installed = isXcodebuildInstalled && isXcrunInstalled && isSimctlInstalled;
+    const installed = await checkXcodeExists();
     const errorMessage =
-      "Xcode was not found. [Install Xcode from the Mac App Store](https://apps.apple.com/us/app/xcode/id497799835?mt=12) and have Xcode Command Line Tools enabled.";
+      "Xcode was not found. If you are using alternative Xcode version you can find out more in troubleshooting section of our documentation. Otherwise, [Install Xcode from the Mac App Store](https://apps.apple.com/us/app/xcode/id497799835?mt=12) and have Xcode Command Line Tools enabled.";
     this.webview.postMessage({
       command: "isXcodeInstalled",
       data: {
@@ -428,9 +425,12 @@ export class DependencyManager implements Disposable {
 export async function checkIfCLIInstalled(cmd: string, options: Record<string, unknown> = {}) {
   try {
     // We are not checking the stderr here, because some of the CLIs put the warnings there.
-    const { stdout } = await command(cmd, { encoding: "utf8", ...options });
-    return !!stdout.length;
+    const { stdout } = await command(cmd, { encoding: "utf8", quiet: true, ...options });
+    const result = !!stdout.length;
+    Logger.debug(`CLI: ${cmd} ${result ? "" : "not"} installed `);
+    return result;
   } catch (_) {
+    Logger.debug(`CLI: ${cmd} not installed `);
     return false;
   }
 }
@@ -459,6 +459,13 @@ export function checkMinDependencyVersionInstalled(dependency: string, minVersio
     Logger.debug(message, "Module not found.");
     return "not_installed";
   }
+}
+
+export async function checkXcodeExists() {
+  const isXcodebuildInstalled = await checkIfCLIInstalled("xcodebuild -version");
+  const isXcrunInstalled = await checkIfCLIInstalled("xcrun --version");
+  const isSimctlInstalled = await checkIfCLIInstalled("xcrun simctl help");
+  return isXcodebuildInstalled && isXcrunInstalled && isSimctlInstalled;
 }
 
 export async function checkAndroidEmulatorExists() {
--- a/packages/vscode-extension/src/dependency/DependencyManager.ts
+++ b/packages/vscode-extension/src/dependency/DependencyManager.ts
@@ -426,7 +426,7 @@ export async function checkIfCLIInstalled(cmd: string, options: Record<string, u
   try {
     // We are not checking the stderr here, because some of the CLIs put the warnings there.
     const { stdout } = await command(cmd, { encoding: "utf8", quiet: true, ...options });
-    const result = !!stdout.length;
+    const result = stdout.length > 0;
     Logger.debug(`CLI: ${cmd} ${result ? "" : "not"} installed `);
     return result;
   } catch (_) {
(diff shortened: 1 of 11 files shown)


packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -428,9 +425,12 @@ export class DependencyManager implements Disposable {
 export async function checkIfCLIInstalled(cmd: string, options: Record<string, unknown> = {}) {
   try {
     // We are not checking the stderr here, because some of the CLIs put the warnings there.
-    const { stdout } = await command(cmd, { encoding: "utf8", ...options });
-    return !!stdout.length;
+    const { stdout } = await command(cmd, { encoding: "utf8", quiet: true, ...options });
```
jakub-gonet, line 428: `quiet` name needs to change eventually to something like `printStdErr`, I'd assume if quiet is true we only get status code on exit.

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -428,9 +425,12 @@ export class DependencyManager implements Disposable {
...
+    const result = !!stdout.length;
```
jakub-gonet, line 429: Please be more explicit:
```
const hasOutput = stdout.lenght > 0;
```

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -428,9 +425,12 @@ export class DependencyManager implements Disposable {
...
+    Logger.debug(`CLI: ${cmd} ${result ? "" : "not"} installed `);
```
jakub-gonet, line 430: Nit: "CLI: " prefix is unnecessary.

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -125,11 +128,11 @@ export class DeviceManager implements DeviceManagerInterface {
     return getAvailableIosRuntimes();
   }
 
-  public async listAllDevices() {
+  public async listAllDevices(force?: boolean) {
```
jakub-gonet, line 131: Maybe `forceReload`?

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -154,9 +167,10 @@ function PreviewView() {
           onZoomChanged={onZoomChanged}
         />
       ) : (
-        <div className="missing-device-filler">
-          {devicesNotFound ? <DevicesNotFoundView /> : <VSCodeProgressRing />}
-        </div>
+        <LoadingComponent
+          finishedInitialLoad={finishedInitialLoad}
```
jakub-gonet, line 171: Why we're checking for finishedInitialLoad twice in the ternary and in the component?

packages/vscode-extension/src/dependency/DependencyManager.ts
filip131311: I think it is useful to have prefixes that allow to serach what part of our code generated the log. We have. a growing code base and most of that log is generated dynamically so searching for it would be hard without it.

packages/vscode-extension/src/webview/views/PreviewView.tsx
filip131311: there is more then one reason for rendering the loading component, but not all of them behave the same.

packages/vscode-extension/src/dependency/DependencyManager.ts
filip131311, line 429:
```suggestion
    const result = stdout.length > 0;
```

packages/vscode-extension/src/devices/DeviceManager.ts
filip131311, line 131:
```suggestion
  public async listAllDevices(forceReload?: boolean) {
   const devices = extensionContext.globalState.get(DEVICE_LIST_CACHE_KEY) as
      | DeviceInfo[]
      | undefined;
    if (devices && !forceReload) {
```

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -104,13 +105,15 @@ export class DeviceManager implements DeviceManagerInterface {
...
     });
-    const simulators =
-      Platform.OS === "macos"
-        ? listSimulators().catch((e) => {
-            Logger.error("Error fetching simulators", e);
-            return [];
-          })
-        : Promise.resolve([]);
+
+    const shouldLoadSimulators = Platform.OS === "macos" && (await checkXcodeExists());
```
kmagiera, line 109: Maybe we can add additional log here when checkXcodeExists fails. For example "Couldn't list iOS simulators as XCode installation wasn't found" – this will allow us to track issues in case those checks aren't accurate

packages/vscode-extension/src/webview/providers/ProjectProvider.tsx
```diff
@@ -26,6 +26,7 @@ const ProjectContext = createContext<ProjectContextProps>({
       longitude: 19.965474,
       isDisabled: false,
     },
+    locale: "en_US",
```
kmagiera, line 29: why is this change here?

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -125,11 +128,11 @@ export class DeviceManager implements DeviceManagerInterface {
     return getAvailableIosRuntimes();
   }
 
-  public async listAllDevices() {
+  public async listAllDevices(forceReload?: boolean) {
     const devices = extensionContext.globalState.get(DEVICE_LIST_CACHE_KEY) as
       | DeviceInfo[]
       | undefined;
-    if (devices) {
+    if (devices && !forceReload) {
```
kmagiera, line 135: this isn't accurate as we *always* load the devices. The difference is that we return the updated list or let it update asynchronously. I'm a bit hesitant about this change as from what I remember, we added device caching as it was slowing down the initial startup significantly. As it isn't documented in PR description I wonder why you're changing this behavior?

packages/vscode-extension/src/webview/providers/ProjectProvider.tsx
filip131311: It was missed before and ts was not happy, it does not effect the webview much at all as it never uses default context.

packages/vscode-extension/src/devices/DeviceManager.ts
filip131311: When user had an iPhone selected when they last used IDE, but something happend to xcode (e.g. user started using beta but not updated the `xcode-select`)  application would select an iPhone anyway as it is the most recent device and would get stuck on loading screen. 

What dou you think about `forceLoad` instead of `forceReload`

kmagiera: I think we should optimize for the most common case in which people don't change their xcode location. If there are cases when it happens, I understand they will get an error when trying to load the simulator, and by that time we will have the list refetched. If that's not the case we should adjust the logic such that the worst thing that can happen if you mess up the simjulator cache or xcode is that you get an error loading the simulator instead of slowing down all the time to interact for everyone else.

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -104,13 +105,20 @@ export class DeviceManager implements DeviceManagerInterface {
...
-      Platform.OS === "macos"
-        ? listSimulators().catch((e) => {
-            Logger.error("Error fetching simulators", e);
-            return [];
-          })
-        : Promise.resolve([]);
+
+    let shouldLoadSimulators = Platform.OS === "macos";
+
+    if (!(await checkXcodeExists())) {
```
kmagiera, line 111: do we even want to run those checks if we're not on a Mac? 

Perhaps doing it this way would be cleaner:
```ts
const shouldLoadSimulators = Platform.OS === "macos" && (await checkXcodeExists());
```

Then you can move the log to be printed in `checkXcodeExists`

packages/vscode-extension/src/project/project.ts
```diff
@@ -201,13 +200,21 @@ export class Project
...
+        return;
+      }
+      if (await selectInitialDevice(newDevices)) {
+        this.deviceManager.removeListener("devicesChanged", listener);
+      }
+    };
+    this.deviceManager.addListener("devicesChanged", listener);
+
+    if (await selectInitialDevicePromise) {
+      this.deviceManager.removeListener("devicesChanged", listener);
```
kmagiera, line 217: This code doesn't do the same thing as the one on the left. In general I'm most afraid of the changes in awaiting that you're adding here. Not saying we shouldn't have them, but they require some explanation in the PR description, such that we can understand why we're changing the await flow.

This part is the most complex change and it introduces some unusual pattern (that we, albeit, use in other places but typically with some explanation) where we create promise and then await inside of an arrow function.

I think that given this method already awaits `selectDevice` (which wasn't the case before this change) , we can just add the listener before calling `selectInitialDevice` ?

kmagiera (commented):
Looks mostly good, left a few comments inline

packages/vscode-extension/src/devices/DeviceManager.ts
filip131311: I did this in this way to address your previous comment about adding logging in !checkXcodeExist() scennario

kmagiera (approved):
lgtm 👍

PR #564, opened 2024-09-19
Change multitouch button for windows and fix misc minor bugs

p-malecki:
This PR:
- changes the button that initiates the multitouch gesture to the `control` key on Windows
- fixes the error message display when the `diagnoseWorkspaceStructure` command cannot find the `appRoot` folder
- corrects warning messages in `Logger`

### DIFF
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -377,7 +377,7 @@ async function findAppRootFolder() {
       });
   }
 
-  if (appRootCandidates) {
+  if (appRootCandidates.length === 1) {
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/extension.ts
```diff
@@ -377,7 +377,7 @@ async function findAppRootFolder() {
       });
   }
 
-  if (appRootCandidates) {
+  if (appRootCandidates.length === 1) {
```
filip131311, line 380: why?

filip131311 (changes requested):

filip131311 (approved):

PR #578, opened 2024-09-30
Shared app template

jakub-gonet:
Adds shared template to all test apps. New screen was enabled only for apps that had default code setup. The rest only have infra for using template.

After merging it's needed to run `npm run copy-shared` script to get the apps to work.
I looked into running the script automatically but didn't find an easy way to do that.

If template change frequently, we may add a script that runs it for every defined app. We should add a way to include components at the top or at the bottom of the template, so each app may add its own set of tests (geolocalization, locale, etc)

## How does it work

In `test-apps/` directory there is `shared/` directory with the common UI code for apps. To use this code in the app, use `copy.sh` script.
It should use minimal set of dependencies, ideally nothing except react-native.
If app uses expo-router, we additionally copy `shared/navigation/` directory which uses expo-router and `@expo/vector-icons`.

If app wants to use this shared code, it should add an entry to `.gitignore` for shared directory under its own `src/` and run copy script. If shared code is updated, every app using it should rerun the copy script. 

<details><summary>Screenshot</summary>

</details> 

## Testing

Ran all apps except RN76.

## Followups

- Some way to include custom components in the main screen.


kmagiera (approved):
Overall I'm ok wit these changes. However, would love to see a bit more details in the PR and possibly some changes to the docs explaining this setup.

For example here are things I don't understand that'd be good to have clarified here:
1) does app template have any dependencies exept for RN itself? If so how these are managed? What is allowed when making changes to the app template
2) Follow up on the above and specifically interested in expo-router. Does it mean now all the apps use expo-router?
3) Where is the script you mentioned in the description defined? Is it in every individual app project, meaning we need to run it separately in every project we test?

PR #579, opened 2024-10-01
Add option to capture replays

kmagiera:
This PR adds replays support.

Replays allows the user to capture a recent moment into a video and preview it instantly within the IDE panel.

The current way this feature has been designed is as follows:
 - in device setting you can enable/disable replays functionality. Disabling hides the replays button from the UI
 - we're adding new "replay" button that's in the main IDE panel and can be use to capture the replay and moves the user into the preview mode
 - in preview mode we're displaying a full screen overlay over the IDE panel with video player
 - the overlay has a seekbar, arrows to move forward/backward by one frame, save button, and the video length selector
 - simulator-server uses a fixed size memory buffer for replays, depending on the frame rate (it is variable) it may store a different lengths of the video depending on the content. When replay is captured we show the past 5 seconds by default and allow user to increase that length using length selector as long as the captured video is of a sufficient duration.

Here's what we are adding as part of this PR:
1) New UI element to PreviewView that displays the preview button
2) New UI screen that's responsible for rendering replay preview with all the controls
3) We're extending the webview<>extension API with methods for capturing video replays and for saving video files
4) We're extensing device settings with the new replay setting. Then we use this setting to start/stop recording depending on device session state. We want the recording to start after every device reload, such that we're not recording reload screens etc.

## Test plan

1. Enable replays in device options -> see it changes the appearance of replay button in top-right
2. Tap on replay after a while, the video should appear in an overlay
3. For longer videos, you should be able to change between lengths
4. Updating length once video finished playback should make it play from a certain moment
5. Updating length to shorter when player is at the start of playback, should move to the start of new length
6. Save button should store the full length video. When one is stored with the same name, you'll get a confirmation dialog


filip131311 (changes requested):
While testing I came across a problem with menage devices menu that is probably caused by the overlay:

https://github.com/user-attachments/assets/1d394de6-acac-4697-8d8d-b32b6b4ac37e

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -38,27 +44,95 @@ export class Preview implements Disposable {
...
+          if (handlers && videoReadyMatch) {
+            // match array looks as follows:
+            // [0] - full match
+            // [1] - URL or error message
+            // [2] - File URL
+            const tempFileLocation = videoReadyMatch[2];
+            const ext = path.extname(tempFileLocation);
+            const fileName = workspace.name
+              ? `${workspace.name} RadonIDE replay${ext}`
+              : `RadonIDE-replay${ext}`;
```
p-malecki, line 82: We should separate words with dashes in filename also when condition is true
```suggestion
              ? `${workspace.name}-RadonIDE-replay${ext}`
              : `RadonIDE-replay${ext}`;
```

filip131311:
review notes: 
- Replays do not work when option was selected while iPhone is actively simulated
- the name of saved file should not include whitespaces
- I believe the setting enabling replay should be located in top settings menu as it is not a device setting, but ide setting

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -153,6 +154,24 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
...
+            <span className="icons-container">
+              <span className="codicon codicon-triangle-left icons-rewind" />
+              <span className="codicon codicon-triangle-left icons-rewind" />
+            </span>
+            Enable Replays
+            <Switch.Root
+              className="switch-root small-switch"
+              id="enable-replays"
+              onCheckedChange={(checked) =>
+                project.updateDeviceSettings({ ...deviceSettings, replaysEnabled: checked })
```
filip131311, line 168: I belive this option should be part of workspace settings (we have a separate interface for that called `WorkspaceConfig` as it is not a device setting also it would be nice if this option synchronized between computers for users that are logged in to microsoft account

packages/vscode-extension/src/webview/components/ReplayUI.css
```diff
@@ -0,0 +1,206 @@
...
...2pGpeEVeCCA7b85BO3F9DzxB3cdqvBzWcmzbyMiqhzuYqtHRVG2y4x+KOlnyqla8AoWWpuBoYRxzXrfKuILl6SfiWCbjxoZJUaCBj1CjH7GIaDbc9kqBY3W/Rgjda1iqQcOJu2WW+76pZC9QG7M00dffe9hNnseupFL53r8F7YHSwJWUKP2q+k7RdsxyOB11n0xtOvnW4irMMFNV4H0uqwS5ExsmP9AxbDTc9JwgneAT5vTiUSm1E7BSflSt3bfa1tv8Di3R8n3Af7MNWzs49hmauE2wP+ttrq+AsWpFG2awvsuOqbipWHgtuvuaAE+A1Z/7gC9hesnr+7wqCwG8c5yAg3AL1fm8T9AZtp/bbJGwl1pNrE7RuOX7PeMRUERVaPpEs+yqeoSmuOlokqw49pgomjLeh7icHNlG19yjs6XXOMedYm5xH2YxpV2tc0Ro2jJfxC50ApuxGob7lMsxfTbeUv07TyYxpeLucEH1gNd4IKH2LAg5TdVhlCafZvpskfncCfx8pOhJzd76bJWeYFnFciwcYfubRc12Ip/ppIhA1/mSZ/RxjFDrJC5xifFjJpY2Xl5zXdguFqYyTR1zSp1Y9p+tktDYYSNflcxI0iyO4TPBdlRcpeqjK/piF5bklq77VSEaA+z8qmJTFzIWiitbnzR794USKBUaT0NTEsVjZqLaFVqJoPN9ODG70IPbfBHKK+/q/AWR0tJzYHRULOa4MP+W/HfGadZUbfw177G7j/OGbIs8TahLyynl4X4RinF793Oz+BU0saXtUHrVBFT/DnA3ctNPoGbs4hRIjTok8i+algT1lTHi4SxFvONKNrgQFAq2/gFnWMXgwffgYMJpiKYkmW3tTg3ZQ9Jq+f8XN+A5eeUKHWvJWJ2sgJ1Sop+wwhqFVijqWaJhwtD8MNlSBeWNNWTa5Z5kPZw5+LbVT99wqTdx29lMUH4OIG/D86ruKEauBjvH5xy6um/Sfj7ei6UUVk4AIl3MyD4MSSTOFgSwsH/QJWaQ5as7ZcmgBZkzjjU1UrQ74ci1gWBCSGHtuV1H2mhSnO3Wp/3fEV5a+4wz//6qy8JxjZsmxxy5+4w9CDNJY09T072iKG0EnOS0arEYgXqYnXcYHwjTtUNAcMelOd4xpkoqiTYICWFq0JSiPfPDQdnt+4/wuqcXY47QILbgAAAABJRU5ErkJggg==);
```
filip131311, line 22: could you add more context to that image? where is it from why url is base 64 encoded etc

kmagiera:
> While testing I came across a problem with menage devices menu that is probably caused by the overlay:
> Screen.Recording.2024-10-07.at.17.47.00.mov

Thanks for testing. This seemed to be some transient issue that got resolved after I merged main. Will look into other comments

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
kmagiera: I thought about it, but feel like it fits the device setting pane better for now. 

The workspace settings doesn't have any configuration options except from the IDE panel location for now. Adding it there would make it an isolated option. For me it still feels like its device related because we're recording the device screen in the end. The only thing is that it affects the panel UI as it adds a new button when replays are enabled.

I'd keep it as is for now, and we can consider moving it in the future or once we have more similar options added that'd make it less of a precedence to have a switch button included in the workspace settings dropdown.

kmagiera:
Seems like @p-malecki resolved other minor comments. 

The only part I don't understand is this comment:
> Replays do not work when option was selected while iPhone is actively simulated

Can you provide more details on what exacly is not working here?

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
filip131311: I do not agree, for the following reasons: 
- `WorkspaceConfig` currently includes showDeviceFrame setting that is affecting user ui in a similar way as replay button does, (it is not included in the settings dropdown but it is accessible through vscode settings panel) 
- it does not change anything on the device were as all other device settings do, I see your point that it records device but the recording does not happen on the device itself which means that it won't affect how device works during testing. As a user I would see the device setting menu as a menu i use often while testing my application and all other options as dose i set up only when something goes wrong. 
- this is a setting that is rarely used (probably only once per user) as a User does not have much reason to use it except their preference  which I expect does not change often
- workspace settings persist across devices for logged in user which is a useful functionality that we get for free 
- I'm not even sure if this setting needs to exist in the UI because of how rarely it should be used and it's location in the vscode settings would make it easy to remove it from UI settings if they get to cluttered

filip131311 (approved):

PR #586, opened 2024-10-02
Refactor dependency manager proxy

jakub-gonet:
- replaced vscode event emit calls with listeners and proxies abstracting them
- cleaned up the API of both dependency manager and provider on frontend
- changed incorrect approach to detecting pods change (it should be improved in the future though)
- allowed to package manager be not installed and not crashing the IDE when installing node modules

## Testing

Ran expo-51-dev-client, tested autoinstallation, ran expo-go, tested missing `npm` command.

### DIFF
--- a/packages/vscode-extension/src/builders/BuildManager.ts
+++ b/packages/vscode-extension/src/builders/BuildManager.ts
@@ -73,17 +73,11 @@ export class BuildManager {
         });
         buildResult = await buildIos(
           deviceInfo,
-          getAppRootFolder(),
           forceCleanBuild,
           cancelToken,
           this.buildOutputChannel,
           progressListener,
-          () => {
-            return this.dependencyManager.checkPodsInstalled();
-          },
-          (appRootFolder: string, cleanBuild: boolean, token: CancelToken) => {
-            return this.dependencyManager.installPods(appRootFolder, cleanBuild, token);
-          }
+          this.dependencyManager
         );
       }
       await storeCachedBuild(platform, {
--- a/packages/vscode-extension/src/builders/BuildManager.ts
+++ b/packages/vscode-extension/src/builders/BuildManager.ts
@@ -73,11 +73,17 @@ export class BuildManager {
         });
         buildResult = await buildIos(
           deviceInfo,
+          getAppRootFolder(),
           forceCleanBuild,
           cancelToken,
           this.buildOutputChannel,
           progressListener,
-          this.dependencyManager
+          () => {
+            return this.dependencyManager.isInstalled("pods");
+          },
+          () => {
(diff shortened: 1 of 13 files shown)


packages/vscode-extension/src/builders/BuildManager.ts
```diff
@@ -73,17 +73,11 @@ export class BuildManager {
...
           cancelToken,
           this.buildOutputChannel,
           progressListener,
-          () => {
-            return this.dependencyManager.checkPodsInstalled();
-          },
-          (appRootFolder: string, cleanBuild: boolean, token: CancelToken) => {
-            return this.dependencyManager.installPods(appRootFolder, cleanBuild, token);
-          }
+          this.dependencyManager
```
kmagiera, line 80: this will conflict with my changes from #587
Maybe you can revert this and changes from `buildIOS` as they seem out of scope of this PR

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -1,440 +1,264 @@
...
       encoding: "utf8",
       quietErrorsOnExit: true,
-      ...options,
+      env: { ...process.env, LANG: "en_US.UTF-8" },
     });
-    const result = stdout.length > 0;
-    Logger.debug(`CLI: ${cmd} ${result ? "" : "not"} installed `);
-    return result;
+    const installed = stdout.length > 0;
+    return installed;
```
kmagiera, line 260: shouldn't we check exit code instead of testing whether output is non-empty?

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -1,440 +1,264 @@
...
-        error: undefined,
-      },
-    });
-    Logger.debug("Project pods installed");
+    return "notInstalled";
   }
 }
 
-export async function checkIfCLIInstalled(cmd: string, options: Record<string, unknown> = {}) {
+async function isCommandInstalled(cmd: string) {
```
kmagiera, line 251: I'd  just rename to `testCommand`,`probeCommand` or `checkCommand` as the method doesn't verify whether something is installed but only checks whether the provided command runs sucesfully

kmagiera (approved):
The small changesin BuildManager and buildIOS will cause conflicts with #587 but apart from minor renaming comments this looks good.

packages/vscode-extension/src/dependency/DependencyManager.ts
jakub-gonet: I was wondering the same thing but decided to not touch this code – `xcrun simctl` is a subcommand that cannot be tested easily. If we can drop it then we can only test exit code.

PR #587, opened 2024-10-02
Fix ios rebuild infinite cycle when opening IDE panel

kmagiera:
This PR fixes an issue with the IDE constantly attempting to rebuild an app for iOS on every start of the IDE (or when switching between devices).

The issue was there because of failing pods check, which was triggering pods installation on every build. This on its own wasn't enough to trigger the cycle. On top of the above, the pods install was affecting the fingerprint of the workspace by generating xcworkspace and other files. As a consequence, when caching iOS builds, we'd use a stale fingerprint from before the pods installation. Because of that, on the subsequent restarts, we'd always detect fingerprint change and try to rebuild the app which in turn would again store the build under stale fingerprint cache key.

The above issue has been resolved by the following:
1) Upgrading @expo/fingerprint library – the new version no longer detects the change as long as pod install was not making any actual changes (the previous version would check the modification date of files in pods directory I believe which remained unchanged)
2) Despite the main source being addressed by the fingerprint update, we still need to consider the scenario in which the fingerprint changes due to pods. This can happen when new dependency is added that requires autolinking. Therefore we are changing the logic such that we update the current fingerprint after pods installation (when triggered).

This PR also resolves some other build related issues:
1) We're updating xcodebuild command to use `arch` and `sdk` options instead of relying on temporary simulator created with the exact same settings. I validated that properly setting these options doesn't impact the build time and that when using `-showBuildSettings` options, the xcode generates the same settings in both cases. We want to avoid having temporary simulators as they'd sometimes appear in xcode UI.
2) We're updating some places where logging wasn't accurate. For example, we only print "Installing node_modules" when we're actually triggering the installation rather than when checking if installation is necessary and then optionally installing.
3) We're reporting build times now to logger so that we can better diagnose similar issues in the future
4) We're also adding build time reports option to xcode builds for the similar reason.

## Test plan
1) With expo-router, try running `pod install` from terminal. Then open project in IDE – this will trigger native build. Then subsequent re-opening of the project should no longer trigger build and there's no fingerprint mismatch in logs.
2) Verify the clean build times are the same with previous and new commands (for me they both take around 90seconds)

### DIFF
--- a/packages/vscode-extension/src/builders/BuildManager.ts
+++ b/packages/vscode-extension/src/builders/BuildManager.ts
@@ -43,7 +43,7 @@ export class BuildManager {
     const cancelToken = new CancelToken();
 
     const buildApp = async () => {
-      const newFingerprint = await generateWorkspaceFingerprint();
+      let newFingerprint = await generateWorkspaceFingerprint();
       if (forceCleanBuild) {
         // we reset the cache when force clean build is requested as the newly
         // started build may end up being cancelled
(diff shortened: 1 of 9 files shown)


packages/vscode-extension/src/builders/BuildManager.ts
```diff
@@ -71,19 +71,25 @@ export class BuildManager {
...
           log: true,
         });
+        const appRootFolder = getAppRootFolder();
+        const installPodsIfNeeded = async () => {
+          const podsInstalled = await this.dependencyManager.checkPodsInstalled();
+          if (!podsInstalled) {
+            Logger.info("Pods installation is missing or outdated. Installing Pods.");
+            // installing pods may impact the fingerprint as new pods may be created under the project directory
+            // for this reason we need to recalculate the fingerprint after installing pods
+            await this.dependencyManager.installPods(appRootFolder, cancelToken);
```
jakub-gonet, line 81: I wonder if it would be easier if we moved this function to be inside dependency manager (same goes for installNodeModules(ifNeeded)) and returned new fingerprint.

jakub-gonet (approved):

packages/vscode-extension/src/builders/BuildManager.ts
kmagiera: I'd keep it as is for now. We only call it from one place right now

PR #589, opened 2024-10-03
Add custom fingerprinting and build cache

jakub-gonet:
- Extracts logic responsible to writing build info from build manager so we always use correct fingerprints
- Replaces `buildScript` in configuration with `customBuild` object that allows using custom build and fingerprinting scripts

## Testing

Checked force rebuilds, closing IDE while building, and detecting changed fingerprints when file change for regressions.
Checked custom fingerprints when closing and opening IDE and when project files change (installing new native dependency, changing fingerprint and saving file).


packages/vscode-extension/src/project/project.ts
```diff
@@ -585,17 +585,22 @@ export class Project
...
+      // instance every time
+      const platform = this.projectState.selectedDevice.platform;
+      const buildCache = new PlatformBuildCache(platform, cancelToken);
+
+      if (await buildCache.isCacheInvalidated()) {
+        this.buildCacheInvalidated = true;
         this.eventEmitter.emit("needsNativeRebuild");
       }
     }
   }, 300);
```
jakub-gonet, line 600: Build cache shouldn't live here but I considered few options and all of them made structure more complex.

packages/vscode-extension/src/builders/PlatformBuildCache.ts
```diff
@@ -0,0 +1,152 @@
...
+        return build;
+      }
+    } catch (e) {
+      // we only log the error and ignore it to allow new build to start
+      Logger.error("Error while attempting to load cached build: ", e);
+      return undefined;
+    }
+  }
+
+  public async isCacheInvalidated() {
```
kmagiera, line 97: "invalidating" here is not the right term. It refers to the fact of cleaning the cached data. In this case, it'd be checking if the stored cache key still exists. This method however, checks whether the stored cache is up to date. I'd rename to `hasUpToDateBuild` or something along these lines

packages/vscode-extension/src/builders/PlatformBuildCache.ts
```diff
@@ -0,0 +1,152 @@
...
+  buildHash: string;
+  buildResult: AndroidBuildResult | IOSBuildResult;
+};
+
+export class PlatformBuildCache {
+  private cacheKey: string;
+
+  constructor(
+    private readonly platform: DevicePlatform,
+    private readonly cancelToken: CancelToken
```
kmagiera, line 39: cancel tokens are transient and shouldn't be kept like that. Their purpose is to be able to pass them to async methods, such that they can be intercepted. Storing cancelToken like that defeats this purpose, because who is now in the control of cancelling it? 

We should likely pass cancelTokens to methods that needs them directly, such that the caller is responsible for cancelling. 
Alternatively we can create new cancel tokens for the calls that require them and expose some cancel method from this class which may be called by the caller.

packages/vscode-extension/src/project/project.ts
```diff
@@ -585,17 +585,22 @@ export class Project
...
-  private checkIfNativeChanged = throttle(async () => {
-    if (!this.detectedFingerprintChange && this.projectState.selectedDevice) {
-      if (await didFingerprintChange(this.projectState.selectedDevice.platform)) {
-        this.detectedFingerprintChange = true;
+  private checkIfNativeChanged = throttle(async (cancelToken: CancelToken) => {
+    if (!this.buildCacheInvalidated && this.projectState.selectedDevice) {
+      // TODO: should be refactored to not create a new PlatformBuildCache
+      // instance every time
+      const platform = this.projectState.selectedDevice.platform;
+      const buildCache = new PlatformBuildCache(platform, cancelToken);
```
kmagiera, line 593: responding to your comment here: maybe we could have a static method in `PlatformBuildCache.get(platform)` that returns an instance. This way we can hide the fact it is a stateless object and avoid constructing it in multiple places

kmagiera (changes requested):

packages/vscode-extension/src/project/project.ts
jakub-gonet: This works quite nicely but we now split getting fingerprint and storing the build (which is correct) so the value of the class is reduced. I think it's still marginally better than standalone functions but not by much.

packages/docs/docs/launch-configuration.md
```diff
@@ -129,15 +129,23 @@ Below is an example of how the `launch.json` file could look like with android v
...
-The requirement for scripts is to output the absolute path to the built app as the
-last line of the standard output.
-
-Both `buildScript` and `eas` are objects having `ios` and `android` optional
+The requirement for scripts is to output the absolute path to the built app as
+the last line of the standard output. If custom fingerprint script is used, it
+should output fingerprint as the last line of the standard output. When
+fingerprint changes between invocations, RN IDE will rebuild the project. The
+IDE runs fingerprint quite frequently (i.e., on every file save), so this
+process should be fast and avoid over the network communication.
```
kmagiera, line 137: We should explain the concept of fingerprint here. Specifically that it is a string that allows us to tell whether new native build needs to be run

packages/docs/docs/launch-configuration.md
```diff
@@ -146,8 +154,12 @@ command used to build the app. Example below:
       "type": "radon-ide",
       "request": "launch",
       "name": "Radon IDE panel",
-      "buildScript": {
-        "android": "npm run build:ftp-fetch-android"
+      "customBuild": {
+        "android": { "buildScript": "npm run build:ftp-fetch-android" }
```
kmagiera, line 158: This isn't a script, so maybe we should rename to `buildCommand`? On the other hand in `package.json` you also have `scripts` sections where you can put commands and not just scripts.

packages/vscode-extension/src/project/project.ts
```diff
@@ -586,9 +586,14 @@ export class Project
   };
 
   private checkIfNativeChanged = throttle(async () => {
-    if (!this.detectedFingerprintChange && this.projectState.selectedDevice) {
-      if (await didFingerprintChange(this.projectState.selectedDevice.platform)) {
-        this.detectedFingerprintChange = true;
+    if (!this.isCachedBuildStale && this.projectState.selectedDevice) {
+      // TODO: should be refactored to not create a new PlatformBuildCache
+      // instance every time
```
kmagiera, line 591: Is this still relevant? We're not creating it here directly now

kmagiera (approved):
Accepting but added some comments re. the naming. You may want to address those prior to merging.

PR #590, opened 2024-10-03
Add Localization on Android

filip131311:
This PR adds localization on android. 

https://github.com/user-attachments/assets/f077b194-d79e-4ca3-bdd6-806536d1627b

#### How Android changes the locale settings when the user does it in the settings app?
Android changes the memory responsible for holding information about locale and tigers a change in two types of persistent storage, one is system prop `persist.sys.locale` and the second one is `system_locales` prop in system level settings. 
Before user changes the settings both of those fields are empty. When Android boots it either sets a locale to the default one or to the one specified in one of the persistent locations, it seems that if only one location exist android will still retrive the information, but conflict between `persist.sys.locale` and `system_locales` leads to unpredictable behavior. 
#### How to change the properties with adb? 
adb allows us to change system properties including `persist.sys.locale` granted we have a root access, this is achieved either by using `adb root` command or `su` command inside shell. The root access is only possible on emulators without GMS (google mobile services). adb also allows us to chage system settings using `adb shell settings` command and it does not require root access, but permanent state change of system_settings require `-writable-system` flag to be added on emulator boot, so we do it. 
#### How emulator command changes settings when using `-change-locale` option? 
Underneath emulator calls `adb shell su 0 setprop persist.sys.locale ${locale}`.
#### Uncovered case: 
Because of the above I did not find a way to reliably change `persist.sys.locale` and scenario in which user manually changes locale (in the settings app).

### Test Plan: 
1) Changing localization while using the device
- turn on IDE and select android device 
- check if localization settings are disabled during boot process: 
Screenshot 2024-10-11 at 18 13 35

- change devices localization after it was booted up 
2) Changing localization while using other available device 
- turn on IDE and select ios device 
- change devices localization after it was booted up 
- switch devices to android and check if the localization was configured correctly


packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -117,10 +199,29 @@ export class AndroidEmulatorDevice extends DeviceBase {
         settings.location.latitude.toString(),
       ]);
     }
-    return false;
+    return shouldRestart;
+  }
+
+  // it is necessary to use SIGTERM 9 as it would take to much time otherwise and would degrade user experience
+  // having said that use this function with caution only in scenarios when you are sure that no user data will be lost
+  private async resetDevice() {
```
jakub-gonet, line 207: `forcefullyResetDevice` to signify this isn't typical reboot sequence?

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -70,7 +70,89 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+
+  // this method changes device locale in two ways, firstly it modifies system_locales setting,
+  // which in most cases is enough, to change locale. Unfortunately if persist.sys.locale exist
+  // it will conflict with system_locales and make device behavior unpredictable, so in that case
+  // we need to modify if as well. persist.sys.locale is added to the device when a user changes locale
+  // in the settings. Unfortunately this method does not handle all possible scenarios.
+  // If the device uses google mobile services persist.sys.locale can not be changed, as it requires a root access
+  // TODO:  Find a way to change or remove persist.sys.locale without root access. note: removing the whole
+  // data/property/persistent_properties file would also work as we persist device settings globally in Radon IDE
+  private async changeLocale(newLocale: Locale): Promise<boolean> {
```
jakub-gonet, line 101: Isn't it always returning true? What does the return value mean?

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -70,7 +70,89 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+    const { stdout } = await exec(ADB_PATH, [
+      "-s",
+      this.serial!,
+      "shell",
+      "getprop",
+      "persist.sys.locale",
+    ]);
+
+    if (!stdout) {
+      return true;
```
kmagiera, line 124: add some logging here such that we can diagnose it for the users who may've set the language using settings app

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -70,7 +70,89 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+        locale,
+      ]);
+    } catch (e) {
+      // this is expected if device is using google mobile services
+      Logger.debug(
+        "CHANGE LOCALE - Device is not allowing root access, moving on without modifying persist.sys.locale"
+      );
+    }
+
+    return true;
```
kmagiera, line 146: I think it should be ok to remove this part. It is unlikely that someone would change the language via settings, and even if they do, it is unlikely you can actually use sudo on Android devices as it is only available on pure AOSP builds which are rarely used in practice

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -70,7 +70,89 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+      this.serial!,
+      "shell",
+      "settings",
+      "get",
+      "system",
+      "system_locales",
+    ]);
+
+    // if user did not use the device before it might not have system_locales property
+    // as en-US is the default locale we assume that no value is the same as en-US
```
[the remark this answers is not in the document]
filip131311: If there is no value and the user selected locale is en-US I don't want to trigger locale change as it is already a correct one

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -70,7 +70,74 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+    const { stdout } = await exec(ADB_PATH, [
+      "-s",
+      this.serial!,
+      "shell",
+      "getprop",
+      "persist.sys.locale",
+    ]);
+
+    // this is needed to make sure that changes will persist
+    await exec(ADB_PATH, ["shell", "sync"]);
```
kmagiera, line 125: why we don't pass `"-s", this.serial` here?

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -70,7 +70,74 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+
+    if (stdout) {
```
kmagiera, line 127: would feel more natural if we made this check before the other `exec` call given you use a transient stdout variable

kmagiera (changes requested):
Please see my inline comments and add test plan to PR description.

I believe the way you tested this was by changing it while the Android device was booted. But what happen when you change language on iOS and then switch to Android, or when you change the language before the device is fully booted.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
filip131311: it was a mistake thank you

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -161,10 +161,12 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
 
 const LocalizationItem = () => {
   const { openModal } = useModal();
+  const { projectState } = useProject();
   return (
     <>
       <Label>Device Localization</Label>
       <DropdownMenu.Item
+        disabled={projectState.status !== "running"}
```
kmagiera, line 169: similar change already landed in #643 – we likely want to use that one instead

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -70,7 +70,74 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+
+    // if user did not use the device before it might not have system_locales property
+    // as en-US is the default locale we assume that no value is the same as en-US
+    if ((stdout ?? "en-US") === locale) {
+      return false;
+    }
+    return true;
+  }
+
+  // this method changes device locale in two ways, firstly it modifies system_locales setting,
```
kmagiera, line 93: please read my other comment about method level comments I posted below.

kmagiera (changes requested):
Almost there, I left a few more comments inline

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -78,7 +78,68 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+      this.serial!,
+      "shell",
+      "settings",
+      "get",
+      "system",
+      "system_locales",
+    ]);
+
+    // if user did not use the device before it might not have system_locales property
+    // as en-US is the default locale we assume that no value is the same as en-US
```
kmagiera, line 94: I don't understand what does it mean "no value is the same as en-US"

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -78,7 +78,68 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+    if ((stdout ?? "en-US") === locale) {
+      return false;
+    }
+    return true;
+  }
+
+  /**  This method changes device locale by modifying system_locales system setting. */
```
kmagiera, line 101:
```suggestion
  /**
   * This method changes device locale by modifying system_locales system setting.
   */
```

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -78,7 +78,68 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+      "-s",
+      this.serial!,
+      "shell",
+      "getprop",
+      "persist.sys.locale",
+    ]);
+
+    if (stdout) {
+      Logger.warn(
+        "persist.sys.locale detected while changing locale. It might indicate that the user has changed locale settings in the settings application, which will prevent localization change."
```
kmagiera, line 131:
```suggestion
        "Updating locale will not take effect as the device has altered locale via system settings which always takes precedence over the device setting the IDE uses."
```

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -125,10 +186,27 @@ export class AndroidEmulatorDevice extends DeviceBase {
         settings.location.latitude.toString(),
       ]);
     }
-    return false;
+    return shouldRestart;
+  }
+
+  /** This method restarts the emulator process using SIGKILL signal.
+   *  Should be used for the situations when quick reboot is necessary
+   *  and when we don't care about the emulator's process state */
```
kmagiera, line 194:
```suggestion
  /**
   * This method restarts the emulator process using SIGKILL signal.
   * Should be used for the situations when quick reboot is necessary
   * and when we don't care about the emulator's process state
   */
```

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -144,6 +222,7 @@ export class AndroidEmulatorDevice extends DeviceBase {
         "-no-boot-anim",
         "-grpc-use-token",
         "-no-snapshot-save",
+        "-writable-system",
```
kmagiera, line 225: what is this option? why are we adding it here? It should be explained in PR description

kmagiera (approved):
Added a couple more inline comments Accepting to unblock this but please address the comments before merging

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
filip131311: I modified the comment to explain it better

Issue #545, opened 2024-09-14
JAVA_HOME is set to an invalid directory


abdul-razaq:
**ERROR*
I keep getting the error below when I try to boot up the Android emulator in the react native IDE.

**ERROR: JAVA_HOME is set to an invalid directory: C:\Program Files\Android\Android Studio\jbr
Please set the JAVA_HOME variable in your environment to match the 
2024-09-14 15:34:47.193 [info] location of your Java installation. **

This error occurs even tho my JAVA_HOME points to the right java installation and works well when opening my emulator in the normal usual way.

**Environment**
- Version 0.0.18
- React native 0.74.5.
- Expo: ~51.0.31
- Java version "17.0.11" 2024-04-16 LTS, Java(TM) SE Runtime Environment (build 17.0.11+7-LTS-207), Java HotSpot(TM) 64-Bit Server VM (build 17.0.11+7-LTS-207, mixed mode, sharing).

cacacaio:
I have the same issue on windows, the env variable is working if I print it via powershell and cmd. But on IDE it always points to C:\Program Files\Android\Android Studio\jbr

What I did as a workaround is copy all the JDK files to this folder so at least it works, but it would be good to have a permanent fix

abdul-razaq:
Actually, this worked for me. I hope there is a permanent fix for this as soon as possible. Thank you!

p-malecki:
Hi, thanks for reporting. Radon uses JAVA_HOME env var, so there's no need for workarounds from now on.

Issue #581, opened 2024-10-01
[windows] keybindings.json file not found


erickriva:
**Description**
When opening `Settings`, an error was shown in logs. Searched path is from Mac OS, not Windows.

```
[error] Error while parsing keybindings.json A system error occurred
(ENOENT: no such file or directory, open 'C:\Users\USERNAME\Library\Application Support\Code\User\keybindings.json')
```

`keybindings.json` is found in this path in my Windows PC: `C:\Users\USERNAME\AppData\Roaming\Code\User\keybindings.json`

![image](https://github.com/user-attachments/assets/30844c05-55ba-45bf-b683-f1824420d24b)

**Environment**

- OS: **Windows 11**
- Version of currently used Radon IDE: **v0.0.19**

p-malecki:
Hello, thanks for reporting, Radon should use correct keybindings paths from now on.

PR #592, opened 2024-10-04
Fix JAVAHOME, ANDROIDHOME, keybindings.json paths on Windows

p-malecki:
This PR :
- fixes invalid directory errors caused by using the hardcoded path "(...)Android\Android Studio\jbr" to find the JAVA_HOME env variable. The solution is to use an already existing env variable in the first place and to find the path only when the variable does not exist
- changes slashes to backslashes in the Windows path to ANDROID_HOME
- fixes the path to keybindings.json on Windows

resolves: #545 , #581


packages/vscode-extension/src/utilities/utils.ts
```diff
@@ -24,7 +25,10 @@ export class Utils implements UtilsInterface {
     try {
       keybindingsJsonPath = path.join(
         homedir(),
-        "Library/Application Support/Code/User/keybindings.json"
+        Platform.select({
+          macos: "Library/Application Support/Code/User/keybindings.json",
+          windows: "AppData\\Roaming\\Code\\User\\keybindings.json",
```
jakub-gonet, line 30: Still breaks on Cursor :c

jakub-gonet (approved):

packages/vscode-extension/src/utilities/android.ts
```diff
@@ -7,7 +7,7 @@ export const ANDROID_HOME =
   process.env.ANDROID_HOME ??
   Platform.select({
     macos: path.join(os.homedir(), "Library/Android/sdk"),
-    windows: path.join(os.homedir(), "AppData/Local/Android/Sdk"),
+    windows: path.join(os.homedir(), "AppData\\Local\\Android\\Sdk"),
```
filip131311, line 10: why not use path.join all the way to avoid \\ / difrence?

filip131311 (changes requested):
I left some inline changes

packages/vscode-extension/src/utilities/utils.ts
p-malecki: We can deduce used IDE by checking value of `vscode.env.appName`, so now it also works on Cursor c:

filip131311 (approved):
I aprove but would you mind adding a path.join method to macos paths as well?

PR #593, opened 2024-10-04
Include stack trace for unhandled promises and error logging

kmagiera:
This PR fixes the logging of errors to include stack traces in the log output.

Before this change, we'd append error as last element to log output channel, whereas VSCode API expects the error to be passed as first item. As a consequence, error would just be treated as strings and we'd only get error.message in the output.

This PR reorders the elements if error object is passed as the last item to Logger.error (we do it in a number of places throught the codebase). Error messages are then provided as the first item which makes VSCode format them properly and also display stack trace in the output.

This is particularily usecul for unhandled exception handlers where we have no clue of where a given error could be thrown.

This PR also disables expected rejections in certain CDP calls that we are making as they were polluting our log output.
We are also now wrapping CDP calls in Error objects such that we can see proper stack traces in the output.

## Test plan
1. Disable catch() calls in Adapter
2. See the unhandled exception with stack trace (there was only message before this change)

### DIFF
--- a/packages/vscode-extension/src/Logger.ts
+++ b/packages/vscode-extension/src/Logger.ts
@@ -24,7 +24,12 @@ const logger = {
   },
 
   error(message: string, ...args: any[]) {
-    outputChannel.error(message, ...args);
+    if (args.length > 0 && args[0] instanceof Error) {
+      const error = args[0] as Error;
+      outputChannel.error(error, message, ...args.slice(0, -1));
+    } else {
+      outputChannel.error(message, ...args);
+    }
   },
 
   openOutputPanel() {
--- a/packages/vscode-extension/src/Logger.ts
+++ b/packages/vscode-extension/src/Logger.ts
@@ -24,8 +24,8 @@ const logger = {
   },
 
   error(message: string, ...args: any[]) {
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/Logger.ts
```diff
@@ -24,7 +24,12 @@ const logger = {
   },
 
   error(message: string, ...args: any[]) {
-    outputChannel.error(message, ...args);
+    if (args.length > 0 && args[args.length - 1] instanceof Error) {
+      const error = args[args.length - 1] as Error;
```
filip131311, line 28: is the as Error necessary, I think `&& args[args.length - 1] instanceof Error` should cause type cast by itself

filip131311 (approved):
I aproved left one question in line

packages/vscode-extension/src/Logger.ts
kmagiera: no, apparently cannot infer the type this way. If I first stored it in variable and only have `instanceof` there, it'd work. We actually don't need to make it of Error type, as the logger then accepts "any" as input. But this way it is more explicit.

PR #594, opened 2024-10-07
Add build and launch telemetry

kmagiera:
This PR adds more loggind and telemetry around build and launch process.

Building and launching the app is one of the most fragile element of the IDE pipeline and we get numerous reports around that. While there are a few known issues around it that we are fixing and investigating. I figured it'd be good to also add telemetry around that such that we can track how it works for a larger group of people and also be able to detect regressions.

## Test plan

Just run the example app with clean build and make sure nothing breaks.

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -17,6 +17,8 @@ import { getLaunchConfiguration } from "../utilities/launchConfiguration";
 import { DebugSession, DebugSessionDelegate } from "../debugging/DebugSession";
 import { throttle } from "../utilities/throttle";
 import { DependencyManager } from "../dependency/DependencyManager";
+import { getTelemetryReporter } from "../utilities/telemetry";
+import { platform } from "os";
 
 type StartOptions = { cleanBuild: boolean };
 
@@ -100,6 +102,11 @@ export class DeviceSession implements Disposable {
   }
 
   private async launchApp() {
+    const launchReequestTime = Date.now();
+    getTelemetryReporter().sendTelemetryEvent("app:launch:requested", {
+      platform: this.device.platform,
+    });
+
     // FIXME: Windows getting stuck waiting for the promise to resolve. This
     // seems like a problem with app connecting to Metro and using embedded
     // bundle instead.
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -100,6 +101,11 @@ export class DeviceSession implements Disposable {
   }
 
   private async launchApp() {
+    const launchReequestTime = Date.now();
```
p-malecki, line 104: typo:
launchReequestTime -> launchRequestTime

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -111,10 +117,30 @@ export class DeviceSession implements Disposable {
...
+      const reportWaitingStuck = setTimeout(() => {
+        Logger.info("App is taking very long to boot up, it might be stuck");
+        getTelemetryReporter().sendTelemetryEvent("app:launch:waiting-stuck", {
+          platform: this.device.platform,
+        });
+      }, 30000);
+      waitForAppReady.then(() => clearTimeout(reportWaitingStuck));
+    }
+
     const [previewUrl] = await Promise.all([this.device.startPreview(), waitForAppReady]);
```
p-malecki, line 131: Shouldn't we wait for waitForAppReady once?
```suggestion
    }

    const [previewUrl] = await Promise.all([this.device.startPreview(), waitForAppReady]);
    if (shouldWaitForAppLaunch) {
       clearTimeout(reportWaitingStuck);
    }
```

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -111,10 +117,30 @@ export class DeviceSession implements Disposable {
...
     Logger.debug("App and preview ready, moving on...");
     this.eventDelegate.onStateChange(StartupMessage.AttachingDebugger);
     await this.startDebugger();
+
+    const launchDurationSec = (Date.now() - launchReequestTime) / 1000;
```
p-malecki, line 136: typo:
launchReequestTime -> launchRequestTime

p-malecki (approved):
Left some comments inline

packages/vscode-extension/src/project/deviceSession.ts
kmagiera: I like my approach moare as it doesn't mix the reporting code with the existing logic. Whatever is placed inside that if block I added, is just for reporting and can be deleted. Otherwise I'd have to keep the `reportWaitingStuck` in main scope, while we'd need to have `shouldWaitForAppLaunch` checks both before and after the await.

On top of that this isn't really an equivalent, as the await is for both startPreview and app ready event.

PR #595, opened 2024-10-08
Fix android studio path after #592

kmagiera:
In #592 we broke the Android Studio path on Mac while replacing the actual path with `path.join` calls. Apparently, `path.join` doesn't start at the file system root, but creates relative path if no lash at the start is provided.

This PR updates the logic of the code that finds java home:
1. We no longer use path.join where it doesn't make sense. Specifically given we have Platform.select we now exacly how paths are constructed on each individual system. Using path.join just makes it more difficult to process this code.
2. We move constants that doesn't need to live under global scope into findJavaHome method (specifically path for android studio)
3. We extract fs root on windows using cwd path root attribute to make it more resilient depending on the users' filesystem setup.

### Test plan
1. Make sure android builds can run on Mac (they were broken before)
2. Test on windows as well (don't have it at hand right now but we will test it later today)

### DIFF
--- a/packages/vscode-extension/src/utilities/android.ts
+++ b/packages/vscode-extension/src/utilities/android.ts
@@ -11,7 +11,7 @@ export const ANDROID_HOME =
   });
 
 const ANDROID_STUDIO_PATH = Platform.select({
-  macos: path.join("Applications", "Android Studio.app"),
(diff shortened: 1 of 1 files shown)


p-malecki (approved):
Tested on Windows, works correctly.

PR #597, opened 2024-10-08
Add custom device names

p-malecki:
This PR introduces following enhancements:
- Adds visual differentiation to the active device in `DevicesSelect` and `ManageDevicesView` by bolding the device name and using a brighter icon
- Allows to change the active device directly from the `ManageDevicesView` 
- Adds an option to assign a `device name` during the creation of a new device using the `CreateDeviceView` (supported for both Android and iOS devices)
- Adds an option to rename a `device name` in the `ManageDevicesView`
- Renames `displayName` to `deviceName` in `DeviceInfo`
- Update text and background colors in disabled fields in the `CreateDeviceView` 

Custom names are limited to alphanumeric chars, spaces, dashes, and underscores and length between 1 and 30 characters.

In Android's config.ini, the "avd.ini.displayname" key is used to store the device name, which can store any string. During emulator creation we pass to 'hw.device.name' key model's identifier which later is used by emulator to display device status bar layout correctly.  The `mapIdToModel` function is used to translate model identifiers into names within the webview.

⚠️ When you need to add a new supported device, make sure to also include the device identifier in the `consts.ts`. Identifiers are sourced from `hw.device.name` in `config.ini` for Android and `deviceType` in `device.plist` for iOS.

Test plan:
1. Switch the active device in the `DevicesSelect` and `ManageDevicesView`.
2. Create new iOS and Android devices with names different than device model names, try to include illegal characters or names longer than max acceptable length.
3. Add a second device with the same name.
4. Rename iOS and Android device' names.
5. Restart the development environment (RNIDE) and verify if all newly added devices are present and accessible.

screenshots:
Screenshot 2024-10-29 at 12 48 37
Screenshot 2024-10-15 at 10 58 11
Screenshot 2024-10-28 at 12 33 49
Screenshot 2024-10-28 at 12 35 27
Screenshot 2024-10-29 at 12 49 26
Screenshot 2024-10-29 at 12 48 51


packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -508,6 +516,7 @@ async function listEmulatorsForDirectory(avdDirectory: string) {
         avdId,
         name: displayName,
         systemName: systemImageName ?? "Unknown",
+        customName: customName ?? "",
```
jakub-gonet, line 519: I'd fallback when displaying, not here.

packages/vscode-extension/src/webview/views/ManageDevicesView.tsx
```diff
@@ -59,8 +82,12 @@ function ManageDevicesView() {
     reload();
   }, []);
 
-  const androidDevices = devices.filter((device) => device.platform === DevicePlatform.Android);
-  const iosDevices = devices.filter((device) => device.platform === DevicePlatform.IOS);
+  const iosDevices = devices.filter(
+    ({ platform, name }) => platform === DevicePlatform.IOS && name.length > 0
```
jakub-gonet, line 86: Why we need to filter out devices without name?

p-malecki: Similar filter is used in `DeviceSelect`, which was added in Windows Integration PR. I believe it is related to emulator files left after device removal, which happens sometimes on Windows.

filip131311:
Is it possible to add editing device name after it was created? I think it might be useful when you add your second device for the first time.

p-malecki:
> Is it possible to add editing device name after it was created? I think it might be useful when you add your second device for the first time.

Thanks for pointing it out. I initially considered adding this feature in future updates, but I agree that it's best to include it in this PR.

kmagiera (changes requested):
Haven't reviewed the whole thing but the comments I left inline should be enough for the next iteration.

On top of that I'd like to suggest some changes to the UI:
1) When creating new device, the name field shouldn't be pre-filled with "My new iOS Device" but have a placeholder that uses the name we'd use if you don't specify anything there
2) Name change button should use codicon-edit instead of "Aa" – I know we only allow editing the name, but the other icon look very unintuitive
3) When you use the device select button (play icon), it should change the device and close the dialog

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -438,7 +439,7 @@ export async function createEmulator(
     ["hw.cpu.ncore", "4"],
     ["hw.dPad", "no"],
     ["hw.device.manufacturer", "Google"],
-    ["hw.device.name", deviceName],
+    ["hw.device.name", mapModelToId(modelName)],
```
kmagiera, line 442: I wonder if this is even necessary to have a proper device name there, have you tested this? Maybe there's no reason to introduce this mapping and this additional complexity here if it doesn't make any difference what we put there

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -464,26 +482,29 @@ export async function listSimulators(
...
-          UDID: device.udid,
-          name: device.name,
-          systemName: runtime?.name ?? "Unknown",
-          available: device.isAvailable ?? false,
-          deviceIdentifier: device.deviceTypeIdentifier,
-          runtimeInfo: runtime!,
-        };
-      });
+      const deviceInfos = await Promise.all(
+        devices.map(async (device) => {
```
kmagiera, line 490: why is this async if it doesnt do anything that's asynchronous?

This way you could eliminate Promise.all above and the one from line 485

packages/vscode-extension/src/webview/components/ZoomControls.css
```diff
@@ -12,11 +12,11 @@
 }
 
 .zoom-controls .icon-button.zoom-out-button {
-  border-radius: 18px 18px 0 0;
+  border-radius: 0 0 18px 18px;
```
kmagiera, line 15: I think there was similar change in #643

packages/vscode-extension/src/webview/components/ZoomControls.tsx
```diff
@@ -19,7 +19,7 @@ type ZoomControlsProps = {
 const ZoomLevelSelect = ({ zoomLevel, onZoomChanged }: ZoomControlsProps) => {
   const onValueChange = useCallback(
     (e: string) => {
-      if (e == "Fit") {
+      if (e === "Fit") {
```
kmagiera, line 22: this also seem oddly similar to changes from #643

packages/vscode-extension/src/webview/utilities/consts.ts
```diff
@@ -26,6 +25,7 @@ export type DeviceProperties = {
   maskImage: string;
 };
 
+// When including a new device update the `supportedDevices.ts` in extension with the model identifier.
```
kmagiera, line 28: not sure what you mean here, this doesn't seem to be a grammatically correct sentence

packages/vscode-extension/src/webview/views/CreateDeviceView.tsx
```diff
@@ -143,13 +158,15 @@ function CreateDeviceView({ onCreate, onCancel }: CreateDeviceViewProps) {
...
-        <div className="form-label" />
         {systemImagesOptions.length > 0 ? (
           <Select
             disabled={!platformSelected}
             className="form-field"
             value={selectedSystemName ?? ""}
-            onChange={(newValue) => selectSystemName(newValue)}
+            onChange={(newValue) => {
+              selectSystemName(newValue);
+              inputRef.current!.value = deviceModel ?? "";
```
kmagiera, line 168: this is a controlled component, so this shouldn't be here, why have you added this?

packages/vscode-extension/src/webview/views/DevicesNotFoundView.tsx
```diff
@@ -9,8 +9,8 @@ import { useDevices } from "../providers/DevicesProvider";
 import { AndroidSupportedDevices, iOSSupportedDevices } from "../utilities/consts";
 import { IOSDeviceTypeInfo, IOSRuntimeInfo } from "../../common/DeviceManager";
 import { useDependencies } from "../providers/DependenciesProvider";
-import { vscode } from "../utilities/vscode";
 import { Platform, useUtils } from "../providers/UtilsProvider";
+import { mapModelToId } from "../../devices/supportedDevices";
```
kmagiera, line 13: webview should only reference code from under webview and common. this way we have clear separation on what gets bundled where. If you want to use it here, it should live under common

kmagiera (changes requested):
Reviewed most part of this, but it seem like there are some changes that aren't necessarily related. Can you extract these as this PR is already pretty huge and difficult to review

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
filip131311: @km1chno tested this in https://github.com/software-mansion/radon-ide/pull/601 and without this prop device displayed wrong status bar layout

packages/vscode-extension/src/common/Project.ts
```diff
@@ -124,6 +124,7 @@ export interface ProjectInterface {
   restart(forceCleanBuild: boolean): Promise<void>;
   goHome(homeUrl: string): Promise<void>;
   selectDevice(deviceInfo: DeviceInfo): Promise<boolean>;
+  updateSelectedDevice(deviceInfo: DeviceInfo): void;
```
kmagiera, line 127: how is this different from `selectDevice`defined above ??

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -480,7 +481,7 @@ export class AndroidEmulatorDevice extends DeviceBase {
 }
 
 export async function createEmulator(
-  displayName: string,
+  modelName: string,
```
kmagiera, line 484: so it appears we are effectively renaming `displayName` to `deviceName` ?? If so I don't think I like that. `deviceName` suggest this is something that's associated with the device itself, while `displayName` makes it pretty obvious this can be user defined and doesn't have any hidden meaning

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -622,6 +627,29 @@ async function ensureOldEmulatorProcessExited(avdId: string) {
   }
 }
 
+export async function renameEmulator(avdId: string, newDisplayName: string) {
```
kmagiera, line 630: and here we actually use "displayName` to name the same parameter we refer to as "deviceName" iin other places?

kmagiera (changes requested):
Added some inline comments about renaming displayName.

Also, If I read correclty `modelName` seems to be only used in the webview code? So maybe there's no reason to add the whole mapping thing, but instead we could just rely on modelId (the field you remove in favor of adding a map). Then the extension code wouldn't need to perform any mapping, and we could use the names from consts.ts to perform only one-way mapping in the webview code

packages/vscode-extension/src/common/Project.ts
p-malecki: `selectDevice` changes active device, so sets different project status and starts a new device session, while updateSelectedDevice simply updates deviceInfo in updateProjectState with changed device name

kmagiera: can we only update the selected device?

packages/vscode-extension/src/webview/components/DeviceRenameDialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+          className="device-rename-button"
+          type="ternary"
+          disabled={!isDisplayNameValid}
+          onClick={async () => {
+            const newDisplayName = displayName.trim();
+            try {
+              await deviceManager.renameDevice(deviceInfo, newDisplayName);
+            } finally {
+              deviceInfo.displayName = newDisplayName;
+              project.updateSelectedDevice(deviceInfo);
```
kmagiera, line 74: I understand that you can update any device from the manage devices list. Why are we updating the selected one here then?

packages/vscode-extension/src/webview/utilities/consts.ts
```diff
@@ -108,3 +113,14 @@ export const AndroidSupportedDevices: DeviceProperties[] = [
     maskImage: pixel6amask,
   },
 ] as const;
+
+export function mapIdToModel(deviceId: string): string {
+  const device = iOSSupportedDevices
```
kmagiera, line 118: would be easier to just call `find` on one collection and then on the other. This will clone arrays and given it runs in a cell render function it may be executed a large number of times

packages/vscode-extension/src/webview/views/CreateDeviceView.tsx
```diff
@@ -95,44 +102,60 @@ function CreateDeviceView({ onCreate, onCancel }: CreateDeviceViewProps) {
 
     setLoading(true);
     try {
+      const modelId = iOSSupportedDevices.concat(AndroidSupportedDevices).find((device) => {
```
kmagiera, line 105: can't we use `mapIdToModelName` here?

packages/vscode-extension/src/webview/views/CreateDeviceView.tsx
```diff
@@ -95,44 +102,60 @@ function CreateDeviceView({ onCreate, onCancel }: CreateDeviceViewProps) {
...
+
   return (
     <div className="edit-device-form">
       <div className="form-row">
         <Label>Device Type</Label>
         <Select
           className="form-field"
           value={`${
-            devicePlatform && displayName ? `${devicePlatform}:${displayName}:${deviceName}` : ""
+            devicePlatform && deviceModelName ? `${devicePlatform}:${deviceModelName}` : ""
```
kmagiera, line 149: can't we just use modelId here for the value?

packages/vscode-extension/src/webview/utilities/consts.ts
```diff
@@ -108,3 +113,16 @@ export const AndroidSupportedDevices: DeviceProperties[] = [
     maskImage: pixel6amask,
   },
 ] as const;
+
+export function mapIdToModel(deviceId: string): string {
+  let device = iOSSupportedDevices.find((d) => d.modelId === deviceId);
+  if (!device) {
+    device = AndroidSupportedDevices.find((d) => d.modelId === deviceId);
+  }
```
kmagiera, line 121:
```suggestion
  let device = iOSSupportedDevices.find((d) => d.modelId === deviceId) || AndroidSupportedDevices.find((d) => d.modelId === deviceId);
```

packages/vscode-extension/src/webview/views/CreateDeviceView.tsx
```diff
@@ -100,39 +91,45 @@ function CreateDeviceView({ onCreate, onCancel }: CreateDeviceViewProps) {
...
-            const [newPlatform, newDisplayName, newDeviceName] = newValue.split(":", 3);
-            assertPlatform(newPlatform);
-            setDisplayName(newDisplayName);
-            setDeviceName(newDeviceName);
-            setDevicePlatform(newPlatform);
+          value={deviceModelId ?? ""}
+          onChange={(modelId: string) => {
+            setDevicePlatform(modelId.startsWith("com.apple") ? "ios" : "android");
+            setDeviceModelName(mapIdToModel(modelId));
+            setDeviceModelId(modelId);
```
kmagiera, line 129: seems like it'd be much simpler to just keep DeviceProperties object in the state. Then we can have a method that finds it based on the model id and updates the state. We can extract devicePlatform, modelId and modelName from it

packages/vscode-extension/src/webview/components/DeviceRenameDialog.tsx
kmagiera: Feels like this isn't the right place to update the selected device. First of all, it still seems to assume that we can only edit the selected device which is weird. We're also introducing a new method to project interface with unclear semantics (it actually changes the selected device info object but does not trigger the device change which can be very confusing).

I think the whole renaming of the current device should be entirely handled by the extension code and there should be no additional code for it in the UI. When someone renames the selected device, we can trigger `updateProjectState` from within the project and this should propagate down to the UI normally.

kmagiera (changes requested):

packages/vscode-extension/src/webview/components/DeviceRenameDialog.tsx
p-malecki: I agree that the webview shouldn't trigger updateProjectState, the naming of updateSelectedDevice is unclear, and its purpose isn't intuitive. Instead updateSelectedDevice method, we should use the new method renameDevice also from project.ts, which will call renameDevice from DeviceManager.ts to save the name change and update the project state only if it’s the currently selected device that is being renamed.

kmagiera (approved):

PR #598, opened 2024-10-09
Use metro message endpoint to dispatch dev menu request

kmagiera:
This PR uses a different method for opening dev menu.

Previously we'd call native dev menu module, which was causing problems, specifically in expo-go where the module resolution was resulting in bundle error. Since it turned out to be easier to use the standard method that metro uses for triggering the menu than to debug the old method, I decided to rework the underlying implementation.

Now, we use metro /message endpoint that is used to proxy messages between dev tools and device. One of the messages is "devMenu" which opens the dev menu on the device side (which is what we wanted).

This PR also removes old code which was placed in the wrapper, but also for android we were using a different method that was relying on adb. The new method works on both devices, hence we don't need separate implementation and can delete both of them.

## Test plan
1. Use dev menu option on Android and iOS across all configurations we have in the repo (expo-go, react-native-75, expo-51 etc)

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -269,7 +269,7 @@ export class DeviceSession implements Disposable {
   }
 
   public async openDevMenu() {
-    this.metro.openDevMenu();
+    await this.metro.openDevMenu();
   }
 
   public startPreview(previewId: string) {
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -269,17 +269,7 @@ export class DeviceSession implements Disposable {
   }
 
   public async openDevMenu() {
-    // on iOS, we can load native module and dispatch dev menu show method. On
-    // Android, this native module isn't available and we need to fallback to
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -269,17 +269,7 @@ export class DeviceSession implements Disposable {
...
-    // adb to send "menu key" (code 82) to trigger code path showing the menu.
-    //
-    // We could probably unify it in the future by running metro in interactive
-    // mode and sending keys to stdin.
-    if (this.device.platform === DevicePlatform.IOS) {
-      this.devtools.send("RNIDE_iosDevMenu");
-    } else {
-      await (this.device as AndroidEmulatorDevice).openDevMenu();
-    }
+    this.metro.openDevMenu();
```
jakub-gonet, line 272: `await`?

jakub-gonet (approved):

packages/vscode-extension/src/project/deviceSession.ts
kmagiera:
```suggestion
    await this.metro.openDevMenu();
```

PR #599, opened 2024-10-09
Suppress expected errors for certain CDP commands

kmagiera:
Previously, we added `catch` statements to avoid certain, expected errors from being reported in the log output. However, this turns out to not work, as it seems the catch with no handler method doesn't seem to catch the exception. 

This PR adds a no-op handler for catch statement which makes the error no longer appear in the extension output.

### DIFF
--- a/packages/vscode-extension/src/debugging/DebugAdapter.ts
+++ b/packages/vscode-extension/src/debugging/DebugAdapter.ts
@@ -98,12 +98,15 @@ export class DebugAdapter extends DebugSession {
     this.connection = new WebSocket(configuration.websocketAddress);
 
     this.connection.on("open", () => {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -98,12 +98,15 @@ export class DebugAdapter extends DebugSession {
     this.connection = new WebSocket(configuration.websocketAddress);
 
     this.connection.on("open", () => {
-      this.sendCDPMessage("FuseboxClient.setClientMetadata", {}).catch(); // ignore error as it will throw on old debugger and is not critical
+      // the below catch handler is used to ignore errors coming from non critical CDP messages we
+      // expect in some setups to fail
+      const ignoreError = () => {};
```
jakub-gonet, line 103: Should be defined outside of class.

jakub-gonet (approved):
You can merge even without addressing the comment.

PR #601, opened 2024-10-09
Add Pixel 8 and 9 skins

### LINKED ISSUE #583: Add new android device types (pixel 8,9)

km1chno:
Resolves #583


filip131311 (changes requested):
I'm not sure why you did not expirience this problem but it seems that mask size is missaligned on pixel9

km1chno:
interesting, I didn't have this issue
i will look into it

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -428,9 +429,8 @@ export async function createEmulator(displayName: string, systemImage: AndroidSy
     ["hw.cpu.arch", getNativeQemuArch()],
     ["hw.cpu.ncore", "4"],
     ["hw.dPad", "no"],
-    ["hw.device.hash2", "MD5:3db3250dab5d0d93b29353040181c7e9"],
     ["hw.device.manufacturer", "Google"],
-    ["hw.device.name", "pixel_7"],
```
km1chno, line 433: I managed to reproduce the issue with misaligned front camera, but it was flaky - sometimes it worked, sometimes not, but I believe this change fixed it as I'm not able to reproduce it now :P 

As for other misalignments, it looks like there is an issue with skin/mask in Android Studio itself, when I run Pixel 9 emulator from command line it also looks not that well:

Zrzut ekranu 2024-10-10 o 16 58 09

Zrzut ekranu 2024-10-10 o 16 52 17

Zrzut ekranu 2024-10-10 o 16 55 11

Zrzut ekranu 2024-10-10 o 16 57 49

filip131311 (approved):
I approve, but please let @p-malecki take a look before merging, as he is currently working on devices names, and I am not sure if the name you use will still exist.

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -487,7 +487,7 @@ function Preview({
   }, [project, openRebuildAlert, projectStatus]);
 
   const device = iOSSupportedDevices.concat(AndroidSupportedDevices).find((sd) => {
-    return sd.name === projectState?.selectedDevice?.name;
+    return sd.displayName === projectState?.selectedDevice?.name;
```
p-malecki, line 490: we want to accept only supported devices detecting them by device names, not the display names, which users will be able to change after #597

packages/vscode-extension/src/webview/utilities/consts.ts
```diff
@@ -24,7 +29,7 @@ export type DeviceProperties = {
 // iOS devices names should match supportedDeviceTypes inside the runtime
 export const iOSSupportedDevices: DeviceProperties[] = [
   {
-    name: "iPhone 15 Pro",
+    displayName: "iPhone 15 Pro",
```
p-malecki, line 32: we should store the device names in the deviceName variable with the value "iPhone 15 Pro" in this case. The displayName should be determined later based on the deviceName or the user's own device name

p-malecki (commented):
As @filip131311 mentioned in the last comment, I am working on PR #597 in which we add the option for users to set and change device display names. In that PR, I use displayName for storing the user's custom device name and deviceName for detecting the device type and determining whether it is supported. There are some differences in the way you use them in this PR, but I think it's best to accept it as is and make changes later in #597.

I left some comments but you do not have to address them, they are changes that I intend to implement in #597.

packages/vscode-extension/src/webview/utilities/consts.ts
```diff
@@ -9,7 +13,8 @@ import iphoneSEmask from "../../assets/iphone_SE/mask.png";
 import { DevicePlatform } from "../../common/DeviceManager";
 
 export type DeviceProperties = {
-  name: string;
+  displayName: string;
+  deviceName?: string; // only needed for Android to set hw.device.name in config.ini
```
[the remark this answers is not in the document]
km1chno: we decided to rename the essential value to `modelName`
63f3196c82019876b7f05fbf99e3b924d98643cb

packages/vscode-extension/src/webview/utilities/consts.ts
km1chno: we decided to store the value `iPhone 15 Pro` in `modelName` 
63f3196c82019876b7f05fbf99e3b924d98643cb

PR #604, opened 2024-10-09
Add extension typecheck workflow and enhance lint workflow

km1chno:
## Changes

Enhanced linting and typecheck in `vscode-extension`

#### Typecheck 

- typecheck CI workflow to run on changes in `vscode-extension`
- added default `replaysEnabled: false` to `ProjectContextProps` in `ProjectProvider.tsx` (typecheck was failing because it was missing)

#### Lint

- `eslint` is now run also on `.tsx/.js/.jsx` files, so I fixed issues in them
- fixed `prettier --check` in CI which was not working
- fixed some lint warnings about `!=` instead of `!==`
- added `no-unused-vars` rule to `eslint` to prevent unused imports and variables, fixed issues that have been found
- set `{ args: none }` in `no-unused-vars` rule because we have some functions with arguments that are not used - but I didn't want to change any interfaces


packages/vscode-extension/src/webview/providers/DependenciesProvider.tsx
```diff
@@ -58,11 +58,11 @@ export default function DependenciesProvider({ children }: PropsWithChildren) {
   }, []);
 
   useEffect(() => {
-    const dependencies: DependencyRecord = {};
+    const newDependencies: DependencyRecord = {};
```
kmagiera, line 61: While I agree with most "oldSth" and "newSth" renames, here the dependencies is actually the main source of data that is then propagated to state. This object is created once per component lifetime as it is in useEffect with [] passed, and then it is used to accumulate all the updates.

kmagiera (approved):
Accepting this, just added one inline comment as I don't understand why we are renaming one variable. Otherwise looks good. Enable Replays should be false by default.

packages/vscode-extension/src/webview/providers/DependenciesProvider.tsx
km1chno: you are right, i reverted this change and instead changed the name of constant that was shadowed by this one to `dependenciesDomain`, as it looks like it is an array used only for defining `Dependency` type

dd1023a88cec7a993b4ba0f4ac9e1dd67ede9a5d

PR #585, opened 2024-10-02
feat: show dimensions of inspected element

km1chno:
Resolves #584 

## Changes

Adds `{width} x {height}` label in inspector mode when an element is inspected (position fixed to left-bottom of preview). The measure is in `px`.

## Demo

https://github.com/user-attachments/assets/1642fdad-29f5-4c11-86d5-efe195d32622

PR #605, opened 2024-10-09
Fix small problems

filip131311:
This PR fixes a few small problems: 
1) it formats files added in #585
2) awaits isExpoGoProject in dependencyManager, before it was always false causing useDiagnosticAlert in expo go projects
3) adds new checks for android and ios native directories; before this change users that did not run `expo prebuild` were met with a generic build error alert that did not point them to the problem, now the information about the problem is added to the diagnostics panel, Radon IDE logs and a new message is added to build error alert to point user in the right direction 

Test plan: 
- run expo go project to test if IDE shows any unnecessary alerts 
- run non expo go application and remove android directory, then run android emulator and see the alerts that are shown by IDE


kmagiera (approved):
Please expand the description a bit to explain the goals of the changes as opposed to just listing what this PR is adding. Also please add test plan to the description.

PR #606, opened 2024-10-10
Let the release build use signed prebuilds from sim-server repo

kmagiera:
This PR changes the build script we use for simulator server release builds.

Since in the sim-server repo we have automated action that generates properly built and signed binaries, we should use that instead of building release version locally (especially that we need binaries for all the platforms).

The new script uses github CLI to download the pre-built binaries from sim-server repo releases, and to work requires that github CLI is installed and authenticated with an account that has access to the repo.

The script also verifies that the binaries we use for sim-server are from a tagged version. We do that by checking the tag of sim-server submodule. This way we get the tag for the release from which we should download the binaries.

As a consequence, the release builds of the extension can only use a tagged version of the sim-server (which is good, as would let as keep better track of the updates done between the two repos).

## Test plan
1. Make release build, make sure it runs ok


filip131311 (approved):
I approve, but also wonder if there is a better `curl` based approach that do not require gh CLI, if we published it publicly? As it is a binary to be published along with the extension, we don't loose any information that way, but users of `install-sim-server-release-build.sh` won't be required to has gh CLI setup.

kmagiera:
Utlimately, we will change the workflow such that sim-server repo publishes the artifacts in some publically available place. This is a temporary solution that just simplifies the workflow as otherwise I had to manually add windows binaries when building on mac and vice versa.

PR #608, opened 2024-10-10
Disable default context menu on windows

filip131311:
This PR disables default context menu on windows, as it clashed with right-click inspector functionality. 

Test plan: 
- open Radon IDE 
- right click inside app previe to trigger inspector

### DIFF
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -417,7 +417,7 @@ function Preview({
     // this is a fix that disables context menu on windows https://github.com/microsoft/vscode/issues/139824
     // there is an active backlog item that aims to change the behavior of context menu, so it might not be necessary
     // in the future https://github.com/microsoft/vscode/issues/225411
-    function onContextMenu(e) {
+    function onContextMenu(e: any) {
       e.stopImmediatePropagation();
     }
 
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -413,11 +413,16 @@ function Preview({
     }
   }
 
-  function onContextMenu(e: MouseEvent<HTMLDivElement>) {
-    e.preventDefault();
-  }
-
   useEffect(() => {
+    // this is a fix that disables context menu on windows https://github.com/microsoft/vscode/issues/139824
+    // there is an active backlog item that changes the bahaviour of context menu, so it might not be necessery
```
filip131311, line 418:
```suggestion
    // there is an active backlog item that aims to change the behavior of context menu, so it might not be necessary
```

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -426,7 +431,10 @@ function Preview({
       }
     }
     addEventListener("blur", onBlurChange, true);
-    return () => removeEventListener("blur", onBlurChange, true);
+    return () => {
+      window.removeEventListener("contextmenu", onContextMenu);
+      removeEventListener("blur", onBlurChange, true);
```
kmagiera, line 436: can we update `removeEventListener` here and also corresponding `addEventLIstener` call to specify whether it operates on document or window (or something else?)

kmagiera (approved):
Please add test plan to the PR description

PR #609, opened 2024-10-10
Rewrite copy shared script

filip131311:
This PR rewrites `copy.sh` script that was used to copy shared components to test-apps. 
The rewrite is necessary to work on windows without resulting to using tools that might not be present on user machine.


kmagiera (approved):
I'm ok with merging this. However, I'm now leaning towards just keeping copies of the shared stuff in the repo, and perhaps only have the script to synchronize these copies. In anyway, this is outside of this PR's scope

PR #617, opened 2024-10-13
Use bundle and add proper logging for pod installation

kmagiera:
This is a draft as it is build on top of #615

This PR refactors the way we install pods in the IDE. The main objective is to optionally use `bundle` command. This is similar to the changes proposed in #304 but apart from using bundle for the diagnostics check, we also use it when installing (which is a mechanism that wasn't in place when #304 was created).

The second main change that we're making is that we trigger pod installation when clean build is requested. Before that wasn't happening.

On top of that, we are including pod command output in the iOS build log. This will help diagnose potential problems better and will also provide some feedback for the scenarios when installation take very long.

Finally, we are fixing the startup message component such that it resets long wait flag and we are also adding a more prominent message "open logs" for the native build phase. This way it should be easier to spot for the users that they can see the build output in case the build takes long:
image

### How Has This Been Tested: 
1. Run clean build in on of the test apps for iOS. See cocoapods in the log output
2. Run RN 76 example on iOS to see bundle being used there instead of pod install (check logs)
3. When building on iOS, see that the "open logs" text appears in the startup message and you can click it to open the logs output.


packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -155,7 +155,7 @@ export class DependencyManager implements Disposable, DependencyManagerInterface
     return true;
   }
 
-  public async installPods(cancelToken: CancelToken) {
+  public async installPods(buildOutpuChannel: OutputChannel, cancelToken: CancelToken) {
```
p-malecki, line 158: typo:
buildOutpuChannel -> buildOutputChannel

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -164,16 +164,15 @@ export class DependencyManager implements Disposable, DependencyManagerInterface
...
+      const shouldUseBundle = await this.shouldUseBundleCommand();
+      const process = command(shouldUseBundle ? "bundle exec pod install" : "pod install", {
         cwd: iosDirPath,
         env: { ...env, LANG: "en_US.UTF-8" },
       });
-    };
-
-    try {
-      await cancelToken.adapt(commandInIosDir("pod install"));
+      lineReader(process).onLineRead((line) => buildOutpuChannel.appendLine(line));
```
p-malecki, line 174: typo:
buildOutpuChannel -> buildOutputChannel

p-malecki (approved):
Looks good, I left inline comment about typo.

PR #619, opened 2024-10-14
Make select content background darker in dark mode for distinction

### LINKED ISSUE #618: Manage devices - No visible border between manage devices window and drop-down list
Description:
No visible border between manage devices window and drop-down list
When a gray window appears on a gray window, sometimes the text is less readable because it is not known which window it belongs to, some distinction would be useful, e.g.

Image

km1chno:
Fixes #618 

#### Before: 

Zrzut ekranu 2024-10-14 o 09 40 03

#### After:

Zrzut ekranu 2024-10-14 o 09 39 50

#### On light color scheme (no problem, no changes)

Zrzut ekranu 2024-10-14 o 09 40 20

### How Has This Been Tested: 

Tested select menus in "Manage Device" modal and "Launch Configuration" modal on both dark and light color scheme.


kmagiera (approved):
👍

PR #624, opened 2024-10-15
Use bundle install before pod install when bundle is used

kmagiera:
In #617 we added an option for the IDE to use bundle command instead of pods directly as it is now recommended in new react native templates and setups. However, for bundle to work properly, we need to `bundle install` first. This was missed in that PR as I was testing a project where bundle install has already run.

### How Has This Been Tested: 
1. Delete rn 75 project and checkout a fresh copy
2. Open the fresh RN 75 project with the IDE

### DIFF
--- a/packages/vscode-extension/src/dependency/DependencyManager.ts
+++ b/packages/vscode-extension/src/dependency/DependencyManager.ts
@@ -167,10 +167,14 @@ export class DependencyManager implements Disposable, DependencyManagerInterface
     try {
       const env = getLaunchConfiguration().env;
       const shouldUseBundle = await this.shouldUseBundleCommand();
-      const process = command(shouldUseBundle ? "bundle exec pod install" : "pod install", {
-        cwd: iosDirPath,
-        env: { ...env, LANG: "en_US.UTF-8" },
-      });
+      const process = command(
+        shouldUseBundle ? "bundle install && bundle exec pod install" : "pod install",
+        {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -167,10 +167,14 @@ export class DependencyManager implements Disposable, DependencyManagerInterface
...
-      const process = command(shouldUseBundle ? "bundle exec pod install" : "pod install", {
-        cwd: iosDirPath,
-        env: { ...env, LANG: "en_US.UTF-8" },
-      });
+      const process = command(
+        shouldUseBundle ? "bundle install && bundle exec pod install" : "pod install",
+        {
+          shell: shouldUseBundle, // when using bundle, we need shell to run multiple commands
+          cwd: iosDirPath,
+          env: { ...env, LANG: "en_US.UTF-8" },
```
jakub-gonet, line 175: Would be nice to add LANG automatically.

jakub-gonet (approved):

p-malecki (approved):

packages/vscode-extension/src/dependency/DependencyManager.ts
kmagiera: I don't think we actually want to add it in all scenarios. I think we analyze the printed output here and so we add it such that we don't get the error messages translated.

PR #629, opened 2024-10-16
Fix `ReplayOverlay` rewind timer

balins:
This PR fixes a bug introduced when adding video replays in #579 that was making the timer freeze and sometimes show negative values during rewind.

The issue was caused by updating only the video player's current time and not propagating the change to the element rendering the overlay. We've seemed to rely on `handleTimeUpdate` callback registered on `"timeupdate"` event, but this event is not emitted when updating the current time of the video programmatically.
Sometimes, the timer indicated negative values during rewind, and this was also caused by setting the current time on video without updating the overlay's `time` accordingly, which in turn led to `time = 0` (default on mount) and `startTime = [video duration - 5]`, causing the timer to show negative values when `video duration > 5` (we calculate the time displayed on timer as `time - startTime`).

The proposed changes include passing a callback that allows the rewind function not only for updating the video's current time, but also the current time of the overlay component.
Another change is that we won't show `• REPLAY 0:00` until video metadata is loaded to avoid flickering from `0:00` to `video duration` when starting the rewind.

https://github.com/user-attachments/assets/fec826d1-b677-40fb-84cb-a13a792fe0ac

https://github.com/user-attachments/assets/b5f07a42-647d-4c34-8fc0-126df618a637

### How Has This Been Tested: 

Tested that with introduced changes, the timer behaves as expected both in rewind and normal playback.

### DIFF
--- a/packages/vscode-extension/src/webview/components/ReplayOverlay.tsx
+++ b/packages/vscode-extension/src/webview/components/ReplayOverlay.tsx
@@ -13,9 +13,9 @@ const INITIAL_REPLAY_LENGTH_SEC = 5;
  * the rewinding speeds up over time and also takes a fixed amount of time to rewind.
  */
 function acceleratedRewind(
-  video: HTMLVideoElement,
   fromTime: number,
   toTime: number,
+  setTimeCallback: (time: number) => void,
   readyCallback: () => void
 ) {
   const rewindTimeSec = 1.6;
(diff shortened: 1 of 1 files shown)


kmagiera (approved):
Code looks good. Can you just expand in the description why was the issue occurring and why the proposed changes fix it?

balins:
> Code looks good. Can you just expand in the description why was the issue occurring and why the proposed changes fix it?

Updated the description 👍

PR #630, opened 2024-10-16
Improve error recovery after metro process terminates

kmagiera:
Under some (yet uncertain) circumstances, metro process may be terminated. This causes the restart functionality to stop working and the only recovery option was clean rebuild or restarting the IDE tab.

The reason was that we had a `ready` call in metro, that would always use a single promise instance that is created when metro starts. When the startup was successful, the promise resolves and future call to `ready` would always be successful. This logic wouldn't consider the case when the process starts up fine but terminates later on. As a consequence, we'd always assume the metro process is healthy which resulted in the app being stuck in "waiting for app to load" state.

After fixing the above logic, by resetting the startup promise to reject once the metro process is terminated. I found some further flaws in the resetting logic.
1. When we call `selectDevice` from the restart flow. This codepath would never check for metro health assuming it is past this step and that metro should be always ready.
2. Even if restarting the app process would fail because of metro, we'd still only call `selectDevice` as an ultimate fallback in the reset flow. This way, we'd never end up restarting metro.

This PR fixes the following things:
1. When metro process is terminated, the startup promise is updated to always reject, this way, future attempts of waiting for metro will fail and allow for different recovery mechanism to activate
2. When selectDevice is called, along with waiting for devtools, or device preview, we also do metro.ready check which determines metro's process status
3. Finally, in `restart` call, we are reshuffling the calls, such that if any of the happy-path restart mechanisms fails, we always go into fallback that restarts all subprocesses and only then starts up the device. Before, we'd only restart the device which wasn't enough to recover from certain types of errors (i.e. when metro has died).

BTW: we are also adding a log statement in runtime that would help us diagnose problems where runtime couldn't be loaded.

This fixes an issue I encountered when testing the app. While the app was running, the metro process got terminated for unknown reason. This doesn't seem to happen very often but we should still have a more resilient mechanism for recovering from errors like that when the user clicks "restart" button.

Fixes # (issue)

### How Has This Been Tested: 
1. Boot up any test project
2. After it is started, use "Process Explorer" from VSCode cmd+P menu to terminate metro process
3. Click "restart" – before this change, the app would end up in "waiting for device to load" and we wouldn't be able to recover using "restart" button. Now "restarting" will spin up new metro instance and boot normally.

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -11,8 +11,6 @@ import {
   StartupMessage,
   TouchPoint,
 } from "../common/Project";
-import { DevicePlatform } from "../common/DeviceManager";
-import { AndroidEmulatorDevice } from "../devices/AndroidEmulatorDevice";
 import { getLaunchConfiguration } from "../utilities/launchConfiguration";
 import { DebugSession, DebugSessionDelegate } from "../debugging/DebugSession";
 import { throttle } from "../utilities/throttle";
@@ -94,8 +92,12 @@ export class DeviceSession implements Disposable {
         return true;
       case "reloadJs":
         if (this.devtools.hasConnectedClient) {
-          await this.metro.reload();
-          return true;
+          try {
+            await this.metro.reload();
+            return true;
+          } catch (e) {
+            Logger.error("Failed to reload JS", e);
+          }
         }
         return false;
     }
@@ -138,6 +140,7 @@ export class DeviceSession implements Disposable {
     }
 
     await Promise.all([
+      this.metro.ready(),
       this.device.startPreview().then((url) => (previewURL = url)),
       waitForAppReady,
     ]);
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -29,7 +29,6 @@ import { DebugSessionDelegate } from "../debugging/DebugSession";
 import { Metro, MetroDelegate } from "./metro";
 import { Devtools } from "./devtools";
 import { AppEvent, DeviceSession, EventDelegate } from "./deviceSession";
-import { CancelToken } from "../builders/cancelToken";
 import { PlatformBuildCache } from "../builders/PlatformBuildCache";
 
 const DEVICE_SETTINGS_KEY = "device_settings_v4";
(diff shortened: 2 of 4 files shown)


packages/vscode-extension/src/project/project.ts
```diff
@@ -282,32 +281,30 @@ export class Project
...
         return;
       }
-    }
 
-    // otherwise we try to restart the device session
-    try {
+      if (onlyReloadJSWhenPossible) {
+        // if reloading JS is possible, we try to do it first and exit in case of success
+        // otherwise we continue to restart using more invasive methods
+        if (await this.reloadMetro()) {
```
filip131311, line 303: I don't like that logic, as it is right now unsuccessful metro reload causes full device reboot, which seems unnecessary if we know that the reason it is failing is because metro was terminated.

It seems to me that `onlyReloadJSWhenPossible` option should first cause `${metroURL}/reload` as it is right now, then if it fails it should run `this.metro.start()` and only if it fails it should reboot the device

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -138,6 +140,7 @@ export class DeviceSession implements Disposable {
     }
 
     await Promise.all([
+      this.metro.ready(),
```
filip131311, line 143: why is it here? isn't `this.devtools.appReady()` promise dependent on metro process?

filip131311 (changes requested):
I left some inline comments about the fallback flow

packages/vscode-extension/src/project/deviceSession.ts
kmagiera: it is not, I added a note about that in the PR description

packages/vscode-extension/src/project/project.ts
kmagiera: From my testing the metro process crashes are extremely rare, so it's not very often that it'll restart the device because of that.

Note that it isn't as simple to restart the metro w/o running a series of other reload tasks, because it launches on ephemeral port. So if we just restart it, the existing process wouldn't be able to connect to it anyway, as it'd bind on a different port.

Handling this properly would require some more in-depth refactor of device session and reset logic in general that I'd rather avoid, as we are aiming to stabilize the current codebase.

filip131311: I agree it would be to complex of a change for now

filip131311 (approved):
I approve and add a ticket for future refactor that will prevent full device reboot.

PR #634, opened 2024-10-16
Fix misaligned lines in debugger

filip131311:
This PR fixes a problem with debugger on expo projects, caused by prelude lines added to the entry bundle file added by expo. The solution is t0 add the required offset to the prelude causing the problem and scanning for it when receiving source map in the debuger.

It is possible that the changes will no longer be needed in future versions of react native as the expo team tries to correct source map generation to include extra lines, for more information read those PRs:
- [expo side ](https://github.com/expo/expo/pull/29463)
- [metro side](https://github.com/facebook/metro/pull/1284)

This is why we add version check before adding lineOffset to initial source map.

### How Has This Been Tested: 
- open expo project and set a breakpoint, before the changes it would trigger 2 lines above the place it's been set. 
- save any change in the file containing breakpoint and make sure it still works.
- check if links to files generated next to console logs are pointing in to the correct place, again both before and after making changes to the file.
- check if error position indicator (when uncaught exception  is raised) points to the correct position. 

### Additional changes: 
This PR also fixes a minor mistake with `getReactNativeVersion` utility that does not need to be async and as it is used, we make it synchronous as part of this PR.

### DIFF
--- a/packages/vscode-extension/lib/metro_helpers.js
+++ b/packages/vscode-extension/lib/metro_helpers.js
@@ -41,14 +41,17 @@ function adaptMetroConfig(config) {
         module.output[0].data.code = `${preludeCode};var __REACT_DEVTOOLS_PORT__=${process.env.RCT_DEVTOOLS_PORT};`;
       }
     } else if (module.path === "__env__") {
-      // this handles @expo/env plugin, which is used to expose the number of lines in the prelude. 
-      // This is used to calculate the line number offset when reporting line numbers for 
-      // console.logs, breakpoints and uncaught exceptions. The reason why this is needed, is that
+      // this handles @expo/env plugin, which is used to inject environment variables
+      // the code below instantiates a global variable __EXPO_ENV_PRELUDE_LINES__ that stores
+      // the number of lines in the prelude. This is used to calculate the line number offset
+      // when reporting line numbers from the JS runtime, breakpoints
+      // and uncaught exceptions. The reason why this is needed, is that
       // metro doesn't include __env__ prelude in the source map resulting in the source map
       // transformation getting shifted by the number of lines in the prelude.
-      const expoEnvPreludeLines = module.output[0].data.lineCount;
-      process.stdout.write(JSON.stringify({ type: "RNIDE_Expo_Env_Prelude_Lines", expoEnvPreludeLines }));
-      process.stdout.write("\n");      
+      const expoEnvCode = module.output[0].data.code;
+      if (!expoEnvCode.includes("__EXPO_ENV_PRELUDE_LINES__")) {
+        module.output[0].data.code = `${expoEnvCode};var __EXPO_ENV_PRELUDE_LINES__=${module.output[0].data.lineCount};`;
+      }
     }
     return origProcessModuleFilter(module);
   };
--- a/packages/vscode-extension/lib/metro_helpers.js
+++ b/packages/vscode-extension/lib/metro_helpers.js
@@ -41,16 +41,14 @@ function adaptMetroConfig(config) {
         module.output[0].data.code = `${preludeCode};var __REACT_DEVTOOLS_PORT__=${process.env.RCT_DEVTOOLS_PORT};`;
       }
     } else if (module.path === "__env__") {
-      // this handles @expo/env plugin, which is used to inject environment variables
-      // the code below instantiates a global variable __EXPO_ENV_PRELUDE_LINES__ that stores
-      // the number of lines in the prelude. This is used to calculate the line number offset
-      // when reporting line numbers from the JS runtime. The reason why this is needed, is that
+      // this handles @expo/env plugin, which is used to expose the number of lines in the prelude. 
+      // This is used to calculate the line number offset when reporting line numbers for 
+      // console.logs, breakpoints and uncaught exceptions. The reason why this is needed, is that
       // metro doesn't include __env__ prelude in the source map resulting in the source map
       // transformation getting shifted by the number of lines in the prelude.
-      const expoEnvCode = module.output[0].data.code;
-      if (!expoEnvCode.includes("__EXPO_ENV_PRELUDE_LINES__")) {
-        module.output[0].data.code = `${expoEnvCode};var __EXPO_ENV_PRELUDE_LINES__=${module.output[0].data.lineCount};`;
-      }
+      const expoEnvPreludeLines = module.output[0].data.lineCount;
+      process.stdout.write(JSON.stringify({ type: "RNIDE_Expo_Env_Prelude_Lines", expoEnvPreludeLines }));
+      process.stdout.write("\n");      
     }
     return origProcessModuleFilter(module);
   };
--- a/packages/vscode-extension/lib/metro_helpers.js
+++ b/packages/vscode-extension/lib/metro_helpers.js
@@ -48,7 +48,12 @@ function adaptMetroConfig(config) {
       // and uncaught exceptions. The reason why this is needed, is that
       // metro doesn't include __env__ prelude in the source map resulting in the source map
       // transformation getting shifted by the number of lines in the expo generated prelude.
-      process.stdout.write(JSON.stringify({ type: "RNIDE_expo_env_prelude_lines", lineOffset: module.output[0].data.lineCount }));
+      process.stdout.write(
+        JSON.stringify({
+          type: "RNIDE_expo_env_prelude_lines",
+          lineCount: module.output[0].data.lineCount,
+        })
+      );
       process.stdout.write("\n");
     }
     return origProcessModuleFilter(module);
--- a/packages/vscode-extension/lib/metro_helpers.js
+++ b/packages/vscode-extension/lib/metro_helpers.js
@@ -42,16 +42,14 @@ function adaptMetroConfig(config) {
       }
     } else if (module.path === "__env__") {
       // this handles @expo/env plugin, which is used to inject environment variables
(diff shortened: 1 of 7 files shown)


packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -150,7 +153,22 @@ export class DebugAdapter extends DebugSession {
...
+            // in upcoming versions and if the changes are still not added bump the version below.
+            // more over offset should only be applied to the main bundle sourcemap as other files
+            // (generated during reload) do not contain the prelude causing the issue
+            const shouldApplyOffset =
+              semver.lte(getReactNativeVersion(), "0.76.0") && this.sourceMaps.length === 0;
+            this.sourceMaps.push([
+              message.params.url,
+              message.params.scriptId,
+              consumer,
+              shouldApplyOffset ? this.lineOffset : 0,
```
kmagiera, line 170: this seems to assume that we will get some console log call before source maps are provided such that this.lineOffect can be set.

kmagiera (changes requested):
I don't like the coupling between console calls that are also meant to provide the offset and the source map parser that seems to expect the console call prior to the "source maps ready" call.

I think we should instead restructure it in the following way:
1. Remove `__EXPO_ENV_PRELUDE_LINES__` entirely
2. Instead of it, we should log a command from metro process that will be captured by line reader. Similarily to how we do with RNIDE_initialize_done, but it will also contain the offset number
3. Read the offset in metro.ts using line reader and expose it such that we can provide it as a parameter to the debug session

filip131311:
@kmagiera  you were right, I changed it according to your suggestions.

packages/vscode-extension/lib/metro_helpers.js
```diff
@@ -41,16 +41,14 @@ function adaptMetroConfig(config) {
...
+      // This is used to calculate the line number offset when reporting line numbers for 
+      // console.logs, breakpoints and uncaught exceptions. The reason why this is needed, is that
       // metro doesn't include __env__ prelude in the source map resulting in the source map
       // transformation getting shifted by the number of lines in the prelude.
-      const expoEnvCode = module.output[0].data.code;
-      if (!expoEnvCode.includes("__EXPO_ENV_PRELUDE_LINES__")) {
-        module.output[0].data.code = `${expoEnvCode};var __EXPO_ENV_PRELUDE_LINES__=${module.output[0].data.lineCount};`;
-      }
+      const expoEnvPreludeLines = module.output[0].data.lineCount;
+      process.stdout.write(JSON.stringify({ type: "RNIDE_Expo_Env_Prelude_Lines", expoEnvPreludeLines }));
```
kmagiera, line 50: For consistency with RNIDE_initialize_done I wouldn't use capitalized words here

packages/vscode-extension/src/project/metro.ts
```diff
@@ -64,6 +68,7 @@ type MetroEvent =
 export class Metro implements Disposable {
   private subprocess?: ChildProcess;
   private _port = 0;
+  private _initialBundleLineOffset = 0;
```
kmagiera, line 71: I don't like "initial" word here as it isn't accurate. I think the offset applies for the whole bundle not just for the first bundle that gets loaded as it is currently implemented.

We should rename "mainBundleLineOffset"

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -150,7 +154,22 @@ export class DebugAdapter extends DebugSession {
...
-            this.sourceMaps.push([message.params.url, message.params.scriptId, consumer]);
+
+            // This line is here because of a problem with sourcemaps for expo projects,
+            // that was addressed in this PR https://github.com/expo/expo/pull/29463,
+            // unfortunately it still requires changes to metro that were attempted here
+            // https://github.com/facebook/metro/pull/1284 we should monitor the situation
+            // in upcoming versions and if the changes are still not added bump the version below.
+            // more over offset should only be applied to the main bundle sourcemap as other files
+            // (generated during reload) do not contain the prelude causing the issue
+            const shouldApplyOffset =
```
kmagiera, line 165: This check isn't accurate I think, we don't want to use the offset for the first source map only but for any source map that covers the prelude bit. Meaning that all JS reloads would also have this issue, while we may not handle them correctly. Have you tried using reloadJS with your changes now? Note that when we calculate generated position, we do a full loop over all the maps and if the entry is found in multiple maps, we take the last record. So the newer source map will be considered in that case.

kmagiera (changes requested):

kmagiera:
I'm now thinking that it may actually be the easiest solution to still include the offset in the bundle, and when we get source map we'd check if the offset variable exists there and read its value. If this would work, that'd be the best solution in my option. The only reason why it may not work is that I'm not sure if the preamble actually gets included in the source map at all. Tthe reason we're adding this offset was that there was an issue with that. But, perhaps, the issue was only because it is considered to have 0 length rather than just being completely omitted. You'll need to check the bundle for that to tell

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -150,7 +153,31 @@ export class DebugAdapter extends DebugSession {
             const decodedData = Buffer.from(base64Data, "base64").toString("utf-8");
             const sourceMap = JSON.parse(decodedData);
             const consumer = await new SourceMapConsumer(sourceMap);
-            this.sourceMaps.push([message.params.url, message.params.scriptId, consumer]);
+
+            // This is a heuristic that checks if the source map should contain __env__
+            // module that is added by expo, but not reported in the source map
+            const isFileWithOffset = sourceMap.sources.includes("__prelude__");
```
kmagiera, line 159: rename `isMainBundle` or `isFullBundle`

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -150,7 +153,31 @@ export class DebugAdapter extends DebugSession {
             const decodedData = Buffer.from(base64Data, "base64").toString("utf-8");
             const sourceMap = JSON.parse(decodedData);
             const consumer = await new SourceMapConsumer(sourceMap);
-            this.sourceMaps.push([message.params.url, message.params.scriptId, consumer]);
+
+            // This is a heuristic that checks if the source map should contain __env__
+            // module that is added by expo, but not reported in the source map
```
kmagiera, line 158:
```suggestion
            // We detect when a source map for the entire bundle is loaded by checking
            // if __env__ module is present in the sources.
```

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -96,6 +98,7 @@ export class DebugAdapter extends DebugSession {
     this.absoluteProjectPath = configuration.absoluteProjectPath;
     this.projectPathAlias = configuration.projectPathAlias;
     this.connection = new WebSocket(configuration.websocketAddress);
+    this.lineOffset = configuration.lineOffset;
```
kmagiera, line 101: This parameter should really be called `expoPreludeExtraLines` or something along these lines instead of being so generic

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -150,7 +153,31 @@ export class DebugAdapter extends DebugSession {
...
+            // module that is added by expo, but not reported in the source map
+            const isFileWithOffset = sourceMap.sources.includes("__prelude__");
+
+            // This line is here because of a problem with sourcemaps for expo projects,
+            // that was addressed in this PR https://github.com/expo/expo/pull/29463,
+            // unfortunately it still requires changes to metro that were attempted here
+            // https://github.com/facebook/metro/pull/1284 we should monitor the situation
+            // in upcoming versions and if the changes are still not added bump the version below.
+            const shouldApplyOffset =
+              semver.lte(getReactNativeVersion(), "0.76.0") && isFileWithOffset;
```
kmagiera, line 167: should we check for expo version here and not react native?

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -150,7 +153,31 @@ export class DebugAdapter extends DebugSession {
...
+
+            // This is a heuristic that checks if the source map should contain __env__
+            // module that is added by expo, but not reported in the source map
+            const isFileWithOffset = sourceMap.sources.includes("__prelude__");
+
+            // This line is here because of a problem with sourcemaps for expo projects,
+            // that was addressed in this PR https://github.com/expo/expo/pull/29463,
+            // unfortunately it still requires changes to metro that were attempted here
+            // https://github.com/facebook/metro/pull/1284 we should monitor the situation
+            // in upcoming versions and if the changes are still not added bump the version below.
```
kmagiera, line 165:
```suggestion
            // When using expo <${PUT_VERSION_HERE}, source maps skip the prelude module which is 
            // included in the main bundle at the start. As a result, all lines in the main bundle are shifted by
            // the amount of lines the prelude file adds. To offset this, we use the lineOffset parameter that we pass
            // to the source maps list that is used later on to correct line numbers when translating from generated
            // to the original positions.
```

kmagiera (approved):
Left a few more comments inline, but since only related to variable/parameter names and comments I'm accepting, as otherwise the code looks good.
The only comment that may require more attention is about whether we should test agains expo or react native version when determining when to use offset parameter and when not

PR #636, opened 2024-10-17
Fix Inspector right click 

filip131311:
This PR fixes the behavior of right click when using inspector tool, it removes the old `ContexMenu` in favor of `DropdownMenu`, which is more appropriate as it does not require `contextmenu` event to be activated, which since #608  is disabled. 

### How Has This Been Tested: 

- run any test application and use right click to inspect a component
- use inspector button and right click while it is active


kmagiera (approved):
I remember there was some issue with the right-click on windows which was also activaing the built-in context menu. Therefore this needs to be tested on Windows as well.

PR #638, opened 2024-10-18
Show unsupported platforms error message

### LINKED ISSUE #620: command RNIDE.diagnose not found
**Description**
after installing the extension the UI to start using it doesn't appear. the docs recommend using the diagnose command but i get this error

https://github.com/user-attachments/assets/0eb87fd0-1f0a-41e3-86f4-bf1c070f94a7

**Environment**

Expo project (SDK 51)
latest version of vscode 1.94
latest version of extension .19

p-malecki:
This PR fixes fixes issue where error messages for unsupported platforms were not displayed and prevents Xcode errors from appearing on Windows by ensuring the checkXcodeExists function is only executed on macOS.

Previously, an error related to the OS const in `platform.ts` was always thrown before the check in `extension.ts`. This change ensures the verification process happens correctly.

Fixes #620 

### How Has This Been Tested: 
- run extension on currently unsupported platform (linux)
- run extension and check the logs for any errors caused by xcrun on Windows

image

### DIFF
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -63,6 +63,11 @@ export function deactivate(context: ExtensionContext): undefined {
 export async function activate(context: ExtensionContext) {
   handleUncaughtErrors();
 
+  if (Platform.OS !== "macos" && Platform.OS !== "windows") {
+    window.showErrorMessage("Radon IDE works only on macOS and Windows.", "Dismiss");
+    return;
+  }
+
   setExtensionContext(context);
   if (context.extensionMode === ExtensionMode.Development) {
     enableDevModeLogging();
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/extension.ts
```diff
@@ -63,11 +63,6 @@ export function deactivate(context: ExtensionContext): undefined {
 export async function activate(context: ExtensionContext) {
   handleUncaughtErrors();
 
-  if (Platform.OS !== "macos" && Platform.OS !== "windows") {
```
filip131311, line 66: can you throw an error here after sending a popup message instead of throwing it in the platform utility? 

you could change te default to "unsupported" or something like that. I don't like the idea of utility that seems to be a static value crashing the application

filip131311 (changes requested):
inline comments

PR #640, opened 2024-10-21
Add Open Deep Link modal

### LINKED ISSUE #631: Add Deep Link dialog

km1chno:
Resolves #631 

## Changes

- Adds shared component `<SearchSelect/>`, which is used in Open Deep Link modal.
- The component allows to select option from scrollable list and provides search bar, which performs some simple version of fuzzy search, it doesn't support matching with misspellings, but allows for holes (basically for search phrase `abc` this is matching to regex `*a*b*c*`).
- Adds "Open Deep Link" modal that allows to input a deep link to open in app and browse history of recently opened deep links, which is cached in workspace state (only last 50 entries are cached).

## Demos (dark and light mode)

https://github.com/user-attachments/assets/2ac2018f-66c5-4c65-8a84-326b26e4106e

https://github.com/user-attachments/assets/0e65ccd7-1fbe-457f-9ad0-1228079a277a

### How Has This Been Tested: 

Tested in light and dark mode, on Iphone 15 Pro sim and Pixel 9 emulator, on projects `expo-router` and `react-native-75`.


kmagiera (changes requested):
While I haven't reviewed the code yet, wanted to add some UX related comments based on the videos you attached:
1) It doesn't feel good when the height of the dialog changes when you type (there are less choices as you narrow down the search, but shifting the modal makes it difficult to type or follow).
2) I think it'd make sense for this to work similarily to the browser URL bar, specifically it doesn't seem like we really need to input fields for the URL and for search separately. The main one should be a typeahead (using past history) input

kmagiera:
From the UX perspective it makes little sense to have the magnifier glass icon there. It is an input field that also narrows down the history list.

I'd perhaps just leave it without any icon and instead add placeholder "Enter deeplink or search history"

I'd also experiment with adding some label for the list of previous entries such that it is clear what we show there. For example "History" or "Previously used URLs"

km1chno:
@kmagiera what do you think about something like this?

Zrzut ekranu 2024-10-30 o 11 36 32

kmagiera:
Looks good!

packages/vscode-extension/src/project/project.ts
```diff
@@ -378,6 +381,22 @@ export class Project
...
+  async openDeepLink(link: string) {
+    const history = await this.getDeepLinksHistory();
+    if (history.length === 0 || link !== history[0]) {
+      extensionContext.workspaceState.update(
+        DEEP_LINKS_HISTORY_KEY,
+        [link, ...history.filter((s) => s !== link)].slice(0, DEEP_LINKS_HISTORY_LIMIT)
+      );
+    }
+
+    this.deviceSession?.sendDeepLink(link);
```
kmagiera, line 397: do we want to await here for consistency?

kmagiera (approved):
Looks good. Left some tiny inline comments to address before merging

km1chno:
I will get back to this once #675 is merged so we can use app level cache for storing recently used deep links

kmagiera:
There's no reason to wait, this is accepted for a few days already. #675 only changes the storage for build result metadata, we're still storing other settings and data in workspace-specific storage. It doesn't make sense to put URLs storage in that cache leaving other settings in workspace cache as it is conceptually closer to device settings than it is to native builds.

km1chno:
Ok then I'm merging this one. When the app level cache is created and we feel a need to move app level settings there from workspace state, then we can move cached deep links too.

PR #643, opened 2024-10-21
Disable device settings during booting and UI zoom changes

p-malecki:
This PR includes several updates:

- disables ability to modify any device settings during device booting to resolve `adb: device 'undefined' not found` error
- fixes unchanging background color on hover in zoom controls
- swaps the positions of the zoom in and zoom out buttons in the zoom controls
- corrects the alignment issues in dropdown items containing `KeybindingInfo`

**Testing:**
Verify that device settings are disabled during the booting process.

before:
imageř
Screenshot 2024-10-21 at 18 57 03

after:
Screenshot 2024-10-21 at 18 56 27
Screenshot 2024-10-21 at 18 56 06
Screenshot 2024-10-22 at 15 37 07

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -127,7 +127,7 @@ export interface ProjectInterface {
   updatePreviewZoomLevel(zoom: ZoomLevelType): Promise<void>;
 
   getDeviceSettings(): Promise<DeviceSettings>;
-  updateDeviceSettings(deviceSettings: DeviceSettings): Promise<void>;
+  updateDeviceSettings(deviceSettings: DeviceSettings, sendToDevice?: boolean): Promise<void>;
   sendBiometricAuthorization(match: boolean): Promise<void>;
 
   resumeDebugger(): Promise<void>;
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -127,7 +127,7 @@ export interface ProjectInterface {
   updatePreviewZoomLevel(zoom: ZoomLevelType): Promise<void>;
 
   getDeviceSettings(): Promise<DeviceSettings>;
-  updateDeviceSettings(deviceSettings: DeviceSettings, sendToDevice?: boolean): Promise<void>;
+  updateDeviceSettings(deviceSettings: DeviceSettings): Promise<void>;
   sendBiometricAuthorization(match: boolean): Promise<void>;
 
   resumeDebugger(): Promise<void>;
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -287,7 +287,7 @@ export class DeviceSession implements Disposable {
   }
 
   public async changeDeviceSettings(settings: DeviceSettings): Promise<boolean> {
-    const onlyReplaySettingsChanged =
+    const didOnlyReplaySettingsChanged =
       this.deviceSettings?.replaysEnabled !== settings.replaysEnabled;
     this.deviceSettings = settings;
 
@@ -297,7 +297,7 @@ export class DeviceSession implements Disposable {
       this.device.stopReplays();
     }
 
-    if (onlyReplaySettingsChanged) {
+    if (didOnlyReplaySettingsChanged) {
       // return value tells if device needs a restart
       return false;
     }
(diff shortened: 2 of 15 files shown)


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -286,14 +286,21 @@ export class DeviceSession implements Disposable {
...
       this.device.startReplays();
     } else {
       this.device.stopReplays();
     }
-    return this.device.changeSettings(settings);
+
+    if (sendToDevice) {
+      return this.device.changeSettings(settings);
+    }
+    return false; // return value tells if device needs a restart
```
km1chno, line 303: maybe lets put this comment above function header? it regards the function in general, not only this specific line

km1chno (approved):
looks good, just one little nit

packages/vscode-extension/src/common/Project.ts
```diff
@@ -127,7 +127,7 @@ export interface ProjectInterface {
   updatePreviewZoomLevel(zoom: ZoomLevelType): Promise<void>;
 
   getDeviceSettings(): Promise<DeviceSettings>;
-  updateDeviceSettings(deviceSettings: DeviceSettings): Promise<void>;
+  updateDeviceSettings(deviceSettings: DeviceSettings, sendToDevice?: boolean): Promise<void>;
```
kmagiera, line 130: Why do we need this extra argument? It appears like we should be able to tell which setting changed based on the deviceSettings object, and it should be up to the implementation to decide when the restart is needed and not up to the UI.

kmagiera (changes requested):

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -287,12 +287,21 @@ export class DeviceSession implements Disposable {
   }
 
   public async changeDeviceSettings(settings: DeviceSettings): Promise<boolean> {
+    const didOnlyReplaySettingsChanged =
+      this.deviceSettings?.replaysEnabled !== settings.replaysEnabled;
```
km1chno, line 291: you don't check if field `replaysEnabled` is the **only** difference, there might be other

if this is what you want to achieve, then please rename the reference to `didReplaySettingsChanged` or something like that

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -165,7 +165,7 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
               className="switch-root small-switch"
               id="enable-replays"
               onCheckedChange={(checked) =>
-                project.updateDeviceSettings({ ...deviceSettings, replaysEnabled: checked })
+                project.updateDeviceSettings({ ...deviceSettings, replaysEnabled: checked }, false)
```
kmagiera, line 168: revert this change

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -287,12 +287,23 @@ export class DeviceSession implements Disposable {
...
+
     this.deviceSettings = settings;
+
     if (settings.replaysEnabled && !this.isLaunching) {
       this.device.startReplays();
     } else {
       this.device.stopReplays();
     }
+
+    if (didOnlyReplaySettingsChanged) {
```
kmagiera, line 302: I looked up implementation of changeSettings and it seem like it only returns `true` when the parameter in question changes. So the only thing that this changes is the problem where we can trigger some commands during boot that the device doesn't like. 

Given that's the case, I don't think the introduced changes address the root cause of that, because even if we can enable replays during boot, changing any other device setting during that time would still result in that error.

I think this has been discussed in #590 already and the consensus was to block device settings entirely for the duration of boot which would also elliminate the need for this change.

kmagiera (changes requested):
Please review the PR description as there are some outdated statements there

kmagiera:
I'd suggest that this PR only focuses on the slight UI enhancements related to zoom controls and menu. The changes related to the logic for updating. the settings doesn't fit with the other minor changes. Also, as suggested in mycomment, it'd make more sense to block device setting interactions entirely like it's done in #590 – we can extract that code from that diff and have it make it to the next release

packages/vscode-extension/src/project/deviceSession.ts
p-malecki: You're correct, disabling all device settings during boot is better solution, so I've added it in update.

kmagiera (changes requested):
Instead of disabling every single item on the list, why don't we just disable the device settings menu button? The disabled option doesn't look very nice. For the button we can also add a tooltip that'd explain why its disabled.

kmagiera (approved):
Looks good. Next time I'd prefer that this type of changes are split into different PRs. The style changes were relatively straightforward and would've been accepted many revisions ago, whereas the device settings logic require some back and forth.

PR #644, opened 2024-10-21
Fix breakpoints in RN 76 apps

kmagiera:
This PR fixes an issue with breakpoints in RN 76 apps.

There were two core issues related to the breakpoints:
1) We'd not process metro-watchFolders aliases properly, the support for this source map alias was missing
2) We were not updating breakpoints with hermes after hot reload of individual files

As the reason for (1) was simply that we didn't add the implementation for watch folders, the reason for the bug related to (2) is more evolved: the culprit was in `updateBreakpointsInSource` method, were we'd use `mapping.source` directly from the source-map consumer to locate existing breakpoints. Because the stored breakpoints would always use local filesystem URLs, with the source using aliased name, we'd never locate the breakpoint that we should update. As a consequence, the breakpoints that were set for the files which then got hot reloaded, would never be set to stop at those newly loaded scripts.

To solve (2), we are now transforming source from the source map consumer into the absolute file path before we perform breakpoint lookup.

In order to fix (1), we are adding a more sophisticated mapping mechanims:
1) instead of `absoluteProjectPath` and `projectPathAlias` we now keep an array of source map aliases pairs
2) we use that new array to map from source map aliased path into filesystem paths
3) the map is constructed at the moment of creating debug session and following the way metro does it we set aliases for metro-project and metro-watchFolders
4) we extract the watch folders using metro config overrides, then use metro's stdout to pass that back to the extension process (same technique that we use for passing metro's port numer). In addition we accommodate for metro internals that prepend project root to watch folder after the config is loaded.

### How Has This Been Tested: 
1) Run RN 76 example (need #641 for this)
2) Set breakpoint somewhere in the code
3) Check it stops at the breakpoint
4) Add some lines before the breakpoint
5) Check again to see that the debugger stops at the altered breakpoint location (this was broken before this change)
6) Test debugger in RN 75 to verify this didn't break old debugger


p-malecki (approved):
I approve. It should also be tested on Windows, as the debugger and breakpoints did not function correctly there in projects with RN 75 and RN 76.

p-malecki:
It appears the problem still exists on Windows, debugger does not stop on breakpoints. Tested on RN 76 app after #641 update.

https://github.com/user-attachments/assets/b321aaee-f54a-40bb-97a7-1a6f3b19b416

PR #653, opened 2024-10-23
Support text paste on iOS

kmagiera:
This PR adds text paste support on iOS.

For Android, we use simulator-server to pass paste command. On iOS that was tricky to do and since there's a simctl command for paste we can use that instead.

In this change, we are overriding the sendPaste implementation for iOS that uses simctl pbcopy command that expects the pasted text to be sent via standard input.

In addition, we disable sending cmd+v key combination as HID keys to the simulator. It turned out to result in a buggy behavior when sometimes the key combination would trigger paste action and actually paste text into a focused text input, while most of the times it'd just do nothing. There were also instances where the previous text was pasted instead. Blocking this behavior isn't ideal but it puts iOS functionality in pair with Android while also avoiding a number of issues. It appears like the bugs are not caused by the IDE since we also observe these when testing paste functionality with the standard iOS simulator app.

Closes: #553

### How Has This Been Tested: 
1. Open any text example, add TextInput field
2. Copy some text using cmd+c and use cmd+v with the iOS simulator focused
3. Now when long-pressing the input field you can paste the copied text

### DIFF
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -207,6 +207,20 @@ export class IosSimulatorDevice extends DeviceBase {
     ]);
   }
 
+  public async sendPaste(text: string) {
+    const deviceSetLocation = getOrCreateDeviceSet(this.deviceUDID);
+    const subprocess = exec("xcrun", [
+      "simctl",
+      "--set",
+      deviceSetLocation,
+      "pbcopy",
+      this.deviceUDID,
+    ]);
+    subprocess.stdin?.write(text);
+    subprocess.stdin?.end();
+    await subprocess;
+  }
+
   private async changeLocale(newLocale: Locale): Promise<boolean> {
     const deviceSetLocation = getOrCreateDeviceSet(this.deviceUDID);
     const languageCode = newLocale.match(/([^_-]*)/)![1];
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -211,7 +211,8 @@ export class IosSimulatorDevice extends DeviceBase {
     ]);
   }
 
-  private pressingMetaKeys = 0; // 0 - not pressing, 1+ - how many meta keys are pressed
+  private pressingLeftMetaKey = false;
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -207,6 +211,43 @@ export class IosSimulatorDevice extends DeviceBase {
...
+    // Other times it kicks in before the pasteboard is filled with the content
+    // therefore pasting the previously compied content instead.
+    // As a temporary workaround, we disable sending cmd+V as key combination
+    // entirely to prevent this bugge behavior. Users can still paste content
+    // using the context menu.
+    // This is not an ideal workaround as people may still trigger cmd+v by
+    // pressing V first and then cmd, but it is good enough to filter out
+    // the majority of the noisy behavior since typically you press cmd first.
+    if (keyCode === LEFT_META_HID_CODE || keyCode === RIGHT_META_HID_CODE) {
+      this.pressingMetaKeys += direction === "Down" ? 1 : -1;
```
filip131311, line 228: This is not a good approach as it assumes that meatKey "up" event is going to happen, but if it won't it will brak the v key forever. 

How to reproduce a problem: 
1) Click `cmd` with Simulator focused 
2) still holding `cmd` key click on something outside the ide and the release the `cmd` key 
3) the `v` key is now unusable

filip131311 (changes requested):
left explanation inline

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
kmagiera: Doesn't the described scenario break all the keys though? In this case the simulator would think that the meta key is pressed the whole time so other keys shouldn't be interactive as well. I don't think fixing this is within the scope of this PR

filip131311: you are correct that this is a larger problem, but in the more general example the keyboard recovers after pressing and releasing `cmd` key, but this solution introduces a new bug that in this scenario disables exclusively the `v` key

kmagiera: I updated the code to use bool flags instead of the counter, so now you'll be able to recover from this issue by pressing meta key again

filip131311 (approved):

PR #655, opened 2024-10-23
Fix handling of uncaught errors

filip131311:
This PR fixes the problem with debugger ignoring uncaught errors after reloading JS. Because errors Global Handler  was only set on first connection to the debugger it would not be added after reload. The solution was to move the logic of setting the handler to happen on debugger event that is triggered both on first and consecutive application loads. 

### How Has This Been Tested: 

1. Run Test application and cause error
2. reload JS 
3. repet step 1). Before the change it would pass the error to the debug console and not stop the application
4. Test it both on application using RN version older then 76 and 76. 

Additional changes:
this PR also hides related internal logging from the users debug console

### DIFF
--- a/packages/vscode-extension/src/debugging/DebugAdapter.ts
+++ b/packages/vscode-extension/src/debugging/DebugAdapter.ts
@@ -203,11 +203,12 @@ export class DebugAdapter extends DebugSession {
     // some logs may baypass that, especially when printed in initialization phase, so we
     // need to detect whether the wrapper has added the stack info or not
     // We check if there are more than 3 arguments, and if the last one is a number
-    // We also check if the log is not internal to avoid exposing it as part of
-    // application logs.
+    // We filter out logs that start with __RNIDE_INTERNAL as those are messages
+    // used by IDE for tracking the app state and should not appear in the VSCode
+    // console.
     const argsLen = message.params.args.length;
     let output: OutputEvent;
-    if (message.params.args[0].value === "__RNIDE_INTERNAL") {
+    if (argsLen > 0 && message.params.args[0].value === "__RNIDE_INTERNAL") {
       return;
     } else if (argsLen > 3 && message.params.args[argsLen - 1].type === "number") {
       // Since console.log stack is extracted from Error, unlike other messages sent over CDP
--- a/packages/vscode-extension/src/debugging/DebugAdapter.ts
+++ b/packages/vscode-extension/src/debugging/DebugAdapter.ts
@@ -107,12 +107,6 @@ export class DebugAdapter extends DebugSession {
       this.sendCDPMessage("Debugger.setAsyncCallStackDepth", { maxDepth: 32 }).catch(ignoreError);
       this.sendCDPMessage("Debugger.setBlackboxPatterns", { patterns: [] }).catch(ignoreError);
       this.sendCDPMessage("Runtime.runIfWaitingForDebugger", {}).catch(ignoreError);
-      this.sendCDPMessage("Runtime.addBinding", {
-        name: "__RNIDE_onRuntimeLoaded",
-      });
-      this.sendCDPMessage("Runtime.evaluate", {
-        expression: "__RNIDE_onDebuggerReady()",
-      });
     });
 
     this.connection.on("close", () => {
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -203,9 +203,13 @@ export class DebugAdapter extends DebugSession {
     // some logs may baypass that, especially when printed in initialization phase, so we
     // need to detect whether the wrapper has added the stack info or not
     // We check if there are more than 3 arguments, and if the last one is a number
+    // We also check if the log is not internal to avoid exposing it as part of
+    // application logs.
     const argsLen = message.params.args.length;
     let output: OutputEvent;
-    if (argsLen > 3 && message.params.args[argsLen - 1].type === "number") {
+    if (message.params.args[0].value === "__RNIDE_INTERNAL") {
```
kmagiera, line 210: what if there are 0 args?

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -203,9 +203,13 @@ export class DebugAdapter extends DebugSession {
     // some logs may baypass that, especially when printed in initialization phase, so we
     // need to detect whether the wrapper has added the stack info or not
     // We check if there are more than 3 arguments, and if the last one is a number
+    // We also check if the log is not internal to avoid exposing it as part of
+    // application logs.
```
kmagiera, line 207:
```suggestion
    // We filter out logs that start with __RNIDE_INTERNAL as those are messages
    // used by IDE for tracking the app state and should not appear in the VSCode
    // console.
```

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -203,9 +203,13 @@ export class DebugAdapter extends DebugSession {
...
     const argsLen = message.params.args.length;
     let output: OutputEvent;
-    if (argsLen > 3 && message.params.args[argsLen - 1].type === "number") {
+    if (message.params.args[0].value === "__RNIDE_INTERNAL") {
+      return;
```
kmagiera, line 211: add comment here to explain why we don't do anything with those logs

kmagiera (changes requested):

packages/vscode-extension/src/debugging/DebugAdapter.ts
filip131311: isn't this in this explained by this comment? 
    
    // We filter out logs that start with __RNIDE_INTERNAL as those are messages
    // used by IDE for tracking the app state and should not appear in the VSCode
    // console.

kmagiera:
One more thing that comes to my mind here is that you should test this with RN 76 with the new debugger, as it may handle reloads differently compared to old one.

filip131311:
@kmagiera  you were right, the debugger creates execution context in a different moment in RN 76, I changed the approach to accommodate both versions.

packages/vscode-extension/src/debugging/DebugAdapter.ts
kmagiera: if the log doesn't appear anywhere what's its purpose?

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -144,6 +141,12 @@ export class DebugAdapter extends DebugSession {
         case "Debugger.scriptParsed":
           const sourceMapURL = message.params.sourceMapURL;
 
+          if (message.params.url) {
```
kmagiera, line 144: this deserves some explanation. what does this check do, how often do we execute onDebuggerReady

packages/vscode-extension/src/debugging/DebugAdapter.ts
filip131311: it appears in metro log feed with type: `client_log`, can be useful for us to debug when runtime started.

kmagiera: yes, that's exacly what you should write here to explain why we just return and do nothing with the log.

kmagiera (approved):

Issue #649, opened 2024-10-22
Selecing items from URL bar interfers with device touch system


kmagiera:
**Description**
Interacting with navigation bar (and possible other overlaying menus) can interfere with touch events sent to the device.

In the video below:
1) We position device such that the url bar menu covers the device when opened
2) We open bar and click some option. It is important that the option we click falls withing the device screen bounds that are underneath the menu
3) After the option is selected it appears like the device thinks we're still pressing the key despite we're not
4) After moving the cursor now over the device screen, it will start "swiping" despite the mouse button isn't pressed.

https://github.com/user-attachments/assets/3e126f90-8481-4dd2-b9f8-d41415d1862a

**Environment**
Reproduced this on 0.20 using Android as it seem to be easier to trigger the top control panel this way which makes the bug easily noticeable. But this will likely repro in any environment.

PR #657, opened 2024-10-23
Block touch and keypress events, hide multitouch indicators in replays

p-malecki:
This PR blocks sending of touch and keyboard events and hides multitouch indicators during replays.

#### Test plan:

1. Open a replay and attempt to use multitouch. Confirm that multitouch indicators are not visible.
2. Uncomment the "SIMSERVER_LOG": "debug" flag in the extension's config. Open a replay in the test app, click both inside and outside of the device screen, press any keyboard button and ensure that the logs do not include the "Sending touch" or Sending key" entries.

before:
https://github.com/user-attachments/assets/ba93504a-8008-4b70-9902-ea0f3fb7e2af

after:

https://github.com/user-attachments/assets/ae38420d-660d-4b0e-8811-5277808a2715

### DIFF
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -548,35 +548,33 @@ function Preview({
                 />
                 {replayData && <ReplayUI onClose={onReplayClose} replayData={replayData} />}
 
-                {isMultiTouching && (
-                  <div
-                    style={{
-                      "--x": `${touchPoint.x * 100}%`,
-                      "--y": `${touchPoint.y * 100}%`,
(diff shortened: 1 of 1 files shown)


kmagiera (changes requested):
I wonder whether there's some better way this can be handled such that we don't add more conditionals to this logic. This seems possible as with replays there's an overlay that should capture events and not pass them down. There's one similar issue I reported https://github.com/software-mansion/radon-ide/issues/649 and I was hoping we can have some more generic solution that'd allow us to avoid testing agains every other control that can be opened over the screen

p-malecki:
Suggestions have been applied in #685, so I am closing this PR.

PR #659, opened 2024-10-24
Remove previous app button

filip131311:
This PR removes the "previous app button" from top left corner of the ios simulator. Before this change If the user switched workspaces without shouting down the ios simulator, the button pointing to the previous app would appear in top left corner of the application. Clicking the button would break the IDE functionality as there is no way of navigating through the simulator and the only way to recover would be to restart app process (which would still  leave the button in the corner).

The solution it to terminate any open applications before launching a new one. 

Screenshot 2024-10-24 at 20 34 01

### How Has This Been Tested: 

- open an application ios using IDE
- switch workspaces without closing ide 
- in the new workspace open the ios application using the same device 
- switch workspaces back to the original one without closing ide

### DIFF
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -263,8 +263,43 @@ export class IosSimulatorDevice extends DeviceBase {
     }
   }
 
+  /**
+   * This function terminates any running applications. Might be useful when you launch a new application
+   * before terminating the previous one.
+   */
+  async terminateAnyRunningApplications() {
+    const deviceSetLocation = getOrCreateDeviceSet(this.deviceUDID);
+    const { stdout } = await exec("xcrun", [
+      "simctl",
+      "--set",
+      deviceSetLocation,
+      "listapps",
+      this.deviceUDID,
+    ]);
+
+    const regex = /ApplicationType = User;\s*[^{}]*?\bCFBundleIdentifier = "([^"]+)/g;
+
+    let match;
+    while ((match = regex.exec(stdout)) !== null) {
+      const bundleID = match[1];
+      // Terminate the app if it's running:
+      try {
+        await exec(
+          "xcrun",
+          ["simctl", "--set", deviceSetLocation, "terminate", this.deviceUDID, bundleID],
+          { allowNonZeroExit: true }
+        );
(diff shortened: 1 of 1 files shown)


kmagiera (commented):
Not sure if I like this approach, it may terminate some system processes if I understand correctly. Instead of doing that, maybe there is a way to force opening home screen right before launching. This way the handoff button should be gone as well?

filip131311:
This Portion of the regex `ApplicationType = User;` makes sure we don't try to close any system processes/applications and It was tricky to find a way to force simulator to home screen, there might be a way on sim-server side, but I couldn't find a command through `xcrun`.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -263,8 +263,43 @@ export class IosSimulatorDevice extends DeviceBase {
...
+    let match;
+    while ((match = regex.exec(stdout)) !== null) {
+      const bundleID = match[1];
+      // Terminate the app if it's running:
+      try {
+        await exec(
+          "xcrun",
+          ["simctl", "--set", deviceSetLocation, "terminate", this.deviceUDID, bundleID],
+          { allowNonZeroExit: true }
+        );
```
kmagiera, line 291: would likely be better to use Promise.all to await on all the commands

kmagiera (changes requested):

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
filip131311: done

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -263,8 +263,53 @@ export class IosSimulatorDevice extends DeviceBase {
...
+      } catch (e) {
+        // terminate will exit with non-zero code when the app wasn't running. we ignore this error
+      }
+    };
+
+    await Promise.all(
+      matches.map((bundleID) => {
+        return terminateApp(bundleID);
+      })
+    );
```
kmagiera, line 305:
```suggestion
    await Promise.all(matches.map(terminateApp));
```

kmagiera (approved):

PR #661, opened 2024-10-25
Set minimum size for the device in zoom controls

### LINKED ISSUE #654: Set some minimum size for the device when in "fit" zoom mode
**Description**
Currently "fit" causes the device to shrink indefinitely. This isn't very practical as, for example, here's how it look like when opening logs on quite typical screen resolution (13inch MBP):
image

Instead, we should have some minimum size to which it resizes to still allow for the device screen to be readable.

**Environment**
0.21

p-malecki:
This PR fixes the issue where the "fit" mode causes indefinite shrinking of the device display. Included changes:
- sets a minimum device height of 350px in the CSS, which restricts the "fit" mode
- adds minimum zoom level of 0.4 (`MIN_ZOOM_LEVEL`) to prevent excessive zooming out

Fixes #654 

### How Has This Been Tested: 
- `MIN_ZOOM_LEVEL` and minimal device height in CSS settings were chosen and tested on a 14-inch MacBook 
- Select "fit" mode and adjust the tab's height. Use the zoom out button to zoom out as far as possible. In both cases, ensure that the minimum device size display remains readable.

before:
https://github.com/user-attachments/assets/5fbb86e1-ad9f-4338-9070-6679d9b26d08

after:
https://github.com/user-attachments/assets/8fa2437d-fa49-4702-8464-89652a21e8f6

### DIFF
--- a/packages/vscode-extension/src/webview/hooks/useResizableProps.tsx
+++ b/packages/vscode-extension/src/webview/hooks/useResizableProps.tsx
@@ -2,7 +2,11 @@ import { ResizeCallback } from "re-resizable";
 import { CSSProperties, useCallback, useEffect, useState } from "react";
 import { DeviceProperties } from "../utilities/consts";
 import { ZoomLevelType } from "../../common/Project";
-import { DEVICE_DEFAULT_SCALE } from "../components/ZoomControls";
+import {
+  DEVICE_DEFAULT_SCALE,
+  MIN_FIT_ZOOM_LEVEL,
+  FIT_MODE_WINDOW_HEIGHT_THRESHOLD,
+} from "../components/ZoomControls";
 
 type UseResizableProps = {
   wrapperDivRef: React.RefObject<HTMLDivElement>;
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/webview/hooks/useResizableProps.tsx
```diff
@@ -31,10 +35,17 @@ export function useResizableProps({
   const calculatePhoneDimensions = useCallback(
     (delta = 0) => {
       if (zoomLevel === "Fit") {
-        setPhoneHeight("100%");
-        setMaxWidth("100%");
+        const windowHeight = document.getElementById("root")?.clientHeight ?? 0;
```
filip131311, line 38: I think it would be nice to use media query in css here, aside from code aesthetic that  I think would improve, if you aggressively resize the hight in the current implementation you can see a little "jump" back up in size to the `MIN_FIT_ZOOM_LEVEL` when the device changes sizes between `(windowHeight < FIT_MODE_WINDOW_HEIGHT_THRESHOLD)` condition and other scenarios and I  think it could be avoided with media query.

filip131311 (commented):
I left inline comment

packages/vscode-extension/src/webview/hooks/useResizableProps.tsx
p-malecki: I don't think that a media query is a solution here unless we want to enforce MIN_FIT_ZOOM_LEVEL regardless of zoom mode. If yes then in that case media query is not necessary.

filip131311 (approved):
LGTM

PR #662, opened 2024-10-25
Add scrollbars to device select and device settings

p-malecki:
This PR adds scrollbars to the `device select` and the `device settings` dropdown.

### How Has This Been Tested:
- Added a large number of devices to extend the device list, then reduced the window heigh to test the scrollbar in the device select.
- Opened the device settings to verify that the scrollbar operates as expected.

after:
https://github.com/user-attachments/assets/144a1d58-9a74-4d1b-b61b-61455c026052

### DIFF
--- a/packages/vscode-extension/src/webview/components/DeviceSelect.css
+++ b/packages/vscode-extension/src/webview/components/DeviceSelect.css
@@ -123,7 +123,7 @@
   justify-content: center;
   border-radius: 16px;
   color: var(--swm-default-text);
-  background-color: var(--swm-device-icon-background);
+  background-color: var(--swm-device-select-rich-item-icon);
 }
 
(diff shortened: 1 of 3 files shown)


kmagiera:
Do the scrollbars appear when there's enough room to display all entries?

p-malecki:
> Do the scrollbars appear when there's enough room to display all entries?

The scrollbar in the device select is actually not visible at all, there are scrollUp and scrollDown buttons that become visible only when there is no room for all device records. For the device settings, the scrollbar only becomes visible when there isn't enough space to display all options.

Screenshot 2024-10-28 at 09 15 43

https://github.com/user-attachments/assets/4cb8fe50-6536-4e28-a038-f1a43e1edd98

PR #663, opened 2024-10-25
MultiRoot Workspace support

filip131311:
1) This PR refactors how we handle the `appRoot` to allow for changing it in runtime. To make sure that the app root is unified across all the components we now store it in `project.ts` and move all objects dependent on `appRoot` to project as well. We listen for launchConfiguration changes and restart everything dependent on appRoot after it changes. 

2) We refactor how appRoot is detected: 
after analyzing the old approach that was using `workspace.findFiles` we realized that on more elaborate setups it took almost 2 seconds to traverse all of the workspace directories. To combat that we introduce a new `findAppRootCandidates` implementation that takes only a couple of ms in most setups. 
*note*: the new function is not perfectly equivalent to the old one as it removes a check if `node_modules` contains `react-native` it is done on purpose as it has a following drawbacks:
- generetes false positives as some not applications directories (e.g. libries) might contain `react-native` in its node modules,
- mono repositories might generate false positives at top level  depending on the implementation
- it requires searching through `node_modules` which are usually quite large
- it is not even that useful, because it requires `node_modules` to be installed and first time users could encounter problems if we relied on this check.

3) Additionally we enhance the Lunch Configuration View to allow the user for selecting the appRoot from the list. 

### How has dis been tested? 
- Run the mono repo with 2 application and switch between them.

- Run `react-native-svg` repo and select test apps in it 
- Run `react-native-reanimated` repo and select test apps in it, additionally check if cashes are stored properly after switching between applications.
- To benchmark the `findAppRootCandidates` I added the following code to at the end of the `activate` function in the extension: 

```js
const start = performance.now();

  await findAppRootFolder() 

  const end = performance.now()

  Logger.debug("My benchmark result:", end-start);
```

note: that the benchmark make sens only if you don't have appRoot set in launchConfiguration.

Results for `react-native-reanimated` repository:

| old    | new |
| -------- | ------- |
| 1479.05  ms| 1.27  ms |


packages/vscode-extension/src/builders/BuildCache.ts
```diff
@@ -37,30 +37,31 @@ function makeCacheKey(platform: DevicePlatform, appRoot: string) {
...
-  constructor(private readonly platform: DevicePlatform, private readonly appRoot: string) {
-    this.cacheKey = makeCacheKey(platform, appRoot);
-  }
+  constructor(private readonly appRootFolder: string) {}
 
   /**
    * Passed fingerprint should be calculated at the time build is started.
    */
-  public async storeBuild(buildFingerprint: string, build: BuildResult) {
+  public async storeBuild(buildFingerprint: string, build: BuildResult, platform: DevicePlatform) {
```
kmagiera, line 45: can't we use `buiild.platform` here instead of providing platform as extra argument?

packages/vscode-extension/src/builders/customBuild.ts
```diff
@@ -18,9 +17,10 @@ export async function runExternalBuild(
   cancelToken: CancelToken,
   buildCommand: string,
   env: Env,
-  platform: DevicePlatform
+  platform: DevicePlatform,
+  appRoot: string
```
kmagiera, line 21: nit: in this particular place I'd just call it `cwd` here as the method is pretty abstract, so it doesn't need to have a context that it runs in an app root IMO

packages/vscode-extension/src/builders/customBuild.ts
```diff
@@ -64,16 +64,21 @@ export async function runExternalBuild(
   return binaryPath;
 }
 
-export async function runfingerprintCommand(externalCommand: string, env: Env) {
-  const output = await runExternalScript(externalCommand, env);
+export async function runfingerprintCommand(externalCommand: string, env: Env, appRoot: string) {
```
kmagiera, line 67: same here

packages/vscode-extension/src/builders/customBuild.ts
```diff
@@ -64,16 +64,21 @@ export async function runExternalBuild(
...
   }
   return output.lastLine;
 }
 
-async function runExternalScript(externalCommand: string, env: Env, cancelToken?: CancelToken) {
-  let process = command(externalCommand, { cwd: getAppRootFolder(), env });
+async function runExternalScript(
+  externalCommand: string,
+  env: Env,
+  appRoot: string,
```
kmagiera, line 78: and here

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -69,24 +70,23 @@ export class DependencyManager implements Disposable, DependencyManagerInterface
...
     });
 
     this.checkProjectUsesExpoRouter();
     this.checkProjectUsesStorybook();
     this.checkEasCliInstallationStatus();
   }
 
-  public async validateNodeVersion() {
-    const appRootFolder = getAppRootFolder();
+  public async validateNodeVersion(appRoot: string) {
```
kmagiera, line 87: given DependencyManager already stores appRoot, we shouldn't require it as argument.

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -203,9 +208,9 @@ export class DependencyManager implements Disposable, DependencyManagerInterface
     Logger.debug("Project pods installed");
   }
 
-  private async getPackageManager() {
+  private async getPackageManager(appRoot: string) {
```
kmagiera, line 211: same here, we should use `this.appRootFolder` here, why isn't this consistent?

kmagiera (commented):
Haven't finished revieweing everything but left a few comments already

packages/vscode-extension/src/extension.ts
```diff
@@ -446,8 +297,9 @@ async function captureScreenshot() {
 }
 
 async function diagnoseWorkspaceStructure() {
-  const appRootFolder = await configureAppRootFolder();
+  const appRootFolder = findAppRootFolder() !== undefined;
   if (appRootFolder) {
+    commands.executeCommand("setContext", "RNIDE.extensionIsActive", true);
```
kmagiera, line 302: I don't like that the diagnose method has a side effect like this. We should restructure it and set the context elsewhere. Now running diagnostics will actually fix the editor context ?

packages/vscode-extension/src/project/project.ts
```diff
@@ -101,20 +142,83 @@ export class Project
...
+    );
+  }
+
+  private async setupAppRoot() {
+    const appRoot = findAppRootFolder();
+    if (!appRoot) {
+      return;
+    }
+
+    this.appRootFolder.setAppRoot(appRoot);
```
kmagiera, line 209: It doesn't seem like having the AppRootFolder class makes sense any longer. We just use it in project and when the app root changes we also only call a code that is defined in the project class.

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -15,15 +19,187 @@ export const extensionContext = new Proxy<ExtensionContext>({} as ExtensionConte
   },
 });
 
-let _appRootFolder: string | null = null;
+export class AppRootFolder {
```
kmagiera, line 22: I think it'd be cleaner if we get rid of this class and just rely on project controlling changes to appRootFolder value

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -15,15 +19,187 @@ export const extensionContext = new Proxy<ExtensionContext>({} as ExtensionConte
...
-  return _appRootFolder;
+
+  const { android, appRoot, ios, isExpo, metroConfigPath, env } = RNIDEConfiguration;
+
+  return { android, appRoot, ios, isExpo, metroConfigPath, env };
+};
+
+export function findAppRootCandidates(maxSearchDepth: number = 3): string[] {
+  const candidates: string[] = [];
+  const searchedFileNames = [
```
kmagiera, line 78: in the previous approach we were also looking for folders that have node_modules with react-native installed. Do we want to change this?

packages/vscode-extension/src/utilities/reactNative.ts
```diff
@@ -1,8 +1,6 @@
 import path from "path";
-import { getAppRootFolder } from "./extensionContext";
 
-export function getReactNativeVersion() {
-  const workspacePath = getAppRootFolder();
+export function getReactNativeVersion(workspacePath: string) {
```
kmagiera, line 3: maybe we can rename to appRootFolder here?

kmagiera (commented):
Left a couple more inline comments

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -15,15 +19,187 @@ export const extensionContext = new Proxy<ExtensionContext>({} as ExtensionConte
...
+    "app.json",
+    "app.config.js",
+    "app.config.ts",
+  ];
+
+  // In order to optimize the search time we exclude directories,
+  // that shouldn't contain applications.
+  const excludedDirectoryPatterns: RegExp[] = [/^node_modules$/, /^ios$/, /^android$/, /^\..+/];
+
+  const searchQueue: [string, number][] | undefined = workspace.workspaceFolders?.map(
```
jwajgelt, line 90: I'd rather have this be a record rather than a tuple.
```suggestion
  const searchQueue: {path: string, searchDepth: number}[] | undefined = workspace.workspaceFolders?.map(
```
Right now, you have to read the whole function to figure out what the `string` and `number` here are.

Or even better, make this a named type, since you refer to it in other places.

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -15,15 +19,187 @@ export const extensionContext = new Proxy<ExtensionContext>({} as ExtensionConte
...
+    (workspaceFolder) => {
+      return [workspaceFolder.uri.path, 0];
+    }
+  );
+
+  if (searchQueue === undefined) {
+    Logger.warn("[FindFiles] Could not determine active workspace");
+    return [];
+  }
```
jwajgelt, line 99: I think it'd make more sense to instead early return on `workspace.workspaceFolders` being undefined.
```suggestion
  if (workspace.workspaceFolders === undefined) {
    Logger.warn("[FindFiles] Could not determine active workspace");
    return [];
  }

  const searchQueue: [string, number][] = workspace.workspaceFolders.map(
    (workspaceFolder) => {
      return [workspaceFolder.uri.path, 0];
    }
  );
```

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -15,15 +19,187 @@ export const extensionContext = new Proxy<ExtensionContext>({} as ExtensionConte
...
+      searchQueue.push([currentDir![0] + "/" + dirEntry.name, currentDir![1] + 1]);
+    });
+    currentDir = searchQueue.shift();
+  }
+
+  if (candidates.length > 1) {
+    Logger.warn(
+      `Found multiple directories containing one or more of ${searchedFileNames} files in the workspace`
+    );
+  }
```
jwajgelt, line 133: Is this something we want to `warn` about? Isn't this expected in a monorepo?

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -15,15 +19,187 @@ export const extensionContext = new Proxy<ExtensionContext>({} as ExtensionConte
...
+      }
+
+      if (excludedDirectoryPatterns.some((pattern) => pattern.test(dirEntry.name))) {
+        return;
+      }
+
+      searchQueue.push([currentDir![0] + "/" + dirEntry.name, currentDir![1] + 1]);
+    });
+    currentDir = searchQueue.shift();
+  }
```
jwajgelt, line 127: NIT:
I'd move this to a separate function for clarity.

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -15,15 +19,187 @@ export const extensionContext = new Proxy<ExtensionContext>({} as ExtensionConte
...
+    }
+  );
+
+  if (searchQueue === undefined) {
+    Logger.warn("[FindFiles] Could not determine active workspace");
+    return [];
+  }
+
+  let currentDir: [string, number] | undefined = searchQueue.shift();
+  while (currentDir !== undefined) {
```
jwajgelt, line 102: I'd condition the loop on `searchQueue.length` and declare `currentDir` inside its body instead.

packages/vscode-extension/src/utilities/extensionContext.ts
jwajgelt, line 127: I think the body of this `forEach` callback is pretty complex.
I'd rewrite it into two array operations instead, one for checking if you're the candidate and another for selecting children to search. Something along the
```suggestion
    const isCandidate = filesAndDirs.some(dirEntry => {
      return dirEntry.isFile() && searchedFileNames.includes(dirEntry.name);
    });
    
    if (isCandidate) { candidates.push(currentDir[0]) }
    
    const childrenToSearch = filesAndDirs.filter(dirEntry => {
      return !excludedDirectoryPatterns.some((pattern) => pattern.test(dirEntry.name))
    }).map(dirEntry => [currentDir![0] + "/" + dirEntry.name, currentDir![1] + 1]);
    searchQueue.push(...childrenToSearch);

    currentDir = searchQueue.shift();
  }
```

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -15,15 +19,187 @@ export const extensionContext = new Proxy<ExtensionContext>({} as ExtensionConte
...
+    return {};
   }
-  return _appRootFolder;
+
+  const { android, appRoot, ios, isExpo, metroConfigPath, env } = RNIDEConfiguration;
+
+  return { android, appRoot, ios, isExpo, metroConfigPath, env };
+};
+
+export function findAppRootCandidates(maxSearchDepth: number = 3): string[] {
```
jwajgelt, line 76: Nit: this shouldn't be hard to rewrite as a [generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*) instead.
This would allow you to only look for the first candidate when activating the extension, and get all of the candidates when generating the list of possible appRoots with the same implementation.

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -201,8 +201,11 @@ async function isBunModulesInstalled(): Promise<boolean> {
   return false;
 }
 
-export async function isNodeModulesInstalled(manager: PackageManagerInfo): Promise<boolean> {
-  const workspacePath = manager.workspacePath ?? getAppRootFolder();
+export async function isNodeModulesInstalled(
+  manager: PackageManagerInfo,
+  appRoot: string
```
jwajgelt, line 206: Maybe `PackageManagerInfo` should hold the `appRoot` instead? Isn't the "Package Manager Info" only relevant in the context of a specific "app root"?

packages/vscode-extension/src/webview/providers/LaunchConfigProvider.tsx
```diff
@@ -63,9 +76,22 @@ export default function LaunchConfigProvider({ children }: PropsWithChildren) {
     [config, setConfig]
   );
 
+  const addCustomApplicationRoot = (appRoot: string) => {
+    const newState = [...applicationRoots, appRoot];
+    setApplicationRoots(newState);
+    launchConfig.addCustomApplicationRoot(appRoot);
```
jwajgelt, line 82: Could we just do:
```suggestion
    launchConfig.addCustomApplicationRoot(appRoot);
```
and rely on the `launchConfig` listener to update it for us?

jwajgelt (changes requested):

packages/vscode-extension/src/extension.ts
```diff
@@ -248,33 +242,16 @@ export async function activate(context: ExtensionContext) {
   );
 
   context.subscriptions.push(
-    workspace.onDidChangeConfiguration((event: ConfigurationChangeEvent) => {
+    workspace.onDidChangeConfiguration(async (event: ConfigurationChangeEvent) => {
```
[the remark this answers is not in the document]
filip131311: this is a remnant of a previous approach thank you for pointing it out.

packages/vscode-extension/src/extension.ts
filip131311: It is the old behavior of this command. I agree that it inappropriately named but I would like to constrain the out facing changes to things strictlly related to setting and selecting appRoot as this PR is quite large any way and this seems like a scope creep. I will adres this comment in the fallow up.

packages/vscode-extension/src/builders/customBuild.ts
```diff
@@ -64,16 +64,21 @@ export async function runExternalBuild(
...
+  cancelToken?: CancelToken
+) {
+  let process = command(externalCommand, { cwd: appRoot, env });
```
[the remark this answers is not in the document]
filip131311: This seems to be a serious problem, that we didn't think about before serious addressing multiRoot configuration, there are more similar problems as well for example right now our configuration is workspace specific for all configuration values so there is now way of setting appRoot specific env. But will adress it in future PRs

packages/vscode-extension/src/utilities/extensionContext.ts
filip131311: nice catch i was thinking in therm of the old aproach

packages/vscode-extension/src/utilities/extensionContext.ts
filip131311: what do you mean by `this`

filip131311: The point of creating this function was to optimize and iterators in js are not lazy, so this is not a zero-cost abstraction, I agree this would be more readable, but It is a worthy trade of in my opinion

packages/vscode-extension/src/utilities/packageManager.ts
filip131311: I'm not sure if it ever gonna be useful in the future, but it is not. You theoretically can call `nmp install` in any workspace directory and with the correct setiting all of the will be installed

packages/vscode-extension/src/webview/providers/LaunchConfigProvider.tsx
filip131311: yes, thank you

packages/vscode-extension/src/utilities/extensionContext.ts
filip131311: type script is not to happy when you do that I'll leave it as it is

packages/vscode-extension/src/project/project.ts
```diff
@@ -163,14 +143,14 @@ export class Project
...
+      workspace.onDidChangeConfiguration((event: ConfigurationChangeEvent) => {
         if (event.affectsConfiguration("launch")) {
           const config = getCurrentLaunchConfig();
-          const oldAppRoot = this.appRootFolder?.getAppRoot();
+          const oldAppRoot = this.appRootFolder;
           if (config.appRoot === oldAppRoot) {
             return;
           }
-          const newAppRoot = await this.setupAppRoot();
+          const newAppRoot = this.setupAppRoot();
```
jwajgelt, line 153: I think `setupAppRoot` shouldn't return anything -- I think it's a bit confusing when a function is used both for side effects (here, updating the object state) and returning a value.
Just call `setupAppRoot` here and use `this.appRootFolder` instead.

packages/vscode-extension/src/project/project.ts
```diff
@@ -76,11 +75,11 @@ export class Project
...
 
-  public launchConfig: LaunchConfigController;
-  public dependencyManager: DependencyManager;
+  public launchConfig?: LaunchConfigController;
+  public dependencyManager?: DependencyManager;
 
-  private appRootFolder: AppRootFolder;
-  private buildCache: BuildCache;
+  private appRootFolder?: string;
+  private buildCache?: BuildCache;
```
jwajgelt, line 83: Right now, the type system allows setting these to undefined, which means that:
1. we have to handle this case everywhere, even though it should be unreachable
2. it won't stop us from accidentally setting these fields to `undefined`.

I think we could work around that by "bundling" these dependencies into a separate object (`ProjectContext`, `ApplicationContext`, I don't know) that:
1. in the constructor, sets up the things that depend on the `appRoot`
2. has a `dispose()` method that cleans them up?
Then we could replace calling `setupAppRoot()` with calling 
```
this.applicationContext.dispose();
this.applicationContext = new ApplicationContext(newAppRoot);
```
and we would have to deal with the `undefined` case here.
The `launchConfig` and `dependencyManager` can be exposed via `get`ters to preserve current API.

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -1,11 +1,13 @@
 import fs from "fs";
-import { commands, ExtensionContext, Uri, workspace, window, Disposable } from "vscode";
+import { commands, ExtensionContext, Uri, workspace, window } from "vscode";
 import { Logger } from "../Logger";
 import { getLaunchConfiguration } from "./launchConfiguration";
 import { LaunchConfigurationOptions } from "../common/LaunchConfig";
 
 let _extensionContext: ExtensionContext | null = null;
 
+type searchItem = { path: string; searchDepth: number };
```
jwajgelt, line 9: NIT:
```suggestion
type SearchItem = { path: string; searchDepth: number };
```
Capitalizing types makes the code easier to read at a glance.

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -87,31 +55,57 @@ export function findAppRootCandidates(maxSearchDepth: number = 3): string[] {
...
+
+  let currentDir: searchItem | undefined = searchQueue.shift();
   while (currentDir !== undefined) {
-    if (currentDir[1] > maxSearchDepth) {
+    if (currentDir.searchDepth > maxDepth) {
       break;
     }
 
-    const filesAndDirs = fs.readdirSync(currentDir[0].toString(), { withFileTypes: true });
+    const filesAndDirs = fs.readdirSync(currentDir.path.toString(), { withFileTypes: true });
```
jwajgelt, line 101: Is there a good reason for this to be `sync`? Do we want to block the whole extension while we're looking for the appRoot?

jwajgelt (changes requested):

packages/vscode-extension/src/utilities/packageManager.ts
jwajgelt: Yes, but `PackageManagerInfo` contains the information on which package manager should be used which, if I understand correctly, depends on the contents of `appRoot`, no? Can't changing the app root could cause the selected package manager to change?

packages/vscode-extension/src/utilities/extensionContext.ts
jwajgelt: How so?

```tsx
  const searchQueue: searchItem[] = searchDirectories;

  while (searchQueue.length > 0) {
    // NOTE: this is safe because we've just checked `searchQueue` is non-empty
    const { path, searchDepth } = searchQueue.shift()!;
    if (searchDepth > maxDepth) {
      break;
    }

    const filesAndDirs = fs.readdirSync(path.toString(), { withFileTypes: true });

    let matched = false;

    filesAndDirs.forEach((dirEntry) => {
      if (dirEntry.isFile()) {
        if (!matched && searchedFileNames.includes(dirEntry.name)) {
          results.push(path);
          matched = true;
        }
        return;
      }

      if (excludedDirectoryPatterns.some((pattern) => pattern.test(dirEntry.name))) {
        return;
      }

      searchQueue.push({
        path: path + "/" + dirEntry.name,
        searchDepth: searchDepth + 1,
      });
    });
  }

  return results;
```

this typechecks correctly for me.

jwajgelt: And if you don't trust the non-null assertion next to shift, you can always check for `undefined` and `break` explicitely.

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -15,15 +21,176 @@ export const extensionContext = new Proxy<ExtensionContext>({} as ExtensionConte
...
+  if (appRootCandidates.length > 1) {
+    const openLaunchConfigButton = "Open Launch Configuration";
+    window
+      .showWarningMessage(
+        `Multiple react-native applications were detected in the workspace. "${appRootCandidates[0]}" was automatically chosen as your application root. To change that or remove this warning in the future, you can setup a permanent appRoot in Launch Configuration.`,
+        openLaunchConfigButton
+      )
+      .then((item) => {
+        if (item === openLaunchConfigButton) {
+          commands.executeCommand("workbench.action.debug.configure");
```
jwajgelt, line 169: Is this the correct action for this? For me it opens the command palette, while I'd expect it to open _our_ launch configuration menu (where I can actually select the app root).

packages/vscode-extension/src/project/project.ts
jwajgelt: Honestly, I feel like the "`Project`" should just take the `appRoot` in the constructor, and be created again whenever the app root changes, but I feel like that would break some other assumptions elsewhere at this point.

filip131311: > Honestly, I feel like the "Project" should just take the appRoot in the constructor, and be created again whenever the app root changes, but I feel like that would break some other assumptions elsewhere at this point.

We were discussing that option but there is just too much te project class do, includeing the device management which should not be restarted on app root change, we will work towards untangling those functionalities in the future and then It will be posible

packages/vscode-extension/src/utilities/extensionContext.ts
filip131311: I don't know why it opens a command palet for you, It works for me and opens a correct configuration file, In the future i would love to migrete to opening our configuration panel but i feel like it is not stable enough yet.

filip131311: could you run this comand from command palet and see what happens? do you use cursor or vs code?

packages/vscode-extension/src/utilities/extensionContext.ts
filip131311: making it async would require as to chege how we build large part of the extension and refactor multiple clases to use builder pattern instead of constructor, we might do it in the future if it turns out to be benefitial but it is a large change for a small improvement

packages/vscode-extension/src/utilities/extensionContext.ts
filip131311: good catch

packages/vscode-extension/src/utilities/packageManager.ts
filip131311: It sometimes could, but it is unlikely, but at the end of the day the important question is not if those values are sometimes related but if there might be the situation ever of changing only one of them and as I have written above there might be

packages/vscode-extension/src/project/project.ts
filip131311: overall the sugestion to move those objects to the separet class was exactly what I was missing thank you

packages/vscode-extension/src/utilities/extensionContext.ts
jwajgelt: VSCode.
It's probably because I don't have the file in that project.

packages/vscode-extension/src/utilities/packageManager.ts
jwajgelt: But this is what I mean -- if you change the `appRoot` you probably want to recreate the `PackageManagerInfo`, right?
The way it's written right now, you supply them separately, so you may provide the "wrong" `PackageManagerInfo`, i.e. one unrelated to the passed `appRoot`.

packages/vscode-extension/src/project/applicationContext.ts
```diff
@@ -0,0 +1,60 @@
...
+    this.buildCache = new BuildCache(newAppRoot);
+
+    this.disposables.push(this.launchConfig, this.dependencyManager);
+  }
+
+  public dispose() {
+    this.disposables.forEach((disposable) => {
+      disposable.dispose();
+    });
+  }
```
jwajgelt, line 59: We have a helper `disposeAll` function in utils, use that instead.

packages/vscode-extension/src/utilities/extensionContext.ts
```diff
@@ -84,43 +84,44 @@ export function findAppRootCandidates(maxSearchDepth: number = 3): string[] {
...
+      .filter((dirEntry) => {
+        return (
+          !dirEntry.isFile() &&
+          !excludedDirectoryPatterns.some((pattern) => pattern.test(dirEntry.name))
+        );
+      })
+      .forEach((dir) => {
+        searchQueue.push({
+          path: currentDir!.path + "/" + dir.name,
+          searchDepth: currentDir!.searchDepth + 1,
```
jwajgelt, line 122:
```suggestion
          path: currentDir.path + "/" + dir.name,
          searchDepth: currentDir.searchDepth + 1,
```

packages/vscode-extension/src/project/applicationContext.ts
jwajgelt, line 1: Shouldn't the file name be capitalized?

packages/vscode-extension/src/project/applicationContext.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (!newAppRoot) {
+      window.showErrorMessage(
+        "Failed to determine any application root candidates, you can set it up manually in launch configuration",
+        "Dismiss"
+      );
+      Logger.error("[Project] The application root could not be found.");
+      throw Error(
+        "Couldn't find app root folder. The extension should not be activated without reachable app root."
+      );
+    }
```
jwajgelt, line 28: NIT: I think I'd prefer if the `appRootFolder` was passed as an argument to the constructor to remove as much "logic" from it, but that's just my opinion.

jwajgelt (changes requested):

jwajgelt (approved):

PR #667, opened 2024-10-28
Use websocket /message endpoint to deliver reload request

### LINKED ISSUE #664: Reloads are broken on Expo 52-beta app
1. Create new Expo 52-beta project
2. Open in IDE
3. Click reload button
4. It'll be stuck in "Restarting" state

kmagiera:
In Expo 52, the metro /reload endpoint has been moved to a new location. As a consequence we couldn't perform JS reloads which resulted in the app being stuck in "reloading" state.

This PR updates the way we trigger reload action. A while ago, we switched the mechanism for triggering devMenu to use /message endpoint. This endpoint is also capable of handling reload command. Here, we are swapping out the reload implementation to use the /message websocket endpoint to deliver it. This method is more portable and work across older version while also being supported by expo SDK 52 (given https://github.com/expo/expo/issues/32399 is published)

Fixes #664

### How Has This Been Tested: 
1. Use reloads and dev menu across different projects, specifically older ones like RN 73, expo-go and on Expo 52 (not yet added here as test project)

### DIFF
--- a/packages/vscode-extension/src/project/metro.ts
+++ b/packages/vscode-extension/src/project/metro.ts
@@ -280,7 +280,7 @@ export class Metro implements Disposable {
     ws.send(
       JSON.stringify({
         version: 2 /* protocol version, needs to be set to 2 */,
-        method: message,
+        method,
       })
     );
     // we disconnect immediately after sending the message as there's no need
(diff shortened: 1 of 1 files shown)


filip131311 (commented):
I tested expo-go and RN76 apps and it worked well, but after installing the newest expo template app with `npx create-expo-app --template default@beta` reloading app caused problems: 


https://github.com/user-attachments/assets/e0b5877c-1081-4b12-b396-bd7f40d4e2bc


Also left some inline comments

filip131311 (approved):
It seems that problems are not related to this PR and Expo 52 template app works on IOS.

PR #668, opened 2024-10-28
Refactor throttling logic for async function

kmagiera:
When testing Radon I encountered an issue with newly cloned repo in which the IDE would generate an excessive number of npm subprocesses. The issue turned out to be related to fingerprint library, which on some occasions can spawn at least three npm processes. The main problem though was that this problem was amplified by the fact we'd request fingerprint calculation upon the watched files update. Even though the requests were throttled, because they'd take significant time to compute, we'd end up spawning many concurrent processes. This was specifically noticeable at the moment when the project was installing node modules which generates a ton of fs events that resulted in the fingerprint calculation getting triggered.

The solution this PR is implementing is to add a new throttling mechanism for async functions. It is a non-standard mechanism hence we're introducing a new method. The new mechanism is then used to throttle async code that calculates the fingerprint.

The mechanism works by keeping the following constraints:
1) it never allows for the method to be called more often than every X millisecond
2) it never allows for the method to be executed concurrently
3) when called multiple times, it is guaranteed that the latest call with the most recent arguments will be executed

### How Has This Been Tested: 
1. Clone some open source RN project
2. Without installing node modules just open it with the IDE
3. On my laptop before this change, I'd observe an excessive number of npm processes being spawned.

### DIFF
--- a/packages/vscode-extension/src/builders/PlatformBuildCache.ts
+++ b/packages/vscode-extension/src/builders/PlatformBuildCache.ts
@@ -112,16 +112,18 @@ export class PlatformBuildCache {
   }
 
   public async calculateFingerprint() {
+    Logger.info("Calculating fingerprint");
     const customFingerprint = await this.calculateCustomFingerprint();
 
     if (customFingerprint) {
+      Logger.info("Using custom fingerprint", customFingerprint);
       return customFingerprint;
     }
 
     const fingerprint = await createFingerprintAsync(getAppRootFolder(), {
       ignorePaths: IGNORE_PATHS,
     });
-    Logger.log(`Workspace fingerprint: '${fingerprint.hash}'`);
+    Logger.info("App folder fingerprint", fingerprint.hash);
     return fingerprint.hash;
   }
 
--- a/packages/vscode-extension/src/builders/PlatformBuildCache.ts
+++ b/packages/vscode-extension/src/builders/PlatformBuildCache.ts
@@ -112,18 +112,18 @@ export class PlatformBuildCache {
   }
 
   public async calculateFingerprint() {
-    Logger.info("Calculating fingerprint");
+    Logger.debug("Calculating fingerprint");
     const customFingerprint = await this.calculateCustomFingerprint();
 
     if (customFingerprint) {
-      Logger.info("Using custom fingerprint", customFingerprint);
+      Logger.debug("Using custom fingerprint", customFingerprint);
       return customFingerprint;
     }
 
     const fingerprint = await createFingerprintAsync(getAppRootFolder(), {
       ignorePaths: IGNORE_PATHS,
     });
-    Logger.info("App folder fingerprint", fingerprint.hash);
+    Logger.debug("App folder fingerprint", fingerprint.hash);
     return fingerprint.hash;
   }
 
--- a/packages/vscode-extension/src/builders/PlatformBuildCache.ts
+++ b/packages/vscode-extension/src/builders/PlatformBuildCache.ts
@@ -123,7 +123,7 @@ export class PlatformBuildCache {
     const fingerprint = await createFingerprintAsync(getAppRootFolder(), {
       ignorePaths: IGNORE_PATHS,
     });
-    Logger.debug("App folder fingerprint", fingerprint.hash);
+    Logger.log("App folder fingerprint", fingerprint.hash);
     return fingerprint.hash;
   }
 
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/utilities/throttle.ts
```diff
@@ -28,3 +28,44 @@ export function throttle<T extends AnyFn>(func: T, limitMs: number): WithForce<T
...
+              timeout = null;
+              recentArgs = null;
+            } else {
+              // if new arguments were provided while the function was executing,
+              // we need to run it again to ensure the last call is executed.
+              // we use setTimeout to avoid potentially infinite recursion, and
+              // to ensure the function is not run more often than once every
+              // `limitMs` milliseconds, we schedule it to run after the provided
+              // limit.
+              setTimeout(execute, limitMs);
```
filip131311, line 63: nit: if you are not using the timeout to cancel it, but just to check if it was completed or not it might be better to keep a simple boolean value instead, but if you want to keep the current timeout in the value `timeout` you should assign it here as well

packages/vscode-extension/src/utilities/throttle.ts
```diff
@@ -28,3 +28,44 @@ export function throttle<T extends AnyFn>(func: T, limitMs: number): WithForce<T
...
+            }
+          });
+      };
+      timeout = setTimeout(execute, limitMs);
+    }
+    recentArgs = args;
```
filip131311, line 69: I believe that without this change the function runs without arguments the first go around (wouldn't come up during testing because `checkIfNativeChanged` does not take any)

```suggestion
  return async function (...args: any) {
    recentArgs = args;
    if (!timeout) {
      const execute = () => {
        const currentArgs = recentArgs;
        const result = func(...recentArgs);
        result
          .catch(() => {})
          .then(() => {
            if (recentArgs === currentArgs) {
              timeout = null;
              recentArgs = null;
            } else {
              // if new arguments were provided while the function was executing,
              // we need to run it again to ensure the last call is executed.
              // we use setTimeout to avoid potentially infinite recursion, and
              // to ensure the function is not run more often than once every
              // `limitMs` milliseconds, we schedule it to run after the provided
              // limit.
              setTimeout(execute, limitMs);
            }
          });
      };
      timeout = setTimeout(execute, limitMs);
    }
```

packages/vscode-extension/src/builders/PlatformBuildCache.ts
```diff
@@ -112,16 +112,18 @@ export class PlatformBuildCache {
   }
 
   public async calculateFingerprint() {
+    Logger.info("Calculating fingerprint");
```
filip131311, line 115: Why is this log not debug level?

filip131311 (commented):
Left some inline comments

packages/vscode-extension/src/builders/PlatformBuildCache.ts
kmagiera: we rarely use Logger.debugg as currently we redirect almost all logs to info anyway to make it easier for us to trace down reported issues

packages/vscode-extension/src/utilities/throttle.ts
kmagiera: that's a good catch too. I copied it from regular throttle where we have "force" method that'd cancel the timeout. This isn't needed here though

packages/vscode-extension/src/utilities/throttle.ts
kmagiera: we assign recentArgs at the end already, there is no difference whether we call it before or later as we will use them in function that runs with timeout

packages/vscode-extension/src/builders/PlatformBuildCache.ts
filip131311: That's is true that we redirect most logs to info but we do it on the logger level in `Logger.ts` file the code looks like this: 

```javascript
debug(message: string, ...args: any[]) {
    // in the initial phase od development, we want to surface debug messages such that
    // they can be seen by the extension user. This is useful for collecting feedback and
    // being able to better troubleshoot issues. In vscode, the default logging level is set
    // to INFO. We don't want to ask users to change their logging level as it impacts all
    // the logs vscode generates. Also, when changing log level, one need to reload vscode
    // window for the extension to pick up the change.
    outputChannel.info(message, ...args);
  },
```

Eventually those logs do not seem info level as they do not provide information to a user that is not familiar with our code base, so I would mark the as `debug`

kmagiera: If I had this logs I would've fixed this problem long time ago as they'd result in an excessive number of entries getting printed out. I don't really care whether its debug or info since they ultimately result in the same behavior at the moment, so not going to insist to keep it one way of the other, but I don't agree they don't provide any information as at least they would've prevented the issue I'm fixing here

filip131311 (approved):
LGTM

PR #673, opened 2024-10-29
Prevent clicks on device screen behind the UrlBar and SelectDevice

### LINKED ISSUE #649: Selecing items from URL bar interfers with device touch system

p-malecki:
This PR addresses the issue where selecting items from UrlBar or DeviceSelect results in touch events interacting with elements underneath. This is especially problematic when the touch position is above the device screen, unintentionally sending touch to the device.

The likely cause is the Select.Item implementation in Radix's Select. A similar problem was reported in [this Radix issue](https://github.com/radix-ui/primitives/issues/1658). That issue was closed as it was confirmed to be solved since the react-select 2.1.2-rc.8 version. However, updating the package didn't resolve our issue.

The root of the problem is that when an item is clicked, the select value changes so quickly that the popper disappears, causing the touch event to simultaneously click on the element behind it.

To solve it, an `isOpen` state has been added which manages the opening and closing of the Select.Content. This state is modified using `setTimeout` with a duration of `0 ms`, delaying the state update (`setOpen(!isOpen)`) to the end of the JS call stack, which allows the touch event to complete before closing the Select.Content, preventing any accidental clicks.

Fixes  #649 

### How Has This Been Tested: 
- Followed test steps from issue #649, also confirmed on iOS devices, and similarly tested DeviceSelect.

### DIFF
--- a/packages/vscode-extension/src/webview/components/DeviceSelect.tsx
+++ b/packages/vscode-extension/src/webview/components/DeviceSelect.tsx
@@ -1,4 +1,4 @@
-import React, { PropsWithChildren } from "react";
+import React, { PropsWithChildren, useState } from "react";
 import * as Select from "@radix-ui/react-select";
 import { DeviceInfo, DevicePlatform } from "../../common/DeviceManager";
 import "./DeviceSelect.css";
@@ -100,6 +100,8 @@ interface DeviceSelectProps {
 }
 
 function DeviceSelect({ onValueChange, devices, value, label, disabled }: DeviceSelectProps) {
+  const [isOpen, setOpen] = useState<boolean>(false);
+
   const { projectState } = useProject();
   const selectedProjectDevice = projectState?.selectedDevice;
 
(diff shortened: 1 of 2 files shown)


kmagiera:
I'd have hoped that instead of doing this we can have some more generic solution that isn't associated with the controls that display over the device. For example, the same problem applies to device settings menu and other menus, we don't want to keep adding workarounds to all the UI elements we have now and will add in the future.

I think the generic solution could be based on focus or event target mechanism for the events we receive. I also think that we should consider it for both mouse and keyboard events.

p-malecki:
Suggestions have been applied in #685, so I am closing this PR.

PR #674, opened 2024-10-29
Update build scripts to adapt to sim-server build pipeline changes

kmagiera:
In sim-server repo, we're updating the build pipleine to distribute linux binaries. Because of that, we're adaping a new naming scheme, where each binary gets platform specific suffix (simulator-server-macos, simulator-server-windows.exe and simulator-server-linux). Along with those changes we are making some housekeeping tasks related to build scripts like removing the need for passing debug/release config as we already migrated to a process where release builds are only made on CI while the local builds are scoped to development only.

With the changes made, we will no longer require people to manually download simulator server binaries as now the dev build script can fetch them from the releases page.

As part of this PR we're also removing code that'd copy binaries during extension initialization. It is no longer necessary because when installing the VSIX file, the binary file permission no longer occurs.

What this PR is not solving (but we didn't have a way of doing that before anyway) is to give ability to use simulator-server versions updated in between IDE releases. Now, users are only able to use binaries that are included in IDE release, but if some code change that is in main requires updated simulator-server, there's no way in the current flow to publish that and let the people use it.

### How Has This Been Tested: 
1. Build release package and test it with cursor and vscode
2. Clone new repo w/o sim-server and test development builds
3. Test dev builds in normal environment


packages/docs/docs/development.md
```diff
@@ -36,23 +36,17 @@ npm install
...
 npm run build:sim-server-debug
 ```
 
-In case of any errors, please read the output of this command before proceeding.
+Note that some changes in the extension may depend on changes made to the simulator server, so you occasionally may need to re-run this script, for example when switching git branches.
 
 ### 4. Open extension project in Visual Studio Code
 
-It is necessary that you open that exact folder rather than the whole repository, as it contains project specific run configuration for launching the extension in development mode.
+It is necessary that you open exacly the main extension folder rather than the whole repository, as it contains project specific run configuration for launching the extension in development mode.
```
balins, line 49:
```suggestion
It is necessary that you open exactly the main extension folder rather than the whole repository, as it contains project specific run configuration for launching the extension in development mode.
```

packages/vscode-extension/src/extension.ts
```diff
@@ -461,24 +453,3 @@ function migrateOldConfiguration() {
     Logger.error("Error when migrating old configuration", e);
   }
 }
-
-async function fixMacosBinary(context: ExtensionContext) {
```
balins, line 465: Out of curiosity, why is this no longer required?

balins (approved):

packages/vscode-extension/src/extension.ts
kmagiera: good point, forgot to mention this in the description, will do

PR #675, opened 2024-10-29
Fix build cache management in workspaces

filip131311:
This PR changes a way we manage build caches. Previous approach was dependent on current vscode workspace, which led to  build caches not being accessible when switching from working in the root of the workspace to the root of the application (or doing the same in reverse). The new approach saves this data in global storage, and adds a appRoot identifier to the cache key.  

This PR is a dependency for #663

### How Has This Been Tested: 

- Open a project in the root of the workspace and run some application that is part of it in the IDE.
- Open IDE in the root of that application and see if the build cache is loaded. 
- repeat the proces in reverse
- open an application that had a build made before this change and check if migration process works as expected

Verify that the build cache migration work:
- Build project without these changes
- Open the same project with no changes but with the new version of extension – expect the cache hit for native build


kmagiera:
Is this really a scenario we want to optimize for? When I work with monorepo setups I never switch between opening the workspace from different places. I either have the app bit open, or I have the whole monorepo opened. I never switch between these two approaches for a given project. Do you have any insights whether this is what people tend to do?

packages/vscode-extension/src/utilities/common.ts
```diff
@@ -94,6 +94,14 @@ export function getOrCreateAppDownloadsDir() {
   return downloadsDirLocation;
 }
 
+export function getOrCreateBuildCachesDir() {
```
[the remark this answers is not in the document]
filip131311: You are right it would be great to it build like that, thank you

filip131311:
@kmagiera This PR as I eluded in description is necessary to enable switching between applications in https://github.com/software-mansion/radon-ide/pull/663, so the fact that it will help some people that like to change set up is just a bonus, but to answer your question when I talked to @jakex7 about his monorepo set up he seemed interested in being able to use booth approaches.

packages/vscode-extension/src/utilities/appCaches.ts
```diff
@@ -0,0 +1,97 @@
...
+ * @returns the new value that was set
+ */
+export function setAppCache(key: string, value: string) {
+  const appCachesPath = getAppCachesPath();
+  let oldCaches;
+  if (fs.existsSync(appCachesPath)) {
+    try {
+      oldCaches = JSON.parse(fs.readFileSync(appCachesPath).toString());
+    } catch (e) {
+      Logger.warn(`Error parsing cold caches file for app ${getAppRootFolder()}`, e);
```
km1chno, line 51:
```suggestion
      Logger.warn(`Error parsing old caches file for app ${getAppRootFolder()}`, e);
```

packages/vscode-extension/src/utilities/appCaches.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+export function removeAppCache(key: string) {
+  const appCachesPath = getAppCachesPath();
+  let oldCaches;
+  try {
+    oldCaches = JSON.parse(fs.readFileSync(appCachesPath).toString());
+  } catch (e) {
+    Logger.warn(`Error parsing old caches file for app ${getAppRootFolder()}`, e);
+    return;
+  }
```
km1chno, line 85: this logic is repeated several times, you can extract it to new function `getAppCaches()` or something like that (without key, it returns the whole json object)

packages/vscode-extension/src/utilities/appCaches.ts
```diff
@@ -0,0 +1,97 @@
...
+  if (!caches[key]) {
+    Logger.debug(`No app caches found for key: ${key}}`);
+    return undefined;
+  }
+  return caches[key];
+}
+
+/**
+ * This function sets app level caches from caches storage.
```
kmagiera, line 39: What does that even mean?

packages/vscode-extension/src/builders/BuildManager.ts
```diff
@@ -135,7 +135,7 @@ export class BuildManager {
         );
       }
 
-      await buildCache.storeBuild(buildFingerprint, buildResult);
+      await buildCache.storeCache(buildFingerprint, buildResult);
```
kmagiera, line 138: I don't like the naming being so mixed up in this whole PR. We should have naming that better maps to the intuition what cache is rather than trying to provide a super-accurate naming.

Here, it sound very weird that we are "storing cache" using "build cache". Intuitively, what this line does is it stores some build metadata (build result) in a cache storage. IMO "storeBuild" just much better describes what this method does than "storeCache".

packages/vscode-extension/src/builders/PlatformBuildCache.ts
```diff
@@ -53,21 +54,34 @@ export class PlatformBuildCache {
   /**
    * Passed fingerprint should be calculated at the time build is started.
    */
-  public async storeBuild(buildFingerprint: string, build: BuildResult) {
+  public async storeCache(buildFingerprint: string, build: BuildResult) {
```
kmagiera, line 57: why has this been renamed? The method takes fingerprint and build result that gets stored. We don't store any cache here

packages/vscode-extension/src/builders/PlatformBuildCache.ts
```diff
@@ -53,21 +54,34 @@ export class PlatformBuildCache {
...
     const appPath = await getAppHash(getAppPath(build));
-    await extensionContext.workspaceState.update(this.cacheKey, {
+
+    const cache = JSON.stringify({
       fingerprint: buildFingerprint,
       buildHash: appPath,
       buildResult: build,
     });
+
+    setAppCache(this.cacheKey, cache);
```
kmagiera, line 66: In contrast, this method actually stores stuff, so why is it called "set" – this is a frequently used prefix for setting class members and using it for method that writes to disk is very confusing.

packages/vscode-extension/src/utilities/appCaches.ts
```diff
@@ -0,0 +1,97 @@
...
+ * @param key
+ * @param value
+ * @returns the new value that was set
+ */
+export function setAppCache(key: string, value: string) {
```
kmagiera, line 44: this whole module doesn't have a clear contract. It seem to assume that the values it saves or retrieves are strings, but internal implementation parses those files expecting this is actually a structured data

kmagiera (changes requested):

packages/vscode-extension/src/utilities/appCaches.ts
filip131311: I stores strings for given key but the whole caches file is a json

kmagiera (changes requested):
After reviewing this I think it'd be best to restructure this a bit. Below are some things that I think we should do:

1) I don't think we need the whole logic for managing app caches directory etc. We've been using workspaceState,  but if we cannot rely on this data on being workspace-specific, we should instead use globalState. For the keys, we could just use the app root directory instead of a hash which should make it more readable and allow for cleaning it up at later time.

2) we should settle on avoiding the use of getAppRoot going forward to make it easier to support cases where we want to switch app roots. The first step we can make could be by adding appRoot as constructor parameter to `PlatformBuildCache` – this way the build cache can reference the right place in the storage and we don't need to add more references to getAppRoot.

packages/vscode-extension/src/builders/BuildCache.ts
```diff
@@ -30,44 +30,34 @@ export type BuildCacheInfo = {
...
-  private constructor(private readonly platform: DevicePlatform) {}
+export class BuildCache {
+  constructor(private readonly platform: DevicePlatform, private readonly appRoot: string) {}
 
   get cacheKey() {
-    return this.platform === DevicePlatform.Android ? ANDROID_BUILD_CACHE_KEY : IOS_BUILD_CACHE_KEY;
+    const platformKey =
+      this.platform === DevicePlatform.Android ? ANDROID_BUILD_CACHE_KEY : IOS_BUILD_CACHE_KEY;
+    const appRoot = this.appRoot;
+    return appRoot + platformKey;
```
kmagiera, line 40: would prefer to add some delimiter here, would be cleaner to read in case we actually need to access this in the future.

Also, because of that, it'd be better to have the path at the end, otherwise it would be harder to get the path as you couldn't tell where the file path actually ends.

Finally, lets make this `private` as it is not being accessed from the outside, and I don't think it needs to be exposed.

So I'd go for:
```
private get cacheKey() {
  const keyPrefix = this.platform === DevicePlatform.Android ? ANDROID_BUILD_CACHE_KEY : IOS_BUILD_CACHE_KEY;
  return `${keyPrefix}:{this.appRoot}`
}
```

packages/vscode-extension/src/project/project.ts
```diff
@@ -671,7 +671,8 @@ export class Project
   private checkIfNativeChanged = throttleAsync(async () => {
     if (!this.isCachedBuildStale && this.projectState.selectedDevice) {
       const platform = this.projectState.selectedDevice.platform;
-      const isCacheStale = await PlatformBuildCache.forPlatform(platform).isCacheStale();
+      const buildCache = new BuildCache(platform, getAppRootFolder());
```
kmagiera, line 674: would it be possible to restructure the code in a way that we don't have to instantiate the new cache object here? This doesn't matter at this point because the class is stateless, but may be a source of errors in the future if we end up changing that.

packages/vscode-extension/src/builders/BuildCache.ts
```diff
@@ -160,3 +148,20 @@ function getAppPath(build: BuildResult) {
...
+
+  platformKeys.forEach((platformKey) => {
+    const cache = extensionContext.workspaceState.get<BuildCacheInfo>(platformKey);
+    if (!cache) {
+      return;
+    }
+
+    extensionContext.globalState.update(appRoot + platformKey, cache);
+
+    // remove the old cache afterwords
```
kmagiera, line 164:
```suggestion
    // remove the old cache afterwards
```

PR #680, opened 2024-10-31
Implement reboot functionality

filip131311:
This PR add a possibility to reboot the reboot  device, metro and devtools processes without triggering clean build. It was previously only possible by closing and reopening the IDE panel itself. 

This also will provide necessary functionality for  #663

### How Has This Been Tested: 

- run an application and pick "reboot" option from reload dropdown menu

### DIFF
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -336,6 +336,14 @@ export class Project
 
   public async reload(type: ReloadAction): Promise<boolean> {
     this.updateProjectState({ status: "starting" });
+
+    if (type === "reboot") {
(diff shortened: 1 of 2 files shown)


kmagiera (commented):
I'm fine with adding this as an option in general, but not a fan of the naming here. On top of that this list is getting a bit long and making it more important for the options to be clear for the users.

Note that on top of the existing option I'd like to also add a way to clear metro cache, which seems like a necessary recovery option that would be helpful in some scenarios that people already reported when metro mixes up things when switching between console and IDE-based workflows.

filip131311:
@kmagiera  i changed the name to "Reboot Simulated Device" which seems to capture what the option does

kmagiera (approved):

PR #685, opened 2024-11-01
Block unwanted touch and keyboard interactions in PreviewView and Replays

### LINKED ISSUE #649: Selecing items from URL bar interfers with device touch system

p-malecki:
This PR resolves the issue where selecting items from UrlBar or DeviceSelect results in touch events interacting with device screen below. The issue was caused by calling onMouseDown from PreviewView, originally added in #551 to enable pressing from outside the device screen and frame. The solution is to move the onMouseDown, onMouseUp, and onMouseLeave handlers from PreviewView to the Preview component and place them in `wrapperTouchHandlers` in `phone-wrapper`.
Additionally, this PR:

- blocks the sending of touch and keyboard events during replays
- hides multitouch indicators as a result of the above
- deactivates multitouch indicators when the mouse leaves the phone screen

Fixes #649 

### How Has This Been Tested: 
- Followed test steps from issue #649 , also confirmed on iOS devices, and similarly tested DeviceSelect.
- Enabled "SIMSERVER_LOG": "debug", clicked and pressed keys in and around the device screen during a replay, and confirmed no touch or key logs appeared.
- Confirmed that multitouch indicators are not visible during a replay.


packages/vscode-extension/src/webview/components/ReplayOverlay.tsx
```diff
@@ -274,6 +274,25 @@ export default function ReplayOverlay({
...
   const timeFormat = `${Math.floor(timeSec / 60)}:${(timeSec % 60).toString().padStart(2, "0")}`;
+
+  useEffect(() => {
+    function handleEvents(e: MouseEvent | KeyboardEvent) {
+      e.preventDefault();
+      e.stopPropagation();
+    }
+
+    document.addEventListener("mousedown", handleEvents, true);
+    document.addEventListener("mouseup", handleEvents, true);
```
kmagiera, line 285: why is this needed here and not needed in other UI items that displays over phone-wrapper?

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -147,17 +148,32 @@ function PreviewView() {
 
   function onMouseDown(e: MouseEvent<HTMLDivElement>) {
     e.preventDefault();
+    const elementName = (e.target as Element).className;
```
kmagiera, line 151: I missed that in #551 we added this extra layer of capturing mouse events. What is the reason for it. The pressing state is not read in this component except from passing it down to Preview

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -147,17 +148,32 @@ function PreviewView() {
...
+    if (elementName !== "phone-wrapper") {
```
kmagiera, line 152: This adds another layer of indirection as we rely on a specific class name to be used by a specific element that the component we render here renders – this makes it very complicated to follow in code and I wonder if there better way to avoid this.

On top of this, I wonder if this doesn't break the swipe from edge functionality that we have, where you can start swipe gesture at the phone frame and drag front the side (i.e. to go back on iOS).

One way I was hoping this could be solved was by using [contains()](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains) along with react refs

packages/vscode-extension/src/webview/components/ReplayOverlay.tsx
maciekstosio, line 1: I wonder if that part couldn't be solved by checking for `replayData` in https://github.com/software-mansion/radon-ide/blob/2d3d6b6c143b9f825087198179c89745aa2fe9f4/packages/vscode-extension/src/webview/components/Preview.tsx#L498

maciekstosio: Or if all of the elements should be in `.touch-area`, maybe there should be a separate wrapper that contains all of preview related elements (including touch area with `MjpegImg`)

packages/vscode-extension/src/webview/views/PreviewView.tsx
p-malecki: In #551, the pressing state was moved to previewView so it remains true when the mouse leaves and reenters device screen during multitouch gesture. Before this change if mouse reentered multitouch wouldn't send touches unless mouse button was released and clicked again within the phone frame. However, I can revert this if necessary.

packages/vscode-extension/src/webview/components/ReplayOverlay.tsx
p-malecki: Using `shouldPreventTouchInteraction` is a good idea. Thanks!

packages/vscode-extension/src/webview/views/PreviewView.tsx
p-malecki: The simplest solution is to remove `onMouseDown` from PreviewView. It was causing unwanted touch down signals. Removing it should fix the issue if we want to start single and multitouch gestures from the edge of the phone rather than outside the touch area.

maciekstosio:
I wonder if the touches shouldn't be handled the other way around. My feeling is that we catch them too low (too close to the root) thus we need hacks to disable them to not intercept with other elements. Is there any reason why we don't handle touches just on some wrapper around `MjpegImg` with some constant padding around? I get that for multitouch we would like to be able to move the mouse out of the phone, but maybe when multitouch is active there should be some invesible overlay with high zIndex to cover everything else. Not sure if that's in the scope of this task tho.

PR #686, opened 2024-11-01
Disable react compiler for app wrapper to workaround runtime issues

kmagiera:
This PR fixes issue with Radon IDE not working with recent versions of react compiler.

We've been testing Radon with some previous versions of the compiler where this issue didn't occur. I only recently stumbled upon it when testing an app that was using a most recent version of the compiler.

The issue I started seeing is when compiler was enabled, I would get errors about react-compiler-runtime not being a function when accessed from the wrapper component. I'm assuming this has to do with wrapper component being injected by the IDE babel plugin which may confuse babel which while processes the file with react runtime plugin may not properly link the runtime imports which are necessary for React 17 and 18.

For the time being, the fix is to disable react compiler at the wrapper file level. This prevents the file from being processed by the compiler plugin as as a consequence, no runtime imports are added.

This PR also adds compiler to react-native-76 test app so that we can test this easier and avoid potential regressions related to react compiler in the future.

### How Has This Been Tested: 
1. Use RN 76 app with react compiler enable and see it runs ok
2. Before the change in wrapper, the app would get stuck on loading and errors about react-compiler-runtime not being a function were shown in the logs.

### DIFF
--- a/packages/vscode-extension/lib/wrapper.js
+++ b/packages/vscode-extension/lib/wrapper.js
@@ -1,3 +1,5 @@
+"use no memo";
+
 const { useContext, useState, useEffect, useRef, useCallback } = require("react");
 const {
   LogBox,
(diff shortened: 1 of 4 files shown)


packages/simulator-server
filip131311, line 1: This seems like an accidental addition here

PR #691, opened 2024-11-04
Fix auto select in InspectDataMenu when opened above cursor

### LINKED ISSUE #687: Inspector context menu closes immediately when displayed above the cursor
**Description**

When using inspector context menu while in setting when it displays above the cursor, it immediately closes selecting the bottom most item from the menu.

This seems to be due to the fact the menu displays over the area where we click.

This issue can be reproduced in any test app when you inspect close to the bottom forcing the menu to render above the cursor:

https://github.com/user-attachments/assets/37a0fa7c-44fe-4434-94fb-1031d6130b6d

This issue is much more amplified on real-life projects though where the inspector stacks often have 10+ items. In that scenario the menu covers a big chunk of screen and displays over the pointer which has the same result.

**Environment**
- Radon 0.21

p-malecki:
This PR addresses an issue where the inspector context menu automatically selects and closes upon appearing above the cursor. The problem was caused by the menu opening  upon mouseUp immediately after a mouseDown event.

The solution ensures the menu displays adjacent to the cursor with an offset by introducing a `span` that positions the `Trigger` component at the mouse location, ensuring the menu opens next to the `Trigger`.

Additionally, this PR:
- limits the number of inspect items displayed in `InspectDataMenu` to 5 using `MAX_INSPECT_ITEMS`
- adds a "Show all" button to reveal all items
- adds ellipsis truncation for long filenames in `InspectDataMenu`
- adjusts menu alignment based on cursor position, displaying to the right if the cursor is on the left side of the screen and to the left if on the right

Fixes #687 

### How Has This Been Tested: 
- Tested by right-clicking near the screen's bottom, ensuring the menu opens above the cursor and doesn't auto-select any items.

After:
Screenshot 2024-11-04 at 20 39 32
Screenshot 2024-11-05 at 18 33 04

https://github.com/user-attachments/assets/b09a6fdb-23a6-4ecd-b707-980f7074a652


kmagiera (commented):
I don't think this is the right fix for the issue. Instead of working around button events, I believe it should be enough to offset the menu such that it appears over the cursor rather than directly under. This would also be better from UX perspective, as now, the menu covers the element we are inspecting, which isn't the case when the menu displays under.

I think we should change from using an outer-level div for positioning, to adjusting the position of the Trigger component. This seems to be the way based on the examples from https://www.radix-ui.com/primitives/docs/components/dropdown-menu – right now the trigger component is empty while it seems the offset is added automatically such that the menu doesn't overlap the trigger.

Menu shown under:
image

Menu shown over:
image

kmagiera (approved):
Looks good. As a final touch you may want to change the logic slightly such that if there are exacly LIMIT+1 elements, you still show them all, as LIMIT+1 items occupy exactly the same amount of space as LIMIT + "show more" button. So when you click "show more" you don't actually expand the menu at all.

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
```diff
@@ -37,40 +42,72 @@ export function InspectDataMenu({
...
+        }
+      }}>
+      <DropdownMenu.Trigger asChild>
+        <span
+          style={{
+            position: "absolute",
+            left: inspectLocation.x,
+            top: inspectLocation.y,
+            opacity: 0,
+          }}></span>
```
maciekstosio, line 67:
```suggestion
          }}/>
```

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
```diff
@@ -37,40 +42,72 @@ export function InspectDataMenu({
...
+          className="inspect-data-menu-content"
+          sideOffset={5}
+          align={inspectMenuAlign}
+          collisionPadding={5}>
+          <DropdownMenu.Label className="inspect-data-menu-label">
+            {displayDimensionsText}
+          </DropdownMenu.Label>
+          {inspectItems.map((item, index) => {
+            // extract file name from path:
+            const fullFileName = item.source.fileName.split("/").pop() ?? "";
```
maciekstosio, line 80: I would extract it to separate Component or at least helper for the strings

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
```diff
@@ -37,40 +42,72 @@ export function InspectDataMenu({
...
+            const lastDotIndex = fullFileName.lastIndexOf(".");
+            const fileName = fullFileName.substring(0, lastDotIndex);
+            const fileExtension = fullFileName.substring(lastDotIndex);
+            return (
+              <DropdownMenu.Item
+                className="inspect-data-menu-item"
+                key={index}
```
maciekstosio, line 87: In general, `index` as a key is a bad practice, I think in our case when we don't have state it shouldn't cause bugs, but maybe you could use something as unique ID here? For example filePath with code line

maciekstosio (commented):
Some nitpicks

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
```diff
@@ -37,40 +77,57 @@ export function InspectDataMenu({
...
+          <DropdownMenu.Label className="inspect-data-menu-label">
+            {displayDimensionsText}
+          </DropdownMenu.Label>
+          {inspectItems.map((item) => (
+            <InspectItem item={item} onSelected={onSelected} onHover={onHover} />
+          ))}
+          {isOverMaxItems && !shouldShowAll && (
+            <DropdownMenu.Item
+              className="inspect-data-menu-item show-all"
+              key={"show-all"}
```
maciekstosio, line 121: I think now the `key` is unnecessary as the elements here differ

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
```diff
@@ -1,10 +1,48 @@
...
+const MAX_INSPECT_ITEMS = 5;
+
+interface InspectItemProps {
+  item: InspectDataStackItem;
+  onSelected: (item: any) => void;
+  onHover: (item: any) => void;
+}
+
+const InspectItem = React.forwardRef<HTMLDivElement, InspectItemProps>(
+  ({ item, onSelected, onHover }, forwardedRef) => {
```
maciekstosio, line 18: Why `forwardRef`? As far as I can tell, it's not used

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
```diff
@@ -1,10 +1,48 @@
...
+    const extractFileDetails = (filePath: string) => {
```
maciekstosio, line 19: Imo, now you don't need a helper (I meant component or helper, but I prefer the component) - just fyi, when you define a method in render, it's created on every rerender

maciekstosio (changes requested):

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
p-malecki: I think it's the way for creating custom components in radix https://www.radix-ui.com/primitives/docs/components/dropdown-menu

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
p-malecki: Its needed to avoid warrning `[WEBVIEW LOG] Warning: Each child in a list should have a unique "key" prop.%s%s` in logs.

maciekstosio: Interesting

maciekstosio (approved):
LGTM 🚀

PR #693, opened 2024-11-04
Fix inspect stack not returning all ancestors

### LINKED ISSUE #692: Inspector stack is not returning all ancestors
**Description**

Seems like the inspector stack is not always complete and lacks some ancestors of inspected component. For example, in `react-native-75`, after right-clicking on `Try and Authorize` button, the `<View>` component that the button is nested in is not displayed in the inspector data menu.

https://github.com/user-attachments/assets/d1b370e3-5bd7-432a-a220-99b8a914245b

**Steps to reproduce**

- open `react-native-75` app on iOS or Android emulator
- right click on the "Try and Authorize" button
- Result: `<View>` in which the button is nested is not displayed on the list
- Expected: it should be visible

**Environment**

- Version of currently used Radon IDE `0.0.22`
- Version of React Native/Expo used in the project. `react-native 75`
- ~~Any other versions of tools related to the problem.~~

km1chno:
## Changes

Fixes #692 

This PR is also necessary for #666 

This PR aims to fix the issue with inspector stack that not all ancestors of inspected elements are discovered. It turns out that the hierarchy returned by `getInspectorData()` is already missing these ancestors, so we have to traverse the component tree ourselves.

The first node is the `closestInstance` in object returned by `getInspectorDataViewAtPoint`. To access parent node of a node, we look at `node.return`. This way we can simply traverse the whole path from first node up to the root. Along the way, we collect all necessary information and create a stack that is returned at the end. We use `getInspectorDataForInstance(node)` renderer function to obtain inspector data for given fiber node.

This version is a lot more accurate, from my testing it returns all the most used visible components (like `<View>`, `<Text>`, `<Button>`), which is already an improvement compared to the current solution.

## Demos

The following are demos how the inspect stack looks like before and after changes from this PR. The component hierarchy of the app consists of several nested components, which have border for visualization.

#### Before

https://github.com/user-attachments/assets/39501ea3-9c95-46bb-9a06-9b1db9c6123f

#### After

https://github.com/user-attachments/assets/414f43f8-62a6-429f-a25e-6ab374dfcab9

### How Has This Been Tested: 

**Test 1:**
- Open shared app 
- Expand step
- Inspect (right-click) text in step
- Make sure all components are in the trace (for example ScrollView)

step

**Test 2:**

- Open shared app 
- Add after last `Step`:
```
<View>
  <View>
    <View>
      <Text>Nested texts</Text>
    </View>
  </View>
</View>
```
- Inspect (right click) `Nested texts` 
- Make sure all 4 Views are visible 

nested


packages/vscode-extension/lib/wrapper.js
```diff
@@ -190,53 +221,47 @@ export function AppWrapper({ children, initialProps, ..._rest }) {
...
               frame: scaledFrame,
-              stack: stack,
             });
-          });
+            return;
+          }
+
+          const inspectNode = (node, stack) => {
+            // Optimization: we break after reaching fiber node corresponding to OffscreenComponent (with tag 22).
+            if (!node || node.tag === 22) {
```
kmagiera, line 237: why is it 22? Is it documented somewhere or at least there's some code that sets it we can link to?

I recall that there's some numbering convention for root view instances, but from what I remember it changes in between JS reloads (it gets incremented) and also used to be different between Android and iOS. Can you check if this works well after JS reload too?

packages/vscode-extension/lib/wrapper.js
```diff
@@ -190,53 +221,47 @@ export function AppWrapper({ children, initialProps, ..._rest }) {
...
+                  x: pageX / width,
+                  y: pageY / height,
+                  width: viewWidth / width,
+                  height: viewHeight / height
+                };
+
+                const element = (inspectorData.source) ? 
+                  createStackElement(stackElementFrame, item.name, inspectorData.source) : undefined;
+                
+                inspectNode(node.return, [...stack, element]);
```
kmagiera, line 259:
```suggestion
                if (element) {
                  stack.push(element);
                }
                return inspectNode(node.return, stack);
```

packages/vscode-extension/lib/wrapper.js
```diff
@@ -190,53 +221,47 @@ export function AppWrapper({ children, initialProps, ..._rest }) {
...
+          }
+
+          const inspectNode = (node, stack) => {
+            // Optimization: we break after reaching fiber node corresponding to OffscreenComponent (with tag 22).
+            if (!node || node.tag === 22) {
+              devtoolsAgent._bridge.send("RNIDE_inspectData", {
+                id: payload.id,
+                frame: scaledFrame,
+                stack: Array.from(stack).filter(Boolean)
+              });
```
kmagiera, line 242: it'd be easier to follow if we just `return stack` here and move this after `inspectNode` call from line 264. This way it'd be clear that this is a closing statement of the node stack processing logic.

I also added another comment to the recursive call which uses conditional push, therefore eliminating the need to filter here.

packages/vscode-extension/lib/wrapper.js
```diff
@@ -190,53 +221,47 @@ export function AppWrapper({ children, initialProps, ..._rest }) {
...
+
+                const element = (inspectorData.source) ? 
+                  createStackElement(stackElementFrame, item.name, inspectorData.source) : undefined;
+                
+                inspectNode(node.return, [...stack, element]);
+              });
+            }
+          };
+          
+          inspectNode(viewData.closestInstance, []);
```
kmagiera, line 264:
```suggestion
          const stack = inspectNode(viewData.closestInstance, []);
          devtoolsAgent._bridge.send("RNIDE_inspectData", {
            id: payload.id,
            frame: scaledFrame,
            stack:,
          });
```

kmagiera (changes requested):
Please see my inline comments. I think it'd be good to test this on RN76 on both Android and iOS, as this is the only project that we have that's currently setup with the new architecture where some of the internals that you're accessing here may work differently.

packages/vscode-extension/lib/wrapper.js
km1chno: From what I see, the tag does not change between reloads. What changes is the unique node identifier, in some places it's called `_nativeTag`, in other `handle`.

These constants are defined here: https://github.com/facebook/react/blob/c3570b158d087eb4e3ee5748c4bd9360045c8a26/packages/react-reconciler/src/ReactWorkTags.js#L62

I will add this link to the comment

km1chno:
I tested this on rn76, it looks like it's working 😄

packages/vscode-extension/lib/wrapper.js
```diff
@@ -173,12 +173,43 @@ export function AppWrapper({ children, initialProps, ..._rest }) {
     [openPreview, closePreview, requestNavigationChange]
   );
 
+  const getInspectorDataForInstance = (node) => {
```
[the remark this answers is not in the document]
km1chno: I agree, I applied all your suggestions and moved this function along with the logic of traversing tree up to new file `lib/inspector.js`, so that the logic in `wrapper.js` is minimal

packages/vscode-extension/lib/inspector.js
```diff
@@ -0,0 +1,112 @@
+const { Dimensions, findNodeHandle } = require("react-native");
+const getInspectorDataForViewAtPoint = require("react-native/Libraries/Inspector/getInspectorDataForViewAtPoint");
+
+const getInspectorDataForInstance = (node) => {
```
kmagiera, line 4:
```suggestion
function getInspectorDataForInstance(node) {
```

The same goes for other methods defined here

packages/vscode-extension/lib/inspector.js
```diff
@@ -0,0 +1,112 @@
+const { Dimensions, findNodeHandle } = require("react-native");
+const getInspectorDataForViewAtPoint = require("react-native/Libraries/Inspector/getInspectorDataForViewAtPoint");
```
kmagiera, line 2: There's a reason why we use inline require for this method in wrapper.js

In short: since wrapper can be loaded out of normal order, we want to avoid a chain reaction of other modules being loaded that way. Hence, we should only load it when we need it (in getInspectorDataForCoordinates method) – at that point the whole bundle should be fully loaded and executed hence the order should no longer matter.

packages/vscode-extension/lib/inspector.js
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+const traverseComponentsTreeUp = (node, stack) => {
+  // Optimization: we break after reaching fiber node corresponding to OffscreenComponent (with tag 22).
+  // https://github.com/facebook/react/blob/c3570b158d087eb4e3ee5748c4bd9360045c8a26/packages/react-reconciler/src/ReactWorkTags.js#L62
+  if (!node || node.tag === 22) {
+    return stack;
+  } else {
+    const data = getInspectorDataForInstance(node);
+    const item = data.hierarchy[data.hierarchy.length - 1];
```
kmagiera, line 38: getInspectorDataForInstance can return empty objects in which case this line will crash

packages/vscode-extension/lib/inspector.js
```diff
@@ -0,0 +1,112 @@
...
+    const inspectorData = item.getInspectorData((arg) => findNodeHandle(arg));
+
+    const stackElementPromise = new Promise((resolve, _) => {
+      inspectorData.measure((_x, _y, viewWidth, viewHeight, pageX, pageY) => {
```
kmagiera, line 42: we used to have `try catch` block around the measure call. I suspect it was there for a reason

kmagiera (commented):
I'm not a fan of using arrow function when without a specific reason. Specifically when defining top level methods.

kmagiera (changes requested):
This PR seems to refactor some bits of the old codebase w/o qestioning why they've been implemented in the certain way. Specifically:
1) inline requies
2) try catch around measure
3) Promise.all approach

The use of promises here doesn't seem to make a lot of sense. Before the role of promises was such that we can call async measure in parallel. This PR changes it and makes all measure calls sequential and thus potentially slower.

packages/vscode-extension/lib/inspector.js
km1chno: right, to be honest I'm not sure what it was for, I thought that if an exception is thrown inside promise, it rejects with this error 

but I put it back as I'm not sure, maybe there is something weird inside the `measure` function that I don't know about

packages/vscode-extension/lib/inspector.js
km1chno: thanks for this remark, I feel like I'm learning a lot about javascript in general working on this project:D 

I changed it back to inline require

packages/vscode-extension/lib/inspector.js
```diff
@@ -0,0 +1,112 @@
...
+  const { width: screenWidth, height: screenHeight } = Dimensions.get("screen");
+
+  const scaleFrame = (frame) => ({
+    x: frame.x / screenWidth,
+    y: frame.y / screenHeight,
+    width: frame.width / screenWidth,
+    height: frame.height / screenHeight
+  });
+
+  return new Promise((resolve, _) => {
```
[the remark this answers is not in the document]
km1chno: this was my way of getting around the fact that `getInspectorDataForViewAtPoint` does not return and it seems to not bind to names in upper scope in its callback function, but I changed it a bit so that `getInspectorDataForCoordinates` obtains callback that is called inside the `getInspectorDataForViewAtPoint` callback. I believe now it is cleaner

km1chno:
@kmagiera As for the `Promise.all` - I missed the fact that it makes the promise resolution parallel. I put it back because why not, it could be potentially faster this way in some scenarios, as you said

kmagiera (approved):
Thanks for addressing all the comments. This looks good now

packages/vscode-extension/lib/inspector.js
```diff
@@ -0,0 +1,118 @@
+const { Dimensions, findNodeHandle } = require("react-native");
+
+function requireGetInspectorDataForViewAtPoint() {
+  return require("react-native/Libraries/Inspector/getInspectorDataForViewAtPoint");
```
kmagiera, line 4: We should just put require call where this method is used. Very little benefit from extracting it to a method

km1chno: moved to the method

packages/vscode-extension/lib/wrapper.js
```diff
@@ -179,69 +177,20 @@ export function AppWrapper({ children, initialProps, ..._rest }) {
...
-          }
-          stackPromise.then((stack) => {
-            devtoolsAgent._bridge.send("RNIDE_inspectData", {
-              id: payload.id,
-              frame: scaledFrame,
-              stack: stack,
-            });
+      const { id, x, y, requestStack } = payload;
+      
+      InternalImports.getInspectorDataForCoordinates(
```
km1chno, line 182: I added the `require('./inspector').getInspectorDataForCoordinates` to `InternalImports` because if it was just a regular import on top level, I got error `cannot read getInspectorDataForCoordinates of undefined`. It seemed to be exclusive for `react-native-76` app. On all other test apps it worked fine.

Now it works everywhere

kmagiera (changes requested):
Appears like there are some issues with this change on React Native 76. Marking this as 'request changes" to avoid it being accidentally merged.

packages/vscode-extension/lib/inspector.js
```diff
@@ -0,0 +1,116 @@
...
+  let node = startNode;
+
+  // Optimization: we break after reaching fiber node corresponding to OffscreenComponent (with tag 22).
+  // https://github.com/facebook/react/blob/c3570b158d087eb4e3ee5748c4bd9360045c8a26/packages/react-reconciler/src/ReactWorkTags.js#L62
+  while (node && node.tag !== 22) {
+    const data = getInspectorDataForInstance(node);
+
+    data.hierarchy?.length && stackPromises.push(new Promise((resolve, reject) => {
+      const item = data.hierarchy[data.hierarchy.length - 1];
+      const inspectorData = item.getInspectorData((arg) => findNodeHandle(arg));
```
maciekstosio, line 43:
```suggestion
      const inspectorData = item.getInspectorData(findNodeHandle);
```

packages/vscode-extension/lib/wrapper.js
```diff
@@ -69,6 +69,118 @@ function useAgentListener(agent, eventName, listener, deps = []) {
   }, [agent, ...deps]);
 }
 
+function getInspectorDataForInstance(node) {
+  const renderers = Array.from(window.__REACT_DEVTOOLS_GLOBAL_HOOK__?.renderers?.values());
```
kmagiera, line 73: It seems like we should get the renderer at the start of the traversal. There's no need to do all of that at every iteration of the loop

packages/vscode-extension/lib/wrapper.js
```diff
@@ -69,6 +69,118 @@ function useAgentListener(agent, eventName, listener, deps = []) {
...
+  if (!renderers) {
+    return {};
+  }
+  for (const renderer of renderers) {
+    if (renderer.rendererConfig?.getInspectorDataForInstance) {
+      const data = renderer.rendererConfig.getInspectorDataForInstance(node);
```
kmagiera, line 79: we will find the exact same `rendererConfig` object at every iteration of the loop from traversal. It'd seem to make more sense to do it once and not separately for every single node

packages/vscode-extension/lib/wrapper.js
```diff
@@ -69,6 +69,118 @@ function useAgentListener(agent, eventName, listener, deps = []) {
...
+function traverseComponentsTreeUp(startNode) {
+  const stackPromises = [];
+  let node = startNode;
+
+  // Optimization: we break after reaching fiber node corresponding to OffscreenComponent (with tag 22).
+  // https://github.com/facebook/react/blob/c3570b158d087eb4e3ee5748c4bd9360045c8a26/packages/react-reconciler/src/ReactWorkTags.js#L62
+  while (node && node.tag !== 22) {
+    const data = getInspectorDataForInstance(node);
+
+    data.hierarchy?.length && stackPromises.push(new Promise((resolve, reject) => {
```
kmagiera, line 110: that's a very obfuscated way to write an `if` statement. I'm fine with doing that for one liners, but not when there's a huge block of code afterwards

kmagiera (changes requested):
Left some more inline comments in the code.

In general what worries me is that I don't understand what this PR is actually fixing. It links to an empty issue but there are no reproducable steps that would allow us to test before/after this change and see if it fixes anything. This is important as in the future, if someone wants to change something in this code, they may want to test again the scenario that this PR claims to fix.

packages/vscode-extension/lib/wrapper.js
km1chno: changed to obtain the `getInspectorDataForInstance` method only once at the beginning of `traverseComponentsTreeUp`

packages/vscode-extension/lib/wrapper.js
km1chno: as above

packages/vscode-extension/lib/wrapper.js
km1chno: changed to regular if statement

kmagiera (approved):
Looks good now

packages/vscode-extension/lib/wrapper.js
```diff
@@ -69,6 +69,126 @@ function useAgentListener(agent, eventName, listener, deps = []) {
...
+        try {
+          inspectorData.measure((_x, _y, viewWidth, viewHeight, pageX, pageY) => {
+            const stackElementFrame = {
+              x: pageX,
+              y: pageY,
+              width: viewWidth,
+              height: viewHeight
+            };
+
+            stackElement = (inspectorData.source) ?
```
kmagiera, line 130:
```suggestion
            const stackElement = (inspectorData.source) ? 
```

kmagiera: Also, it appears like we can tell whether source is set before we call measure, so maybe we can skip calling measure altogether in such a case?

km1chno: added an if statement before measure to check if source is absent and resolve the promise early in such case

kmagiera:
Will re-test again after changes and try to land this

PR #695, opened 2024-11-05
Add option to display screen touches

kmagiera:
This PR add option to show touches on the device screen and on replays.

As part of this PR:
1) We're updating simulator server version to include changes necessary for drawing touch pointers
2) We're adding new device setting for showing touches
3) We're renaming "screen recording" section to "screen settings" as it now includes both "screen recording" and "show touches" options.

### How Has This Been Tested: 
1. Run one of the test apps
2. Enable show touches -> start pressing the screen to see touch indicator
3. Record video to see the indicator being visible on the recording
4. Disable show touches option to see them gone


km1chno (approved):
looks good, just one inline comment

PR #696, opened 2024-11-05
Bring back ability to reveal preview when stuck on waiting

kmagiera:
In #475 we accidently broke the functionality that'd allow users to tap on "waiting for app to load" text in order to reveal the device preview.

This option was pretty handy specifically to troubleshoot issues when app gets stuck because of expo devclient onboarding dialog.

This PR brings back the callback-based code that got removed in #475 with small adjustments as the underlying implementation from deviceSession got split into two separate methods in the meantime.

### How Has This Been Tested: 
1. Run expo dev client 51
2. Use "reinstall app" option, the app would get stuck on the waiting to load screen
3. Click on the "waiting to load" text to reveal preview – after this change you should be able to see the preview, without this change the button would do nothing as the frontend code doesn't have access to the preview URL until the app is fully launched.

### DIFF
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -597,8 +597,8 @@ export class Project
 
       const previewURL = await newDeviceSession.start(this.deviceSettings, {
         cleanBuild: forceCleanBuild,
-        previewReadyCallback: (previewURL) => {
-          this.updateProjectStateForDevice(deviceInfo, { previewURL });
+        previewReadyCallback: (url) => {
+          this.updateProjectStateForDevice(deviceInfo, { previewURL: url });
         },
       });
       this.updateProjectStateForDevice(this.projectState.selectedDevice!, {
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/project/project.ts
```diff
@@ -597,6 +597,9 @@ export class Project
 
       const previewURL = await newDeviceSession.start(this.deviceSettings, {
         cleanBuild: forceCleanBuild,
+        previewReadyCallback: (url) => {
+          this.updateProjectStateForDevice(deviceInfo, { previewURL: url });
+        },
       });
       this.updateProjectStateForDevice(this.projectState.selectedDevice!, {
```
balins, line 604: This is called twice with `previewURL`, some explanation why is that could be useful here (I assume that ideally, the second call would just set the status to `"running"`, since the `previewURL` has been already set at this point but I don't know the codebase).

balins (approved):
Looks good!

packages/vscode-extension/src/project/project.ts
km1chno: +1, couldn't the second one update only "status"? then the `previewURL` local variable is redundant too

PR #697, opened 2024-11-05
Add test app for Expo 52

kmagiera:
This PR adds test app with Expo 52 (prebuild).

### How Has This Been Tested: 
1. Run the newly added app on Android and iOS


maciekstosio (changes requested):
Shouldn't the folder be called `expo52-dev-client` to keep it consistent?

maciekstosio:
And the test app seems to be cut. But it doesn't seem to be a problem with IDE.
|CLI|Radon|
|-|-|
|expo52-CLI|expo52-RADON|

kmagiera:
this isn't a dev-client based project hence the name is not expo52-dev-client

maciekstosio:
Right, I got confused by the ios/android folders. But running prebuild gives us a different setup than "default" expo project. What do you think about calling it `expo52-cng` or `expo52-prebuild`?

maciekstosio (approved):
Other then above it looks ok I was able to run it on both platforms

kmagiera:
I agree, we may want to unify it somehow eventually.

PR #699, opened 2024-11-05
Fix metro cache issue when running project from CLI and IDE interchangeably

kmagiera:
When working on a single project with IDE and from CLI interchangeably, metro cache may cause bundling issue because of the way we override files in babel transformer. This can happen when transformed output of certain files, for which we modify the source code is cached. Although I couldn't find a reliable way of reproducing this problem, it seem to randomly pop up when building from CLI and then using the IDE, and can surface via the CLI bundler or the IDE bundler.

To workaround this problem, we are using cacheVersion metro configuration option. This lets the IDE to generate its own, unique cache keys such that to avoid cache collisions between the metro started via CLI and in the IDE. 

Here's a breakdown of changes we're making here:
1) We're adding cacheVersion config option to metro
2) We're adding blockList option to metro's resolver, as fixing the cache issue revealed another issue with metro resolving files from under the extension directory when used with expo's fast resolver option (EXPO_USE_FAST_RESOLVER)
3) We're adding an option to restart the project with metro cache reset – this way to provide an easy troubleshooting step for the cases when this or some other metro cache related issue occurs

We're also renaming some env variables starting with REACT_NATIVE_IDE to RADON_IDE that I discovered in the metro config code.

### How Has This Been Tested: 
1) I used a new create-expo-stack project with tailwind, expo-dev-client (SDK 51) and EXPO_USE_FAST_RESOLVER option
2) Build the project from CLI and IDE interchangeably several times until you start getting weird bundle errors
3) After this change, it should no longer happen
4) Test "Clear metro cache" option and see that it restarts the process, make sure the native build doesn't trigger and also that it takes longer to bundle first time it loads after this option is selected

### DIFF
--- a/packages/vscode-extension/lib/metro_helpers.js
+++ b/packages/vscode-extension/lib/metro_helpers.js
@@ -23,7 +23,7 @@ function overrideModuleFromAppDir(moduleName, exports) {
   }
 }
 
-const extensionLib = process.env.REACT_NATIVE_IDE_LIB_PATH;
+const extensionLib = process.env.RADON_IDE_LIB_PATH;
 
 function adaptMetroConfig(config) {
   // We use processorModuleFilter to inject some code into the bundle prelude.
@@ -71,7 +71,7 @@ function adaptMetroConfig(config) {
   } else {
     const originalResolveRequest = config.resolver?.resolveRequest;
     if (originalResolveRequest) {
-      // Some storybook setups relie on resolveRequest being overridden
+      // Some storybook setups rely on resolveRequest being overridden
       // in order to exclude storybook files from being imported into the bundle.
       // The files are only included when STORYBOOK_ENABLED environment variable
       // is set. Apparently, we can't set that variable for the whole metro process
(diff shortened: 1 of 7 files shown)


packages/vscode-extension/lib/metro_helpers.js
```diff
@@ -128,10 +128,26 @@ function adaptMetroConfig(config) {
...
+  if (config.resolver.blockList) {
+    // if block list is array, we use it as original block list
+    if (Array.isArray(config.resolver.blockList)) {
+      origBlockList = config.resolver.blockList;
+    } else {
+      // otherwise we create a new array containing the original block list
+      origBlockList = [config.resolver.blockList];
+    }
+  }
+  config.resolver.blockList = [new RegExp(extensionLibNodeModules + "/.*"), ...origBlockList];
```
balins, line 149: Are you sure this works for Windows? I mean, if `RegExp` properly handles backslashes in `extensionLibNodeModules`, especially that we potentially mix it with a forward slash

packages/vscode-extension/src/webview/components/UrlBar.tsx
```diff
@@ -18,7 +18,8 @@ function ReloadButton({ disabled }: { disabled: boolean }) {
         "Reload JS": () => project.reload("reloadJs"),
         "Restart app process": () => project.reload("restartProcess"),
         "Reinstall app": () => project.reload("reinstall"),
-        "Clean rebuild": () => project.restart(true),
+        "Clear metro cache": () => project.restart("metro"),
```
balins, line 21: nit: `Metro` should be capitalized

balins (approved):

maciekstosio (approved):

packages/vscode-extension/lib/metro_helpers.js
kmagiera: This is actually only needed in extension development process. When extension is installed from the store, it doesn't have node_modules folder under the extension lib location. I maybe able to fix this by adding ".*" as path element when joining.

PR #709, opened 2024-11-07
Fix appearance on high contrast themes

### LINKED ISSUE #702: Dialogs and popups are hardly visible when using high contrast theme
Here's an example how things look like when using default "Dark High Contrast" theme:

image

All dialogs and popups have transparent background making it impossible to read their content. This also includes inspector context menu.

To reproduce:
1) Select "Color theme" from command palette and select "Dark High Contrast" (it is a theme that's installed by default)

km1chno:
Fixes #702 

### Screens

Zrzut ekranu 2024-11-7 o 18 36 29
Zrzut ekranu 2024-11-7 o 18 35 49

### How Has This Been Tested: 

Set theme to `Dark High Contrast` or `Light High Contrast` in vscode and open any app in extension. Check appearance of component:
- any dialog/modal
- any popup
- tooltips
- buttons
- inspector menu
- inspector overlay
- device settings
- manage devices
- replay overlay
- replay UI

### DIFF
--- a/packages/vscode-extension/src/webview/styles/theme.css
+++ b/packages/vscode-extension/src/webview/styles/theme.css
@@ -137,7 +137,8 @@
 }
 
 /* Light theme */
-body[data-vscode-theme-kind="vscode-light"] {
+body[data-vscode-theme-kind="vscode-light"],
+body[data-vscode-theme-kind="vscode-high-contrast-light"] {
   --swm-active-item: var(--blue-dark-100);
   --swm-preview-background: var(--white);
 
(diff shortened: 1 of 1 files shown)


kmagiera:
Have you tested things like manage device dialogs, settings menu, etc? These all look bad with high contrast themes. Also I wonder if there are other "data-vscode-theme-kind"s that we should consider?

p-malecki:
> Have you tested things like manage device dialogs, settings menu, etc? These all look bad with high contrast themes. Also I wonder if there are other "data-vscode-theme-kind"s that we should consider?

I believe @km1chno covered all kinds of themes according to:
https://code.visualstudio.com/docs/getstarted/themes#:~:text=To%20customize%20the%20themes%20that,defaults%20to%20Light%20High%20Contrast
Screenshot 2024-11-08 at 11 09 56

kmagiera (approved):
Add testing dialogs and menus to test plan. Otherwise looks good

PR #715, opened 2024-11-08
Fix missing key mappings and warning from DeviceRenameDialog

p-malecki:
This PR resolves an issue where missing mappings for `ArrowUp`, `ArrowDown`, and `IntlBackslash` in `keyboardEventToHID` led to null values being sent to the sim-server and following error message:

```
[ERROR simulator_server] Error when handling input command: Error parsing positional argument 'code' with value 'null': invalid digit found in string)
```

The second issue fixed was a warning in `DeviceRenameDialog` about an input element having both `value` and `defaultValue` props, which led to another warning.

### How Has This Been Tested:
- Tested by pressing all MacBook keyboard keys to prevent sim-server errors.
- Opened DeviceRenameDialog and verified that no warning messages appear.


maciekstosio (approved):
I would consider throwing a warning in the Radon IDE output instead of passing null to sim-server. But other than that looks ok.

PR #716, opened 2024-11-08
Store temporary mouse position in ref to prevent unnecessary rerenders

maciekstosio:
We store the position of the touchPoint in the state; we need it to display touch indicators while multitouching. Currently, the state is updated on mouse move to prevent a jump (see the attached video), even when multitouch is inactive. This causes rerenders even when we don't need to render touch indicators in new positions. This PR moves the tracking mouse position to the ref when the multitouch is inactive. This is a quick fix, as I believe in the future, we should split Preview into smaller pieces, so those rerenders are contained in a small Component. 

<details>
<summary>touch indicators jump problem</summary>

|Without continuos  tracking|With continuos tracking|
|-|-|
|<video src="https://github.com/user-attachments/assets/a1a37356-4bea-4100-99cc-f0b48d0f7ac7" />|<video src="https://github.com/user-attachments/assets/df51f852-4fbe-4ff5-9955-43793b33ac15" />|

</details>

### How Has This Been Tested: 

Add `console.log('render') to the `Preview.tsx` before `return`. Then focus IDE preview and move the mouse around. Before `Preview.tsx` constantly rerenders, now, it happens only when multitouch is active.


p-malecki (approved):
LGTM

PR #718, opened 2024-11-08
Unclick all keys on IDE panel focus change

### LINKED ISSUE #710: Unclick all meta keys on IDE panel focus change
When sending keyboard events we sometimes miss seinding up event in case user changes window focus.

This causes the simulator to think that some keys are pressed forever and may lead to bugs like the one reported in #326

I dont think there's an ideal solution to this, as when regaining window focus we can't access the keys that are being pressed at that moment. I believe that a good approximation would be to emit up events. As in the worst case when the user expects some key to be pressed and it doesn't behave like it, they can press it again.

We should *at least* emit up events for meta keys (ctr/cmd/alt) that are used as modifier keys. Or perhaps we can remember all down events without corresponding up and emit up on window focus change.

To reporduce the problem you can focus a TextInput, hold "a" and focus other window - the preview thinks "a" is still hold.

maciekstosio:
Fixes #710 

Currently, when we send down an event for a key and change focus to another window the up event is never sent. Best to observe the following steps from how this has been tested. But this might be especially confusing while sending modifier keys. 

This PR adds keeping track of pressed keys and sending up events when losing focus. 

### How Has This Been Tested: 

Focus TextInput press "d", focus another window while pressing "d". Before it keeps adding after it stops when losing focus. 

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/9d503abf-97c0-4f43-8786-8d9e1bbb5ac4" />|<video src="https://github.com/user-attachments/assets/720aefbf-a2f5-4de1-acbc-f8830e854e1b" />|

### DIFF
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -1,4 +1,4 @@
-import { useState, useRef, useEffect, MouseEvent, forwardRef, RefObject, ReactNode } from "react";
+import { useState, useRef, useEffect, MouseEvent, forwardRef, RefObject, ReactNode, useCallback } from "react";
 import clamp from "lodash/clamp";
 import { VSCodeProgressRing } from "@vscode/webview-ui-toolkit/react";
 import { Resizable } from "re-resizable";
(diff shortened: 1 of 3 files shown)


maciekstosio:
Changed to draft, found another issue.

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -37,6 +46,40 @@ declare module "react" {
 
 const HIDE_ZOOM_CONTROLS_DELAY = 2000;
 
+const useKeyPresses = () => {
```
kmagiera, line 49:
```suggestion
function useKeyPresses() {
```

kmagiera (approved):

PR #724, opened 2024-11-13
Handle nx monorepo setup

### LINKED ISSUE #683: Radon does not use root node modules in nx workspace
**Description**

When using NX monorepos, Radon will try to use apps root path to install latest packages instead of using root node_modules and specified versions.

Here is the reproducible: https://github.com/matinzd/radon-ide-bug-nx

Try running it with Radon and it will try to install RN 0.76.1 although the specified version is 0.74.5 in root `package.json`.

**Environment**

- Radon: 0.0.21
- React Native: 0.75.4 and 0.75.4
- Any other versions of tools related to the problem.

maciekstosio:
Fixes #683

This PR adds nx monorepo setup handling by: 
- adding recognizing nx monorepo root 
- installing dependencies in the workspace root 
- adding nx-monorepo test-app 

**Remarks**:
- I created PR (https://github.com/nrwl/nx/pull/28931) to nx so their template doesn't fail in Radon IDE, we don't set their envs, my understanding is that those are used internally by their CLI tool, and I don't think want to implement that
- ~~iOS fails on `bundle install && bundle exec pod install` with `EACCES: Permission denied` I'm investigating if it's related to IDE or something else (i.e. my setup), manual `pod install` solves the issue~~ seems like problem with my setup (those files were owned by root, reinstalling ruby using rbenv solved the problem) 

### How Has This Been Tested: 
**Scenario:** Open iOS and Android, check if it builds.
**Tested on:** 
- workspace-example
- expo-52
- nx-monorepo


packages/vscode-extension/src/utilities/common.ts
```diff
@@ -16,16 +16,24 @@ export function getDevServerScriptUrl() {
 
 export function isWorkspaceRoot(dir: string) {
   const packageJsonPath = path.join(dir, "package.json");
-  let workspaces;
+  const nxJsonPath = path.join(dir, "nx.json");
+
   try {
-    workspaces = require(packageJsonPath).workspaces;
+    require(nxJsonPath);
```
filip131311, line 22: Is there a scenario in which require fails for reasons unrelated to nx.json existence?
Could we use `fs.existSync()` here?

filip131311 (approved):
I tested it and it works, but please look at inline comments before you merge

packages/vscode-extension/src/utilities/common.ts
maciekstosio: Changed in: https://github.com/software-mansion/radon-ide/pull/724/commits/7264f94c25a04bb0854ce8bd2beb27333d41a839

Issue #678, opened 2024-10-30
EAS custom build requires prebuild, also android emulator doesn't work.


aka241193:
**Description**
EAS build requires `expo prebuild` to be run and contain android and ios folder explicitly. Which I guess shouldn't be the case since the build is being used from eas as seen the log.

launch.json
```
    {
      "type": "radon-ide",
      "request": "launch",
      "name": "Radon IDE panel",
      "isExpo": true,
      "eas": {
        "ios": {
          "profile": "development-simulator"
        },
        "android": {
          "profile": "development-simulator", 
        }
      },
    },
```

Further on executing the command `expo prebuild` and build being used from eas following are the failures for:
android
```
Error: Command failed with exit code 1: /Users/someUser/Library/Android/sdk/platform-tools/adb -s emulator-5554 uninstall com.app.some
Failure [DELETE_FAILED_INTERNAL_ERROR]
adb: filename doesn't end .apk or .apex
```

ios works fine.

I have followed all the troubleshooting steps mentioned here: https://ide.swmansion.com/docs/troubleshooting

**Environment**

- Version of currently used Radon IDE (Extensions tab -> click on Radon IDE extension -> version is located next to the extension name) v0.0.21
- Version of React Native/Expo used in the project.
"expo": "~51.0.28",
"react-native": "0.74.5",
- Any other versions of tools related to the problem.

pietgk:
please note that eas.json probably does not have an android simulater build, so make sure eas.json builds match your profile name in your launch configuration

pietgk:
i could partly solve the android issue by manually installing the build, after renaming it to have the .apk extension
(the adb command is copied from the logs).

```
╰─➤  adb -s emulator-5554 install -d -r /var/folders/02/lddpjzr501b5y5t3x_70dwr40000gn/T/rn-ide-eas-build-1PLeB0/7834fb61-afb8-4269-9cf0-57b52ae65f77
Performing Streamed Install
adb: filename doesn't end .apk or .apex: /var/folders/02/lddpjzr501b5y5t3x_70dwr40000gn/T/rn-ide-eas-build-1PLeB0/7834fb61-afb8-4269-9cf0-57b52ae65f77
╭─grop@MB-Piet ~/dev/app/development ‹3.3.0› ‹development*› 
╰─➤  cd  /var/folders/02/lddpjzr501b5y5t3x_70dwr40000gn/T/rn-ide-eas-build-1PLeB0                                     
[...]

PR #727, opened 2024-11-13
Fix Eas builds on Android

filip131311:
This PR adds additional user facing information about eas builds, to let users know when production build is used instead of development one. More over eas build crashes are getting a new build error alert description to redirect users to Radon IDE log instead of non existing build logs.

Finally this PR adds .apk file extension to android packages to avoid errors with tools we use to install and uninstall application on device.

Fixes android portion of  #678

### How Has This Been Tested: 

Run an application using eas on android using production and development profiles.

### DIFF
--- a/packages/vscode-extension/src/builders/eas.ts
+++ b/packages/vscode-extension/src/builders/eas.ts
@@ -64,6 +64,13 @@ async function fetchBuild(config: EasConfig, platform: DevicePlatform) {
 
   const build = maxBy(builds, "completedAt")!;
 
+  if (!build.binaryUrl.endsWith(".apk") && !build.binaryUrl.endsWith(".apex")) {
+    Logger.error(
+      `EAS build artefact needs to be in .apk or .apex format to work with the Radon IDE, make sure you set up eas to use "development" profile`
+    );
+    return undefined;
+  }
+
   Logger.debug(`Using EAS build artifact with ID ${build.id}.`);
   return build;
 }
@@ -80,7 +87,10 @@ async function downloadAppFromEas(
   const { id, binaryUrl } = build;
 
   const tmpDirectory = await mkdtemp(path.join(os.tmpdir(), "rn-ide-eas-build-"));
-  const binaryPath = path.join(tmpDirectory, id);
+  const binaryPath =
+    platform === DevicePlatform.Android
+      ? path.join(tmpDirectory, `${id}.apk`)
+      : path.join(tmpDirectory, id);
 
   const success = await downloadBinary(binaryUrl, binaryPath);
   if (!success) {
--- a/packages/vscode-extension/src/builders/eas.ts
+++ b/packages/vscode-extension/src/builders/eas.ts
@@ -66,7 +66,7 @@ async function fetchBuild(config: EasConfig, platform: DevicePlatform) {
 
   if (!build.binaryUrl.endsWith(".apk") && !build.binaryUrl.endsWith(".apex")) {
     Logger.error(
-      `EAS build artefact needs to be in .apk or .apex format to work with the Radon IDE, make sure you set up eas to use "development" profile`
+      `EAS build artifact needs to be a development build in .apk or .apex format to work with the Radon IDE, make sure you set up eas to use "development" profile`
     );
     return undefined;
   }
--- a/packages/vscode-extension/src/webview/hooks/useBuildErrorAlert.tsx
+++ b/packages/vscode-extension/src/webview/hooks/useBuildErrorAlert.tsx
@@ -30,7 +30,8 @@ function BuildErrorActions({
         onClick={() => {
           if (logsButtonDestination === "extension") {
             project.focusExtensionLogsOutput();
-            return;
+          } else {
+            project.focusBuildOutput();
           }
           project.focusBuildOutput();
         }}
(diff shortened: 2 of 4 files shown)


packages/vscode-extension/src/builders/eas.ts
```diff
@@ -64,6 +64,13 @@ async function fetchBuild(config: EasConfig, platform: DevicePlatform) {
 
   const build = maxBy(builds, "completedAt")!;
 
+  if (!build.binaryUrl.endsWith(".apk") && !build.binaryUrl.endsWith(".apex")) {
+    Logger.error(
+      `EAS build artefact needs to be in .apk or .apex format to work with the Radon IDE, make sure you set up eas to use "development" profile`
```
kmagiera, line 69:
```suggestion
      `EAS build artifact needs to be a development build in .apk or .apex format to work with the Radon IDE, make sure you set up eas to use "development" profile`
```

packages/vscode-extension/src/builders/eas.ts
```diff
@@ -80,7 +87,10 @@ async function downloadAppFromEas(
   const { id, binaryUrl } = build;
 
   const tmpDirectory = await mkdtemp(path.join(os.tmpdir(), "rn-ide-eas-build-"));
-  const binaryPath = path.join(tmpDirectory, id);
+  const binaryPath =
+    platform === DevicePlatform.Android
+      ? path.join(tmpDirectory, `${id}.apk`)
```
kmagiera, line 92: The error message from above says it can be in apex format too?

packages/vscode-extension/src/webview/hooks/useBuildErrorAlert.tsx
```diff
@@ -24,6 +28,10 @@ function BuildErrorActions() {
       <IconButton
         type="secondary"
         onClick={() => {
+          if (logsButtonDestination === "extension") {
+            project.focusExtensionLogsOutput();
+            return;
+          }
```
kmagiera, line 34:
```suggestion
          } else {
            project.focusBuildOutput();
          }
```

packages/vscode-extension/src/webview/hooks/useBuildErrorAlert.tsx
```diff
@@ -68,11 +77,20 @@ export function useBuildErrorAlert(shouldDisplayAlert: boolean) {
       'Your project does not have "ios" directory. If this is an Expo project, you may need to run `expo prebuild` to generate missing files, or configure external build source using launch configuration.';
   }
 
+  const isEasBuild =
+    (!!eas?.android && projectState.selectedDevice?.platform === DevicePlatform.Android) ||
+    (!!eas?.ios && projectState.selectedDevice?.platform === DevicePlatform.IOS);
```
kmagiera, line 82: wouldn't it be easier to pass information whether its eas or not via project? It seems like we have those checks already to tell which build to start, so duplicating it here would just increase the coupling

packages/vscode-extension/src/builders/eas.ts
filip131311: according to the [documentation](https://source.android.com/docs/core/ota/apex#apex-are-apk) are apk files so adding less specific extension won't break anything

packages/vscode-extension/src/webview/hooks/useBuildErrorAlert.tsx
filip131311: the check if the build should come from eas is very similar to the check here, just 
```
 if (eas?.android) {
    const apkPath = await fetchEasBuild(cancelToken, eas.android, DevicePlatform.Android);
    if (!apkPath) {
      throw new Error("Failed to build Android app using EAS build.");
    }
```

I think it is okay to use Launch Configuration here as we use it on webview side anyway

packages/vscode-extension/src/webview/providers/LaunchConfigProvider.tsx
```diff
@@ -18,11 +19,16 @@ const launchConfig = makeProxy<LaunchConfig>("LaunchConfig");
 type LaunchConfigContextType = LaunchConfigurationOptions & {
   update: LaunchConfigUpdater;
   xcodeSchemes: string[];
+  eas?: {
```
kmagiera, line 22: is eas optional given it is set as empty object in default context?

kmagiera (approved):

packages/vscode-extension/src/webview/providers/LaunchConfigProvider.tsx
filip131311: thx i removed it from default context

packages/vscode-extension/src/builders/eas.ts
```diff
@@ -64,6 +64,13 @@ async function fetchBuild(config: EasConfig, platform: DevicePlatform) {
 
   const build = maxBy(builds, "completedAt")!;
 
+  if (!build.binaryUrl.endsWith(".apk") && !build.binaryUrl.endsWith(".apex")) {
+    Logger.error(
+      `EAS build artifact needs to be a development build in .apk or .apex format to work with the Radon IDE, make sure you set up eas to use "development" profile`
+    );
+    return undefined;
+  }
```
EricZeiberg, line 72: I'm getting this error when trying to build my app on iOS, and from looking at the code, it seems that this check is run even on iOS EAS builds?

filip131311: thank you I added a platform check to avoid that in #796

Issue #708, opened 2024-11-07
Add controls for normal screen recording (start / stop)


kmagiera:
We currently only have replays functionality. Sim-server can already support regular recording that writes to a file, but we don't yet have an interface for it. We'd like to add some sensible controls for that (note that replay button already takes up a lot of space on the panel). After recording is done we may either use replay overlay for previewing, or trigger save dialog immediately. We should also have some safe-guards for recording such that it stops after some time to prevent cases when someone forgot to stop recording.

PR #733, opened 2024-11-15
Add recording controls enhancement

p-malecki:
This PR introduces recording controls enhancement: 
- introduces a toggle button to start/stop recordings
- opens a save dialog immediately after pressing stop (in future, we could reuse replayOverlay for video previews and edits)
- ignores and does not save recordings under 1 second
- adds a maximum limit of 10 minutes for recordings, after which it cancels the recording
- change default save location to Desktop or homedir (on Windows) and add timestamp to filename
- eliminates text labels from the Replay and Logs buttons to create more space in the top button group
- updates the Replay button icon

### Test plan:
1. Start a recording, interact with the test app, then stop the recording after a few seconds and save the file to a selected location. Open the saved recording to verify that the app's screen was correctly captured.
2. Start a recording and stop it immediately to confirm that the save dialog does not appear for recordings under 1 second.
3. Test the maximum recording time to ensure that it automatically stops and opens the save dialog after the 5-minute limit is reached.
4. Ensure that the Replay button is disabled during recording.
5. After capturing a recording, test the Replay functionality to ensure it still works correctly.
6. Verify that the timestamps on the recorded video filenames reflect the correct date and time, matching the system clock.

Before:
Screenshot 2024-11-15 at 17 22 48

After:
Screenshot 2024-11-15 at 17 23 34
Screenshot 2024-11-18 at 18 52 57
<video src="https://github.com/user-attachments/assets/03e4422f-829c-4b42-aa75-b8306ee96d91" />

resolves: #708


packages/vscode-extension/src/devices/preview.ts
```diff
@@ -102,23 +116,23 @@ export class Preview implements Disposable {
     this.subprocess?.stdin?.write("pointer show false\n");
   }
 
-  public startReplays() {
+  public startVideoRecording(videoId: string) {
```
kmagiera, line 119: It is a wrong level of abstraction here. We assume in the callback that IDs can be "recording" and "replay" so we shouldn't allow any string as videoID, as if anything else is passed, the callback would never trigger.

One way would be to specify allowed values in typescript, but I think that this method shouldn't expose videoID as a parameter and just make it an implementation detail.

packages/vscode-extension/src/common/Project.ts
```diff
@@ -145,6 +145,9 @@ export interface ProjectInterface {
   getDeepLinksHistory(): Promise<string[]>;
   openDeepLink(link: string): Promise<void>;
 
+  startRecording(): void;
+  stopRecording(): void;
+  captureRecording(): Promise<RecordingData>;
```
kmagiera, line 150: can we only have one method for stopping that returns the data? What is the point of having both?

kmagiera (changes requested):
I like the changes UI-wise, one suggestion I'd have is to maybe try moving the red dot and timer into the recording button. That is, when we hit record, the red dot displays there instead of "stop" button. This way it should be cleaner that the recording is in progress as the top bar is much more visible than the recording progress component that suddenly appears in top left edge when no one expects it.

packages/vscode-extension/src/devices/preview.ts
p-malecki: I've made the changes. Let me know if they look as you expected. Additionally, I've added 'sendCommandOrThrow' to reduce code duplication.

packages/vscode-extension/src/common/Project.ts
p-malecki: You're right, one is enough.

packages/vscode-extension/src/common/Project.ts
```diff
@@ -145,6 +145,8 @@ export interface ProjectInterface {
   getDeepLinksHistory(): Promise<string[]>;
   openDeepLink(link: string): Promise<void>;
 
+  startRecording(): void;
+  captureAndStopRecording(): Promise<RecordingData>;
```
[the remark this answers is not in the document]
p-malecki: Actually, simserver allows starting, stopping, and saving video recordings without stopping them. CaptureReplay is responsible only for saving, unlike captureAndStopRecording, which saves and stops recordings, so name captureAndStopRecording is more appropriate

packages/vscode-extension/src/devices/DeviceBase.ts
```diff
@@ -52,8 +52,22 @@ export abstract class DeviceBase implements Disposable {
...
+
+  public stopRecording() {
+    return this.preview?.stopRecording();
+  }
+
+  public async captureRecording() {
+    if (!this.preview) {
+      throw new Error("Preview not started");
+    }
+    return this.preview.captureRecording();
```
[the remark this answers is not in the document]
p-malecki: `stopRecording` is also called when device settings change or the app is reloaded, to stop recording in simserver if user forgets to pause it by pressing the recording button.

packages/vscode-extension/src/devices/DeviceBase.ts
```diff
@@ -63,6 +77,10 @@ export abstract class DeviceBase implements Disposable {
     return this.preview.startReplays();
   }
 
+  public stopReplays() {
+    return this.preview?.stopReplays();
+  }
```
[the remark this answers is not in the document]
p-malecki: I agree, names `enableReplay`and `disableReplay` seem more accurate.

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -6,22 +6,57 @@ import { Logger } from "../Logger";
...
+    let resolvePromise: (value: RecordingData) => void;
+    let rejectPromise: (reason?: any) => void;
+    const promise = new Promise<RecordingData>((resolve, reject) => {
+      resolvePromise = resolve;
+      rejectPromise = reject;
+    });
+
+    const lastPromiseKey =
+      promiseType === "recording" ? "lastRecordingPromise" : "lastReplayPromise";
+    const lastPromise = this[lastPromiseKey];
```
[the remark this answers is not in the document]
p-malecki: Right, this mapping could be useful in the future if we decide to introduce additional recordings.

packages/vscode-extension/src/devices/DeviceBase.ts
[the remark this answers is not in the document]
p-malecki: I removed the stopRecording calls after reloading and changing device settings, as the first was causing a bug with recordings not stopping in UI, and the second was unwanted behaviour (we don't want to stop recording on eg. font change), so there is no need for the stopRecording method, and we now use captureAndStopRecording instead.

packages/vscode-extension/src/common/Project.ts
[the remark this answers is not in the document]
kmagiera: what if we change it to `stopRecording(capture: boolean)`

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -6,22 +6,69 @@ import { Logger } from "../Logger";
...
+
+interface VideoRecordingPromiseHandlers {
   resolve: (value: RecordingData) => void;
   reject: (reason?: any) => void;
 }
 
 export class Preview implements Disposable {
+  private videoRecordingPromises: Map<
+    VideoRecordingPromiseType,
+    VideoRecordingPromiseHandlers | undefined
```
kmagiera, line 19: I think we can just use `string` here as key, the keys that we use are not important here because we don't iterate over the map anywhere and we also don't have any special handling based on the key now.

We should not allow undefined value. Just remove entry from map when we want to reset it.

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -6,22 +6,69 @@ import { Logger } from "../Logger";
...
   private subprocess?: ChildProcess;
   public streamURL?: string;
-  private lastReplayPromise?: ReplayPromiseHandlers;
 
-  constructor(private args: string[]) {}
+  constructor(private args: string[]) {
+    this.videoRecordingPromises = new Map<
+      VideoRecordingPromiseType,
+      VideoRecordingPromiseHandlers | undefined
+    >();
```
kmagiera, line 28: this is an empty map, we can construct it in the line where we define it

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -6,22 +6,69 @@ import { Logger } from "../Logger";
...
+
+    this.videoRecordingPromises.set("recording", undefined);
+    this.videoRecordingPromises.set("replay", undefined);
```
kmagiera, line 31: we don't need undefined value there. If its not there `map.get(key)` will return undefined

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -60,33 +107,43 @@ export class Preview implements Disposable {
...
+          const videoReadyMatch = line.match(/video_ready (\S+) (\S+) (\S+)/);
+          const videoErrorMatch = line.match(/video_error (\S+) (.*)/);
+
+          const videoId = videoReadyMatch
+            ? videoReadyMatch[1]
+            : videoErrorMatch
+            ? videoErrorMatch[1]
+            : "";
+
+          if (!this.videoRecordingPromises.has(videoId as VideoRecordingPromiseType)) {
```
kmagiera, line 123: if we just use string as key, we wouldn't need `as` here

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -60,33 +107,43 @@ export class Preview implements Disposable {
...
+            throw new Error(`Invalid video ID: ${videoId}`);
+          }
 
+          const handlers = this.videoRecordingPromises.get(videoId as VideoRecordingPromiseType);
```
kmagiera, line 127: same comment about `as`. We should just get it earlier and throw when `handlers` is undefined rather than using `.has`

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -60,33 +107,43 @@ export class Preview implements Disposable {
...
+          this.videoRecordingPromises.set(videoId as VideoRecordingPromiseType, undefined);
```
kmagiera, line 128: `.delete(videoId)`

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -102,39 +159,28 @@ export class Preview implements Disposable {
...
-      resolvePromise = resolve;
-      rejectPromise = reject;
-    });
-    if (this.lastReplayPromise) {
-      promise.then(this.lastReplayPromise.resolve, this.lastReplayPromise.reject);
-    }
-    this.lastReplayPromise = { resolve: resolvePromise!, reject: rejectPromise! };
-    stdin.write(`video replay save\n`);
-    return promise;
+    return this.handleVideoRecordingPromise("replay");
```
kmagiera, line 183: `handleVideoRecordingPromise` is a weird name. There's no promise at this point, so what is this method "handling"??

I'd just name it `saveVideo` or `saveVideoWithID`

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -126,6 +130,46 @@ function PreviewView() {
     }
   };
 
+  const MAX_RECORDING_TIME = 300;
```
kmagiera, line 133: append units to this constant such that it is clear, e.g. `MAX_RECORDING_TIME_SEC`

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -126,6 +130,46 @@ function PreviewView() {
...
+    const interval = setInterval(() => {
+      setRecordingTime((prevRecordingTime) => {
+        if (prevRecordingTime >= MAX_RECORDING_TIME - 1) {
+          clearInterval(interval);
+          handleRecording();
+          return MAX_RECORDING_TIME;
+        }
+        return prevRecordingTime + 1;
+      });
+    }, 1000);
```
kmagiera, line 168: why there's an interval here, can't we just use setTimeout here?

kmagiera (changes requested):

packages/vscode-extension/src/devices/DeviceBase.ts
```diff
@@ -52,17 +52,33 @@ export abstract class DeviceBase implements Disposable {
...
+    return this.preview.startRecording();
+  }
+
+  public async captureAndStopRecording() {
+    if (!this.preview) {
+      throw new Error("Preview not started");
+    }
+    const recordingDataPromise = this.preview.captureRecording();
+    this.preview?.stopRecording();
+    return recordingDataPromise;
```
kmagiera, line 68: for symmetry, it'd be better if this method just proxied the call down to preview instead of making two calls

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -60,33 +92,43 @@ export class Preview implements Disposable {
...
+            ? videoReadyMatch[1]
+            : videoErrorMatch
+            ? videoErrorMatch[1]
+            : "";
+
+          if (!this.videoRecordingPromises.has(videoId)) {
+            throw new Error(`Invalid video ID: ${videoId}`);
+          }
 
+          const handlers = this.videoRecordingPromises.get(videoId);
```
kmagiera, line 112:
```suggestion
          const handlers = this.videoRecordingPromises.get(videoId);
          if (!handlers) {
            throw new Error(`Invalid video ID: ${videoId}`);
          }

          
```

kmagiera: After this we don't need to check if handlers is set as we do in the lines below

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -126,6 +130,43 @@ function PreviewView() {
...
+
+    setTimeout(() => {
+      setRecordingTime((prevRecordingTime) => {
+        if (prevRecordingTime >= MAX_RECORDING_TIME_SEC - 1) {
+          handleRecording();
+          return MAX_RECORDING_TIME_SEC;
+        }
+        return prevRecordingTime + 1;
+      });
+    }, 1000);
```
kmagiera, line 167: I don't understand this bit. Maybe it should use interval actually. Now it doesn't do effect cleanup which may result in this timer running indefinitely until the timeout is hit regardless of whether the video is stopped or not.

kmagiera (changes requested):

packages/vscode-extension/src/webview/views/PreviewView.tsx
p-malecki: I've tested it and it behaves weird. After stopping recording before the time limit, setTimeout runs once more with prevRecordingTime set to 0 by handleRecording.
Adding a cleanup function would be a better solution.

kmagiera (approved):

PR #734, opened 2024-11-15
[Android] Add native logs 

maciekstosio:
The motivation of this PR is to pass native logs to Radon IDE. Currently, if the app crashes on the native side developer needs to open Android Studio anyway. 

**How it's done?**
- we create a new output channel called `Radon IDE (Android Native Logs)`
- firstly we run `adb logcat` <s>(without any params)</s> in the subprocess to get the app PID (we don't pass the logs to IDE at this point)
- once we get the PID, we run `adb logcat` for the given PID <s>with the history (`-T` param), to get all logs since the process started, and pass them to IDE</s>

_Changed during review:_
- the initial logcat process is filtered by the same regex to limit the processed output 
- helper that's getting PID has setTimeout to avoid hanging on the promise if something goes wrong 
- both processes are added to cancelToken to make sure everything is cleaned properly 
- we don't use `-T` param as it apperas it's not needed 
- add `-T` param to first adb logcat (searching for pid) so we limit the logs to current run

**Why like this?**
We don't want to pass all `adb logcat` logs, as those are enormous. I didn't find a way to _filter_ logs by packageId (except filtering them with regex), so the most reliable way seems to be to filter them by PID.  
Now, suppose the app crashes at the very beginning (for example, during `MainApplication.onCreate()`). In that case, we have no way of getting PID before (because the process is not created yet) and after (the process is already killed). To solve this case, for a short period of time, we go through all logs with regex to look for app start

```2024-11-15 14:51:24.154   574-606   ActivityManager         system_server                        I  Start proc 4474:com.reactnative76/u0a198 for next-top-activity {com.reactnative76/com.reactnative76.MainActivity}```

 and extract pid. Then we kill this process and spawn one that is properly filtered (with history, so the logs that came up during app start are there too).

**Remarks:**
- If we restart the app multiple times, the previous logs are still preserved despite `channel.clear()` this seems to be VSCode issue: https://github.com/microsoft/vscode/issues/204946

### How Has This Been Tested: 
**Normal**

- run react-native-76 and look for `Radon IDE (Android Native Logs)` output to see the logs
- restart the process few times to see if new logs are comming 

**Crash on startup**

- add `throw IllegalStateException("This line should not be reached")` to MainApplication.onCreate()
- run react-native-76
- make sure the logs from crash are visible in `Radon IDE (Android Native Logs)` output 

### Demo:

<video src="https://github.com/user-attachments/assets/951da67a-fff2-4753-9785-e7a546f230c4" />

### CODE REVIEW

File: packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts [hunk 1]
```diff
@@ -49,7 +51,9 @@ interface EmulatorProcessInfo {
 export class AndroidEmulatorDevice extends DeviceBase {
   private emulatorProcess: ChildProcess | undefined;
   private serial: string | undefined;
-
```
maciekstosio, line 52: I assume we could move it to the abstract class once iOS part is ready.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -49,7 +51,9 @@ interface EmulatorProcessInfo {
...
+  private nativeLogsOutputChannel: OutputChannel | undefined;
+  private nativeLogsProcess: ChildProcess | undefined;
```
maciekstosio, line 55: I assume we could move it to the abstract class once iOS part is ready.

packages/vscode-extension/package.json
```diff
@@ -516,5 +517,8 @@
     "vscode-test": "^1.5.0",
     "ws": "^8.14.2",
     "xml2js": "^0.6.2"
+  },
+  "dependencies": {
+    "strftime": "^0.10.3"
```
kmagiera, line 522: should go to devDependencies I guess?

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -371,6 +376,48 @@ export class AndroidEmulatorDevice extends DeviceBase {
     ]);
   }
 
+  async mirrorNativeLogsDispose() {
+    this.nativeLogsOutputChannel?.dispose();
+    this.nativeLogsProcess?.kill();
+  }
```
kmagiera, line 382: maybe not worth extracting to separate method

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -371,6 +376,48 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+          }
+        });
+      });
+
+    const nativeLogsOutputChannel = window.createOutputChannel("Radon IDE (Android Native Logs)", {
+      log: true,
+    });
+
+    const pid = await extractPidFromLogcat();
+    const process = exec(ADB_PATH, ["logcat", "--pid", pid, "-T", startTime]);
```
kmagiera, line 413: what's the point of using `-T` parameter here? I understand that the time we set is from before the process is launched. Given PID will be unique I'd expect adb to only provide logs that are relevant to us this way and cropping it with -T shouldn't change anything

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -371,6 +376,48 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+            if (pid) {
+              resolve(pid);
+            } else {
+              reject(new Error("PID not found"));
+            }
+          }
+        });
+      });
+
+    const nativeLogsOutputChannel = window.createOutputChannel("Radon IDE (Android Native Logs)", {
```
kmagiera, line 408: we should just assign it to `this.nativeLogsOutputChannel` immediately. In case anything later throws, we want to dispose things properly. We can also make it a bit "safer" and only create it when the variable is not already assigned – this way in case this method is called twice, it will still work properly (otherwise to make things safe we should "dispose" previously set variable and create new one, but this doesn't make sense here because those would be the same instaces)

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -371,6 +376,48 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+    this.nativeLogsOutputChannel?.dispose();
+    this.nativeLogsProcess?.kill();
+  }
+
+  async mirrorNativeLogs(build: AndroidBuildResult) {
+    const startTime = strftime("%F %T.000", new Date());
+
+    const extractPidFromLogcat = async () =>
+      new Promise<string>((resolve, reject) => {
+        const process = exec(ADB_PATH, ["logcat"]);
```
kmagiera, line 389: this process can run indefinitely in case there's some issue launching the app.

We should likely adapt schema that we use in couple of different places in the codebase and use `cancelToken` here instead of retaining only the the final logcat process.

What we should do is to have cancelToken that we'd use in dispose. Cancelling it would kill either the first or the second logcat process.

Also, would it make sense to filter the logs by passing some options like `-e` and also limiting the output with `-t 0` such that we have fewer lines to process? logcat can sometimes generate a large volumes of output and as long as we limit it when filtering by PID in the next command, we're not doing it here.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -389,6 +436,7 @@ export class AndroidEmulatorDevice extends DeviceBase {
       await this.configureMetroPort(build.packageName, metroPort);
       await this.launchWithBuild(build);
     }
+    this.mirrorNativeLogs(build);
```
kmagiera, line 439: given launch adb commands are async, is there a chance that the process would launch before we start listening to "Start proc" log? Maybe we should call this after `force-stop` but before the launch command is triggered?

kmagiera (changes requested):

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
maciekstosio: Changed in: https://github.com/software-mansion/radon-ide/pull/734/commits/0fe9fda34e73e2eedf7acbc3f3aa70434ca395b4

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
maciekstosio: Changed in: https://github.com/software-mansion/radon-ide/pull/734/commits/bfa2d8cf60e5d5278b4150daee0dd02cb93e3778

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
maciekstosio: I didn't use `-t` param, as it looks like it is not counting by the filtered messages. Instead, I added setTimeout for this method to reject if it won't find the pid in 2s. Changed in: https://github.com/software-mansion/radon-ide/pull/734/commits/bfa2d8cf60e5d5278b4150daee0dd02cb93e3778

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -371,6 +377,51 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+
+            if (pid) {
+              resolve(pid);
+            } else {
+              reject(new Error("PID not found"));
+            }
+          }
+        });
+
+        // We should be able to get pid immediately, if we're not getting it in 2s, then we reject to not block the app.
```
kmagiera, line 403: but we're not blocking the app here? How are we blocking it? We never await the mirrorLogs call

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -49,6 +51,8 @@ interface EmulatorProcessInfo {
 export class AndroidEmulatorDevice extends DeviceBase {
   private emulatorProcess: ChildProcess | undefined;
   private serial: string | undefined;
+  private nativeLogsOutputChannel: OutputChannel | undefined;
+  private nativeLogsCancelToken = new CancelToken();
```
kmagiera, line 55: I think one cancel token shouldn't be cancelled more than once. We should create new token in `mirrorLogs` method

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -371,6 +377,51 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+              reject(new Error("PID not found"));
+            }
+          }
+        });
+
+        // We should be able to get pid immediately, if we're not getting it in 2s, then we reject to not block the app.
+        setTimeout(() => {
+          process.kill();
+          reject(new Error("Timeout while waiting for app to start to get the process PID."));
+        }, 2000);
```
kmagiera, line 407: this could be too agressive timeout. This timeout doesn't cause any delays, it is just a safeguard to avoid the logcat process to run very long if things go wrong. Would be safer if we set it to something higher like 10seconds for example.

kmagiera (changes requested):

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -371,6 +378,59 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+    if (this.nativeLogsCancelToken) {
+      this.nativeLogsCancelToken.cancel();
+    }
+
+    this.nativeLogsCancelToken = new CancelToken();
+
+    const extractPidFromLogcat = async (cancelToken: CancelToken) =>
+      new Promise<string>((resolve, reject) => {
+        const regexString = `Start proc ([0-9]{4}):${build.packageName}`;
+        const process = exec(ADB_PATH, ["logcat", "-e", regexString, "-T", startTime]);
```
maciekstosio, line 393: We need the `-T` param, otherwise, it will catch the same pid (when we restart the app) - the first one

kmagiera: `-T` also takes a number, so maybe we could do `-T 1` to not read old logs? (`-T 0` is not legal)

kmagiera (approved):
Looks good. Can you verify we can just use `-T 1` option when scanning for PID? This way we wouldn't need to bring strftime as dependency

PR #735, opened 2024-11-17
Fix package manager detection

filip131311:
This PR changes how we detect package manager used by the application, to better deal with a situation when a user has multiple lock files in their repository. A situation like that is quite common, because  users sometimes accidentally run `npm install` before checking that the project actually uses yarn. 

Solution: 
- information set up in package.json takes priority, as it is unprovable that set it up accidentally
- in case the information is missing from package.json we still perform the old check but, we send a warning to the console if multiple lock files are detected
- finally we allow user to set up the package manager in launch configuration by hand and than it takes priority 

### How Has This Been Tested: 

- run https://github.com/Almouro/react-native-performance-workshop in IDE which has multiple lock files and see it working 
- remove the information from package json and the if logger logs the warning
- run expo 52 test app example and see it working correctlly

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -309,8 +309,12 @@
                   }
                 }
               },
+              "packageManager": {
+                "type": "string",
+                "description": "Package manager used for the project (eg. yarn)."
+              },
               "preview": {
-                "description": "Custommize the behavior of device preview",
+                "description": "Customize the behavior of device preview",
                 "type": "object",
                 "properties": {
                   "waitForAppLaunch": {
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -311,7 +311,7 @@
               },
               "packageManager": {
                 "type": "string",
-                "description": "Package manager used for the project (eg. yarn)."
+                "description": "Package manager used for the project. Available options include: npm, yarn, pnpm, bun."
               },
               "preview": {
                 "description": "Customize the behavior of device preview",
@@ -400,7 +400,7 @@
               },
               "packageManager": {
                 "type": "string",
-                "description": "Package manager used for the project (eg. yarn)."
+                "description": "Package manager used for the project. Available options include: npm, yarn, pnpm, bun."
               },
               "preview": {
                 "description": "Customize the behavior of device preview",
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -398,6 +398,10 @@
                   }
                 }
               },
+              "packageManager": {
+                "type": "string",
+                "description": "Package manager used for the project (eg. yarn)."
+              },
               "preview": {
                 "description": "Customize the behavior of device preview",
                 "type": "object",
--- a/packages/vscode-extension/src/common/LaunchConfig.ts
+++ b/packages/vscode-extension/src/common/LaunchConfig.ts
@@ -25,6 +25,7 @@ export type LaunchConfigurationOptions = {
     buildType?: string;
     productFlavor?: string;
   };
+  packageManager?: string;
   preview?: {
     waitForAppLaunch?: boolean;
   };
--- a/packages/vscode-extension/src/utilities/packageManager.ts
+++ b/packages/vscode-extension/src/utilities/packageManager.ts
@@ -4,7 +4,6 @@ import { command } from "./subprocess";
 import { getAppRootFolder } from "./extensionContext";
 import { isWorkspaceRoot } from "./common";
 import { Logger } from "../Logger";
-import { window } from "vscode";
 
 export type PackageManagerInfo = {
   name: "npm" | "pnpm" | "yarn" | "bun";
@@ -63,10 +62,6 @@ export async function resolvePackageManager(): Promise<PackageManagerInfo | unde
     if (packageManagerCandidates.length > 1) {
       Logger.warn(
         "Your workspace contains multiple package manager lock files, it might cause wrong manager to be used"
-      );
-      window.showWarningMessage(
-        "Your workspace contains multiple package manager lock files, it might cause wrong manager to be used.",
-        "dismiss"
       );
     }
 
--- a/packages/vscode-extension/src/utilities/packageManager.ts
+++ b/packages/vscode-extension/src/utilities/packageManager.ts
@@ -28,6 +28,19 @@ export async function resolvePackageManager(): Promise<PackageManagerInfo | unde
   const workspacePath = findWorkspace(appRootPath);
 
   async function findPackageManager(workspace: string) {
+    try {
+      const packageManager = require(path.join(workspace, "package.json")).packageManager;
+
+      if (packageManager) {
+        // e.g. yarn@3.6.4
+        const match = packageManager.match(/^([a-zA-Z]+)@/);
+        return match ? match[1] : "npm";
+      }
+    } catch (e) {
+      // there might be a problem while reading package.json in which case move to looking 
+      // for lock files matching package managers in the workspace root
+    }
+
     const lockFiles = new Map(
       Object.entries({
         "yarn.lock": "yarn",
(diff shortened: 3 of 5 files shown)


maciekstosio:
image

The `resolvePackageManager` is called multiple times. So this error also might come up multiple times. I wonder if the message shouldn't be sent only once. In this case, it's almost at the same time, so it's not visible. But I imagine we inspect the repo at the beginning and set those values somewhere in the context. If something is wrong, we inform about it - once.

filip131311:
@maciekstosio  I don't know why we run this check multiple times at the start, but in principle i want to run it multiple times if the check is called again, because the conditions might have changed, in fact because we give the user an option to run diagnostics manually, it seems a good idea it consistently remind them about this issue

packages/vscode-extension/package.json
```diff
@@ -309,8 +309,12 @@
                   }
                 }
               },
+              "packageManager": {
+                "type": "string",
```
maciekstosio, line 313: Couldn't that be an enum?

packages/vscode-extension/package.json
maciekstosio, line 1: Don't we need to update it in react-native-ide config as well?

packages/vscode-extension/package.json
maciekstosio, line 313: Couldn't that be an enum?

packages/vscode-extension/package.json
maciekstosio, line 1: Don't we need to update it in react-native-ide config as well?

packages/vscode-extension/src/common/LaunchConfig.ts
```diff
@@ -25,6 +25,7 @@ export type LaunchConfigurationOptions = {
     buildType?: string;
     productFlavor?: string;
   };
+  packageManager?: string;
```
maciekstosio, line 28: Enum too?

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -28,6 +34,25 @@ export async function resolvePackageManager(): Promise<PackageManagerInfo | unde
   const workspacePath = findWorkspace(appRootPath);
 
   async function findPackageManager(workspace: string) {
+    const { packageManager } = getLaunchConfiguration();
+
+    if (packageManager && isPackageManager(packageManager)) {
```
maciekstosio, line 39: Log some error or warning?

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -28,6 +34,25 @@ export async function resolvePackageManager(): Promise<PackageManagerInfo | unde
...
+        return match ? match[1] : "npm";
+      }
+    } catch (e) {
+      // there might be a problem while reading package.json in which case move to looking
+      // for lock files matching package managers in the workspace root
+    }
+
     const lockFiles = new Map(
       Object.entries({
         "yarn.lock": "yarn",
```
maciekstosio, line 58: I see that listing all of them pops up in different places. Maybe let's do some one source of truth.

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -37,24 +62,38 @@ export async function resolvePackageManager(): Promise<PackageManagerInfo | unde
...
+        "Your workspace contains multiple package manager lock files, it might cause wrong manager to be used"
+      );
     }
+
+    if (packageManagerCandidates) {
+      return packageManagerCandidates[0];
+    }
+
+    // when no package manager were detected we default to npm
     return "npm";
```
maciekstosio, line 97: That also could be a const in one shared places, for example if we decide to use something else as defualt in the future.

maciekstosio (dismissed):

packages/vscode-extension/package.json
filip131311: what do you mean?

packages/vscode-extension/package.json
filip131311: it is not supported https://code.visualstudio.com/api/references/contribution-points

packages/vscode-extension/src/common/LaunchConfig.ts
filip131311: we are not sure about what the user has put there that check comes latter

packages/vscode-extension/src/utilities/packageManager.ts
filip131311: this is a different kind of mapping then in other places that is specific for that application

packages/vscode-extension/src/utilities/packageManager.ts
filip131311: good point

packages/vscode-extension/package.json
```diff
@@ -309,8 +309,12 @@
...
+                "description": "Package manager used for the project (eg. yarn)."
```
kmagiera, line 314: if this can't be enum, can we list all allowed values in the description, there are not that many

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -37,25 +70,39 @@ export async function resolvePackageManager(): Promise<PackageManagerInfo | unde
...
-    } catch (e) {
-      // there might be a problem while reading package.json in which case we default to npm
+    if (packageManagerCandidates.length > 1) {
+      Logger.warn(
+        "Your workspace contains multiple package manager lock files, it might cause wrong manager to be used"
+      );
     }
-    return "npm";
+
+    if (packageManagerCandidates) {
```
kmagiera, line 100: this is always true even if `packageManager` is empty

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -37,25 +70,39 @@ export async function resolvePackageManager(): Promise<PackageManagerInfo | unde
       } as const)
     );
 
-    const files = await fs.readdir(workspace);
+    const getDirFilesSortedByModificationDate = async (dir: string) => {
```
kmagiera, line 73: can we have it defined somewhere outside of the scope? this seems like a utility function and doesn't require to be in this function scope

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -37,25 +70,39 @@ export async function resolvePackageManager(): Promise<PackageManagerInfo | unde
...
+        .map((file) => file.name);
+    };
+
+    const files = await getDirFilesSortedByModificationDate(workspace);
+    const packageManagerCandidates = [];
     for (const file of files) {
-      const packageManager = lockFiles.get(file);
-      if (packageManager) {
-        return packageManager;
+      const manager = lockFiles.get(file);
```
kmagiera, line 88: do I see it wrong of lockFiles is a map with single entry "yarn.lock" ? does it even make sense to use it this way in this case?

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -28,6 +36,31 @@ export async function resolvePackageManager(): Promise<PackageManagerInfo | unde
...
+        Logger.warn(
+          `Package manager provided in launch configuration: ${packageManager} is not supported by radon IDE`
+        );
+        return;
+      }
+      return packageManager;
+    }
+
+    try {
+      const manager = require(path.join(workspace, "package.json")).packageManager;
```
kmagiera, line 52: we should use require with no-cache option or just read the file. Using require will cache the package for as long as the extension process exists

kmagiera (changes requested):

packages/vscode-extension/src/utilities/packageManager.ts
filip131311: github cropped the rest: 
```
const lockFiles = new Map(
      Object.entries({
        "yarn.lock": "yarn",
        "package-lock.json": "npm",
        "pnpm-lock.yaml": "pnpm",
        "bun.lockb": "bun",
      } as const)
    );
```

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -1,15 +1,36 @@
...
+
+async function getDirFilesSortedByModificationDate(dir: string) {
+  const files = await fs.promises.readdir(dir);
+
+  return files
+    .map((fileName) => ({
+      name: fileName,
+      time: fs.statSync(`${dir}/${fileName}`).mtime.getTime(),
+    }))
+    .sort((a, b) => a.time - b.time)
```
kmagiera, line 28: this sorts in ascending order. I think we want to select one that's latest modification date not the one with the oldest

packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -1,15 +1,36 @@
...
   name: "npm" | "pnpm" | "yarn" | "bun";
   workspacePath?: string;
 };
 
+function isPackageManager(candidate: string): boolean {
+  const packageManagers = ["npm", "yarn", "pnpm", "bun"];
+  return packageManagers.includes(candidate);
+}
+
+async function getDirFilesSortedByModificationDate(dir: string) {
```
kmagiera, line 20:
```suggestion
async function listFilesSortedByModificationDate(dir: string) {
```

kmagiera (approved):
Looks good, just left a few final comments incline

PR #742, opened 2024-11-18
[iOS] Add native logs 

maciekstosio:
The motivation of this PR is to pass native logs to Radon IDE. Currently, if the app crashes on the native side developer needs to open Xcode anyway.

⚠️ DISCLAIMER: The current approach won't work with Expo Go and Expo Dev Client, as those are run through deep link not `xcrun simctl launch`. 

I tried: `xcrun simctl --set {deviceLocation} spawn booted log stream`, `react-native log-ios`, looking through files in the container if there is something that looks like logs - none of those work. 

<s>What I did:
- add params to `xcrun simctl launch`  that print `stderr` and `stout` to file,
- watch for changes in this log file,
- send new chunks into VSCode output.</s>

**Changed during review:**

What I did:
- add `--console` param to `xcrun simctl launch`, which logs the process output 
- redirect this output to VSCode. 

### How Has This Been Tested: 

**Normal:**
0. Start by adding to AppDelegate.mm, `didFinishLaunchingWithOptions`:
```
  NSLog(@"AppDelegate didFinishLaunchingWithOptions");
```
1. Start the app
2. Go to `Radon IDE (iOS Simulator Logs)` output
4. Verify the logs are ther
5. Restart app process, verify that new logs are comming 

**Crash:**
0. Start by adding to AppDelegate.mm, `didFinishLaunchingWithOptions`:
```

  NSLog(@"AppDelegate didFinishLaunchingWithOptions");

  double delayInSeconds = 5.0;
  dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delayInSeconds * NSEC_PER_SEC)); // 1 
  dispatch_after(popTime, dispatch_get_main_queue(), ^(void){ // 2 
    @throw [NSException exceptionWithName:@"Something is not right exception"
                                reason:@"Can't perform this operation because of this or that"
                              userInfo:nil];
  });

```
1. Start the app, it should crash after 5s
2. Go to `Radon IDE (iOS Simulator Logs)` output
3. Look for the error in the output 

### Demo

<video src="https://github.com/user-attachments/assets/ed065f66-b710-4227-8ec6-7ebb72b3964f" />

### DIFF
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -1,7 +1,7 @@
 import path from "path";
 import fs from "fs";
 import { OutputChannel, window } from "vscode";
-import { ExecaError } from "execa";
+import { ExecaChildProcess, ExecaError } from "execa";
 import { getAppCachesDir, getOldAppCachesDir } from "../utilities/common";
 import { DeviceBase } from "./DeviceBase";
 import { Preview } from "./preview";
@@ -13,7 +13,6 @@ import { BuildResult } from "../builders/BuildManager";
 import { AppPermissionType, DeviceSettings, Locale } from "../common/Project";
 import { EXPO_GO_BUNDLE_ID, fetchExpoLaunchDeeplink } from "../builders/expoGo";
 import { IOSBuildResult } from "../builders/buildIOS";
-import { CancelToken } from "../builders/cancelToken";
 
 const LEFT_META_HID_CODE = 0xe3;
 const RIGHT_META_HID_CODE = 0xe7;
@@ -55,7 +54,7 @@ type PrivacyServiceName =
 
 export class IosSimulatorDevice extends DeviceBase {
   private nativeLogsOutputChannel: OutputChannel | undefined;
-  private lunchedAppCancelToken = new CancelToken();
+  private runningAppProcess: ExecaChildProcess | undefined;
 
   constructor(private readonly deviceUDID: string, private readonly _deviceInfo: DeviceInfo) {
     super();
@@ -78,7 +77,7 @@ export class IosSimulatorDevice extends DeviceBase {
   public dispose() {
     super.dispose();
     this.nativeLogsOutputChannel?.dispose();
-    this.lunchedAppCancelToken.cancel();
+    this.runningAppProcess?.cancel();
     return exec("xcrun", [
       "simctl",
       "--set",
@@ -358,17 +357,20 @@ export class IosSimulatorDevice extends DeviceBase {
 
     await this.terminateAnyRunningApplications();
 
-    if (this.nativeLogsOutputChannel) {
-      this.nativeLogsOutputChannel.dispose();
+    if (this.runningAppProcess) {
+      this.runningAppProcess.kill(9);
     }
 
-    this.nativeLogsOutputChannel = window.createOutputChannel(
-      "Radon IDE (iOS Simulator Logs)",
-      "log"
-    );
+    if (!this.nativeLogsOutputChannel) {
+      this.nativeLogsOutputChannel = window.createOutputChannel(
+        "Radon IDE (iOS Simulator Logs)",
+        "log"
+      );
+    }
+
     this.nativeLogsOutputChannel.clear();
 
-    const process = exec("xcrun", [
+    this.runningAppProcess = exec("xcrun", [
       "simctl",
       "--set",
       deviceSetLocation,
@@ -379,9 +381,7 @@ export class IosSimulatorDevice extends DeviceBase {
       build.bundleID,
     ]);
 
-    this.lunchedAppCancelToken.adapt(process);
-
-    lineReader(process).onLineRead(this.nativeLogsOutputChannel.appendLine);
+    lineReader(this.runningAppProcess).onLineRead(this.nativeLogsOutputChannel.appendLine);
   }
 
   async launchWithExpoDeeplink(bundleID: string, expoDeeplink: string) {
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/utilities/common.ts
maciekstosio, line 1: I have a feeling that it might be over-engineered. I'm open to other approaches if you have a better solution.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -346,16 +350,36 @@ export class IosSimulatorDevice extends DeviceBase {
     await Promise.all(matches.map(terminateApp));
   }
 
+  private getNativeLogsPath() {
+    const deviceSetLocation = getOrCreateDeviceSet(this.deviceUDID);
+    return path.join(deviceSetLocation, this.deviceUDID, "data", "radon_ide_ios_process.log");
+  }
+
+  mirrorNativeLogs(logFile: string) {
+    this.nativeLogsOutputChannel = window.createOutputChannel("Radon IDE (iOS Native Logs)", "log");
```
kmagiera, line 359: I think I'd like "Simulator Logs" better than "Native Logs" – we have too many "native" terms in RN already

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -346,16 +350,36 @@ export class IosSimulatorDevice extends DeviceBase {
...
     await this.terminateAnyRunningApplications();
+    this.mirrorNativeLogs(logFile);
 
     await exec("xcrun", [
       "simctl",
       "--set",
       deviceSetLocation,
       "launch",
+      `--stdout=${logFile}`,
+      `--stderr=${logFile}`,
```
kmagiera, line 382: wouldn't it make more sense to use `--console` option which prints stuff to processes stdout?

kmagiera (changes requested):
Have you considererd using `--console` option when launching? This way we wouldn't need temporary log file and the watcher

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
maciekstosio: Changed in: https://github.com/software-mansion/radon-ide/pull/742/commits/1d7e107217d8e645ab8ce51d70ed2ac6876713dd

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
maciekstosio: Changed in: https://github.com/software-mansion/radon-ide/pull/742/commits/1d7e107217d8e645ab8ce51d70ed2ac6876713dd

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -351,15 +355,28 @@ export class IosSimulatorDevice extends DeviceBase {
 
     await this.terminateAnyRunningApplications();
 
-    await exec("xcrun", [
+    if (this.nativeLogsOutputChannel) {
+      this.nativeLogsOutputChannel.dispose();
+    }
```
kmagiera, line 360: I think it is fine to just not create the channel if one already exists. We can still clear it regardless

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -351,15 +355,28 @@ export class IosSimulatorDevice extends DeviceBase {
...
+      "log"
+    );
+    this.nativeLogsOutputChannel.clear();
+
+    const process = exec("xcrun", [
       "simctl",
       "--set",
       deviceSetLocation,
       "launch",
+      "--console",
```
kmagiera, line 373: AFAIK using `--console` option would make the launch command hang there for as long as the app process is running. I think we should persist `process` in that case and terminate it in `dispose`. It will likely be terminated by the `shutdown` command we execute there but feel like it'd be safer to do it this way in case we ever change the shutdown procedure.

kmagiera (changes requested):

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -72,6 +77,8 @@ export class IosSimulatorDevice extends DeviceBase {
 
   public dispose() {
     super.dispose();
+    this.nativeLogsOutputChannel?.dispose();
+    this.lunchedAppCancelToken.cancel();
```
kmagiera, line 81: cancel token can only be cancelled once. We should create new one for each process launch

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -351,15 +357,31 @@ export class IosSimulatorDevice extends DeviceBase {
 
     await this.terminateAnyRunningApplications();
 
-    await exec("xcrun", [
+    if (this.runningAppProcess) {
+      this.runningAppProcess.kill(9);
```
maciekstosio, line 361: Just ensuring the process is killed - the weird logs won't come up after the new lunch starts. Thus, I can reuse the channel - recreating it requires the user to select output each time, which is lame.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
maciekstosio: Changed in: https://github.com/software-mansion/radon-ide/pull/742/commits/8b9e8dfc74c040bc8c9405b0330e87b0668af9b2

kmagiera (approved):

PR #743, opened 2024-11-19
Shorten duration of replay rewind animation

p-malecki:
This PR shortens duration of rewind animation to improve the usability of the rewind feature.

### Testing:
Ensure that the duration of the replay takes 0.6sec.

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/68c8354f-49b7-4d1e-af49-11230540591e"/>|<video src="https://github.com/user-attachments/assets/c03b42f3-ec96-4174-8418-bf596b62a2a2" />|

### DIFF
--- a/packages/vscode-extension/src/webview/components/ReplayOverlay.tsx
+++ b/packages/vscode-extension/src/webview/components/ReplayOverlay.tsx
@@ -18,7 +18,7 @@ function acceleratedRewind(
   setTimeCallback: (time: number) => void,
   readyCallback: () => void
 ) {
-  const rewindTimeSec = 1.6;
+  const rewindTimeSec = 0.6;
 
   const v0 = 0.1;
   const vFinal = 2 / rewindTimeSec - v0;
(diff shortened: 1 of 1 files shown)


kmagiera (changes requested):
Don't think there's a lot of value in adding extra persistent state just to show longer animation. We should just decide on the length (1.5sec should be ok I believe) and use it at all times (or remove this animation entirely)

p-malecki:
> Don't think there's a lot of value in adding extra persistent state just to show longer animation. We should just decide on the length (1.5sec should be ok I believe) and use it at all times (or remove this animation entirely)

The previous animation duration was 1.6 sec, so reducing it to 1.5 seconds wouldn't make a significant difference. We agreed experimentally that 0.6 seconds is the best duration for usability, but that may be too fast for the first impression of rewind animation.
If we decide to drop the idea of a longer first animation, I think we should settling on 0.6 sec.

kmagiera (approved):
Ok, I'm fine with shorter animation, just remember you said it was 5 seconds earlier not 1.5 seconds.

Please update the PR description before merging

PR #746, opened 2024-11-19
Enhance zoom controls for better usability

### LINKED ISSUE #730: Tweak zoom widget (on the left) animation timing and trigger condition
**Description**

Screenshot 2024-11-14 at 17 14 58

Currently zoom widget is showing up even if you really fast moved cursor while moving cursor from radon panel to editor panel  - so when you don't have intention to touch this widget but happened that cursor path trajectory crossed this widget. This is a little distracting. I think it would be great if:

1) onhover is only activated if cursor has been in this widget area for small fraction of second like 100ms or 50ms.
2) I would make the show animation duration smaller to make it a little faster
3) I would make hide animation also faster and automatically triggered much faster (currently you have to wait around 3 second to hide this widget), 0.5-1 seconds would be less distracting. 

**Environment**

- Cursor IDE: Version: 0.42.4, VSCode Version: 1.93.1, Radon: 0.0.22

p-malecki:
In this PR, we enhance the user experience of the zoom controls:
- added a 100ms hover delay before displaying the zoom controls to prevent them from appearing too unintentionally
- accelerated the `translateX` transition, reducing its duration from 0.5sec to 0.2sec
- shortened the delay for hiding the zoom controls from 2sec to 0.4sec
- increased the area that detects mouse events, which improves accessibility
- fixed the visibility of the "x" multiplier in the zoom component by reducing the text width to conceal the "x" letter when not hovered over
- replaced `onMouseOut` with `onMouseLeave` to prevent triggering events when moving the mouse out of children elements of the `button-group-left-wrapper` (similar adjustment with `onMouseEnter`)

Fixes #730 

### Test plan: 
- tested regular and rapid interaction with the zoom controls to ensure smooth functionality
- ensured that the zoom controls no longer appears when the cursor quickly crosses its area unintentionally

<table>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td><video src="https://github.com/user-attachments/assets/c073f363-f121-4c81-9c1b-84a3d0194383" /></td>
        <td><video src="https://github.com/user-attachments/assets/9822b390-7056-4853-b272-d03476d127a5" /></td>
    </tr>
    <tr>
        <td colspan="2">Area that detects mouse events:</td>
    </tr>
    <tr>
        <th>Before</th>
        <th>After</th>
    </tr>
    <tr>
        <td><video src="https://github.com/user-attachments/assets/3192e543-c816-4775-a812-64d13b4a65c4" /></td>
        <td><video src="https://github.com/user-attachments/assets/c14a38cf-d195-4c14-95b3-dfd402d1437a" /></td>
</tr>
</table>


packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -44,7 +44,8 @@ declare module "react" {
   }
 }
 
-const HIDE_ZOOM_CONTROLS_DELAY = 2000;
+const SHOW_ZOOM_CONTROLS_DELAY_MS = 100;
+const HIDE_ZOOM_CONTROLS_DELAY_MS = 400;
```
maciekstosio, line 48: IMO it could be 100 for both of them. This delay when you move out seems weird.

packages/vscode-extension/src/webview/components/Preview.tsx
maciekstosio, line 1: I would extract this ButtonGroupLeft component, I don't see a reason why it's in Preview.

packages/vscode-extension/src/webview/components/Preview.tsx
maciekstosio, line 48: IMO it could be 100 for both of them. This delay when you move out seems weird.

maciekstosio (changes requested):

maciekstosio:
<video src="https://github.com/user-attachments/assets/330c8f8b-ccc5-4d61-bf28-1abdb5571d74" />

When you select the zoom outside the observed area it stays open. I don't think it's intended.

p-malecki:
> When you select the zoom outside the observed area it stays open. I don't think it's intended.
> 
>  Screen.Recording.2024-11-20.at.08.13.04.mov

I'm aware of this issue, it was present before this PR. I've created issue #747 and will address it in a future.

packages/vscode-extension/src/webview/components/Preview.tsx
p-malecki: Fair point but I think HIDE_ZOOM_CONTROLS_DELAY_MS should be a bit longer then SHOW_ZOOM_CONTROLS_DELAY_MS just to give more time for recognising the current zoom level state. SHOW_ZOOM_CONTROLS_DELAY_MS with 200ms seems better.

maciekstosio (approved):
Just please remember to extract this component in next PR.

PR #748, opened 2024-11-20
Record user and session data properly for telemetry

kmagiera:
This PR fixes the issue with telemetry module that wasn't collecting user-id and session-id properly resulting in all the telemetry events being squashed under a single user and session.

The main fix for user-id reporting has already landed in the new version of telemetry package https://github.com/microsoft/vscode-extension-telemetry/pull/210
We also submitted a PR that has been merged to address the session ID issue: https://github.com/microsoft/vscode-extension-telemetry/pull/215

Because there's been no release since that update, we're installing a fork of telemetry module with our changes included from https://github.com/software-mansion-labs/vscode-extension-telemetry
Since the package requires a build-step, we've uploaded bundled JS files into the fork repository and install the dependency using github link.

The new version of telementy module now requires a so-called "connection string" instead of just the instrumentation key as the first argument. Because of that we're also updating the public key value.

### How Has This Been Tested: 
1. Change panelOpen event to panelOpen-test20
2. Open extension
3. Check in Application Insights that the new event arrived and it has distinct user id and session set.


filip131311 (approved):
I ran the extension and changed the event to `panelOpened Filip131311`, if it appeared in application insights, you can merge it.

PR #750, opened 2024-11-20
Fix Expo Go on SDK 52 by selecting the right hermes runtime when connecting debugger

kmagiera:
This PR fixes issue with Expo Go on SDK 52 where the new debugger is enabled.

The issue there is that new debugger lists two hermes runtimes, one is used by the Expo Go shell, and the second one is used by the actual app. The fix is to detect that Expo Go is used with the new debugger, and in such case select second runtime from the list instead of the first one.

Note: While this PR fixes the debugger connection and the console logging mechanism, the breakpoints are still non-functional. This is due to the fact that expo go bundle doesn't use file path aliases ([metro-projectRoot] etc)

### How Has This Been Tested: 
1. Use this reproducer repository which is an Expo Go SDK 52 project: https://github.com/friyiajr/debugger-reproducer
2. Run app and click the only button that's there
3. See logs in the console

### DIFF
--- a/packages/vscode-extension/src/project/metro.ts
+++ b/packages/vscode-extension/src/project/metro.ts
@@ -9,6 +9,7 @@ import { extensionContext, getAppRootFolder } from "../utilities/extensionContex
 import { shouldUseExpoCLI } from "../utilities/expoCli";
 import { Devtools } from "./devtools";
 import { getLaunchConfiguration } from "../utilities/launchConfiguration";
+import { EXPO_GO_BUNDLE_ID, EXPO_GO_PACKAGE_NAME } from "../builders/expoGo";
 
 export interface MetroDelegate {
   onBundleError(): void;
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/project/metro.ts
```diff
@@ -350,11 +351,21 @@ export class Metro implements Disposable {
...
+      (page) => page.reactNative && page.title.startsWith("React Native Bridge")
+    );
+    if (newDebuggerPages.length > 0) {
+      // Expo go apps would report at least two pages, first one being the Expo Go host runtime
+      // If we detect Expo Go package (using description field), we want to pick the second page
+      // otherwise we pick the first one
+      const description = newDebuggerPages[0].description;
+      const isExpoGo = description === EXPO_GO_BUNDLE_ID || description === EXPO_GO_PACKAGE_NAME;
+      if (isExpoGo && newDebuggerPages.length > 1) {
+        return newDebuggerPages[1].webSocketDebuggerUrl;
```
filip131311, line 364: How do you know that the second runtime is the correct one? Could you provide source?

packages/vscode-extension/src/project/metro.ts
```diff
@@ -350,11 +351,21 @@ export class Metro implements Disposable {
...
-        return page.webSocketDebuggerUrl;
+    const newDebuggerPages = listJson.filter(
+      (page) => page.reactNative && page.title.startsWith("React Native Bridge")
+    );
+    if (newDebuggerPages.length > 0) {
+      // Expo go apps would report at least two pages, first one being the Expo Go host runtime
+      // If we detect Expo Go package (using description field), we want to pick the second page
+      // otherwise we pick the first one
+      const description = newDebuggerPages[0].description;
+      const isExpoGo = description === EXPO_GO_BUNDLE_ID || description === EXPO_GO_PACKAGE_NAME;
```
filip131311, line 362: Why can it be either of them?

filip131311 (changes requested):
I tested this branch on bare expo 52 project created with: `npx create-expo-app --template default@beta` and although it fixed the console logs it did not fix the break points: 

Screenshot 2024-11-21 at 00 19 56


The break point is not added to the debbuger

packages/vscode-extension/src/project/metro.ts
kmagiera: BUNDLE_IDE is iOS concept and PACKAGE_NAME is Android one. It can be either depending on platform we run on

packages/vscode-extension/src/project/metro.ts
kmagiera: There's no source and it is based on observation. What we observe is documented in the comment that expo go spins up two runtimes. The first one is host runtime and it seem logical that it will be created first

kmagiera:
Thanks, I checked the breakpoints and understand what's going on there. Would prefer to fix it in a follow up

filip131311 (approved):

PR #751, opened 2024-11-20
Fix log counter multiplying itself after reload 

filip131311:
This PR fixes log counter multiplication after, either `restartProcess`  or `reinstall` reload options. It was caused by the old DebugSession not being disposed of. 

To solve the issue we dispose the old debag session before creating a new one.

### How Has This Been Tested: 

- run any application and reload it using  `restartProcess`  or `reinstall` 
- check if the log counter works as expected

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -150,9 +150,6 @@ export class DeviceSession implements Disposable {
     ]);
     Logger.debug("App and preview ready, moving on...");
     this.eventDelegate.onStateChange(StartupMessage.AttachingDebugger);
-    if (this.debugSession) {
-      this.debugSession.dispose();
-    }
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -150,6 +150,9 @@ export class DeviceSession implements Disposable {
     ]);
     Logger.debug("App and preview ready, moving on...");
     this.eventDelegate.onStateChange(StartupMessage.AttachingDebugger);
+    if (this.debugSession) {
+      this.debugSession.dispose();
+    }
```
kmagiera, line 155: would feel more correct for this to be in `startDebugger` method where we also re-assign debugSession variable immediately after disposing the previous one

kmagiera (approved):
Looks good, please just address the inline comment

PR #753, opened 2024-11-22
Fallback to non-aliased source map paths in new debugger

kmagiera:
This PR fixes issue with debugger not mapping breakpoint locations correctly in Expo 52.

While Expo 52 uses the new debugger, it does not follow the newly added project root aliasing in all possible settings. We discussed this with Expo team and it is and although this isn't deliberate it is known to work this way and since it has no impact on the regular debugger workflow there are no immediate plans on adapting the aliasing approach.

The current flow when it comes to mapping breakpoint locations for the new debugger is that when we detect that new debugger is used, we provide aliasing config for project root and other watch folders. Later, when we receive breakpoint request, we transform absolute file name into the aliased version and use source maps to find the generated location for the breakpoint. The step that is problematic is that we always assume that the file name is aliased.

In this PR we're changing this strategy:
1) we collect all the file paths from loaded source maps into a set
2) when mapping location, we test whether the aliased version exists in the source map
3) if it does, we use aliased file path to find generated breakpoint location
4) otherwise we fallback to the absolute path and use it to locate generated breakpoint location.

### How Has This Been Tested: 
1) I used this expo go 52 repo: https://github.com/friyiajr/debugger-reproducer
2) before this change, setting breakpoints wasn't possible – we'd always get a greyed out breakpoint (non-verified)
3) after this change you can set breakpoints and have debugger stop at those
4) also tested scenarios where we append some lines before breakpoint and save and scenarios where we refresh the app
5) tested on both Android and iOS (note that we have to set breakpoint after the app is launch which is due to another issue we're fixing)

### DIFF
--- a/packages/vscode-extension/src/debugging/DebugAdapter.ts
+++ b/packages/vscode-extension/src/debugging/DebugAdapter.ts
@@ -83,6 +83,7 @@ export class DebugAdapter extends DebugSession {
   private sourceMapAliases?: Array<[string, string]>;
   private threads: Array<Thread> = [];
   private sourceMaps: Array<[string, string, SourceMapConsumer, number]> = [];
+  private sourceMapFilePaths: Set<string> = new Set();
   private expoPreludeLineCount: number;
   private linesStartAt1 = true;
   private columnsStartAt1 = true;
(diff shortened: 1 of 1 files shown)


p-malecki (approved):
LGTM

filip131311 (approved):

PR #755, opened 2024-11-22
Refactor debug adapter

filip131311:
This PR refactors debug adapter to handle two issues: 
(1) - the `debugAdapter.ts` - file was becoming hard to manage due to it's versatility of use
(2) - solve problems with multiple sources triggering breakpoint updates at the same time

to solve (1) we separate out: `CDPComunicator`, `SourceMapController` and `breakpointController`
to solve (2) we move the responsibility of updating the breakpoints from `breakpointController` to `MyBreakpoint` class itself and add incoming update requests to the queue.

### How Has This Been Tested: 

- Run Radon IDE Demo and test breakpoints and console.log
- in the same application add a breakpoint before application starts and then check if removing it in vs code actually removes it 
- run Expo 52 test app and check if break points work as expected

### DIFF
--- a/packages/vscode-extension/src/debugging/DebugAdapter.ts
+++ b/packages/vscode-extension/src/debugging/DebugAdapter.ts
@@ -1,4 +1,3 @@
-import path from "path";
 import { DebugConfiguration } from "vscode";
 import {
   DebugSession,
@@ -10,12 +9,10 @@ import {
   Thread,
   TerminatedEvent,
   ThreadEvent,
-  Breakpoint,
   Source,
   StackFrame,
 } from "@vscode/debugadapter";
 import { DebugProtocol } from "@vscode/debugprotocol";
-import { NullablePosition, SourceMapConsumer } from "source-map";
 import { formatMessage } from "./logFormatting";
 import { Logger } from "../Logger";
 import {
@@ -25,9 +22,9 @@ import {
   CDPRemoteObject,
 } from "./cdp";
 import { VariableStore } from "./variableStore";
-import { MyBreakpoint } from "./MyBreakpoint";
 import { CDPCommunicator } from "./CDPCommunicator";
 import { SourceMapController } from "./SourceMapsController";
+import { BreakpointsController } from "./BreakpointsController";
 
 function typeToCategory(type: string) {
   switch (type) {
@@ -41,8 +38,12 @@ function typeToCategory(type: string) {
 
 export class DebugAdapter extends DebugSession {
   private variableStore: VariableStore = new VariableStore();
+
   private CDPCommunicator: CDPCommunicator;
   private sourceMapController: SourceMapController;
+
+  private breakpointController: BreakpointsController;
+
   private threads: Array<Thread> = [];
   private linesStartAt1 = true;
   private columnsStartAt1 = true;
@@ -59,9 +60,15 @@ export class DebugAdapter extends DebugSession {
       },
       this.handleIncomingCDPMethods
     );
+
     this.sourceMapController = new SourceMapController(
       configuration.expoPreludeLineCount,
       configuration.sourceMapAliases
+    );
+
+    this.breakpointController = new BreakpointsController(
+      this.sourceMapController,
+      this.CDPCommunicator
     );
   }
 
@@ -90,7 +97,7 @@ export class DebugAdapter extends DebugSession {
             });
           }
 
-          this.updateBreakpointsInSource(message.params.url, consumer);
+          this.breakpointController.updateBreakpointsInSource(message.params.url, consumer);
         }
 
         this.sendEvent(new InitializedEvent());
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -93,132 +53,84 @@ export class DebugAdapter extends DebugSession {
...
+        const threadId = context.id;
+        const threadName = context.name;
+        this.sendEvent(new ThreadEvent("started", threadId));
+        this.threads.push(new Thread(threadId, threadName));
+        break;
+      case "Debugger.scriptParsed":
+        const sourceMapURL = message.params.sourceMapURL;
+
+        if (sourceMapURL?.startsWith("data:")) {
+          const { consumer, isMainBundle } = await this.sourceMapController.consumeNewSourceMap(
```
kmagiera, line 88: consume would suggest that the map is gone after the method call. Can we rename to "addSourceMap" or "registerSourceMap"

kmagiera: Also it doesn't feel like sourceMapController should really understand the concept of main bundle. It wouldn't hurt to just keep the line that determines isMainBundle value here instead of moving it into source map controller and making this method return an object that carries that information.

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -57,34 +36,15 @@ function typeToCategory(type: string) {
...
-
 export class DebugAdapter extends DebugSession {
   private variableStore: VariableStore = new VariableStore();
-  private connection: WebSocket;
-  private sourceMapAliases?: Array<[string, string]>;
+
+  private CDPCommunicator: CDPCommunicatorInterface;
+  private sourceMapController: SourceMapController;
+
+  private breakpointController: BreakpointsController;
```
kmagiera, line 45: is it for breakpoint or for breakpoints?

packages/vscode-extension/src/debugging/MyBreakpoint.ts
```diff
@@ -0,0 +1,135 @@
...
+  public async delete() {
+    return this.addTaskToQueue(async () => {
+      await this.deleteCDPBreakpoint();
+    });
+  }
+
+  public reset(sourceMapPath: string) {
+    return this.addTaskToQueue(async () => {
+      await this.resetCDPBreakpoint(sourceMapPath);
+    });
```
kmagiera, line 35: shouldn't this just call `delete` and then `add` ? it seems like that's exactly what is happening under the hood.

kmagiera (changes requested):
adding some comments inline after first pass over this

packages/vscode-extension/src/debugging/MyBreakpoint.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  public async add(sourcePath: string) {
```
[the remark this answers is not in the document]
filip131311: Line and column are stored, but not the sourcePath

packages/vscode-extension/src/debugging/MyBreakpoint.ts
filip131311: almost but we additionally change the verification status of the breakpoint to false after deleting it

```
if (this.verified) {
      await this.cdpSession.sendCDPMessage("Debugger.removeBreakpoint", {
        breakpointId: this.getId(),
      });
      this.verified = false; // <- here is the difference 
    }
    await this.setCDPBreakpoint(sourceMapPath);
```

It is usually unnecessary to do that after deleting

I'll change the code to: 

```
private async resetCDPBreakpoint(sourceMapPath: string) {
    if (this.verified) {
      await this.deleteCDPBreakpoint();
      this.verified = false;
    }
    await this.setCDPBreakpoint(sourceMapPath);
  }
``` 
to make the difference clearer

packages/vscode-extension/src/debugging/SourceMapsController.ts
```diff
@@ -0,0 +1,164 @@
...
+    const secondURL = new URL(url2);
+    firstURL.hostname = secondURL.hostname = "localhost";
+    firstURL.port = secondURL.port = "8080";
+    return firstURL.href === secondURL.href;
+  } catch (e) {
+    return false;
+  }
+}
+
+export class SourceMapController {
```
kmagiera, line 17: file is called "SourceMapsController" and class is "SourceMapController"

packages/vscode-extension/src/debugging/CDPBreakpoint.ts
```diff
@@ -0,0 +1,133 @@
...
+  public readonly column: number | undefined;
+  private _id: number | undefined;
+  private executingQueue: boolean = false;
+  private cdpCommunicationQueue: Array<() => Promise<void>> = [];
+
+  constructor(
+    private cdpSession: CDPSession,
+    private sourceMapController: SourceMapController,
+    verified: boolean,
+    line: number,
```
kmagiera, line 17: feels very weird that we pass line number and column but we don't pass the filename. It feels like line number and column makes it much more strict. There's no way breakpoint could move between files keeping its line and column, no?

filip131311: I actually didn't think about it before but the name of the file could change  during fast refresh.
After some initial test it's seems to me that this scenario break fast refresh, because it does not update the files that import the old file. 

The current behavior seems to be that the bundle is not actually updated, which means that the old breakpoint is still there but it is not visible for the user, that would suggest that the break point itself should contain it's sourceFile. 

I am not sure about any of this tho, maybe in this scenario we should induce "Reload JS" ourselves to improve developer experience?  I this scenario we would still like to use old breakpoints but with a new source file. 

Can we leave it as is for now and I'll adres it in a follow up when I'll figure how to best handle the scenario of fast refreshing after changing the file name?

kmagiera (approved):

PR #757, opened 2024-11-22
Add switch for device frame visibility

p-malecki:
This PR introduces a switch for controlling the visibility of the device frame in the IDE panel.

### How This Has Been Tested:
- Toggled the switch to verify that the `Show Device Frame` setting updates correctly and that the device frame visibility is changed accordingly.
- Changed the `Show Device Frame` setting through the VSCode settings and checked if the device settings dropdown reflects this setting correctly after reopening.

Screenshots:

|off|on|
|-|-|
|Screenshot 2024-11-22 at 16 35 33|Screenshot 2024-11-22 at 16 35 39|


filip131311 (approved):
LGTM

PR #765, opened 2024-11-25
Fix inspector on RN 73 after #693

kmagiera:
This PR fixes a regression in element inspector on RN 73 introduced in #693 

Because of the new approach of traversing component hierarchy added there, we broke compatibility with older versions of react native where renderer didn't have `getInspectorDataForInstance` method.

The fix is to restore the old behavior as a fallback and rely on component hierarchy returned by `getInspectorDataForViewAtPoint` when this renderer method isn't available.

### How Has This Been Tested: 
1. Open inspector hierarchy on expo-go app
2. Make sure inspector works on RN 75 and RN 76

### DIFF
--- a/packages/vscode-extension/lib/wrapper.js
+++ b/packages/vscode-extension/lib/wrapper.js
@@ -98,7 +98,7 @@ function extractComponentStack(startNode, viewDataHierarchy) {
 
   let stackItems = [];
   if (rendererConfig) {
-    // when we find renderer copnfig with getInspectorDataForInstance we use fiber node
+    // when we find renderer config with getInspectorDataForInstance we use fiber node
     // "return" property to traverse the component hierarchy
     let node = startNode;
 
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/lib/wrapper.js
```diff
@@ -89,34 +89,42 @@ function getRendererConfig() {
  * Return an array of component data representing a stack of components by traversing
  * the component hierarchy up from the startNode.
  * Each stack entry carries the component name, source location and measure function.
+ * We try to use React's renderer getInspectorDataForInstance to get the details about
+ * each particular component instance. Howeber, with older versions of React Native where
```
maciekstosio, line 93:
```suggestion
 * each particular component instance. However, with older versions of React Native where
```

packages/vscode-extension/lib/wrapper.js
```diff
@@ -89,34 +89,42 @@ function getRendererConfig() {
...
-function extractComponentStack(startNode) {
+function extractComponentStack(startNode, viewDataHierarchy) {
   const rendererConfig = getRendererConfig();
 
-  if (!rendererConfig) {
-    console.warn("Unable to find functional renderer config.");
-    return [];
+  let stackItems = [];
+  if (rendererConfig) {
+    // when we find renderer copnfig with getInspectorDataForInstance we use fiber node
```
maciekstosio, line 101:
```suggestion
    // when we find renderer config with getInspectorDataForInstance we use fiber node
```

maciekstosio (approved):

PR #784, opened 2024-11-27
Fix inspector overlay

filip131311:
This PR resolves a visual misalignment issue observed with the inspector overlay on Android, which was not correctly displaying the inspected component. The root cause of this problem was the method in which the application measured the component's position relative to the window. On Android devices, this calculation is shifted by the status bar displacing the window.

To address this, we have shifted the overlay rendering logic from the WebView to the application wrapper. Within the wrapper, the corrected coordinates provided by the inspector ensure that the overlay aligns accurately with inspected component component. 

However, there is still an issue of accurately identifying the component to be inspected, as we are restricted to using the mouse position data obtained from the WebView.

###Testing:
- Running the react-native-76 test-app on Android devices with and without the status bar.


kmagiera (commented):
I wonder why can't we just use StatusBar.current height to apply the correct offset when sending touch coordinates and later when sending back frame position?

jwajgelt:
Closing, since we're unlikely to merge this after all this time.

PR #785, opened 2024-11-27
Fix log position when using Sentry

maciekstosio:
This PR fixes logs when additional console.log wrappers are added. Currently, we're wrapping `console.log` so we assume that +1 stack frame shows the actual call. When there is an additional Sentry wrapper, all console logs point to a single place in the code - sentry wrapper. To solve that, I added an internal console.log call to count the frames offset. <s>This needs to happen each time, as we don't know when additional wrappers will be initialized.</s> We sample that 5 times, and repeat the process when something changes. 

**How this has been tested**
- Run app in IDE
- Print something using button with `console.log` on onPress
- Make sure the source link points to proper call place
- Add the following fragment into a global context on one of the screens:
```tsx
const origConsoleLog = console.log;
console.log = function (...args) {
  origConsoleLog.apply(console, ['Hello', ...args]);
};
``` 
- Without reloading, make sure the logged lines now include "Hello" prefix but also point to the correct place where console.log is called

On newly added expo-52 (with added sentry https://github.com/software-mansion-labs/radon-ide-test-apps/pull/1) and expo-go

|Before|After|
|-|-|
|SCR-20241128-mivd|SCR-20241128-mjax|


kmagiera (commented):
I was thinking of implementing the following idea that'd provide a more generic solution:

We measure the log stack offset when the first console log gets called in the following way:
1. We parse the current stack
2. We call console.log again but with some modifier that'd prevent that from being sent to actual console. Insead we're only interested in getting stack
3. we compare the top entries from the stack, the top entries that exists on both stacks are the wrapper calls – we need to count these to get the offset
4. we remember this offset for future console calls

packages/vscode-extension/lib/runtime.js
```diff
@@ -25,15 +25,46 @@ global.__RNIDE_onDebuggerReady = function () {
...
+          stackOffset = i;
+          break;
+        }
+      }
+      return;
+    }
+
+    initializationStack = stack;
+    sendInitialLog = true;
+    console.log(); // Recursive dummy console.log to get the source maps
```
kmagiera, line 60: we should avoid doing this for all console logs. We can cache the stack offset value.

kmagiera (changes requested):
Few things:

I don't think we need test-app for this. Perhaps we could add this to one of the existing apps if necessary but having a separate app to test just for this one thing to verify sounds like overkill

The logic should allow for caching. We don't want to recursively call console.log for every single log. I think what we could do is:
1) during first call we should do the recursive call and calculate the top-most index that is a log wrapper
2) during subsequent calls, we verify that the entry at that index is the same (meaning it uses the wrapper)
3) if for some reason the entry at that index is different, we can do recursive call to calculate the new index (this could only happen if someone changes the console config mid-flight)
4) otherwise we just use the frame at index+1 to get log location

 to, during first log call, we should

maciekstosio:
This approach will not work. We can store the index of the last known wrapper, but when we compare it, it will be the same even if another wrapper was added - the new wrapper will be at `index+1`, and we have no way of knowing if it is a wrapper or not without making a recursive call. I proposed a solution where we "sample" wrappers a few times (and repeat sampling when something changes). The idea is that wrappers will most likely be added at the beginning of execution.

packages/vscode-extension/lib/runtime.js
```diff
@@ -25,15 +25,74 @@ global.__RNIDE_onDebuggerReady = function () {
...
+    infoRef !== console.info
+  );
+}
+
+function updateConsoleRefs() {
+  logRef = console.log;
+  warnRef = console.warn;
+  errorRef = console.error;
+  infoRef = console.info;
+}
```
kmagiera, line 48: why can't we just use it for each function separately? This could just be defined inside wrapConsole and only work for the specific console function? Otherwise we're assuming that all log types are wrapped the same way which doesn't have to be the case.

In my optinion it'd be simpler to have it run for each function individually and would also not require this assumption.

packages/vscode-extension/lib/runtime.js
```diff
@@ -25,15 +25,74 @@ global.__RNIDE_onDebuggerReady = function () {
...
+          }
+        }
+
+        updateConsoleRefs();
+        return;
+      }
+
+      initializationStack = stack;
+      sendInitialLog = true;
+      console.log(); // Recursive dummy console.log to get the source maps
```
kmagiera, line 87: Woudl be much cleaner if we set flag immediately before and reset immediately after
```
logFunctionReentryFlag = true;
console.log();
logFunctionReentryFlag = false;

// calculate new offset here
```


when log function is called and `logFunctionReentryFlag` is set, the only thing we do, is we save current stack and exit to keep things simple:

```
let logFunctionReentryStack = null;
let logFunctionReentryFlag = false;

return function (...args) {
    const stack = parseErrorStack(new Error().stack);
   
    if (logFunctionReendrtFlag) {
        logFunctionReentryStack = stack;
        return;
    }
```

packages/vscode-extension/lib/runtime.js
```diff
@@ -25,15 +25,74 @@ global.__RNIDE_onDebuggerReady = function () {
...
     const stack = parseErrorStack(new Error().stack);
-    const expoLogIndex = stack.findIndex((frame) => frame.methodName === "__expoConsoleLog");
-    const location = expoLogIndex > 0 ? stack[expoLogIndex + 1] : stack[1];
+
+    // To get the proper stack frame, so we can display link to the source code
+    // we need to skip wrappers (like wrapConsole below or for example Sentry wrapper)
+    // Otherwise, the stack frame would point to the wrapper and not the actual source code
+    // To do that, we run console.log again in runWrapper, and then compare
+    // first frames to find the offset. We do that when ant of console ref changes
+    if (isConsoleRefChanged()) {
```
kmagiera, line 62: would be better to only check the current console function and not all functions. For that we should change `wrapConsole` to take `logFunctionKey` instead of actual function. It can also assign the wrapped function directly to console object so that we can just call `wrapConsole('log')`

kmagiera (changes requested):

packages/vscode-extension/lib/runtime.js
maciekstosio: Yeah, I think that looks much cleaner, thanks!

packages/vscode-extension/lib/runtime.js
maciekstosio: I think we need to take both. Otherwise, we lose ref to actual log method

maciekstosio: Hmm, I could skip the log function by saving it at the beginning, but the source translations stop working, let me check that.

maciekstosio: Ok, seems to be ok. Changed in https://github.com/software-mansion/radon-ide/pull/785/commits/bdce780109b07255849dc8bfc18718d581804e3c

kmagiera (approved):

PR #789, opened 2024-11-28
Add license registration

filip131311:
This PR adds a way for the user to register device, using a license key, by adding an an "Activate license" button in place of "report issue" button on the main screen. 

Additionally we add a new report issue button to the settings dropdown.

### How Has This Been Tested: 

- An account was created on customer portal and the license was activated.

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -105,6 +105,7 @@ export interface ProjectEventMap {
   log: { type: string };
   projectStateChanged: ProjectState;
   deviceSettingsChanged: DeviceSettings;
+  licenseActivationChanged: boolean;
   navigationChanged: { displayName: string; id: string };
   needsNativeRebuild: void;
 }
@@ -139,6 +140,9 @@ export interface ProjectInterface {
   focusDebugConsole(): Promise<void>;
   openNavigation(navigationItemID: string): Promise<void>;
   openDevMenu(): Promise<void>;
+
+  activateLicense(activationKey: string): Promise<boolean>;
+  isLicenseActivated(): Promise<boolean>;
 
   resetAppPermissions(permissionType: AppPermissionType): Promise<void>;
 
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -101,6 +101,12 @@ export type TouchPoint = {
   yRatio: number;
 };
 
+export enum ActivateDeviceResult {
+  succeeded,
+  notEnoughSeats,
+  unableToVerify,
+}
+
 export interface ProjectEventMap {
   log: { type: string };
   projectStateChanged: ProjectState;
@@ -141,8 +147,8 @@ export interface ProjectInterface {
   openNavigation(navigationItemID: string): Promise<void>;
   openDevMenu(): Promise<void>;
 
-  activateLicense(activationKey: string): Promise<boolean>;
-  isLicenseActivated(): Promise<boolean>;
+  activateLicense(activationKey: string): Promise<ActivateDeviceResult>;
+  hasActiveLicense(): Promise<boolean>;
 
   resetAppPermissions(permissionType: AppPermissionType): Promise<void>;
 
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -104,6 +104,7 @@ export type TouchPoint = {
 export enum ActivateDeviceResult {
   succeeded,
   notEnoughSeats,
+  keyVerificationFailed,
   unableToVerify,
   connectionFailed,
 }
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -105,6 +105,7 @@ export enum ActivateDeviceResult {
   succeeded,
   notEnoughSeats,
   unableToVerify,
+  connectionFailed,
 }
 
 export interface ProjectEventMap {
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -5,6 +5,7 @@ import stripAnsi from "strip-ansi";
 import { minimatch } from "minimatch";
 import { isEqual } from "lodash";
 import {
+  ActivateDeviceResult,
   AppPermissionType,
   DeviceSettings,
   InspectData,
@@ -492,13 +493,13 @@ export class Project
   public async activateLicense(activationKey: string) {
     const computerName = os.hostname();
     const activated = await activateDevice(activationKey, computerName);
-    if (activated) {
+    if (activated === ActivateDeviceResult.succeeded) {
       this.eventEmitter.emit("licenseActivationChanged", true);
     }
     return activated;
   }
 
-  public async isLicenseActivated() {
+  public async hasActiveLicense() {
     return !!(await getLicenseToken());
   }
 
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -32,7 +32,7 @@ import { Devtools } from "./devtools";
 import { AppEvent, DeviceSession, EventDelegate } from "./deviceSession";
 import { PlatformBuildCache } from "../builders/PlatformBuildCache";
 import { PanelLocation } from "../common/WorkspaceConfig";
-import { activateDevice, getLicenseToken, removeLicenseToken } from "../utilities/license";
+import { activateDevice, getLicenseToken } from "../utilities/license";
 
 const DEVICE_SETTINGS_KEY = "device_settings_v4";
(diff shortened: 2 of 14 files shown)


kacperkapusciak (commented):
i can confirm it works with the newest sim server 🎉 
image
I'm missing some indication that everything went okay and i'm successfully activated.

Maybe some faded label in the settings dropdown saying "Radon IDE activated" or something

packages/vscode-extension/src/utilities/license.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Platform } from "./platform";
+import { exec } from "./subprocess";
+
+const TOKEN_KEY = "token_key";
+const BASE_CUSTOMER_PORTAL_URL = "https://portal.ide.swmansion.com/";
+
+export async function activateDevice(licenseKey: string, username: string) {
+  const url = new URL("/api/create-token", BASE_CUSTOMER_PORTAL_URL);
+  const body = {
+    fingerprint: await generateDeviceFingerprint(),
```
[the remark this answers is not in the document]
filip131311: I don't think that will be the problem since will ship it with the new functionality

packages/vscode-extension/src/project/project.ts
```diff
@@ -487,6 +489,19 @@ export class Project
     await this.deviceSession?.openDevMenu();
   }
 
+  public async activateLicense(activationKey: string) {
+    const computerName = os.hostname();
+    const activated = await activateDevice(activationKey, computerName);
```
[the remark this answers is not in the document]
kmagiera: hostname alone isn't ideal but we discussed that with @filip131311 and I believe we should keep it that way for now. The hostname is definitely a safe variable that can be stored in our db and displayed to the license owner. The reasoning is that the hostname would appear to other computers when you join the network, so by design is treated as more public facing information by the computer owner, which isn't necessarily the case with user name (useInfo).

I understand the concern with that name not carrying enough information to identify users but I also don't think it'll be a frequent problem and even when it is, it'll never be critical. Once it becomes a problem at larger scale we will consult the customers affected by to find a better solution.

packages/vscode-extension/src/utilities/license.ts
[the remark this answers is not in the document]
kmagiera: no, we will always package the right version. This is only a problem in development if you didn't build the new version after the update (although this is something we should and can fix too in the dev environment)

packages/vscode-extension/src/utilities/license.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+async function generateDeviceFingerprint() {
+  const simControllerBinary = path.join(
+    extensionContext.extensionPath,
+    "dist",
+    Platform.select({ macos: "simulator-server-macos", windows: "simulator-server-windows.exe" })
+  );
+  const { stdout } = await exec(simControllerBinary, ["fingerprint"]);
+  return stdout;
```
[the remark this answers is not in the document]
filip131311: yes

packages/vscode-extension/src/webview/views/PreviewView.css
```diff
@@ -89,3 +89,7 @@
   border-radius: 50%;
   margin-right: 6px;
 }
+
+.activate-license-button {
+  color: var(--swm-error-color);
```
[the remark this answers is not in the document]
filip131311: Added

packages/vscode-extension/src/utilities/license.ts
```diff
@@ -0,0 +1,68 @@
...
+  } catch (e) {
+    Logger.warn("Creating license token with license key failed", e);
+    return ActivateDeviceResult.unableToVerify;
+  }
+
+  const responseBody = await response.json();
+
+  if (response.ok) {
+    const newToken = responseBody.token as string;
+    extensionContext.secrets.store(TOKEN_KEY, newToken ?? "");
```
kmagiera, line 41: shouldn't we await here?

packages/vscode-extension/src/webview/views/ActivateLicenseView.tsx
```diff
@@ -0,0 +1,108 @@
...
+            ) If you don't have a license, you can purchase it{" "}
+            <a href="https://ide.swmansion.com/pricing" target="_blank" rel="noopener noreferrer">
+              here
+            </a>
+            .
+          </div>
+        )}
+        {wasRejected && wasEnoughSeats && (
+          <div className="error-text">
+            Unable to verify the key. Please ensure your license key is correct. Check this{" "}
```
kmagiera, line 68: We should either distinguish the case where license servers cannot be reached (e.g. no connection) vs the case when license key is invald or make this message inform the user that network connection is needed for activation.

filip131311: done

packages/vscode-extension/src/common/Project.ts
```diff
@@ -101,10 +101,18 @@ export type TouchPoint = {
   yRatio: number;
 };
 
+export enum ActivateDeviceResult {
+  succeeded,
+  notEnoughSeats,
+  unableToVerify,
+  connectionFailed,
```
kmagiera, line 108: I still don't think thats and exhausting list, or rather maybe doesn't cover some important aspects of verification.

There should be a clear state indicating the key is invalid, e.g. keyVerificationFailed which should be used when we are certain that this is the reason the request failed (perhaps we'd get a specific http return code in this case).

Using `unableToVerify` as a wildecard for other types of errors is good, but you should keep in mind when it happens the message to the user should be different than in the case they put in an invalid license key. Specifically the issue could be that our servers are down and we get 500 errors from load balancer. This will still not cause fetch to throw

packages/vscode-extension/src/utilities/license.ts
```diff
@@ -0,0 +1,68 @@
+import fetch from "node-fetch";
+import { extensionContext } from "./extensionContext";
+import { exec } from "./subprocess";
+import { Logger } from "../Logger";
+import { simulatorServerBinary } from "./simulatorServerBinary";
+import { ActivateDeviceResult } from "../common/Project";
+
+const TOKEN_KEY = "token_key";
```
kmagiera, line 8: where is this value stored? Maybe we can use some key that includes radon-ide in the name, such that it could be easier to locate the source of this data or avoid other extensions reading it?

packages/vscode-extension/src/webview/views/ActivateLicenseView.tsx
kmagiera: As pointed out earlier, we need a way to tell whether someone has put in a correct key vs when there was some possibly random issue validating it (i.e. servers are down)

In the first scenario we can write:
> Provided key is not a valid license key. You can find your key on the [Radon IDE customer portal]. Please follow the [license activation instructions] if you need further assistance.

For the case of some generic error in which we know that this isn't key validation response from the server we could write:
> We were unable to verify the license key at this point. Please consult the [license activation instructions] and try again once you check the provided key is valid.

packages/vscode-extension/src/webview/views/ActivateLicenseView.tsx
```diff
@@ -0,0 +1,111 @@
...
+            Radon IDE customer portal (
+            <a href="https://portal.ide.swmansion.com/" target="_blank" rel="noopener noreferrer">
+              link
+            </a>
+            ).
+          </div>
+        )}
+        {activateDeviceResult === ActivateDeviceResult.connectionFailed && (
+          <div className="error-text">
+            We ware unable to register your device due to the connection failing.
```
kmagiera, line 88: > Activating license key requires a network connection. Make sure you are connected and try activating the key again. In case of any issues please consult our [license activation instructions]

kmagiera (commented):
Added some final comments and copy suggestions

packages/vscode-extension/src/utilities/license.ts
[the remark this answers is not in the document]
balins: If for some reason simserver fails with exit code != 0, does this function fail immediately? We'd like to avoid setting the machine fingerprint to some error message or empty string

packages/vscode-extension/src/webview/views/ActivateLicenseView.tsx
```diff
@@ -0,0 +1,129 @@
...
+          </div>
+        )}
+        {activateDeviceResult === ActivateDeviceResult.notEnoughSeats && (
+          <div className="error-text">
+            Your organization does not have any available seats left, you can purchase more on the
+            Radon IDE customer portal (
+            <a href="https://portal.ide.swmansion.com/" target="_blank" rel="noopener noreferrer">
+              link
+            </a>
+            ).
```
kmagiera, line 98: > Your license does not have any available seats. Log to the [Radon IDE customer portal] to manage seats assigned to the license seats or purchase additional seats.

kmagiera (approved):

PR #790, opened 2024-11-28
Fix sourcemap RangeError issue by properly freeing source map consumers

kmagiera:
This PR fixes exception reported by a number of users and also seen in telemetry:
`RangeError: Start offset -2033400168 is outside the bounds of the buffer`

The root cause of this issue was in misuse of [source-map](https://github.com/mozilla/source-map) package which requires destroy method to be called on `SourceMapConsumer` instances when they are no longer used.

The reason why `destroy` call is necessary, is that source-map package uses wasm under the hood where it allocates data structures that allows for fast processing of source map translation queries. Since the wasm memory space is limited, when too many maps were loaded in the memory, the allocate call from [parseMappings](https://github.com/mozilla/source-map/blob/3cb92cc3b73bfab27c146bae4ef2bc09dbb4e5ed/lib/source-map-consumer.js#L312) would return negative values indicating a failed alloc attempt (and that's the number that'd show up in the logs).

Thanks to the help from community members we managed to isolate and narrow down the issue providing a reproducible scenario. In order to get this issue one would need to reload JS several times (in some instances I had to reload more than 20 times, but eventually would start getting this issue).

The fix is to call `destroy` on `SourceMapConsumer` instances when clearing JS execution context. This is called specifically when reloading JS and after that point the stored source maps are no longer in use.

### How Has This Been Tested: 
1. Open any test project
2. Reload JS several times
3. Before this change at some point you'd start seeing RangeError in logs and invalid line numbers would show up in console tab + breakpoints would no longer work. After this change the error never happens even after a significant number of reloads.

### DIFF
--- a/packages/vscode-extension/src/debugging/SourceMapsRegistry.ts
+++ b/packages/vscode-extension/src/debugging/SourceMapsRegistry.ts
@@ -24,6 +24,7 @@ export class SourceMapsRegistry {
   ) {}
 
   public clearSourceMaps() {
+    this.sourceMaps.forEach(([_, __, consumer]) => consumer.destroy());
     this.sourceMaps = [];
     this.sourceMapFilePaths.clear();
   }


filip131311 (approved):
I played with it a bit and it seems like a good solution, but I still don't know why problem appears on some reloads and not others, I am worried about it because when we fast refresh we load a new consumer too (altho a single one), if an issue is just the size of a mapping if someone ware to change a very big file (like a library) the problem could show up again.

kmagiera:
I haven't got to the very bottom of this issue, but it is definitely related to the overall size of the source maps bundled. From my testing it seems to happen around the time when more than 100MB of source maps are loaded which is pretty high limit already. Even if you edit some big file like the renderer, despite being very large it still only contributes a small fraction to the overall size, so I wouldn't be worried that much.

In addition, even when this happens due to source map consumers loaded for fast refresh, users would still get a way to recover by simply refreshing the app which will cause all the memory to clean up – this wasn't the case before.

PR #796, opened 2024-11-29
Add Platform check when validating EAS builds

filip131311:
This PR fixes a regression introduced in #727, that performed android specific check regardless of the platform, causing iOS builds to fail.

### How Has This Been Tested: 

-run eas development build on ios
-run eas development build on android


kmagiera (approved):

EricZeiberg:
Thanks for the quick fix!

kmagiera:
This will go out in the next release we plan for 12/2

PR #800, opened 2024-11-29
Fix device android native logs not working when multiple devices are available 

maciekstosio:
When running IDE with Android device and another Android Emulator, the logcat commands fail on `error: more than one device/emulator`. This PR adds selecting a proper device with `-s <serial>` adb param. Similar pattern is used for other adb executions. 

### How Has This Been Tested: 

- Run a standalone Android emulator 
- Run project with IDE using an Android device 
- The native logs should appear


filip131311 (approved):
LGTM

PR #807, opened 2024-12-02
Add close when stale GH Action

maciekstosio:
This PR adds workflow for `Close when stale` functionality. We could use it to reduce the number of issues - tag the ones that won't be resolved without user interaction. 

Nicked from: https://github.com/software-mansion/react-native-reanimated/blob/main/.github/workflows/github-close-when-stale-nightly.yml


kmagiera (approved):

.github/workflows/close-when-stale.yaml
```diff
@@ -0,0 +1,34 @@
...
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install Actions
+        run: yarn install
+
+      - name: Close when stale
+        uses: ./close-when-stale
+        with:
+          close-when-stale-label: 'Close when stale'
```
kmagiera, line 33: We don't have this label in our project though. Also all labels that we have starts with lowercase letters, so perhaps we want to create one that aligns with the existing labels in that matter

maciekstosio, line 33:
```suggestion
          close-when-stale-label: 'close when stale'
```

maciekstosio: True, I created lowercase label and aligned the code in: https://github.com/software-mansion/radon-ide/pull/807/commits/f077e87fa903fa1f8891cc007a6c3397d9df8330

PR #828, opened 2024-12-04
docs: add better analytics

kacperkapusciak:
This PR adds more granular analytics which will enable us to make better informed decisions about the product:

**Landing page**
- `Download` button click
-  `See video` button click

**Pricing page**
- `See compatibility` button click
- Checkout events
  - `checkout.loaded`
  - `checkout.closed`
  - `checkout.completed`
  - `checkout.error`

Implemented thanks to https://vercel.com/docs/analytics/custom-events

### DIFF
--- a/packages/docs/src/hooks/usePaddle.tsx
+++ b/packages/docs/src/hooks/usePaddle.tsx
@@ -6,7 +6,7 @@ const eventCallback = (event) => {
   const { name, data } = event;
   const { items } = data;
   const [item] = items;
-  const payload = { name: item.price_name, interval: item.billing_cycle.interval };
+  const payload = { priceName: item.price_name, interval: item.billing_cycle.interval };
 
   switch (name) {
     case "checkout.loaded":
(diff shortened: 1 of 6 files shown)


packages/docs/src/hooks/usePaddle.tsx
```diff
@@ -1,10 +1,41 @@
 import { useEffect, useState } from "react";
 import { initializePaddle, InitializePaddleOptions, Paddle } from "@paddle/paddle-js";
+import { track } from "@vercel/analytics";
+
+const eventCallback = (event) => {
+  const { name, data } = event;
+  const { items } = data;
+  const [item] = items;
+  const payload = { name: item.price_name, interval: item.billing_cycle.interval };
```
filip131311, line 9: I assume name is something like "business" "personal", if so could you rename this to something more descriptive like "pricingName", "pricingType", "subscribtionType" ?

filip131311 (approved):
I left a small comment, but seems good. 

Can Vercel tracking tell if the same user clicked on Download, as checkout completed?

packages/docs/src/hooks/usePaddle.tsx
kacperkapusciak: Fixed in [bf24b0a](https://github.com/software-mansion/radon-ide/pull/828/commits/bf24b0af8906931c31d66bc78e7a4a000aecd251)

Answering your question: I don't think so.

It tracks individual page views by hashing an incoming request on the server. Tracking specific user actions would require using more advanced analytics with cookies. We would then need to show a cookie banner etc.

https://vercel.com/docs/analytics/privacy-policy#visitor-identification-and-data-storage

PR #833, opened 2024-12-06
Handle eas build locally thorugh buildCommand in launch.json

### LINKED ISSUE #824: Support for EAS local builds
**Description**
It would be very useful if Radon IDE could use EAS local builds, in combination with local fingerprints. We don't use EAS builds due to its pricing model, and have local builds available in the project root directory that can be used by the simulator / emulator. These are built using `eas build --local` but only have to be rebuilt when the fingerprint changes.

**Environment**

- Version of currently used Radon IDE
1.0.0
- Version of React Native/Expo used in the project.
0.75 / Expo 51

maciekstosio:
Fixes #824

`customBuild.buildCommand` assumes the last line of the output is a path to the binary.  This PR changes this behavior to look for all paths in the last line of the output; if it does exist, it takes the first path and extracts the tar file if necessary (this is how iOS is bundled). This seems to be a common use case (especially with eas --local) and, currently, we have no other way of handling `eas build` command locally. In the feature, I believe we should extend `eas.*` launch config to handle `local` flag. 

### How Has This Been Tested: 

- Open Expo 52 Eas from: https://github.com/software-mansion-labs/radon-ide-test-apps/pull/3
- add launch configuration:
```
{
   // Use IntelliSense to learn about possible attributes.
   // Hover to view descriptions of existing attributes.
   // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
   "version": "0.2.0",
   "configurations": [
      {
         "type": "radon-ide",
         "request": "launch",
         "name": "Radon IDE panel",
         "customBuild": {
            "ios": {
               "buildCommand": "npx eas build --platform ios --profile development --local --non-interactive",
            },
            "android": {
               "buildCommand": "npx eas build --platform android --profile development --local --non-interactive",
            }
         },
         "ios": {
            "configuration": "Debug"
         },
         "android": {
            "buildType": "debug"
         }
      }
   ]
}
```
- verify that the app builds (it takes a long time especially on iOS)and IDE works


packages/vscode-extension/src/builders/customBuild.ts
```diff
@@ -1,28 +1,50 @@
...
+import { mkdtemp } from "fs/promises";
 import { Logger } from "../Logger";
 import { command, lineReader } from "../utilities/subprocess";
 import { CancelToken } from "./cancelToken";
 import { getAppRootFolder } from "../utilities/extensionContext";
+import { extractTarApp } from "./utils";
 
 type Env = Record<string, string> | undefined;
 
+const EXPO_LOCAL_BUILD_PATH_REGEX = new RegExp("You can find the build artifacts in (.*)");
```
kmagiera, line 13: if we only scan the last line I think we could relax it much more and just take the file paths that are present in the final line of output

kmagiera (approved):
changes look good, but we also should plan docs updates to explain the changes better in configuration docs.

maciekstosio:
Docs update: https://github.com/software-mansion/radon-ide/pull/897

packages/vscode-extension/src/builders/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  const appPath = path.join(pathToExtract, appName);
+  Logger.debug(`Extracted app archive to '${appPath}'.`);
+  return appPath;
+}
+
+type TarCommandArgs = { archivePath: string; extractDir: string };
+function tarCommand({ archivePath, extractDir }: TarCommandArgs) {
+  return exec("tar", ["-xf", archivePath, "-C", extractDir]);
```
kmagiera, line 51: perhaps would be better to use some existing module like `node-tar` this way making this work cross platform and resistant to some PATH related issues

packages/vscode-extension/src/builders/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+    tarCommand({ archivePath: binaryPath, extractDir: pathToExtract })
+  );
+
+  if (failed) {
+    Logger.error(`Failed to extract archive '${binaryPath}' to '${pathToExtract}'.`);
+    return undefined;
+  }
+
+  // assuming that the archive contains only one app file
+  const appName = (await readdir(pathToExtract)).find(
```
kmagiera, line 32: instead of having this here both here and in runExternalBuild maybe we should just make this metod return a single extracted file and then only make the file extension checks in `runExternalBuild`

kmagiera (approved):
Looks good, left some comments inline

PR #829, opened 2024-12-04
Add Linux support for VSCode extension (#688)

tobisamuel:
This PR adds Linux support to the VSCode extension, implementing platform-specific paths and configurations. The changes follow patterns established in PR #462 (Windows support) and make the extension functional on Linux systems.

- Add Linux as a supported platform in Platform module
- Implement Linux paths for Android SDK and tooling
- Add Linux-specific configurations for file system and processes
- Update UI components to handle Linux keyboard shortcuts

Fixes #688

### How This Has Been Tested

To ensure the changes in this PR work as expected, the following test plan was carried out:
- Run development version of the extension successfully on my Linux Mint (based on Ubuntu 24.04) machine. 
- Ensure the ANDROID_HOME 
environment variable was correctly set on my system 
- A React Native project was initialized, and the extension managed the building and installing the app on the selected emulator.
- Verified that all extension features, such as the inspector, debugger, device settings, and screen recording worked correctly on Linux.

https://github.com/user-attachments/assets/7a08b927-8ad7-4a3a-96f7-0df5dc4a384d

PR #839, opened 2024-12-09
fix(docs): Change support for Linux in compatibility table

tboba:
Since we've added a support of Radon IDE for Linux (https://github.com/software-mansion/radon-ide/pull/829), it's time to change its support in compatibility table 😄 
This PR changes the value of compatibility for Linux, indicating that this system is officially supported from now on.

### How Has This Been Tested: 

Manually, by building the `docs` package and running it in development mode with `yarn start`.

### DIFF
--- a/packages/docs/docs/getting-started/compatibility.md
+++ b/packages/docs/docs/getting-started/compatibility.md
@@ -73,13 +73,10 @@ Due to the complexity and a non-standard nature of the brownfield development th
 |         | iOS    | Android                  |
 | ------- | ------ | ------------------------ |
 | macOS   | <Yes/> | <Yes/>                   |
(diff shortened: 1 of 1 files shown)


tboba:
Also, one question here - should we indicate that Linux support is still in beta (same as on Windows), or can we keep it as it is?

kmagiera (commented):
Linux support should be considered beta. It will be available under beta license (same as windows where it doesn't require commercial license to use). Also, this isn't published to the markeplace yet, so to early to change this in the documentaiton

tboba:
@kmagiera Sure, will make a change later 👍 Let's wait with merging this PR then.

kmagiera (approved):

PR #841, opened 2024-12-10
Fix Node Modules detection without package.json

p-malecki:
This PR resolves the incorrect marking of the Node Modules dependency as uninstalled when a project lacks a `package.json` file.

The issue occurred because the functions `isNpmModulesInstalled` and `isYarnModulesInstalled` did not check for an empty parsedOutput, which is the result of the `npm ls --json` command. Previously these functions mistakenly returned true when parsedOutput was empty. Now, they correctly return false if parsedOutput is empty.

### How Has This Been Tested: 
- Tested scenarios with and without a `package.json`, lock files, and the `node_modules` folder  to ensure accurate dependency status.

|Before|After|
|-|-|
|Screenshot 2024-12-10 at 10 00 39|Screenshot 2024-12-10 at 10 00 04|

(The "After" comparison showed errors for missing Android and iOS directories because it incorrectly detected the app root folder as `expo-router/no_node_modules/@expo/cli/static/template` instead of the correct `expo-router/`.)

### DIFF
--- a/packages/vscode-extension/src/utilities/packageManager.ts
+++ b/packages/vscode-extension/src/utilities/packageManager.ts
@@ -124,8 +124,13 @@ async function isNpmModulesInstalled(workspacePath: string): Promise<boolean> {
       cwd: workspacePath,
       quietErrorsOnExit: true,
     });
-    const parsedJson = JSON.parse(stdout);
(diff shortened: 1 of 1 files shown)


kmagiera (approved):
looks good

Issue #736, opened 2024-11-17
does not work with pnpm workspaces


AyoCodess:
We have a similar error with the use of  ``catalog:``` to reference node modules versions. 
![Screenshot 2024-11-17 at 15 41 17](https://github.com/user-attachments/assets/2b890701-636e-47c7-b32a-b2a78858bcf1)

maciekstosio:
Hi, thanks for the bug report. Could you please provide a repo with a reproduction? I'll try to take a look at it.

Issue #837, opened 2024-12-09
PNPM Monorepo Support


AlvSovereign:
**Description**
PNPM Monorepo supoport would be greatly appreciated. This is stoping me from purchasing a licence.

necmettindev:
bump

necmettindev:
@p-malecki is there a launch.json configuration for multiple expo app mono-repo?

p-malecki:
There is, you should add the main app folder path to `appRoot` setting in `launch.json` (a relative path should be enough). You can find more config details in the docs: https://ide.swmansion.com/docs/guides/configuration

maciekstosio:
@necmettindev is everything working for you as expected?

PR #842, opened 2024-12-10
Add support for PNPM monorepos

p-malecki:
This PR introduces support for PNPM monorepos:
- identifies the monorepo workspace root based on the presence of the `pnpm-workspace.yaml` file
- installs node modules using PNPM if `isPnpmModulesInstalled` detects them as not installed

Resolves: #736 and #837 

### How Has This Been Tested:
- Tested with the new `pnpm-monorepo` test app introduced in [PR](https://github.com/software-mansion-labs/radon-ide-test-apps/pull/2), confirming correct app building both for iOS and Android on macOS.

### DIFF
--- a/packages/vscode-extension/src/utilities/packageManager.ts
+++ b/packages/vscode-extension/src/utilities/packageManager.ts
@@ -153,9 +153,34 @@ async function isYarnModulesInstalled(workspacePath: string): Promise<boolean> {
   }
 }
 
-async function isPnpmModulesInstalled(): Promise<boolean> {
-  // TODO: add pnpm support
-  return false;
+async function isPnpmModulesInstalled(workspacePath: string): Promise<boolean> {
+  try {
+    const { stdout } = await command("pnpm ls --json", {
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/utilities/packageManager.ts
```diff
@@ -153,9 +153,34 @@ async function isYarnModulesInstalled(workspacePath: string): Promise<boolean> {
...
+      if (!pkg || !pkg.dependencies || Object.keys(pkg.dependencies).length === 0) {
+        return false;
+      }
+
+      for (const depInfo of Object.values<{ version: string }>(pkg.dependencies)) {
+        if (!depInfo || !depInfo.version) {
+          return false;
+        }
+      }
+    }
```
kmagiera, line 178: can you add a few lines of explanation in comments on what do we test here and why?

kmagiera (approved):
Looks ok, would like to have some inline comments explaining the install check logic

PR #846, opened 2024-12-12
Enhance Console logs

maciekstosio:
This PR aims to improve the debugging experience by making it similar to Chrome DevTools. 

It improves the preview, formatting and truncating of console logs and splits large arrays into smaller changes, where chunks are fetched on expand - similarly as `vscode-js-debug` does. The more detailed previews are generated based on `preview` property from the debugger, so it may not work on older RN versions.  

The main idea was not to reinvent the wheel but to reuse as much of https://github.com/microsoft/vscode-js-debug as possible. Thus we forked it with small patch https://github.com/software-mansion-labs/vscode-js-debug/commit/a5794161fde3e3b55df4633104d3aba414194aad that allows us to build it - otherwise the TS on our side failed. `vscode-js-debug` sits in our repo as a git submodule and no additional steps should be necessary (preinstall script should take care of pulling it, installing and building). 

|Before|After|
|-|-|
|logs-before|logs-after| 

|Before|After|
|-|-|
|large-arrays-before|large-arrays-after|

### How Has This Been Tested: 

Update `printLogs` method with:

```
function printLogs() {
    const string = 'string';
    const object = { key: 'value', key2: 'value2' };
    const set = new Set([1, 2, 3, 3]);
    const array1000 = new Array(1000).fill(0).map((_, i) => i);
    const array101 = new Array(101).fill(0).map((_, i) => array1000);
    const int = 1;
    const float = 1.1;
    const func = function() {};
    const anonymousFunc = () => {};

    console.log(string, array1000, array101, object, set,
      int,
      float,
      func,
      anonymousFunc);
}
```

Tested on react-native-76, react-native-73.


packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -143,7 +149,64 @@ export class DebugAdapter extends DebugSession {
     }
   };
 
+  async formatDefaultString (
```
maciekstosio, line 152: Based `formatDefaultString` and `formatComplexStringOutput` based on https://github.com/microsoft/vscode-js-debug/blob/9a436b1a3060958110959c590eac9d1a8d548922/src/adapter/console/textualMessage.ts#L83

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -204,21 +286,73 @@ export class DebugAdapter extends DebugSession {
...
 
-  private createVariableForOutputEvent(args: CDPRemoteObject[]) {
-    // we create empty object that is needed for DAP OutputEvent to display
-    // collapsed args properly, the object references the array of args array
-    const argsObjectDapID = this.variableStore.pushReplVariable(
-      args.map((arg: CDPRemoteObject, index: number) => {
+  private async createVariableForOutputEvent(args: CDPRemoteObject[]) {
+    const prepareVariables = await Promise.all(
+      args.map(async (arg: CDPRemoteObject, index: number) => {
+        // if (arg.type === "object") {
```
maciekstosio, line 292: The first approach is to create an object split into chunks - the downside is that we need to get properties recursively

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -413,6 +547,40 @@ export class DebugAdapter extends DebugSession {
         return this.cdpSession.sendCDPMessage("Runtime.getProperties", params);
       }
     );
+
+    if (args.filter === "indexed") {
```
maciekstosio, line 551: Second variable makes the call to CDP to execute method that returns part of the array, based on: https://github.com/microsoft/vscode-js-debug/blob/6902aae25b4970d137041f9a84221eb327dec5a9/src/adapter/variableStore.ts#L1033
and remote call: 
https://github.com/microsoft/vscode-js-debug/blob/8b0bffc47d1836f4357db9d923685e730630a57a/src/adapter/templates/index.ts#L127

packages/vscode-extension/package.json
```diff
@@ -452,9 +452,11 @@
     "typecheck": "tsc --noEmit",
     "format": "prettier --write --list-different src",
     "build:tests": "tsc --project tsconfig.test.json",
-    "test": "npm run build:extension-code && npm run build:webview && npm run build:tests && vscode-test"
+    "test": "npm run build:extension-code && npm run build:webview && npm run build:tests && vscode-test",
+    "preinstall": "cd ../vscode-js-debug && git submodule update && npm install && npx tsc"
```
kmagiera, line 456: we shouldn't use `npx` in scripts like that, this may result in different versions of tsx being run for different people. I'd assume that repo already has tsc as dev-dependency in which case it shouldn't require `npx` to run it

packages/vscode-extension/src/debugging/templates/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from "./getArraySlots";
```
kmagiera, line 1: do we really need this file?

packages/vscode-extension/src/debugging/DebugAdapter.ts
kmagiera: don't put explanation in comments. If something needs to be explain, it should either be in code as inline comment, or in the PR description that gets included in the git history. Inline comments on PRs like this one are not something people typically look at as you typically check the code and then the blame commit + the comments are automatically folded when you click "resolve"

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -407,12 +434,40 @@ export class DebugAdapter extends DebugSession {
...
+        );
+
+        response.body.variables = properties;
+      } catch (e) {
+        Logger.error("[CDP] Failed to retrieve array partially", e);
       }
-    );
+    }
+
+    if (args.filter !== "indexed" && args.filter !== "named") {
```
kmagiera, line 462: maybe this should be an `else if` given we exclude "indexed" here. Also what happens when "named" is used as this doesn't seem to be handled at all?

kmagiera (changes requested):
Left some comments inline.

What I don't see how this PR addresses which was one of the main motivators for doing this is how do we deal with large strings. Is it entirely handled by the imported methods from vscode-js-debug?

packages/vscode-extension/src/debugging/templates/index.ts
maciekstosio: Nah, I added this assuming there might be more in the future and it's better to import from one place. Remove in https://github.com/software-mansion/radon-ide/pull/846/commits/57750e9475178114cf6c386c4a97df2cb75d8682

packages/vscode-extension/package.json
maciekstosio: Changed in: https://github.com/software-mansion/radon-ide/pull/846/commits/21e4c4f7ccd808540321a849248f272eea00ef3b

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -407,12 +449,43 @@ export class DebugAdapter extends DebugSession {
...
+        );
+
+        response.body.variables = properties;
+      } catch (e) {
+        Logger.error("[CDP] Failed to retrieve array partially", e);
       }
-    );
+    } else if (args.filter === "named") {
+      // We do nothing for named variables. We set 'named' and 'indexed' only for arrays in variableStore
+      // so the 'named' here means "display chunks" (which I is handled by Debugger). If we'd get the properties
```
kmagiera, line 484:
```suggestion
      // so the 'named' here means "display chunks" (which is handled by Debugger). If we'd get the properties
```

kmagiera (approved):
Looks good!

PR #849, opened 2024-12-12
[Expo-go][Android] Fix Debugger

filip131311:
This PR solves an issue with new debugger not being attached on android in some scenarios. In  #750 we introduced a way of ignoring expo go host runtime and selecting the second one when project was expo go, but for some reason the behavior of expo go runtimes seems to be inconsistent across platforms and versions and sometime the host runtime does not exist.

The following is the result of `{metro}/json/list` request for android expo go application in expo version: `52.0.18`:
```
[
  {
    "id": "5ca976e985fc33959628c11becb91a7dc7842a8f-1",
    "title": "React Native Bridgeless [C++ connection]",
    "description": "host.exp.exponent",
    "type": "node",
    "devtoolsFrontendUrl": "devtools://devtools/bundled/js_app.html?experiments=true&v8only=true&ws=192.168.18.153%3A8081%2Finspector%2Fdebug%3Fdevice%3D5ca976e985fc33959628c11becb91a7dc7842a8f%26page%3D1",
    "webSocketDebuggerUrl": "ws://192.168.18.153:8081/inspector/debug?device=5ca976e985fc33959628c11becb91a7dc7842a8f&page=1",
    "deviceName": "sdk_gphone64_arm64 - 15 - API 35",
    "reactNative": {
      "logicalDeviceId": "5ca976e985fc33959628c11becb91a7dc7842a8f",
      "capabilities": {
        "prefersFuseboxFrontend": true,
        "nativeSourceCodeFetching": false,
        "nativePageReloads": true
      }
    }
  }
]
```

### Solution
To solve the problem we introduce a new way of checking if the runtime belongs to the host application based on  functionality introduced in this [PR](https://github.com/expo/expo/pull/32322/files). 

### Additional changes
Additionally testing of expo go applications on android reveled that if application is killed, without closing the host process and then lunched again, the old runtime does not leave the `{metro}/json/list` endpoint. This scenario is unlikely as the user would have to do that by hand on the emulator, but to avoid problems with this bug in the future in case of expo go applications we search for the correct runtime starting form the end of the array.

### How Has This Been Tested: 

Run booth expo go and expo prebuilt bare application on both ios and android.


packages/vscode-extension/src/project/metro.ts
```diff
@@ -349,7 +425,20 @@ export class Metro implements Disposable {
...
-  private lookupWsAddressForNewDebugger(listJson: CDPTargetDescription[]) {
+  private async isExpoHostRuntime(webSocketDebuggerUrl: string) {
+    const HIDE_FROM_INSPECTOR_ENV = "globalThis.__expo_hide_from_inspector__";
+    try {
+      const result = await this.evaluateJsFromCdp(webSocketDebuggerUrl, HIDE_FROM_INSPECTOR_ENV);
+      if (result !== undefined) {
+        return true;
+      }
+    } catch (e) {
+      return false;
```
maciekstosio, line 436: It's redundant with the next one

packages/vscode-extension/src/project/metro.ts
```diff
@@ -292,6 +292,82 @@ export class Metro implements Disposable {
...
+          JSON.stringify({
+            id: REQUEST_ID,
+            method: "Runtime.evaluate",
+            params: { expression: source },
+          })
+        );
+      });
+
+      ws.on("error", (e) => {
+        Logger.debug(`[evaluateJsFromCdpAsync] Failed to connect ${wsUrl.toString()}`, e);
```
maciekstosio, line 329: Does this happen only when the connection fails? It's a bit misleading.

packages/vscode-extension/src/project/metro.ts
```diff
@@ -292,6 +292,82 @@ export class Metro implements Disposable {
     ws.close();
   }
 
+  private async evaluateJsFromCdp(
```
maciekstosio, line 295: I wonder if this is some generic method to call JS through CDP, shouldn't we put it in some helper instead of  as part of metro?

packages/vscode-extension/src/project/metro.ts
```diff
@@ -292,6 +292,82 @@ export class Metro implements Disposable {
...
+    webSocketDebuggerUrl: string,
+    source: string,
+    timeoutMs: number = 2000
+  ): Promise<string | undefined> {
+    const REQUEST_ID = 0;
```
maciekstosio, line 300: I wonder if there is a reason why part of the variables is declared in the promise and rest outside, is there any reason?

packages/vscode-extension/src/project/metro.ts
```diff
@@ -292,6 +292,82 @@ export class Metro implements Disposable {
...
+            if (response.error) {
+              reject(new Error(response.error.message));
+            } else if (response.result.result.type === "string") {
+              resolve(response.result.result.value);
+            } else {
+              resolve(undefined);
+            }
+            settled = true;
+            clearTimeout(timeoutHandle);
+            ws.close();
```
maciekstosio, line 359:
```
             settled = true;
             clearTimeout(timeoutHandle);
             ws.close();
 ```
 It repeated x times, let's extract it to some method

File: packages/vscode-extension/src/project/metro.ts [hunk 6]
```diff
@@ -292,6 +292,82 @@ export class Metro implements Disposable {
...
+    wsUrl.port = String(this._port);
+
+    return new Promise((resolve, reject) => {
+      let settled = false;
+      const ws = new WebSocket(wsUrl);
+
+      timeoutHandle = setTimeout(() => {
+        Logger.debug(`[evaluateJsFromCdpAsync] Request timeout from ${wsUrl.toString()}`);
+        reject(new Error("Request timeout"));
+        settled = true;
```
maciekstosio, line 314: That sounds redundant. Shouldn't you just close the websocket, and the `on('close')` with settled will decide if it's "settled" or not?

maciekstosio (changes requested):

packages/vscode-extension/src/project/metro.ts
filip131311: that was my initial approach but unfortunately we need our uniq metro port that is not passed through the metro endpoints.

packages/vscode-extension/src/project/metro.ts
filip131311: if there is no reason to be defined inside the promise as for example websocket connection needs to be opened inside i lest it outside and it seems a good pattern. (I moved timeoutHandle declaration inside the promise now as it was only an empty declaration)

packages/vscode-extension/src/project/metro.ts
filip131311: no if we did that we wouldn't know the close reason in on(close) and we want to log that

packages/vscode-extension/src/project/metro.ts
filip131311: we attach an error message with more details and it is true that we failed to connect

packages/vscode-extension/src/project/metro.ts
filip131311: thx nice catch

packages/vscode-extension/src/utilities/evaluateJsFromCdp.ts
```diff
@@ -0,0 +1,73 @@
...
+          } else if (response.result.result.type === "string") {
+            resolve(response.result.result.value);
+          } else {
+            resolve(undefined);
+          }
+          settleConnection();
+        }
+      } catch (e) {
+        reject(e);
+        settleConnection();
```
maciekstosio, line 69: You always settle the connection so it go to `finally` or just after try catch.

maciekstosio (approved):
One nitpick, the rest looks good, thanks for the changes!

packages/vscode-extension/src/project/metro.ts
```diff
@@ -362,17 +388,21 @@ export class Metro implements Disposable {
...
-        // When only one page is listed, it means that the app runtime hasn't registered yet in which case
-        // we return undefined which triggers a retry mechanism.
-        if (newDebuggerPages.length === 1) {
-          return undefined;
-        } else {
-          return newDebuggerPages[1].webSocketDebuggerUrl;
+        // Expo go apps using the new debugger could report more then one page,
+        // if it exist the first one being the Expo Go host runtime.
+        // more over expo go on android has a bug causing newDebuggerPages
+        // from previously run applications to leek if the host application
```
kmagiera, line 394: leak

packages/vscode-extension/src/project/metro.ts
```diff
@@ -349,7 +362,20 @@ export class Metro implements Disposable {
...
-  private lookupWsAddressForNewDebugger(listJson: CDPTargetDescription[]) {
+  private async isExpoHostRuntime(webSocketDebuggerUrl: string) {
+    const HIDE_FROM_INSPECTOR_ENV = "globalThis.__expo_hide_from_inspector__";
+    try {
+      const result = await this.evaluateJsFromCdp(webSocketDebuggerUrl, HIDE_FROM_INSPECTOR_ENV);
+      if (result !== undefined) {
+        return true;
+      }
+    } catch (e) {
+      // we ignore errors as they wer already logged by the evaluateJsFromCdp method
```
kmagiera, line 373: were

packages/vscode-extension/src/project/metro.ts
```diff
@@ -292,6 +293,18 @@ export class Metro implements Disposable {
     ws.close();
   }
 
+  private async evaluateJsFromCdp(
+    webSocketDebuggerUrl: string,
+    source: string,
+    timeoutMs: number = 2000
```
kmagiera, line 299: timeout is unused

packages/vscode-extension/src/project/metro.ts
```diff
@@ -292,6 +293,18 @@ export class Metro implements Disposable {
...
+  ): Promise<string | undefined> {
+    const wsUrl = new URL(webSocketDebuggerUrl);
+    wsUrl.hostname = "localhost";
+    wsUrl.port = String(this._port);
+
+    return evaluateJsFromCdp(wsUrl.toString(), source);
```
kmagiera, line 305: this function doesn't seem particularily useful and it calls a method with exact same name which is confusing. I'd just move the websocket URL resolution to `isExpoHostRuntime` where it is now called. That method is short enough to accommodate additional logic

packages/vscode-extension/src/project/metro.ts
```diff
@@ -362,17 +388,21 @@ export class Metro implements Disposable {
...
+        // if it exist the first one being the Expo Go host runtime.
+        // more over expo go on android has a bug causing newDebuggerPages
+        // from previously run applications to leek if the host application
+        // was not stopped.
+        // to solve both issues we check if the runtime is part of
+        // the host application process and select the last one that
+        // is not. To perform this check we use expo host functionality
+        // introduced in https://github.com/expo/expo/pull/32322/files
+        for (const newDebuggerPage of newDebuggerPages.reverse()) {
+          if (!(await this.isExpoHostRuntime(newDebuggerPage.webSocketDebuggerUrl))) {
```
kmagiera, line 401: I don't think we only want to eliminate non-host runtimes here. Given we already connect to the websocket here, this method can actually verify if the runtime is active. This specifically makes sense given you write the old runtime can still be listed.

Since `iseExpoHostRuntime` returns `false` both in the case when we fail to connect with the runtime, and also in the case when we connected sucesfully to the runtime but it didn't have that flag set, I think we should actually check for the first case only. In the current form I'd assume that we'd end up selecting old runtime when the new one hasn't yet appeared on the list.

packages/vscode-extension/src/utilities/evaluateJsFromCdp.ts
```diff
@@ -0,0 +1,73 @@
+import WebSocket from "ws";
+import { Logger } from "../Logger";
+
+export async function evaluateJsFromCdp(
```
kmagiera, line 4: this is a weird choice for the method name. We don't evaluate "from" CDP but rather "over" CDP

packages/vscode-extension/src/project/metro.ts
```diff
@@ -362,17 +395,21 @@ export class Metro implements Disposable {
...
-        // When only one page is listed, it means that the app runtime hasn't registered yet in which case
-        // we return undefined which triggers a retry mechanism.
-        if (newDebuggerPages.length === 1) {
-          return undefined;
-        } else {
-          return newDebuggerPages[1].webSocketDebuggerUrl;
+        // Expo go apps using the new debugger could report more then one page,
+        // if it exist the first one being the Expo Go host runtime.
+        // more over expo go on android has a bug causing newDebuggerPages
+        // from previously run applications to leek if the host application
```
maciekstosio, line 401:
```suggestion
        // from previously run applications to leak if the host application
```

maciekstosio (approved):

kmagiera (approved):

PR #852, opened 2024-12-13
Hide scrollbar in narrow IDE window with hidden device frame

p-malecki:
This PR addresses the issue of a scrollbar appearing when the IDE window is narrow and the device frame is hidden.
The scrollbar appeared because the `touch-area` for edge swipe gestures extends beyond the phone content width. To fix this, `overflow: hidden` is applied for windows narrower than 415px to override `overflow: auto` in `phone-wrapper`.

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/ed75fdb6-7090-4e6a-a13e-c8bd864df648" />|<video src="https://github.com/user-attachments/assets/c9ba00e3-1ad9-47ab-8182-f31a0d4de055" />|

### DIFF
--- a/packages/vscode-extension/src/webview/components/Preview.css
+++ b/packages/vscode-extension/src/webview/components/Preview.css
@@ -165,8 +165,8 @@
   transform: translate(0px, 21px);
 }
 
-@media (width <= 450px) {
-  .phone-content {
+@media (width <=415px) {
+  .phone-wrapper {
     overflow: hidden;
   }
 }
--- a/packages/vscode-extension/src/webview/components/Preview.css
+++ b/packages/vscode-extension/src/webview/components/Preview.css
@@ -164,3 +164,9 @@
 .button-group-left:has(.zoom-controls.select-open) {
   transform: translate(0px, 21px);
 }
+
+@media (width <= 450px) {
+  .phone-content {
+    overflow: hidden;
+  }
+}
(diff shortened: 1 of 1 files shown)


kmagiera (changes requested):
Why does this work ok when device frame is present and breaks with frameless mode?

I don't agree swipe gesture isn't practical. It is equally practical regardless of how narrow the screen is

p-malecki:
> Why does this work ok when device frame is present and breaks with frameless mode?
> 
> I don't agree swipe gesture isn't practical. It is equally practical regardless of how narrow the screen is


It works with visible device frame, because it covers `touch-area`. Without it we get extra pixels in each side which cause the scrollbar to appear.


Screenshot 2024-12-16 at 09 38 21Screenshot 2024-12-16 at 09 38 59

To keep the `touch-area` unrestricted by the `phone-content`, I moved the `overflow: hidden` to the `phone-wrapper`, and it works better.

<video src="https://github.com/user-attachments/assets/141a300a-5b83-499b-a00d-4e37726702ec" />

kmagiera:
ok, but why 415px is the breakpoint here?

p-malecki:
> ok, but why 415px is the breakpoint here?

In fact, setting `overflow: hidden` in the `phone-wrapper` class works as well and also prevents from possible bugs.

kmagiera (approved):

PR #853, opened 2024-12-13
Improve the reloading experience

filip131311:
This PR solves an issue with reload, clean rebuild and clear metro cache functionality when debugger was not attached. 
Because the above methods caused debug session to be disposed be triggered a session getter which throws error when debugger does not exist, breaking the flow of the reboot. To solve that issue we use `this.vscSession` instead. 

Additionally debugging this issue reveled that initial state of the `reload` functionality is trigering initially startup message "attaching debugger" because that is the last one that was set by initial startup process. To fix that we immediately set it to "restarting" similarly to how we did it in `restart` method. 

To avoid progress bar jumping back and forth we set the progress of "restarting" to 0.

### How Has This Been Tested: 
- add `return undefind` at the beginning of `lookupWsAddressForNewDebugger` method to simulate debugger not attaching and check restart functionality after fully booting IDE first.

### DIFF
--- a/packages/vscode-extension/src/debugging/DebugSession.ts
+++ b/packages/vscode-extension/src/debugging/DebugSession.ts
@@ -36,7 +36,13 @@ export class DebugSession implements Disposable {
   }
 
   public dispose() {
-    this.session && debug.stopDebugging(this.session);
+    let sessionExists = false;
+    try {
+      sessionExists = Boolean(this.session);
+    } catch {
+      // ignore errors as we are only interested if the session exists
+    }
+    sessionExists && debug.stopDebugging(this.session);
     this.debugEventsListener.dispose();
   }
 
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -36,7 +36,13 @@ export class DebugSession implements Disposable {
...
 
   public dispose() {
-    this.session && debug.stopDebugging(this.session);
+    let sessionExists = false;
+    try {
+      sessionExists = Boolean(this.session);
+    } catch {
+      // ignore errors as we are only interested if the session exists
+    }
+    sessionExists && debug.stopDebugging(this.session);
```
kmagiera, line 45: we should just use `this.vscSession` here this is class's internal method, so there's no reason to use public interface and then workaround side effects

kmagiera (approved):

PR #857, opened 2024-12-16
Add feedback modal

kacperkapusciak:
This PR implements feedback functionality by gathering `sentiment:positive`, `sentiment:negative` telemetry events.

https://github.com/user-attachments/assets/a0fb9e4a-1ae5-4224-a0e0-0ad9f90c9c0b

https://github.com/user-attachments/assets/739526e8-3885-4126-b953-b5e355a36ce0

### DIFF
--- a/packages/vscode-extension/src/webview/components/Feedback.tsx
+++ b/packages/vscode-extension/src/webview/components/Feedback.tsx
@@ -1,4 +1,4 @@
-import { MouseEvent } from "react";
+import { MouseEvent, useState } from "react";
 import "./Feedback.css";
 import classNames from "classnames";
 import { useUtils } from "../providers/UtilsProvider";
@@ -7,16 +7,26 @@ export type Sentiment = "positive" | "negative";
 
 export default function Feedback() {
   const { sendTelemetry } = useUtils();
+  const [sentiment, setSentiment] = useState<Sentiment | undefined>();
 
-  const handleFeedback = (event: MouseEvent<HTMLButtonElement>, sentiment: Sentiment) => {
+  const handleFeedback = (event: MouseEvent<HTMLButtonElement>, pickedSentiment: Sentiment) => {
     event.preventDefault();
-    sendTelemetry(`feedback:${sentiment}`);
+    sendTelemetry(`feedback:${pickedSentiment}`);
+    setSentiment(pickedSentiment);
   };
 
   return (
     <div className="feedback">
-      <FeedbackButton sentiment="positive" onClick={(e) => handleFeedback(e, "positive")} />
-      <FeedbackButton sentiment="negative" onClick={(e) => handleFeedback(e, "negative")} />
+      {Boolean(sentiment) ? (
+        <p className="feedback-prompt">
+          {sentiment === "positive" ? "What went well?" : "Tell us more"}
+        </p>
+      ) : (
+        <>
+          <FeedbackButton sentiment="positive" onClick={(e) => handleFeedback(e, "positive")} />
+          <FeedbackButton sentiment="negative" onClick={(e) => handleFeedback(e, "negative")} />
+        </>
+      )}
     </div>
   );
 }
--- a/packages/vscode-extension/src/webview/components/Feedback.tsx
+++ b/packages/vscode-extension/src/webview/components/Feedback.tsx
@@ -1,0 +1,39 @@
+import { MouseEvent } from "react";
+import "./Feedback.css";
+import { getTelemetryReporter } from "../../utilities/telemetry";
+
+function Feedback() {
+  return (
+    <div className="feedback">
+      <FeedbackButton sentiment="positive" />
+      <FeedbackButton sentiment="negative" />
+    </div>
+  );
+}
+
+type FeedbackButtonProps = {
+  sentiment: "positive" | "negative";
+};
+
+export function FeedbackButton({ sentiment }: FeedbackButtonProps) {
+  const handleFeedback = (event: MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+    getTelemetryReporter().sendTelemetryEvent(`feedback:${sentiment}`);
+  };
+
+  if (sentiment === "positive") {
+    return (
+      <button className="feedback-button feedback-button-positive" onClick={handleFeedback}>
+        <span className="codicon codicon-thumbsup" />
+      </button>
+    );
+  }
+
+  return (
+    <button className="feedback-button feedback-button-negative" onClick={handleFeedback}>
+      <span className="codicon codicon-thumbsdown" />
+    </button>
+  );
+}
+
+export default Feedback;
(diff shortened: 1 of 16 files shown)


kmagiera:
I'd perhaps add some note that when sending the form we won't include any details about the project, user's setup or the logs

kmagiera:
getting the logs is not as straightforward, also this would fit better in bug/issue reporting flow rather than feedback, so I'd leave it out. Just wanted to make sure it is clear the feedback form doesn't include any potentially sensitive info

kacperkapusciak:
@kmagiera added in [3a4d1ab](https://github.com/software-mansion/radon-ide/pull/857/commits/3a4d1ab99d22da7821fdae98c11434f8700666d3)

kacperkapusciak:
@maciekstosio fixed in [0947f14](https://github.com/software-mansion/radon-ide/pull/857/commits/0947f14a80fff6c82dd2b6b67b3bbfce7753708c)

kmagiera (commented):
Since this uses telemetry reporter it may be that some users have it disabled in which case we won't receive their reports. We can perhaps setup it in a way that we ignore telemetry settings but don't think this is the right way to do this.

kacperkapusciak:
alternatively i can gather the feedback in our customer portal database

kmagiera:
I guess this should be fine for now. One simple workaround we can add is to check if user have telemetry on and only then show this button

maciekstosio (changes requested):
I assume this should stay selected 

<video src="https://github.com/user-attachments/assets/d7531d73-3b4a-4ba8-a83a-2a4bc8c4bdbd" />

packages/vscode-extension/src/webview/providers/UtilsProvider.tsx
```diff
@@ -17,12 +17,30 @@ export const Platform = {
...
 const utils = makeProxy<Utils>("Utils");
 
-const UtilsContext = createContext<UtilsInterface>(utils);
+const UtilsContext = createContext<UtilsContextProps>({ utils, telemetryEnabled: false });
 
 export function UtilsProvider({ children }: PropsWithChildren) {
-  return <UtilsContext.Provider value={utils}>{children}</UtilsContext.Provider>;
+  const [telemetryEnabled, setTelemetryEnabled] = useState(false);
+
+  useEffect(() => {
```
maciekstosio, line 32: I think it should be in a separate hook, as a change to the state will cause unnecessary rerenders in all useUtils occurrences

maciekstosio: Something like this will be better imo. 

```
const useTelemntry = () => {
  const utils = useUtils();
  const [telemetryEnabled, setTelemetryEnabled] = useState(false);

  useEffect(() => {
    utils.isTelemetryEnabled().then(setTelemetryEnabled);
    utils.addListener("telemetryEnabledChanged", setTelemetryEnabled);

    return () => {
      utils.removeListener("telemetryEnabledChanged", setTelemetryEnabled);
    };
  }, []);

  return telemetryEnabled;
}
```

maciekstosio (changes requested):

packages/vscode-extension/src/webview/providers/UtilsProvider.tsx
```diff
@@ -17,9 +17,11 @@ export const Platform = {
   },
 };
 
+type UtilsContextProps = UtilsInterface;
+
 const utils = makeProxy<Utils>("Utils");
 
-const UtilsContext = createContext<UtilsInterface>(utils);
+const UtilsContext = createContext<UtilsContextProps>(utils);
```
maciekstosio, line 24: Why type rename?

packages/vscode-extension/src/webview/providers/UtilsProvider.tsx
```diff
@@ -34,6 +36,36 @@ export function useUtils() {
...
+    utils.isTelemetryEnabled().then(setTelemetryEnabled);
+    utils.addListener("telemetryEnabledChanged", setTelemetryEnabled);
+
+    return () => {
+      utils.removeListener("telemetryEnabledChanged", setTelemetryEnabled);
+    };
+  }, []);
+
+  return (
+    <TelemetryContext.Provider value={{ telemetryEnabled }}>{children}</TelemetryContext.Provider>
```
maciekstosio, line 56: `{telemetryEnabled}` is useMemo -> if the value doesn't change, we'll rerender descend components anyway, cause the object will be recreated.

packages/vscode-extension/src/webview/providers/UtilsProvider.tsx
```diff
@@ -34,6 +36,36 @@ export function useUtils() {
...
+  );
+}
+
+export function useTelemetry() {
+  const context = useContext(TelemetryContext);
+
+  if (context === undefined) {
+    throw new Error("useUtils must be used within a TelemetryContextProvider");
```
maciekstosio, line 64:
```suggestion
    throw new Error("useTelemetry must be used within a TelemetryContextProvider");
```

packages/vscode-extension/src/webview/components/Feedback.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { useUtils } from "../providers/UtilsProvider";
+import { Sentiment } from "./SendFeedbackItem";
+
+type FeedbackProps = {
+  sentiment: Sentiment | undefined;
+  setSentiment: Dispatch<SetStateAction<Sentiment | undefined>>;
+};
+
+export function Feedback({ sentiment, setSentiment }: FeedbackProps) {
+  const { sendTelemetry } = useUtils();
```
maciekstosio, line 13: If we have a separate telemetry hook, it sounds like `sendTelemetry` should be there

packages/vscode-extension/src/webview/providers/UtilsProvider.tsx
```diff
@@ -34,6 +36,36 @@ export function useUtils() {
   return context;
 }
 
+type TelemetryContextProps = { telemetryEnabled: boolean };
+
+const TelemetryContext = createContext<TelemetryContextProps>({ telemetryEnabled: false });
+
+export function TelemetryProvider({ children }: PropsWithChildren) {
```
maciekstosio, line 43: IMO, it should be in a separate file. I am not sure why we need to pass utils through the context, but if we cannot just import it in a separate file, you can always use `useUtils` in `TelemetryProvider`. Just remember to order providers properly (imo this is why it would be better to have it just in hook instead of context, especially since we don't use it anywhere else).

maciekstosio (changes requested):

packages/vscode-extension/src/webview/providers/TelemetryProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+    utils.isTelemetryEnabled().then(setTelemetryEnabled);
+    utils.addListener("telemetryEnabledChanged", setTelemetryEnabled);
+
+    return () => {
+      utils.removeListener("telemetryEnabledChanged", setTelemetryEnabled);
+    };
+  }, []);
+
+  return (
+    <TelemetryContext.Provider value={{ telemetryEnabled }}>{children}</TelemetryContext.Provider>
```
maciekstosio, line 22: return just `telemetryEnabled` or wrap with `useMemo`

maciekstosio (approved):

PR #858, opened 2024-12-16
Improve landing performance

kacperkapusciak:
This PR:
- makes sure to only use one H1 tag on the page
- shrinks the hero image from `371 KB` to `77.8 KB` and sets the `width` and `height` attributes to prevent layout shift
- compresses videos used on landing by `50%`
- lazy loads videos using intersection observer – videos will only be loaded when in view
  - during video loading a lightweight placeholder is shown to prevent layout shifts 
  - this is important because we used to load 10MB of video on page load even for users which never scrolled down the page
  
 See before:
image


packages/docs/src/components/Sections/Overview/OverviewItem/index.tsx
```diff
@@ -18,19 +19,25 @@ const OverviewItem = ({ label, title, body, mediaSrc }: Props) => {
...
           href="https://marketplace.visualstudio.com/items?itemName=swmansion.react-native-ide"
           onClick={handleButtonClick}
         />
       </section>
-      <div className={styles.media}>
-        <video autoPlay loop muted playsInline>
-          <source src={mediaSrc} type="video/mp4" />
-        </video>
-      </div>
+      <InView triggerOnce>
```
maciekstosio, line 30: Could you add to the PR description how this helps? Some comparison, so in the future we remember why another package actually helped with performance which might be counterintuitive :D

maciekstosio (approved):
One small think, rest looks ok

packages/docs/src/components/Sections/Overview/OverviewItem/index.tsx
kacperkapusciak: sure

 i'm using intersection observer to load videos only when they are in view. it doesn't make sense to load 10MB of videos for every user even when they never scroll down the page
 
I show a lightweight (~7kb) placeholder image during the video loading

PR #859, opened 2024-12-18
Fix vscode-js-debug submodule initialization

filip131311:
This PR fixes an issue that occurred on fresh clones of the repository. In  #846 we added a new submodule `vscode-js-debug` along with preinstall actions that should build it. This PR adds initialization to that flow so the command works on fresh clones as well.  

### How Has This Been Tested: 

run:
- `git clone git@github.com:software-mansion/radon-ide.git` 
- `cd packages/vscode-extension`
- `npm i` 

everything should work


maciekstosio (approved):
Thanks!

Issue #729, opened 2024-11-14
Allow changing primary background Color of Radon


pzoltowski:
**Description**
It would be great that either radon allow to change primary background color or used the one already set but `sideBar.background`. If someone has completely different dark theme current Radon background feels very off 

<img width="533" alt="Screenshot 2024-11-14 at 17 00 36" src="https://github.com/user-attachments/assets/8a664394-cc0a-4045-9612-819da572e016">

**Environment**

- 0.0.22 inside cursor IDE (Version: 0.42.4, VSCode Version: 1.93.1)
- MacOS 15.1 (24B83)

p-malecki:
Hi, thank you for suggesting this enhancement. We plan to implement it in the future, although it is not currently a priority for us.

PR #866, opened 2024-12-19
Remove background color

piaskowyk:
Hello IDE team 👋 

This PR simply removes the background color of the IDE panel and uses the default one provided by the selected theme. This approach feels more natural - in my opinion. Here is a comparison video:

**Before**
<video src="https://github.com/user-attachments/assets/5970f39a-995b-4c78-852b-b6b36325085c" />

**After**
<video src="https://github.com/user-attachments/assets/7382a08d-8f93-4bcc-a45d-d5bb5b3ce666" />

This is just a suggestion. Feel free to close this PR if you don't like the changes - no hard feelings 😄

### DIFF
--- a/packages/vscode-extension/src/webview/styles/theme.css
+++ b/packages/vscode-extension/src/webview/styles/theme.css
@@ -140,7 +140,6 @@
 body[data-vscode-theme-kind="vscode-light"],
 body[data-vscode-theme-kind="vscode-high-contrast-light"] {
   --swm-active-item: var(--blue-dark-100);
-  --swm-preview-background: var(--white);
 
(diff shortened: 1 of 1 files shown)


kacperkapusciak:
It's a compromise I'm not too happy with.

I agree that when we remove the background the extension blends better with the vscode themes but the buttons, the modals, the dropdowns randomly do not match the rest of the extension. Without the background the whole product feels _incoherent_, _unpolished_, and just _off_.

For now, I'd rather have a consistent yet incomplete product (missing themes). I could spend some time to implement this properly in a future release. 

Related to https://github.com/software-mansion/radon-ide/issues/729

PR #870, opened 2024-12-20
Refactor open preview logic

### LINKED ISSUE #278: Cannot read properties of undefined (reading 'project')
**Description**
When I am launching the IDE window I receive the following error toast in vscode "Cannot read properties of undefined (reading 'project')" and also my log console and breakpoints do not work, I don't see anything attached to the console and breakpoints just do not work. I assume this is why, because of that error.

**Environment**
- v0.0.10-beta
- "react-native": "0.72.6" with expo ejected

maciekstosio:
Fixes #278

This PR refactors logic related to opening previews. We were using `showIDEPanel` for `RNIDE.showPanel`/`RNIDE.openPanel` commands, the launch preview depended on the existence of the `fileName` and `lineNumber`. The problem is that VSCode passes its arguments there, but instead number in lineNumber we get an object with the panel's id, which triggers this logic always. Thus, I refactored the opening preview to use a separate command. 

Additionally, startPreview was running even if the project wasn't set up yet, which led to an error that we used `.project` on `null`. Now, we open the panel and show a warning in that case. 

### How Has This Been Tested: 

Repeat those scenarios for preview both in `tab panel` and `side panel`: 

**Steps:**
- Close the IDE panel
- Reopen (or restart if in debug mode) VSCode 
- Open IDE panel from top right corner 

**Expected:**
The panel opens without error dialog. 

**Steps:**
- Close the IDE panel
- Reopen (or restart if in debug mode) VSCode 
- Click "Open preview" 

**Expected:**
Confirm that panel opens and there is warning message 

**Steps:**
- Make sure IDE Panel is open
- Click "Open preview" 

**Expected:**
Confirm that preview started

_Tested on react-native-76_

### DIFF
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -78,7 +78,7 @@ export async function activate(context: ExtensionContext) {
 
   commands.executeCommand("setContext", "RNIDE.sidePanelIsClosed", false);
 
-  async function showIDEPanel(fileName?: string, lineNumber?: number) {
+  async function showIDEPanel() {
     await commands.executeCommand("setContext", "RNIDE.sidePanelIsClosed", false);
 
     const panelLocation = workspace
@@ -86,9 +86,9 @@ export async function activate(context: ExtensionContext) {
       .get<PanelLocation>("panelLocation");
 
     if (panelLocation !== "tab") {
-      SidePanelViewProvider.showView(context, fileName, lineNumber);
+      SidePanelViewProvider.showView();
     } else {
-      TabPanel.render(context, fileName, lineNumber);
+      TabPanel.render(context);
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/extension.ts
```diff
@@ -78,17 +78,17 @@ export async function activate(context: ExtensionContext) {
...
-  async function showIDEPanel(fileName?: string, lineNumber?: number) {
+  async function showIDEPanel() {
     await commands.executeCommand("setContext", "RNIDE.sidePanelIsClosed", false);
 
     const panelLocation = workspace
       .getConfiguration("RadonIDE")
       .get<PanelLocation>("panelLocation");
 
     if (panelLocation !== "tab") {
-      SidePanelViewProvider.showView(context, fileName, lineNumber);
```
filip131311, line 89: Why is context not necessery in Side panel while it remains necessary in tabpanel?

filip131311 (commented):
I wonder if we shouldn't  update package.json as well? 
als left an inline question

packages/vscode-extension/src/extension.ts
maciekstosio: `showView` doesn't use it at all, so I removed it 💁‍♂️ 

From what I quickly went through, SidePanelViewProvider gets context in the constructor and, being a static method, fails gracefully when it isn't provided. I didn't go deeper, as I assume this is outside the scope of this PR.

filip131311: Fair enough

filip131311 (approved):
LGTM

PR #872, opened 2024-12-20
Improve performance with useMemo

filip131311:
this PR prevents unnecessary rerenders caused by setting states inside objects therefore triggering rerender even f the value didn't change.

### How Has This Been Tested: 

Run test application and see if it works as expected.


kmagiera (approved):
🤷

PR #873, opened 2024-12-20
Fix UrlSelect trigger value reset for non expo router projects

p-malecki:
This PR changes the UrlBar behaviour in projects not using Expo Router navigation.
When the user intends to close the preview by pressing the goHome button (or reload), the UrlSelect's trigger value needs to be reset to a placeholder ("/"). 

Additionally, this update disables the `goBack` button for projects that do not use Expo Router.

### How Has This Been Tested: 
- Tested on expo-router and react-native-76 apps.

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/3fa5b50a-c5c3-4f70-b184-f7a1853f101b" />|<video src="https://github.com/user-attachments/assets/051577d3-5f43-454c-92b8-eefdd1d37426" />|

### DIFF
--- a/packages/vscode-extension/src/webview/components/UrlBar.tsx
+++ b/packages/vscode-extension/src/webview/components/UrlBar.tsx
@@ -3,6 +3,7 @@ import { useProject } from "../providers/ProjectProvider";
 import UrlSelect, { UrlItem } from "./UrlSelect";
 import { IconButtonWithOptions } from "./IconButtonWithOptions";
 import IconButton from "./shared/IconButton";
+import { useDependencies } from "../providers/DependenciesProvider";
 
 function ReloadButton({ disabled }: { disabled: boolean }) {
   const { project } = useProject();
@@ -29,6 +30,7 @@ function ReloadButton({ disabled }: { disabled: boolean }) {
 
 function UrlBar({ disabled }: { disabled?: boolean }) {
   const { project } = useProject();
+  const { dependencies } = useDependencies();
 
   const MAX_URL_HISTORY_SIZE = 20;
   const MAX_RECENT_URL_SIZE = 5;
@@ -79,6 +81,8 @@ function UrlBar({ disabled }: { disabled?: boolean }) {
     return [...urlList].sort((a, b) => a.name.localeCompare(b.name));
   }, [urlList]);
 
+  const isExpoRouterProject = !dependencies.expoRouter?.isOptional;
+
   return (
     <>
       <IconButton
@@ -86,7 +90,7 @@ function UrlBar({ disabled }: { disabled?: boolean }) {
           label: "Go back",
           side: "bottom",
         }}
-        disabled={disabled || urlHistory.length < 2}
+        disabled={disabled || !isExpoRouterProject || urlHistory.length < 2}
         onClick={() => {
           setUrlHistory((prevUrlHistory) => {
             const newUrlHistory = prevUrlHistory.slice(1);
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -15,14 +15,15 @@ const SelectItem = React.forwardRef<HTMLDivElement, PropsWithChildren<Select.Sel
...
+  key: number;
   value: string;
   onValueChange: (newValue: string) => void;
   recentItems: UrlItem[];
   items: UrlItem[];
   disabled?: boolean;
 }
 
-function UrlSelect({ onValueChange, recentItems, items, value, disabled }: UrlSelectProps) {
+function UrlSelect({ onValueChange, recentItems, items, key, value, disabled }: UrlSelectProps) {
```
kmagiera, line 26: I believe that it only has been added in React 19 that you can receive key prop like this. I don't think in the current version in propagates down to the component

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -35,7 +36,7 @@ function UrlSelect({ onValueChange, recentItems, items, value, disabled }: UrlSe
   };
 
   return (
-    <Select.Root onValueChange={handleValueChange} value={value} disabled={disabled}>
+    <Select.Root onValueChange={handleValueChange} key={key} value={value} disabled={disabled}>
```
kmagiera, line 39: It should be enough to do something along these lines, instead of having the whole logic for urlSelectKey in the parent component:

```tsx
<Select.root key={value ? "has-value : "no-value"}>
```

packages/vscode-extension/src/webview/components/UrlBar.tsx
```diff
@@ -117,8 +130,9 @@ function UrlBar({ disabled }: { disabled?: boolean }) {
         }}
         recentItems={recentUrlList}
         items={sortedUrlList}
-        value={urlList[0]?.id}
-        disabled={disabledAlsoWhenStarting || urlList.length < 2}
+        key={urlSelectKey}
```
kmagiera, line 133: setting this here is enough to re-mount the whole component. It doesn't have to be accessed in the underlying component

kmagiera (changes requested):

p-malecki:
I’ve removed the urlSelectKey state and key prop from UrlSelect, which were added in a previous PR version. Setting setUrlSelectValue to an empty string is sufficient to reset the trigger value of UrlSelect to a placeholder.

packages/vscode-extension/src/webview/components/UrlBar.tsx
```diff
@@ -103,6 +109,9 @@ function UrlBar({ disabled }: { disabled?: boolean }) {
       <IconButton
         onClick={() => {
           project.goHome("/{}");
+          if (!isExpoRouterProject) {
```
filip131311, line 112: I don't like this being placed as a dependency of a user action as we'll have to remember to update it every time we add a new navigation option I would much rather have wrapper.js send an event every time application is loaded and then some global logic would handle setting the correct navigation state to the url bar. 

Having said that I think we should move most of the logic in this file to the extension code to allow passing state between diffrent panel locations any wayand my suggestion will be easier to  active then.

filip131311 (approved):
I approve this changes adding some inline comments to adres #887.

PR #874, opened 2024-12-21
Add themes support 

kewinzaq1:
This pull request adds support for vscode themes
Connected (https://github.com/software-mansion/radon-ide/issues/729#issue-2658109777 , https://github.com/software-mansion/radon-ide/pull/866)

The changes add new theme based on vscode theme variables. As being aware that some themes may decrease readability, this PR adds option to use built-in themes. Built-in themes stay mostly not changed, only few additional variables that allow better customization with the new changes. 

https://github.com/user-attachments/assets/b0bb349e-5a96-4daf-888e-4187f4958ecd

https://github.com/user-attachments/assets/e8a0a7a6-5b3c-4a4c-905c-d32194cf79eb

https://github.com/user-attachments/assets/2601e4e2-db56-40ed-ace3-b7705feb50a2

https://github.com/user-attachments/assets/ec935a4a-addc-434b-8824-16f876d2e419


filip131311 (changes requested):
Hello, thank you for your time. 
In general I really like the changes unfortunately there are some parts of the application that lose on readability after the changes, for example lunchConfiguration:
Screenshot 2024-12-23 at 10 08 55
 
Vscode has styled inputs: 
Screenshot 2024-12-23 at 10 11 28
Screenshot 2024-12-23 at 10 11 35

so i believe you can find the styles for them 

I also left one inline question 

Thank u again and merry Christmas

kewinzaq1:
> Hello, thank you for your time. In general I really like the changes unfortunately there are some parts of the application that lose on readability after the changes, for example lunchConfiguration: Screenshot 2024-12-23 at 10 08 55
> 
> Vscode has styled inputs: Screenshot 2024-12-23 at 10 11 28 Screenshot 2024-12-23 at 10 11 35
> 
> so i believe you can find the styles for them
> 
> I also left one inline question
> 
> Thank u again and merry Christmas

Thanks for the review merry Christmas 🎄

kmagiera:
I'd also be in favor of supporting themes properly. The reason we decided to shift away from this approach initially was that it was difficult to control some of the theme parameters under different vscode settings. This resulted in a number of elements being difficult to use as they'd blend inproperly into elements in the background for example. 

One way this can be solved is be giving an option whether the IDE should use the editor theme or the built-in one. I wonder how difficult would it be to incorporate that while still relying mostly on css variables.

kewinzaq1:
> One way this can be solved is be giving an option whether the IDE should use the editor theme or the built-in one. I wonder how difficult would it be to incorporate that while still relying mostly on css variables.

@kmagiera I think that it should go quite smoothly. I'll return with an example of it. One thing that is striking in this approach is that it would hurt UX a bit, forcing users to manually switch to the default theme. You know, I mean the experience that something does not look well and then someone would report this because overlooked this option in workspace config or so. But maybe I exaggerate.

kmagiera:
I agree such a setting would be difficult to discover. Perhaps we should default to the editor theme and only expect the setting to be used as a fallback when the UI doesn't look good in certain theme

kewinzaq1:
> This looks good. Thank you. Anything you're planning to change as this PR is still marked as draft? We won't be merging it for the upcoming release (this week), but may want to include it for the following one

@kmagiera great, I'm not planning to change anything for now, so I think we can remove the draft label

filip131311 (approved):
This is awesome, thank you so much for your contribution! 

We will include this in next release.

PR #875, opened 2024-12-23
Refactor screen recording and add key bindings

### LINKED ISSUE #868: Integrate Screenshot support

filip131311:
This PR aims to add screenshot functionality, keybindings for capturing replays, starting an capturing the recording and capturing screenshots. 

To achieve that we moved the logic for starting and ending recording from webview to the extension code. 

the proposed keybindings are:
- capture replay: shift+cmd+R
- toggle recording: shift+cmd+E
- capture screenshot: shift+cmd+A

fix #868

### How Has This Been Tested: 

- Run extension and check if commands and previously existing buttons work as expected 
- change `MAX_RECORDING_TIME_SEC` to 10 and check if the logic for closing stoping works 
- check screenshot buton and its effects

https://github.com/user-attachments/assets/540fe2ae-90d6-4f76-983b-98aa38e482da


packages/vscode-extension/package.json
```diff
@@ -74,6 +74,30 @@
...
+      },
+      {
+        "command": "RNIDE.startRecording",
+        "title": "Start Recording",
+        "category": "Radon IDE",
+        "enablement": "RNIDE.extensionIsActive && RNIDE.panelIsOpen"
+      },
+      {
+        "command": "RNIDE.captureRecording",
+        "title": "Capture Recording",
```
maciekstosio, line 92: I wonder if End recording wouldn't be more readable. Or maybe just one command for toggle recording?

packages/vscode-extension/package.json
```diff
@@ -97,6 +121,26 @@
         "command": "RNIDE.performFailedBiometricAuthorization",
         "key": "ctrl+shift+N",
         "mac": "cmd+shift+N"
+      },
+      {
+        "command": "RNIDE.captureReplay",
```
maciekstosio, line 126: I don't like those shortcuts. Usually, the keys used are somehow related to what they do, and I think shift is used for something special for given base shortcut `cmd+z` -> undo, `cmd+shift+z` -> redo. All that we use doesn't seem to have any meaning, or I don't see one.
I was thinking about something like:

ctrl + b/a -> biometric/authorization
ctrl + shift + b/a -> perform failed biometric/authorization
ctrl + r -> toggle recording 
ctrl + shift + r -> show replay 
ctrl + s -> screen shot 

wdyt?

packages/vscode-extension/src/common/Project.ts
```diff
@@ -158,8 +160,9 @@ export interface ProjectInterface {
   openDeepLink(link: string): Promise<void>;
 
   startRecording(): void;
-  captureAndStopRecording(): Promise<RecordingData>;
-  captureReplay(): Promise<RecordingData>;
+  captureAndStopRecording(): void;
```
maciekstosio, line 163: just endRecording?

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -32,27 +32,27 @@ export class Preview implements Disposable {
     stdin.write(command);
   }
 
-  private saveVideoWithID(videoId: string): Promise<RecordingData> {
+  private saveMultimediaWithID(type: "video" | "screenshot", id: string): Promise<MultimediaData> {
```
maciekstosio, line 35: Could we have some enum for `type`?
This part:
```
 public captureReplay() {


     return this.saveMultimediaWithID("video", "replay");
   }

   public captureScreenShot() {
     return this.saveMultimediaWithID("screenshot", "screenshot");
   }
```

would become cleaner imo.

packages/vscode-extension/src/project/project.ts
```diff
@@ -185,28 +193,60 @@ export class Project
       throw new Error("No device session available");
     }
     this.deviceSession.startRecording();
+    this.eventEmitter.emit("isRecordingChanged", true);
+
+    this.recordingInterval = setInterval(() => {
```
maciekstosio, line 198: Why not just `setTimeout`?

packages/vscode-extension/src/project/project.ts
```diff
@@ -185,28 +193,60 @@ export class Project
       throw new Error("No device session available");
     }
     this.deviceSession.startRecording();
+    this.eventEmitter.emit("isRecordingChanged", true);
```
maciekstosio, line 196: This event name is a bit confusing for me, maybe we can come up with something better?

maciekstosio: isRecording? RECORDING_STATE_CHAGNED?

maciekstosio (changes requested):

packages/vscode-extension/package.json
filip131311: unfortunately api you proposed clashes with existing keybindings (file save/ resent workspaces) 

I refactored my api to look like this 
cmd+shift+r capture replay
cmd+shift+e toggle recording 
cmd+shift+a capture screenshot

maciekstosio (approved):

PR #878, opened 2024-12-31
Support for launching expo dev plugin tools in separate panels within VSCode

kmagiera:
This PR adds support for redux and react query dev tools expo plugins and makes their UI display as separate panel in VSCode.

Apart from the support for the listed plugins, this change builds an infrastructure for supporting more expo dev plugins and other types of plugins in the future.

Key changes that the PR introduces:
1. We are expanding panels configuration in package.json. Apparently there's no way to define panels dynamically and we need one panel for every dev tool such that they can be run side-by-side when necessary. To avoid having too many panels, every panel is hidden behind its own context variable that we can dynamically set when we detect the user requested to open this panel
2. We introduce a change to babel transformer that allows us to capture the use of expo-dev-plugins we currently support. This way we can detect which plugins the user has installed and only show relevant tools for the user to open.
3. For the above to work correctly, we are adding a new type of event that is passed via the devtools channel (same that we use for events such as "app ready"). We use this event to notify about the plugins that the app uses.
4. We're adding a UI component to the top menu bar with a "tools" dropdown. The dropdown displays the list of available tools and a checkbox next to every tool allowing the user to enable each tool separately. When tools is enabled, the new panel displaying the tool UI will launch.
5. Finally, we're implementing a new "ToolPlugin" interface and its implementation that coordinates all the expo-dev-plugin specific needs. The interface allows for tools to be listed as available and implements a persistent storage for saving which tools are enabled. The expo-dev-plugin implementation of the tool manages the registration of webview panels, listening to events about the plugins being used by the app, and controls the context variables that shows/hide the panel.

As a follow up to this change, we will be adding a docs update that explains how tools can be used. Currently, we only support epxo-dev-plugin tools, so for the whole setup to work corrently the user needs to use Expo and have the specific tools configured according to the documentation here: https://github.com/expo/dev-plugins

### How Has This Been Tested: 
1. I used a sample Expo project and added redux and react query dev tools following the docs here: https://github.com/expo/dev-plugins
2. I tested enabling/disabling each plugin separately. When plugin is enabled we expect the panel to open immediately.
3. I tested resetting metro cache which forces metro to reset at which point it stars at a new port. This is important as the plugin UI is hosted via metro. In this scenario is it expected for the panel to close while the app is reloading.


filip131311 (commented):
I left an inline comment

packages/vscode-extension/src/plugins/expo-dev-plugins/ExpoDevPluginWebviewProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ExpoDevPluginToolName } from "./expo-dev-plugins";
+import { Logger } from "../../Logger";
+
+function generateWebviewContent(pluginName: ExpoDevPluginToolName, metroPort: number): string {
+  const iframeURL = `http://localhost:${metroPort}/_expo/plugins/${pluginName}`;
+  return /*html*/ `
+    <!DOCTYPE html>
+    <html style="height: 100%;">
+      <head/>
+      <body style="height: 100%;">
```
filip131311, line 19:
```suggestion
      <body style="height: 100%; overflow:hidden">
```

this will remove the scroll bar next to the panel.

packages/vscode-extension/src/project/project.ts
```diff
@@ -458,8 +463,11 @@ export class Project
     if (restart) {
       const oldDevtools = this.devtools;
       const oldMetro = this.metro;
+      const oldToolsManager = this.toolsManager;
       this.devtools = new Devtools();
       this.metro = new Metro(this.devtools, this);
+      this.toolsManager = new ToolsManager(this.devtools, this.eventEmitter);
+      oldToolsManager.dispose();
```
filip131311, line 470: I'm not sure if removing the old tools manager and creating a new one is what we should do here: 
this action causes the plug in panel to disappear completely for the time of the restart which if the panel was open looks bad for the user, I think we should pass an information to the tools manager about restart and while plug in is not available display some kind of loading in the panel. We should also add a functionality to the tools manager that allows us to update any information it needs after restart (like metro port)

packages/vscode-extension/package.json
```diff
@@ -175,7 +192,31 @@
...
-          "when": "config.RadonIDE.panelLocation != 'tab' && !RNIDE.sidePanelIsClosed"
+          "when": "RNIDE.extensionIsActive && config.RadonIDE.panelLocation != 'tab' && !RNIDE.sidePanelIsClosed"
+        }
+      ],
+      "RadonIDEToolExpoDevPluginReactQuery": [
+        {
+          "type": "webview",
+          "id": "RNIDE.Tool.ExpoDevPlugin.ReactQuery.view",
+          "name": "",
+          "when": "RNIDE.Tool.ExpoDevPlugin.ReactQuery.available"
```
filip131311, line 203:
```suggestion
          "when": "RNIDE.Tool.ExpoDevPlugin.ReactQuery.available && RNIDE.panelIsOpen"
```

I think we should hide plug in panels when no app is running

filip131311 (changes requested):
I left some inline comments

maciekstosio (approved):
Looks good

packages/vscode-extension/src/project/project.ts
kmagiera: yes, I know this isn't ideal but this is apparently the best we can do without complicating this immensely. The panel will end up loosing its state which is outside of our control here. This is because after restart we will need to connect to the new instance of metro. Ultimately, I'd rather keep it this way for now.

Ultimately, we can add more integrated tools that could work without installing additional plugin modules which I think is going to be more convenient. In that case we will also have better flexibility as to how restarts can be controlled and when to retain the tool state.

PR #882, opened 2025-01-07
Introduce IDE class that manages all necessary class instances

kmagiera:
This PR restructures the logic for creating and accessing some core extension components like Project, DeviceManager, etc. 

Before they were instantiated in webview panel or tabview controller. In this PR we are introducing a new top level class IDE that controls the lifecycle of all the listed core components and hooks them up properly as some instances require other instances to exists.

The main reason for this change is to lay a groundwork for #878 where we are introducing additional webview panels to be used to display different tools. The panels require separate webview controller instances that also need to access some details about project setup that are currently only kept in the main panel webview controller.

As a side effect of this change, we no longer re-create project when switching between UI location for the Radon IDE panel (between tab and side-panel). Therefore this change makes it more seamless for the users to play with the panel location as updating that no longer restarts the project.

For the above to work corrently, we add additional logic for handling `dispose` calls. When the position of the panel is updated, we first get the original panel disposed, and only then the new panel is created. When the main one is disposed, we don't know whether that would be because of the position change or due to the user closing the IDE panel by hand. In particular, it is possible for user to change the location w/o using the provided menu options, so there's no indication that we can have about the fact that the panel location is being updated. In order to handle that properly, the IDE class has static `attach` method used by the webview panels and also `detach` method that the panel uses when they are being disposed. The IDE panel doesn't get disposed immediately when detached, but it waits for 1 second and if it isn't attached to a new panel after that time, only then it calls dispose.

### How Has This Been Tested: 
1. Open app in IDE
2. Change panel location between tab/side bar/ new window – notice that the app doesn't restart when updating the location.
3. Make sure the IDE is functional after the location is changed.
4. Close IDE panel (i.e. just close the tab when panel is displayed as editor tab) and make sure the IDE is disposed (we print a debug log when that happens).


maciekstosio (dismissed):
Changing panel location seems to work as expected, but running preview when in side panel causes following error: 

image

I assume in `src/panels/SidepanelViewProvider.ts`:

```
      SidePanelViewProvider.currentProvider.webviewController.project.startPreview(
        `preview:/${fileName}:${lineNumber}`
      );
```

should be replaced with: 

```
IDE.getOrCreateInstance(context).project.startPreview(`preview:/${fileName}:${lineNumber}`);
```

packages/vscode-extension/src/project/ide.ts
```diff
@@ -0,0 +1,56 @@
...
+      this.dependencyManager,
+      this.project,
+      this.workspaceConfigController,
+      this.launchConfig
+    );
+    // register disposable with context
+    context.subscriptions.push(this);
+  }
+
+  dispose() {
```
filip131311, line 40: This function is called only when the vscode is closed and is not called then user closes the tab with te extension or clicks on the "close IDE" button in side panel.  This is a serious change of behavior (as now emulator is not killed when the ide panel is closed) and one I don't think we should introduce. 

One possible solution would be to call IDE.instance.dispose() in scenarios when IDE is closed by the user.

filip131311 (changes requested):
I left inline comment

packages/vscode-extension/src/project/ide.ts
kmagiera: I agree, noticed that earlier and just added some changes to handle disposing properly. It is quite tricky as when someone changes the panel location we first get the original panel closed and only after hat the new panel is opened. So the way I'm handling this is that I have some timeout after which the IDE instance will be disposed if not re-attached by another panel.

filip131311 (approved):
I approve after checking it locally just two more things: 
1) I introduced additional commands in #875 thats why there are conflicts 
2) Because after the changes changing the IDE panel location does not reset the application we should re-think how we generate/store information about URLs in expo router. Currently we add information about visited paths to an array stored in webview and that array is reset after changing location. i created an issue to track this #887

PR #890, opened 2025-01-09
Update sim-server revision

kmagiera:
Updating sim-server version to include support for screenshots and a number of perf improvements that landed there since the previous update.


balins (approved):
🏎️

PR #891, opened 2025-01-09
Introduce third party notice generation

filip131311:
This PR adds a way for us to generate THIRDPARTYNOTICE.json. It adds plugins to esbuild and vite build processes and also downloads THIRDPARTY.json from simulator-server as part of `build:sim-server` command. 

To use the new functionality run: `npm run generate:third-party-notice`.

### How Has This Been Tested: 

- run `npm vscode:package`
- examine the content of resulting package to make sure license notice file is there


packages/vscode-extension/scripts/generateThirdPartyNotice.js
```diff
@@ -0,0 +1,53 @@
+const fs = require("fs");
+const path = require("path");
+
+const directoryPath = "dist/third-party-licenses";
+const outputFile = "dist/THIRDPARTYNOTICE.json";
+
+function readJsonFiles() {
```
kmagiera, line 7: you can read json using require('file.json')

packages/vscode-extension/package.json
```diff
@@ -469,11 +469,8 @@
   "scripts": {
     "vscode:package": "vsce package",
     "vscode:prepublish": "npm run package",
-    "build:extension-debug": "npm run build:extension-code-debug && npm run build:expo-fingerprint",
-    "build:extension": "npm run build:extension-code && npm run build:expo-fingerprint",
-    "build:extension-code-debug": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node --sourcemap && cp ./node_modules/source-map/lib/mappings.wasm dist/.",
-    "build:extension-code": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node --minify && cp ./node_modules/source-map/lib/mappings.wasm dist/.",
-    "build:expo-fingerprint": "esbuild ./node_modules/@expo/fingerprint/build/sourcer/ExpoConfigLoader.js --bundle --format=cjs --outfile=dist/ExpoConfigLoader.js --platform=node --minify",
+    "build:extension-debug": "node scripts/buildExtensionCode.mjs debug",
+    "build:extension": "node scripts/buildExtensionCode.mjs prod",
```
kmagiera, line 473: can we use "production" to keep it consistent with vite's config mode?

packages/vscode-extension/scripts/buildExtensionCode.mjs
```diff
@@ -0,0 +1,90 @@
...
+              ],
+            };
+          });
+          return JSON.stringify(result, null, 2);
+        },
+      },
+    },
+  };
+
+  await esbuild.build({
```
kmagiera, line 57: Lets restructure it in a way that we only call `build` once. For example: `) we create config object, 2) when mode==production we update the config to include license plugin 3) after config is setup we call build

kmagiera (changes requested):
Left a few comments inline

packages/vscode-extension/scripts/install-sim-server-release-build.sh
```diff
@@ -36,6 +36,28 @@ sim_server_tag=$(git -C ../simulator-server describe --tags)
...
+echo "Looking for THIRDPARTY.json in: $output_dir/third-party-licenses"
+if [ -f "$output_dir/third-party-licenses/THIRDPARTY.json" ]; then
+    echo "Removing existing file $output_dir/third-party-licenses/THIRDPARTY.json"
+    rm -f "$output_dir/third-party-licenses/THIRDPARTY.json"
+else
+    echo "THIRDPARTY.json does not exist."
+fi
+
+gh release download $sim_server_tag -R software-mansion-labs/simulator-server -p "THIRDPARTY.json" -D "$output_dir"/"third-party-licenses"
+chmod +x "$output_dir"/"third-party-licenses"
```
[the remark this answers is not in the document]
filip131311: I would like to keep deleting the files, because it was quite annoying to receive an error every time i tried to run the command a second time and generating the third party notice is not the only reason somebody might try to run the command. (For example this command might be useful to download the current sim-server release after working on different branch that was using other release or locally build version).

packages/vscode-extension/scripts/install-sim-server-release-build.sh
```diff
@@ -36,6 +36,22 @@ sim_server_tag=$(git -C ../simulator-server describe --tags)
...
 gh release download $sim_server_tag -R software-mansion-labs/simulator-server -p "simulator-server*" -D "$output_dir"
 chmod +x "$output_dir"/simulator-server*
+
+echo "Looking for THIRDPARTY.json in: $output_dir/third-party-licenses"
+if [ -f "$output_dir/third-party-licenses/THIRDPARTY.json" ]; then
+    echo "Removing existing file $output_dir/third-party-licenses/THIRDPARTY.json"
+    rm -f "$output_dir/third-party-licenses/THIRDPARTY.json"
+else
+    echo "THIRDPARTY.json does not exist."
+fi
```
kmagiera, line 52: why do we need this? can't we just `rm -f` here?

packages/vscode-extension/scripts/install-sim-server-release-build.sh
```diff
@@ -36,6 +36,22 @@ sim_server_tag=$(git -C ../simulator-server describe --tags)
...
+
+gh release download $sim_server_tag -R software-mansion-labs/simulator-server -p "THIRDPARTY.json" -D "$output_dir"/"third-party-licenses"
+
+echo "Operation completed successfully."
```
kmagiera, line 56: no need for this

packages/vscode-extension/package.json
```diff
@@ -469,11 +469,8 @@
   "scripts": {
     "vscode:package": "vsce package",
     "vscode:prepublish": "npm run package",
-    "build:extension-debug": "npm run build:extension-code-debug && npm run build:expo-fingerprint",
-    "build:extension": "npm run build:extension-code && npm run build:expo-fingerprint",
-    "build:extension-code-debug": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node --sourcemap && cp ./node_modules/source-map/lib/mappings.wasm dist/.",
-    "build:extension-code": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node --minify && cp ./node_modules/source-map/lib/mappings.wasm dist/.",
-    "build:expo-fingerprint": "esbuild ./node_modules/@expo/fingerprint/build/sourcer/ExpoConfigLoader.js --bundle --format=cjs --outfile=dist/ExpoConfigLoader.js --platform=node --minify",
+    "build:extension-debug": "node scripts/buildExtensionCode.mjs debug",
```
kmagiera, line 472: use `./scripts...` for consistency ?

kmagiera (commented):
Left a couple more comments inline. Looks good otherwise. We should also likely modify `.vscodeignore` file which controls what files are packaged into the vsix package. I think we should skip partial notice files from third-party-licenses directory, while we currentlu include the whole dist folder

kmagiera (approved):
I went ahead and applied the changes I suggested and will merge it for the upcoming release

PR #913, opened 2025-01-15
Fix debugger after multiple reloads

### LINKED ISSUE #860: Multiple instances of React Native Preview Debugger
**Description**
There are multiple React Native Preview Debuggers created with consecutive numbers in their names, but logs are only directed to the original one (without a number).  "Open logs panel" button also redirects only to that first one, which may cause confusion.

Additionally, we should probably rename this to Radon Preview Debugger.

<video src="https://github.com/user-attachments/assets/8dc3c588-5b61-49e8-9e6b-86a29848e200" />

**Environment**
- Radon: v1.1.0.rc0
- Test app: expo-50-dev-client

filip131311:
This PR adreses a problem with multiple debugSessions being held by vscode, after performing multiple reloads in quick succession. This was caused by `debug.stopDebugging(this.vscSession)` not removing the previous session if called in to rapid succession. We trigger the `stopDebugging` method is by dispossessing `debugSession`. Because we dispose of the old `debugSession` every time we create a new one, attempt to create multiple instances of debug session in a row leads to the problem. 

In our current flow multiple instances of `debugSession` might be created after multiple restarts happening before, the previous one finishes. To avoid that we introduce a cancel token to `launchApp` method (which is triggered by the restart)  to exclude parallel execution.

Fixes #860

### How Has This Been Tested: 

Performed reproduction from #860:
- force bundle error 
- reload the app multiple times 
- fix bundle error 
- the debugger should work properly after this fix

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -17,6 +17,7 @@ import { throttle } from "../utilities/throttle";
 import { DependencyManager } from "../dependency/DependencyManager";
 import { getTelemetryReporter } from "../utilities/telemetry";
 import { BuildCache } from "../builders/BuildCache";
+import { CancelToken } from "../builders/cancelToken";
 
 type PreviewReadyCallback = (previewURL: string) => void;
 type StartOptions = { cleanBuild: boolean; previewReadyCallback: PreviewReadyCallback };
@@ -95,8 +96,8 @@ export class DeviceSession implements Disposable {
         await this.launchApp();
         return true;
       case "restartProcess":
-        const launchSuccess = await this.launchApp();
-        if (!launchSuccess) {
+        const launchSucceeded = await this.launchApp();
+        if (!launchSucceeded) {
           return false;
         }
         return true;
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/project/project.ts
```diff
@@ -416,10 +416,12 @@ export class Project
 
       // we first check if the device session hasn't changed in the meantime
       if (deviceSession === this.deviceSession) {
-        await this.deviceSession?.perform("restartProcess");
-        this.updateProjectStateForDevice(deviceInfo, {
-          status: "running",
-        });
+        const restartSuccess = await this.deviceSession?.perform("restartProcess");
```
kmagiera, line 419: `restartSucceeded`

kmagiera (changes requested):
I don't understand the changes here. In description you write the problem was with `stopDebugging` call but it isn't explained how those changes affect it. Also, by looking at DebugSession code I don't see how that `stopDebugging` call could do anything wrong here. It seems like it will always stop a newly created session. It'd be good for this to be covered in the PR description.

As for the code, I think it'd be better to stick to patterns we use in other places to exlude parallel execution of certain code paths. Specifically we should be using CancelToken here. And the pattern should look as follows:
1. When launchApp is called we create new cancel token and cancel the previous one (it it exists)
2. Later, we check token after every await (it may only get cancelled while we await)

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -111,12 +115,18 @@ export class DeviceSession implements Disposable {
     throw new Error("Not implemented " + type);
   }
 
+  private lastLaunchCancelToken: CancelToken | undefined;
```
kmagiera, line 118: I'd just call it `launchAppCancelToken` which indicates that this is cancel token used by launchApp method.

kmagiera (approved):
Looks good

PR #914, opened 2025-01-15
Add native logs and build logs docs

maciekstosio:
In #742 and #734 we added native logs, which are not explained in the docs. This PR adds instructions for those as well as for build logs.


kmagiera (approved):
Looks good but would be also great to mention native logs in troubleshooting section where we list "checking logs" as one of the options in `docs/guides/troubleshooting#6-accessing-extension-logs`

PR #915, opened 2025-01-15
Close drop downs on blur 

### LINKED ISSUE #908: Hide device settings and other dropdown menu on window focus change
To make the IDE panel behave more natively we should be dismissing menus that can be opened from the main panel when the user switches focus to editor or some other window. Currently, when you click "device settings" to open menu and then navigate away from VSCode it will be kept open unless you focus back the IDE window and then click away from the menu. The same applies to other menus that can be opened from the main panel UI.

## Steps to reproduce
1. Have some file open in one panel and IDE panel next to it
2. Go to IDE panel and open device settings
3. Click on the code panel now (the device settings menu will be kept open)

filip131311:
This PR forces dropdown menus to close when focus is taken away from the IDE.

Fixes #908

### How Has This Been Tested: 

Run IDE, open dropdown menu, click outside of the ide

### DIFF
--- a/packages/vscode-extension/src/webview/components/DropdownMenuRoot.tsx
+++ b/packages/vscode-extension/src/webview/components/DropdownMenuRoot.tsx
@@ -1,0 +1,26 @@
+import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
+import { PropsWithChildren, useEffect, useState } from "react";
+
+export function DropdownMenuRoot({ children }: PropsWithChildren) {
+  const [open, setOpen] = useState(false);
+
+  useEffect(() => {
+    const blurListener = () => {
+      setOpen(false);
+    };
+    window.addEventListener("blur", blurListener);
+    return () => {
+      window.removeEventListener("blur", blurListener);
+    };
+  }, []);
+
+  return (
+    <DropdownMenu.Root
+      open={open}
+      onOpenChange={() => {
+        setOpen(!open);
+      }}>
+      {children}
+    </DropdownMenu.Root>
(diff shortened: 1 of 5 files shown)


kmagiera (changes requested):
I was wondering whether we could just simulate a click "outside" when window is blured and then don't really need extra handling for individual dropdowns?

I tested that quickly and seem to work:
```
window.addEventListener("blur", () => {
      document.body.dispatchEvent(new PointerEvent("pointerdown", { bubbles: true }));
```

The current version doesn't seem to cover all menus, for example URL dropdown or device selection dropdown.

filip131311:
@kmagiera I opted for wrapping Dropdown menu instead, because select from radix already handles that functionality [select source](https://github.com/radix-ui/primitives/blob/11525108ae531a787cf02dec67a317ed6364c581/packages/react/select/src/Select.tsx#L615C4-L623C21). I thought it would be more consistent behavior. WDYT?

packages/vscode-extension/src/webview/components/DropdownMenuRoot.tsx
```diff
@@ -0,0 +1,26 @@
...
+      window.removeEventListener("blur", blurListener);
+    };
+  }, []);
+
+  return (
+    <DropdownMenu.Root
+      open={open}
+      onOpenChange={() => {
+        setOpen(!open);
+      }}>
```
kmagiera, line 22:
```suggestion
      onOpenChange={setOpen}>
```

kmagiera (approved):

PR #923, opened 2025-01-20
fix: move image to `docs` dir

kewinzaq1:
I put image in a wrong directory https://github.com/software-mansion/radon-ide/pull/874 and this fixes it.

https://ide.swmansion.com/docs/guides/troubleshooting#13-ide-looks-bad-with-my-theme
image


filip131311 (approved):
Thank you

PR #924, opened 2025-01-23
Refactor: extract some some components from Preview 

maciekstosio:
Mini refactor that extracts some components to separate files. For now it's a bit weird that we're moving it to Preview, but my goal is to split entrie Preview.tsx in small chunks in Preview folder. 

### How Has This Been Tested: 

Nothing should changed


filip131311 (approved):
LGTM

PR #925, opened 2025-01-23
Add small border around device

### LINKED ISSUE #909: Add some tiny borders around device for when no device frame mode is in use
It looks a bit ugly that you can't tell where the device edges are. Especially when the dominating color matches the background of the IDE:

Image

maciekstosio:
Fixes #909 

This PR adds a tiny border for a device with `Show Device Frame` disabled. First of all, for each device, we `add screen.png` this is just a mask with all notches and cameras filled. This is displayed between the frame and stream as a background. As for the bezel, we create a separate `bezel.png`. This is necessary as scaling `screen.png` would cause the borders to be uneven with respect to the aspect ratio, and we'd like a constant outline. `bezel.png` is used as a mask for div with a selected color this way, we can manipulate the color of the border programmatically. 

This PR also changes the dimensions of the mask, skin for iOS, because they didn't match what's in the config - this inconsistency is confusing when creating the `bezel.png` as it may lead to different width of the outline across the devices.  

The solution should also provide more flexibility if we decide to add more different frames in the future. 

**Pixel 8**
|Before|After|
|-|-|
|SCR-20250123-krte|SCR-20250123-opwr|

**iPhone 15 Pro**
|Before|After|
|-|-|
|SCR-20250123-kryi|SCR-20250123-opdj|

**iPhone SE**
|Before|After|
|-|-|
|SCR-20250123-ksdf|SCR-20250123-oozg|

### How Has This Been Tested: 

- Go to settings and search Radon IDE
- Toggle `Show Device Frame`
- You should see a tiny border with a separator color around the device
- Holes for notches and cameras should be black, not transparent
- Test is swipe back works 

Tested on all existing devices.


filip131311 (approved):
please address inline comments before merging

PR #928, opened 2025-01-24
Fix infinite loading when cdp does not return a variable

filip131311:
This PR fixes a problem with debugger infinitely awaiting for variable from cdp, blocking the whole debugger process in consequence. The solution is to add a time out to the function that fetches the properties and return empty variable list if it passes. 

### How Has This Been Tested: 

- add a `console.log(devtoolsAgent)` somewhere in `RNIDE_inspect` section of wrapper.js
- click through returned object before the change the `_persistedSelection` filled would load infinitely after it stops awaiting after 3 s.

### DIFF
--- a/packages/vscode-extension/src/debugging/CDPSession.ts
+++ b/packages/vscode-extension/src/debugging/CDPSession.ts
@@ -70,7 +70,7 @@ export class CDPSession {
     this.connection.close();
   }
 
-  public async sendCDPMessage(method: string, params: object) {
+  public async sendCDPMessage(method: string, params: object, timeoutMs?: number) {
     const message = {
       id: ++this.cdpMessageId,
       method: method,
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debugging/variableStore.ts
```diff
@@ -22,70 +23,77 @@ export class VariableStore {
...
-    } else {
-      properties = this.replVariables.get(id) as CDPPropertyDescriptor[];
-      isCDPObject = false;
-    }
-
-    const variables: Variable[] = properties
-      .map((prop) => {
-        if (prop.value === undefined) {
-          return { name: prop.name, value: "undefined", variablesReference: 0 };
+    return new Promise(async (resolve, reject) => {
```
maciekstosio, line 26: Why? if fetchProperties rejects, you'll catch it here just to reject and catch it again in DebugAdapter. Isn't stacktract enough to pinpoint the problem?

maciekstosio (changes requested):

maciekstosio (approved):

PR #929, opened 2025-01-24
Fix Inspector links

filip131311:
This PR is a draft of a fix of inspector links in react 19


packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -461,6 +461,32 @@ export class DebugAdapter extends DebugSession {
     this.sendResponse(response);
   }
 
+  protected async sourceRequest(
```
maciekstosio, line 464: Seems like this shouldn't be part of this PR

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -82,6 +82,14 @@ export class DebugSession implements Disposable {
     return false;
   }
 
+  public async getOriginalSource(
```
maciekstosio, line 85: Same

maciekstosio (approved):
Remove those above; the rest looks good 🚀

PR #931, opened 2025-02-03
First-party redux support

maciekstosio:
This PR adds first-party support for redux. I resued the enhancer part and WebUI from https://github.com/matt-oakes/redux-devtools-expo-dev-plugin - the code seems to do everything we need with nicely separated communication, except that the communication through the `expo/devtools` is hardcode. I created a fork of it that's included as a submodule in our repo, but I'm trying to upstream the changes with communication logic abstraction (https://github.com/matt-oakes/redux-devtools-expo-dev-plugin/pull/19) so we can easily keep the fork up to date. The rest (communication through vscode) can be found in https://github.com/software-mansion-labs/redux-devtools-expo-dev-plugin/pull/2).

To archive zero-config support, we utilize `window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__`, where we pass compose from `redux-devtools-expo-dev-plugin` with our communication injected. Redux dev tools communicate with the extension through the dev tools, and the extension passes messages through webview events to WebUI. 

We need to have `redux-devtools-expo-dev-plugin` built to import through a wrapper in app runtime. Similarly, the webui needs to be bundled, thus I added `packages/vscode-extension/scripts/install-dev-plugins.sh`, currently the product of it is committed, and it needs to be run manually, but this is subjected to change. 

### How Has This Been Tested: 

- Open `expo-52-prebuild-with-plugins`
- Comment out configuration related to expo dev plugins
```
const store = configureStore({
  reducer: rootReducer,
  // devTools: false,
  // enhancers: (getDefaultEnhancers) =>
  //   getDefaultEnhancers().concat(devToolsEnhancer()),
});
```
- Reload the app
- Confirm redux dev tools work


packages/vscode-extension/lib/plugins/redux-devtools.js
```diff
@@ -0,0 +1,89 @@
...
+
+  setDevtoolsAgent = (agent) =>  {
+    if (!agent) {
+      return;
+    }
+    this.devtoolsAgent = agent;
+    this.devtoolsAgent._bridge.addListener(this.scope, this.handleMessages);
+  };
+
+  clearDevToolsAgent = () => {
```
filip131311, line 26: The pattern we are using in most of the codebase is for setter function to return dispose function that is added to disposables that are called after object destruction , this solution is advantageous, because it does not really on `this.scope` and `this. handleMessages ` being stable, more over it would be consistant with the rest of the codebase

packages/vscode-extension/lib/plugins/redux-devtools.js
```diff
@@ -0,0 +1,89 @@
...
+      data,
+    });
+  };
+
+  addMessageListener = (type, listener) => {
+    const currentListeners = this.listeners.get(type) || [];
+    this.listeners.set(type, [...currentListeners, listener]);
+  };
+
+  removeMessageListener = (type, listener) => {
```
filip131311, line 51: this method seems to never be used

packages/vscode-extension/lib/plugins/redux-devtools.js
```diff
@@ -0,0 +1,89 @@
...
+    this.clearDevToolsAgent();
+    this.listeners.clear();
+  };
+}
+
+let proxyClient = null;
+
+export const isProxyClientReady = () => proxyClient !== null;
+
+export const clearProxyClient = () => proxyClient = null;
```
filip131311, line 67: shouldn't this method call `closeAsync` ?

packages/vscode-extension/lib/plugins/redux-devtools.js
```diff
@@ -0,0 +1,89 @@
...
+
+export const createRNIDEProxyClientAsync = async () => {
```
filip131311, line 69: why is it async?

packages/vscode-extension/lib/plugins/redux-devtools.js
```diff
@@ -0,0 +1,89 @@
+const { useEffect } = require("react");
+const {createComposeWithDevTools} = require('./external/redux-devtools-expo-dev-plugin');
+
+class RNIDEProxyClient {
```
filip131311, line 4: nit: 
This name seems almost too generic i'm not sure what it does from the name

packages/vscode-extension/lib/plugins/redux-devtools.js
```diff
@@ -0,0 +1,89 @@
...
+
+  closeAsync = () => {
+    this.clearDevToolsAgent();
+    this.listeners.clear();
+  };
+}
+
+let proxyClient = null;
+
+export const isProxyClientReady = () => proxyClient !== null;
```
filip131311, line 65: Is there ever a situation in which ProxyClient exist but is not ready? if not i would rename this method to `available if it is possible it should be checked here

packages/vscode-extension/lib/wrapper.js
```diff
@@ -384,7 +387,13 @@ export function AppWrapper({ children, initialProps, fabric }) {
           plugins: Array.from(expoDevPlugins.values()),
         });
       };
+      devtoolsAgent._bridge.send("RNIDE_pluginsChanged", {
+        plugins: [
+          isProxyClientReady() && "RNIDE-redux-devtools",
+        ].filter(Boolean),
+      });
       return () => {
+        clearProxyClient();
```
filip131311, line 396: is it called when the user opens preview? could you test that and check if it does not cause multiple proxies being created?

packages/vscode-extension/package.json
```diff
@@ -198,6 +198,11 @@
           "id": "RadonIDEToolExpoDevPluginReduxDevTools",
```
filip131311, line 198: why don't we retire the old panel?

packages/vscode-extension/plugins_dist/redux/AppEntry-9f5185ef27313606ee5279432a7a2e89.js
filip131311, line 1: why dis file has suck a strage name is it commit name? wouldn't it be better to use tags as a naming convention?

packages/vscode-extension/src/plugins/expo-dev-plugins/expo-dev-plugins.ts
```diff
@@ -16,16 +16,16 @@ type ExpoDevPluginInfo = {
...
-    label: "React Query",
+    label: "React Query DevPlugin",
     viewIdPrefix: "RNIDE.Tool.ExpoDevPlugin.ReactQuery",
   },
   "@dev-plugins/react-native-mmkv": {
-    label: "MMKV",
+    label: "MMKV DevPlugin",
     viewIdPrefix: "RNIDE.Tool.ExpoDevPlugin.MMKV",
   },
   "redux-devtools-expo-dev-plugin": {
```
filip131311, line 26: why don't we retire it?

filip131311 (changes requested):
I left some inline comment

packages/vscode-extension/lib/plugins/redux-devtools.js
maciekstosio: We're implementing methods that could be used by the `redux-devtools-expo-dev-plugin` - those method match expo interface.

packages/vscode-extension/src/plugins/expo-dev-plugins/expo-dev-plugins.ts
maciekstosio: We attach to redux like default dev tools on the web, thus if someone have custom configuration, i.e. using dev-plugins, our will stop working. I could try overriding the config, but I'm unsure if I won't break certain setups. So I'd leave it.

packages/vscode-extension/package.json
maciekstosio: We attach to redux like default dev tools on the web, thus if someone have custom configuration, i.e. using dev-plugins, our will stop working. I could try overriding the config, but I'm unsure if I won't break certain setups. So I'd leave it.

packages/vscode-extension/lib/plugins/redux-devtools.js
maciekstosio: Yeah, If there isn't redux, or if redux devtools have different configuration. Basically when __REDUX_DEVTOOLS_EXTENSION_COMPOSE__ is not used.

packages/vscode-extension/lib/plugins/redux-devtools.js
maciekstosio: Just to follow types from `redux-devtools-expo-dev-plugin`

packages/vscode-extension/lib/wrapper.js
maciekstosio: I removed that in https://github.com/software-mansion/radon-ide/pull/931/commits/0ede4246a2d3100dbf3af63ccb69b96acaf1a10e

packages/vscode-extension/lib/plugins/redux-devtools.js
maciekstosio: Changed in: https://github.com/software-mansion/radon-ide/pull/931/commits/0ede4246a2d3100dbf3af63ccb69b96acaf1a10e

packages/vscode-extension/lib/plugins/redux-devtools.js
maciekstosio: Tbh, I'm not sure if I see any benefit in it in this. I can't get rid of `this.handleMessages` as `this.clearDevToolsAgent` is called also in `closeAsync`, if it would be just for set/clearDevToolsAgent I totally get it.

packages/vscode-extension/package.json
filip131311: I was able to reproduce the behavior in which booth panels are open and only one works:

Screenshot 2025-02-07 at 10 15 22

To achieve that just comment and uncomment configuration without restarting the ide: 
```
const store = configureStore({
  reducer: rootReducer,
  // devTools: false,
  // enhancers: (getDefaultEnhancers) =>
  //   getDefaultEnhancers().concat(devToolsEnhancer()),
});
``` 

I don't mind leaving the old panel alive if it is the most straight forward solution but we need a way of establishing which one is used and killing the other one.

filip131311 (changes requested):
I left one more important comment

packages/vscode-extension/package.json
filip131311: After additional thought this edge case might be adressed in followup review

filip131311 (approved):
LGTM

packages/vscode-extension/package.json
maciekstosio: https://github.com/orgs/software-mansion/projects/15?pane=issue&itemId=96936384

PR #932, opened 2025-02-04
Implement GUI for selecting eas build configuration

jwajgelt:
Adds options to the "Launch Configuration" GUI for enabling EAS Build, selecting build profiles and specifying build UUIDs.

### How Has This Been Tested: 
- open [`expo-52-eas`](https://github.com/software-mansion-labs/radon-ide-test-apps/tree/main/expo-52-eas) and open the "Launch Configuration" window
  - the settings for EAS Build should appear
  - changing the settings in GUI should update `launch.config` file appropriately
- open an application without `build` config in `eas.json` (for example, [`react-native-77`](https://github.com/software-mansion-labs/radon-ide-test-apps/tree/main/react-native-77) or remove the `build` entry in `eas.json` in [`expo-52-eas`](https://github.com/software-mansion-labs/radon-ide-test-apps/tree/main/expo-52-eas)) and open the "Launch Configuration" window
  - the settings for EAS Build shouldn't appear

### DIFF
--- a/packages/vscode-extension/src/common/EasConfig.ts
+++ b/packages/vscode-extension/src/common/EasConfig.ts
@@ -1,3 +1,6 @@
+// Describes the shape of the `eas.json` config file's content.
+// Currently, we only include the parts of the object we care about.
+// See https://docs.expo.dev/eas/json/ for details.
 export interface EasConfig {
   build?: EasBuildConfig;
 }
--- a/packages/vscode-extension/src/common/EasConfig.ts
+++ b/packages/vscode-extension/src/common/EasConfig.ts
@@ -1,0 +1,62 @@
+export interface EasConfig {
+  build?: EasBuildConfig;
+}
+
+export type EasBuildConfig = {
+  [key: string]: EasBuildProfile;
+};
+
+export type EasBuildDistributionType = "internal" | "store";
+
+export interface EasBuildProfileIOSSpecific {
+  simulator?: boolean;
+}
+
+export interface EasBuildProfile {
+  distribution?: EasBuildDistributionType;
+  ios?: EasBuildProfileIOSSpecific;
+}
+
+export function isEasConfig(obj: unknown): obj is EasConfig {
+  if (typeof obj !== "object" || obj === null) {
+    return false;
+  }
+
+  return !("build" in obj) || obj.build === undefined || isEasBuildConfig(obj.build);
+}
+
+export function isEasBuildConfig(obj: unknown): obj is EasBuildConfig {
+  if (typeof obj !== "object" || obj === null) {
+    return false;
+  }
+
+  return Object.values(obj).every((v) => v === undefined || isEasBuildProfile(v));
+}
+
+export function isEasBuildProfile(obj: unknown): obj is EasBuildProfile {
+  if (typeof obj !== "object" || obj === null) {
+    return false;
+  }
+
+  if (
+    "distribution" in obj &&
+    (typeof obj.distribution !== "string" || !["internal", "store"].includes(obj.distribution))
+  ) {
+    return false;
+  }
+
+  if ("ios" in obj && obj.ios !== undefined) {
+    if (typeof obj.ios !== "object" || obj.ios === null) {
+      return false;
+    }
+    if (
+      "simulator" in obj.ios &&
+      obj.ios.simulator !== undefined &&
+      !(typeof obj.ios.simulator === "boolean")
+    ) {
+      return false;
+    }
+  }
+
+  return true;
+}
--- a/packages/vscode-extension/src/panels/LaunchConfigController.ts
+++ b/packages/vscode-extension/src/panels/LaunchConfigController.ts
@@ -11,6 +11,7 @@ import { findXcodeProject, findXcodeScheme } from "../utilities/xcode";
 import { Logger } from "../Logger";
 import { getIosSourceDir } from "../builders/buildIOS";
 import { readEasConfig } from "../utilities/eas";
+import { EasBuildConfig } from "../common/EasConfig";
 
 export class LaunchConfigController implements Disposable, LaunchConfig {
   private config: LaunchConfigurationOptions;
@@ -108,11 +109,11 @@ export class LaunchConfigController implements Disposable, LaunchConfig {
     return await findXcodeScheme(xcodeProject);
   }
 
-  async getAvailableEasProfiles(): Promise<string[]> {
+  async getAvailableEasProfiles(): Promise<EasBuildConfig> {
     const appRootFolder = getAppRootFolder();
     const easConfig = await readEasConfig(appRootFolder);
     const easBuildConfig = easConfig?.build ?? {};
-    return Object.keys(easBuildConfig);
+    return easBuildConfig;
   }
 
   async addListener<K extends keyof LaunchConfigEventMap>(
--- a/packages/vscode-extension/src/panels/LaunchConfigController.ts
+++ b/packages/vscode-extension/src/panels/LaunchConfigController.ts
@@ -10,6 +10,7 @@ import { getAppRootFolder } from "../utilities/extensionContext";
 import { findXcodeProject, findXcodeScheme } from "../utilities/xcode";
 import { Logger } from "../Logger";
 import { getIosSourceDir } from "../builders/buildIOS";
+import { readEasConfig } from "../utilities/eas";
 
 export class LaunchConfigController implements Disposable, LaunchConfig {
   private config: LaunchConfigurationOptions;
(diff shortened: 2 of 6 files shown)


packages/vscode-extension/src/utilities/eas.ts
```diff
@@ -0,0 +1,38 @@
...
+  build?: EasBuildConfig;
+}
+
+export type EasBuildConfig = {
+  [key: string]: EasBuildProfile;
+};
+
+export interface EasBuildProfile {}
+
+function isEasConfig(obj: unknown): obj is EasConfig {
```
maciekstosio, line 13: Maybe `isEasBuildConfig`?

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -44,6 +53,26 @@ function LaunchConfigurationView() {
       <IsExpoConfiguration isExpo={isExpo} update={update} />
 
       <div className="launch-configuration-section-margin" />
+
+      {easBuildProfiles.length > 0 && (
+        <>
+          <Label>eas Build</Label>
```
maciekstosio, line 59:
```suggestion
          <Label>EAS Build</Label>
```

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -241,4 +270,93 @@ function IsExpoConfiguration({ isExpo, update }: isExpoConfigurationProps) {
...
+}
+
+interface easBuildConfigurationProps {
+  eas?: EasConfig;
+  platform: EasPlatform;
+  update: LaunchConfigUpdater;
+  easBuildProfiles: string[];
+}
+
+function EasBuildConfiguration({
```
maciekstosio, line 292: I'd change LaunchConfigView into a folder and extract this component. Maybe we could also use `useLaunchConfig` to avoid passing props

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -241,4 +270,93 @@ function IsExpoConfiguration({ isExpo, update }: isExpoConfigurationProps) {
...
+  const DISABLED = "Disabled" as const;
+
+  const profile = eas?.[platform]?.profile;
+  const buildUUID = eas?.[platform]?.buildUUID;
+
+  const buildUUIDInputRef = useRef<HTMLInputElement>(null);
+
+  const onSchemeChange = (newProfile: string) => {
+    if (newProfile === DISABLED) {
+      let newEasConfig: EasConfig | undefined = { ...eas, [platform]: undefined };
```
maciekstosio, line 307:
```suggestion
      const newEasConfig: EasConfig | undefined = { ...eas, [platform]: undefined };
```

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -241,4 +270,93 @@ function IsExpoConfiguration({ isExpo, update }: isExpoConfigurationProps) {
...
+    const newBuildUUID = buildUUIDInputRef.current?.value || undefined;
+
+    update("eas", {
+      ...eas,
+      [platform]: { profile: newProfile, buildUUID: newBuildUUID },
+    });
+  };
+
+  const onBuildUUIDInputBlur = () => {
+    let newBuildUUID = buildUUIDInputRef.current?.value || undefined;
```
maciekstosio, line 321:
```suggestion
    const newBuildUUID = buildUUIDInputRef.current?.value || undefined;
```

maciekstosio (changes requested):

packages/vscode-extension/src/panels/LaunchConfigController.ts
```diff
@@ -106,6 +107,14 @@ export class LaunchConfigController implements Disposable, LaunchConfig {
     );
     return await findXcodeScheme(xcodeProject);
   }
+
+  async getAvailableEasProfiles(): Promise<string[]> {
+    const appRootFolder = getAppRootFolder();
+    const easConfig = await readEasConfig(appRootFolder);
+    const easBuildConfig = easConfig?.build ?? {};
```
filip131311, line 114:
```
"build": {
    "development": {
      "developmentClient": true,
      "distribution": "internal",
      "ios": {
        "simulator": true
      }
    },
    "preview": {
      "distribution": "internal"
    },
    "production": {
      "autoIncrement": true
    }
```
this is an example eas configuration (from `eas.json`) we should only show profiles that have `distribution` field set to "internal". And `simulator` set to true. It is not necessary true, that profiles with those flag will generate dev builds runnable on simulators, but expo asumes so and you would have to do a lot of manula configuration to "cheat" that.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -241,4 +270,93 @@ function IsExpoConfiguration({ isExpo, update }: isExpoConfigurationProps) {
...
+  const availableEasBuildProfiles = easBuildProfiles.map((buildProfile) => {
+    return { value: buildProfile, label: buildProfile };
+  });
+
+  availableEasBuildProfiles.push({ value: DISABLED, label: DISABLED });
+
+  return (
+    <div className="launch-configuration-container">
+      <div className="setting-description">{prettyPlatformName(platform)} Build Profile:</div>
+      <Select
```
filip131311, line 339: We should allow additional manual input of a value that is not on the list for users with exotic configurations, the button that does it might be hidden as a last psoittion in the select list for example.

filip131311 (changes requested):
I left some inline comments

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
jwajgelt: You're right.
I went off the cli utility (`eas`), but in writing they always capitalize it as EAS, so this is the one we should use.

The `<Label>` displays the text capitalized anyway, so there won't be a visual difference, but still.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
jwajgelt: Missed this.
Does `eslint` highlight this for you? I might have a borked config in that case.

packages/vscode-extension/src/utilities/eas.ts
jwajgelt: You have two types here, `EasConfig` and `EasBuildConfig`
- `EasConfig` is supposed to reflect the structure of the whole `eas.json` file
- `EasBuildConfig` is the type of the field named `"build"` in `eas.json`
This function is the type predicate for `EasConfig`, hence `isEasConfig`.

packages/vscode-extension/src/panels/LaunchConfigController.ts
jwajgelt: I've thought about it. I'm not sure we should hide the "invalid" configs.

I feel like _ideally_ we'd show all configs, but disable selecting the "invalid" ones with some information to the user why they are not selectable, but I'm not sure how much work it is to add that functionality to our components.
I'd leave this out of this PR.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
jwajgelt: Possibly.
I feel like if you have an "exotic configuration", you can be asked to edit the `launch.json` manually, and I fear that adding a "Custom" option may clutter the UI for the other users who, _I think_, are in vast majority here.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
jwajgelt: I agree with splitting this file, but I think I'd rather do that refactor in a separate PR.

filip131311 (approved):
I left some additional comments please adres them before merging

packages/vscode-extension/src/common/EasConfig.ts
```diff
@@ -0,0 +1,62 @@
...
+  distribution?: EasBuildDistributionType;
+  ios?: EasBuildProfileIOSSpecific;
+}
+
+export function isEasConfig(obj: unknown): obj is EasConfig {
+  if (typeof obj !== "object" || obj === null) {
+    return false;
+  }
+
+  return !("build" in obj) || obj.build === undefined || isEasBuildConfig(obj.build);
```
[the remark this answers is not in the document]
jwajgelt: It's just checking if the shape of the object matches what we expect it to be -- is it really that puzzling?
I suppose you might want a reference to some schema here for the `eas.json` file -- previously this was in the same place that _read_ that file, so the context was there to see that's what it's doing, but I suppose now that it's an exported part of a module, some comment might be useful.

PR #934, opened 2025-02-05
Improve user messaging when eas-cli is not installed

jwajgelt:
- Improves build error messages when trying to use EAS Build without `eas-cli` installed.
- Adds `eas-cli` installation status to the diagnostics window.

### How Has This Been Tested: 
Try to build an EAS Build project without `eas-cli` installed (`eas` executable should not be in `PATH`).
image


maciekstosio (approved):
LGTM 🚀

Issue #701, opened 2024-11-06
Add bi-directional clipboard support


kmagiera:
Now we only support pasting into the emulator/simulator – would like to have the clipboard work both ways

PR #936, opened 2025-02-06
Add bidirectional clipboard for iOS Simulator, support multiline paste, show toast when copying/pasting

### LINKED ISSUE #160: cannot copy to clipboard from simulator
This maybe not the most common usecase but I have a share button that lets me copy some text from the app and the text doesn't  seem to get added to my mac clipboard even though it normally would with a simulator running outside of the extension.

balins:
This PR introduces bidirectional clipboard support for iOS Simulator. The copying behaviour is similar to how cmd+v works, but instead of transferring hosts' clipboard content to the simulator, cmd+c transfers the simulator's clipboard to host's.

Apart from that, this PR prevents typing 'c' and 'v' when holding cmd and fixes multiline pastes (previously, newlines were recognized as starting separate commands, resulting in `unrecognized command` errors from simserver). It also adds a toast notifying user when copying/pasting is performed.

https://github.com/user-attachments/assets/c4014a7d-6023-4b30-b3ed-32c353fade02

Fixes #160, #701

### How Has This Been Tested: 

Verified that:
- command+C copies the Simulator's clipboard content to host
- command+V pastes the host's clipboard content to the Simulator
- command+C is ignored for Android Emulator (it automatically sends its clipboard to host whenever text is copied)
- command+V pastes the hosts' clipboard content to Android Emulator's clipboard
- typing 'c' and 'v' into Simulator/Emulator is prevented when holding 'command'
- multiline paste works properly
- toast is shown on cmd+c, cmd+v

### DIFF
--- a/packages/vscode-extension/src/devices/DeviceBase.ts
+++ b/packages/vscode-extension/src/devices/DeviceBase.ts
@@ -6,16 +6,24 @@ import { AppPermissionType, DeviceSettings, TouchPoint } from "../common/Project
 import { DeviceInfo, DevicePlatform } from "../common/DeviceManager";
 import { tryAcquiringLock } from "../utilities/common";
 
+const LEFT_META_HID_CODE = 0xe3;
+const RIGHT_META_HID_CODE = 0xe7;
+const V_KEY_HID_CODE = 0x19;
+const C_KEY_HID_CODE = 0x06;
+
 export abstract class DeviceBase implements Disposable {
   protected preview: Preview | undefined;
   private previewStartPromise: Promise<string> | undefined;
   private acquired = false;
+  private pressingLeftMetaKey = false;
+  private pressingRightMetaKey = false;
 
   abstract get lockFilePath(): string;
 
   abstract bootDevice(deviceSettings: DeviceSettings): Promise<void>;
   abstract changeSettings(settings: DeviceSettings): Promise<boolean>;
   abstract sendBiometricAuthorization(isMatch: boolean): Promise<void>;
+  abstract getClipboard(): Promise<string | void>;
   abstract installApp(build: BuildResult, forceReinstall: boolean): Promise<void>;
   abstract launchApp(build: BuildResult, metroPort: number, devtoolsPort: number): Promise<void>;
   abstract makePreview(): Preview;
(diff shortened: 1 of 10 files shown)


packages/vscode-extension/src/devices/DeviceBase.ts
```diff
@@ -96,11 +104,37 @@ export abstract class DeviceBase implements Disposable {
...
+    }
+
+    if (
+      (this.pressingLeftMetaKey || this.pressingRightMetaKey) &&
+      (keyCode === C_KEY_HID_CODE || keyCode === V_KEY_HID_CODE)
+    ) {
+      // ignore sending C and V when meta key is pressed
+    } else {
+      this.preview?.sendKey(keyCode, direction);
+    }
```
filip131311, line 133: Why, did you move this code to `deviceBase` I understand that is does not break android, as you pointed out in your comment, but it is still an ios specific workaround is it not?

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -191,7 +191,9 @@ export class Preview implements Disposable {
     this.subprocess?.stdin?.write(`key ${direction} ${keyCode}\n`);
   }
 
-  public async sendPaste(text: string) {
-    this.subprocess?.stdin?.write(`paste ${text}\n`);
+  public sendClipboard(text: string) {
+    // This is bad, but prevents simulator server going crazy with multiline pastes
+    // If we want to support multiline pastes we need to change the communication protocol
```
filip131311, line 196: what options do we have in regard to other protocols?

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -347,19 +347,24 @@ function Preview({
   }, []);
 
   useEffect(() => {
-    function dispatchPaste(e: ClipboardEvent) {
+    function synchronizeClipboard(e: ClipboardEvent) {
```
filip131311, line 350: I wonder if we should inform user somehow that thay recived a clipboard?

Like in the example video: 

https://github.com/user-attachments/assets/572e2018-be5e-437d-9da9-b5c8bdb924a9

I know that usually cmd+c does not trigger such a microinteraction, but this is a very specific case in witch user does not copy highlithed text but rather the contents of a clipboard on a simulated device.

filip131311: PS. I don't necessary think it needs to be part of this PR

filip131311 (commented):
I left some comments

packages/vscode-extension/src/devices/preview.ts
balins: We can either switch protocol completely, like move to GRPC or stay with communication via stdin-stdout, but add some markers for multiline commands, like this:
```
paste >>>STARTSIMSERVERPASTEhere is the
actual
multiline content
pasted by
the user<<<ENDSIMSERVERPASTE 
```

packages/vscode-extension/src/webview/components/Preview.tsx
balins: To me, a small modal "device clipboard copied to host"/"host clipboard pasted to device" appearing on copy/paste would be sufficient

balins: Added a toast notification for copy/paste

https://github.com/user-attachments/assets/c4014a7d-6023-4b30-b3ed-32c353fade02

packages/vscode-extension/src/project/project.ts
```diff
@@ -252,8 +260,31 @@ export class Project
 
   //#endregion
 
+  async showToast(message: string, timeout: number) {
```
filip131311, line 263: this should be either part of `UtilsInterface` or just an utill imported here

filip131311 (approved):
I left an inline comment, but i am okay with the rest, 
thank you

PR #938, opened 2025-02-06
Fix color glitch at the beginning

maciekstosio:
This PR fixes color glitch at the beginning. It's done in two stages: 
- we don't set the background until we have theme selected 
- we don't display rest of the UI until we know which one to display 
 
### How Has This Been Tested: 

- Restart the extension and see if the glitch exists

### DIFF
--- a/packages/vscode-extension/src/webview/index.jsx
+++ b/packages/vscode-extension/src/webview/index.jsx
@@ -21,10 +21,10 @@ const root = createRoot(container);
 
 root.render(
   <React.StrictMode>
-    <WorkspaceConfigProvider>
-      <ProjectProvider>
-        <UtilsProvider>
-          <TelemetryProvider>
+    <ProjectProvider>
+      <UtilsProvider>
+        <TelemetryProvider>
+          <WorkspaceConfigProvider>
             <LaunchConfigProvider>
               <DevicesProvider>
                 <DependenciesProvider>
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/webview/index.jsx
```diff
@@ -21,10 +21,10 @@ const root = createRoot(container);
...
 root.render(
   <React.StrictMode>
-    <ProjectProvider>
-      <UtilsProvider>
-        <TelemetryProvider>
-          <WorkspaceConfigProvider>
+    <WorkspaceConfigProvider>
+      <ProjectProvider>
+        <UtilsProvider>
+          <TelemetryProvider>
```
jwajgelt, line 27: Just curious: why is this necessary? I wouldn't think that the other providers need to wait for the theme to be set before they can run.

jwajgelt (approved):
Seems to fix the bug and I don't see any issues with the code.

packages/vscode-extension/src/webview/index.jsx
maciekstosio: I wanted to make sure that nothing renders before workspace config, but actually maybe that's good that we set up the providers anyway, let me check

maciekstosio: Reverted in: https://github.com/software-mansion/radon-ide/pull/938/commits/749780a9da1fb7f4853cbd08a4725e9511fcda0e

PR #939, opened 2025-02-07
Improve contrast of log counter

maciekstosio:
This PR inverses the colors, so the log counts are more visible. 

<video src="https://github.com/user-attachments/assets/d55d3e54-22c1-4bb4-b0e6-4f6dc1588dd4" />

### How Has This Been Tested: 

- Run Radon IDE and see :))

### DIFF
--- a/packages/vscode-extension/src/webview/components/shared/IconButton.css
+++ b/packages/vscode-extension/src/webview/components/shared/IconButton.css
@@ -66,6 +66,7 @@
   border-radius: 4px;
   color: var(--swm-button-counter);
   background-color: var(--swm-button-counter-background);
+  border: 1px solid var(--swm-button-counter-border);
   padding: 2px 4px;
   font-size: 10px;
   position: absolute;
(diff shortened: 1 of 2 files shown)


kmagiera (changes requested):
There are some other variables that refer to badges and would perhaps fit better, in particular they are used in different places in vscode already like for example in the side bar: 
image


```
  --vscode-badge-background:#ed1292;
  --vscode-badge-foreground:#f3eff5;
```

or
```
  --vscode-activityBarBadge-background:#ed1292;
  --vscode-activityBarBadge-foreground:#f3eff5;
```


or
```
  --vscode-extensionBadge-remoteBackground:#ed1292;
  --vscode-extensionBadge-remoteForeground:#f3eff5;
```

One of these should work better IMO

kmagiera:
There's this page that has a list of tons of available variables: https://gist.github.com/estruyf/ba49203e1a7d6868e9320a4ea480c27a

kmagiera (approved):
Approve to unblock. Pick the variables that make sense the most

PR #943, opened 2025-02-10
Cleanup dev plugins architecture along with redux plugin

kmagiera:
In #931 we introduces a buil-in support for redux plugin that doesn't rely on expo dev plugins package. This PR however brought a few potential issues to the codebase, so here we're refactoring it aiming to fix those problems while also preparing for more steamlined support for other tools:

1) removing console log statements introduced in #931 that were polluting the app output
2) clearing up redux registration such that it only relies on the "compose" callback and doesn't require subsequent calls for setting devtools agent nor hook-based integration. This is done by accessing the devtools agent in the same way we do that in the wrapper component.
3) getting rid of direct imports of devtools related code – this for unknown reason turns out to be causing issues with metro as direct imports would impact the order in which modules are loaded and hence may impact the app behavior.
4) we're refactoring module registration such that it is no longer specific to expo. Before we had registry for expo-related plugins and now in #931 a separate registration process was introduced for redux plugin specifically. We're making it more generic and porting both expo plugins and the redux plugin to use it.

### How Has This Been Tested: 
1. Fire up example app with plugins
2. Test redux plugin shows up on the list when redux is used and doesn't show up otherwise.
3. Test scenario from #931

### DIFF
--- a/packages/vscode-extension/lib/wrapper.js
+++ b/packages/vscode-extension/lib/wrapper.js
@@ -11,7 +11,6 @@ const {
   findNodeHandle,
 } = require("react-native");
 const { storybookPreview } = require("./storybook_helper");
-const { useReduxDevTools, isReduxDevToolsReady } = require("./plugins/redux-devtools");
 
 // https://github.com/facebook/react/blob/c3570b158d087eb4e3ee5748c4bd9360045c8a26/packages/react-reconciler/src/ReactWorkTags.js#L62
 const OffscreenComponentReactTag = 22;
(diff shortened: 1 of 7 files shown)


packages/vscode-extension/lib/wrapper.js
```diff
@@ -344,6 +348,15 @@ export function AppWrapper({ children, initialProps, fabric }) {
     [showStorybookStory]
   );
 
+  useAgentListener(
+    devtoolsAgent,
+    "RNIDE_enableNetworkInspect",
+    (payload) => {
+      InternalImports.enableNetworkInspect(devtoolsAgent, payload);
```
maciekstosio, line 355: I think this shouldn't be part of this PR

maciekstosio (changes requested):
Other than that, it looks good!

maciekstosio (approved):
LGTM 🚀

PR #945, opened 2025-02-10
Implement Source Request as for debug adapter protocol standard 

filip131311:
This PR implements `SourceRequest` in debug adapter, it is not not currently in use but might be necessary in the future, if we ever get generated source on the extension side. 

### How Has This Been Tested: 

-run `expo-52-eas`
-run `react-native-78`
-the change does not break any of the applications


kmagiera (approved):
Code looks ok, but I don't understand the description. Can you clarify what it is used for and how, in the future, we can use this code or test it when it becomes relevant?

PR #957, opened 2025-02-12
Add scrolling support

balins:
This PR adds support for scrolling in Radon IDE.

The scroll is simulated by simulator-server and reflected as a swipe (the direction of the swipe is opposite to wheel direction). The captured `WheelEvent` deltas are normalized to provide three predefined scroll magnitudes (short, medium and long) and negated before passing to simulator server for better effect (passing raw values makes it hard to simulate a short scroll).

https://github.com/user-attachments/assets/b49dcf90-7ea0-4348-965a-cf1ebec0cb4e

https://github.com/user-attachments/assets/59fd7a74-c555-4ba3-b482-a96e8a68d187

### DIFF
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -326,8 +326,6 @@ function Preview({
 
   function onWrapperMouseWheel(e: WheelEvent<HTMLDivElement>) {
     e.preventDefault();
-    setIsPressing(false);
-    setIsMultiTouching(false);
   }
 
   const wrapperTouchHandlers = shouldPreventInputEvents


packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -312,11 +324,18 @@ function Preview({
     setIsMultiTouching(false);
   }
 
+  function onWrapperMouseWheel(e: WheelEvent<HTMLDivElement>) {
+    e.preventDefault();
+    setIsPressing(false);
+    setIsMultiTouching(false);
```
jwajgelt, line 330: Just curious: why do we want to cancel the touch when a scroll occurs?

jwajgelt (approved):
I don't see any issues with the code.

packages/vscode-extension/src/webview/components/Preview.tsx
balins: Copied from the function above, forgot to verify this before committing. At first I thought it would help prevent zooming when multitouch is active, but just tested that it only hides the multitouch circles, so removed this in e02c7f0. As for pressing, it seems not to interfere with scrolling, so I removed this `set` as well.

PR #959, opened 2025-02-12
Include license notices from dependencies we include via submodules

kmagiera:
This PR adds separate notices file that includes licenses of dependencies that we include via submoduels, specifically vscode-js-debug, chrom-devtools-ui and the redux plugin. Those dependencies were omitted before from being collected by our license extraction scripts as they are not added as npm dependencies but rather are only included via build scripts run by npm.

Going forward we will have to manually maintain separate file that lists all submodule licenses. This PR introduces this file under main directory in vscode-extension folder. Then we expand third-party notices script to also pull in that newly added file.

### How Has This Been Tested: 
1. Run build:third-party-notice script
2. Make sure the licenses for vscode-js-debug, chrom-devtools and redux plugin are included in the output (ThirdPartyNotices.json)


filip131311 (approved):
LGTM

PR #962, opened 2025-02-12
Add buttons with hints on what to do when stuck on Waiting for ap to Load

kmagiera:
This PR adds clear CTA buttons to the loading screen when we detect it is stuck in "Waiting for app to load" state.

The buttons pop up after 12 seconds of waiting – we are reducing it down from 30 seconds.

The possible options are:
 1) open Radon IDE logs
 2) force show device screen
 3) open troubleshooting guide
 4) Clean rebuild

image

### How Has This Been Tested: 
1. Disable "app ready" event to force the app to get stuck on waiting
2. Test each individual option from the new menu


jwajgelt (approved):
Looks good.
I wonder if adding some message along the lines of "the project is taking longer than expected to connect" above the buttons might be helpful, so there's more context to why they appear.

PR #965, opened 2025-02-12
Surface boot errors to the user and avoid booting unavailable devices

kmagiera:
After recent Xcode update, old versions of SDK were purged from my computer. As a result, the devices I have created are now gone. Since in some projects I've used those devices recently, opening these would attempt to launch the project on unavailable device resulting in unintuitive build error despite the fact build hasn't even started as the error occurred while booting.

This PR addresses this issue in two ways:
1) When selecting device (including the initial selection), we verify whether the device is available and prevent from devices with unavailable runtime from being selected
2) We add separate handling for boot error that display more intuitive message about the error and point to the IDE logs instead of the build logs where users are more likely to find the cause.

On top of the above we made some adjustments in initialization logic to avoid "spinning wheel" when selected device is unavailable:
1) We updated Device Selection component to have no direct interface with the PreviewView, this way it can properly manage its own state and update the placeholder when there is no device to be selected
2) We moved logic for displaying empty state to a renamed DevicesNotFound component (now called NoDeviceView). It can now render state when there are other devices but the initial selection has failed.

When last selected device can no longer be used user will get the following prompt:
image

The IDE will not move over to the build phase and will wait for the user to select the device:
image

If there are some paths in which the user can still select the device that would fail on the boot, the user will get the following error banner:
image

### How Has This Been Tested: 
1. To test this I selected iOS device in my project
2. I uninstalled SDK from Xcode
3. I reopened the project where iOS device was selected
4. Now, this should result in device not being selected and user is prompted to select a different one
5. Comment out the added code that checks for device availability when selecting the device
6. Now it should go past the selection step and you should get boot error banner presented


packages/vscode-extension/src/project/project.ts
```diff
@@ -717,6 +717,13 @@ export class Project
 
   //#region Select device
   private async selectDeviceOnly(deviceInfo: DeviceInfo) {
+    if (!deviceInfo.available) {
+      window.showErrorMessage(
+        "Selected device is not available. Perhaps the device runtime is not installed. Please select another device.",
```
jwajgelt, line 722: In "Manage Devices" we call it the "System Image", not "Device Runtime". I think it'd be best to have the name be consistent throughout the UI.

```suggestion
        "Selected device is not available. Perhaps the system image it uses not installed. Please select another device.",
```

jwajgelt (approved):
No issues I can see.

PR #967, opened 2025-02-14
"Outline Renders" tool

jwajgelt:
Adds the "Outline Renders" tool which highlights React components when they are rendered. Based on [react-scan](https://github.com/aidenybai/react-scan) utility.

- we reuse the [`react-scan`](https://github.com/aidenybai/react-scan) implementation of drawing the outlines on the canvas. Accessing it from our frontend code required abstracting the renderer code and exporting it from the package. These changes are available and pulled from a public fork the `react-scan` repository. https://github.com/software-mansion-labs/react-scan/pull/1
- we use the [`bippy`](https://github.com/aidenybai/bippy) library for instrumenting the React rendering process. We include the built library code (extracted from the official release) in this repo in the `packages/vscode-extension/lib/bippy` directory and inject it into the RN Application bundle. 
We do it this way because it needs to be included in the final extension package in order to be injected into the RN application, and only the `lib` directory is available to Metro.

### How Has This Been Tested: 
- Open the "Tools" dropdown and toggle "Outline Renders".
- Trigger component updates in the application.
- The updated components should be briefly highlighted, with their names and update counter displayed.
- Resizing the phone Preview should not break the outlines' positions.
- The tool's enabled status should persist app, extension and IDE launches/reloads.


packages/vscode-extension/lib/instrumentation.js
```diff
@@ -0,0 +1,198 @@
...
+
+export const setInstrumentationOptions = (partialOptions) => {
+  const componentBlocklist = partialOptions.componentBlocklist
+    ? new Set(partialOptions.componentBlocklist)
+    : options.componentBlocklist;
+  options = { ...options, componentBlocklist, ...partialOptions };
+  if (!inited) {
+    inited = true;
+    instrument({
+      onCommitFiberRoot: (rendererID, root) => onCommitFiberRoot(rendererID, root),
```
filip131311, line 195:
```suggestion
      onCommitFiberRoot: onCommitFiberRoot,
```
Why not?

packages/vscode-extension/lib/wrapper.js
```diff
@@ -409,6 +412,28 @@ export function AppWrapper({ children, initialProps, fabric }) {
     }
   }, [!!devtoolsAgent && hasLayout]);
 
+  useEffect(() => {
+    if (!devtoolsAgent) {
+      return;
+    }
+    InternalImports.setInstrumentationOptions({
```
filip131311, line 419: nit: this function takes partial options as an argument could you modify the name to point to that? 
for example `setPatrialOptions`

filip131311 (commented):
I left some inline comments. I also have additional question regarding how this works: 

If i click multiple tiems on something that changes UI (as in example video) the count will count all the renders of the element that shows up. My question is "when is `onCommitFiberRoot` triggered? if it counts those multiple changes as one commit? Is it something that is normal for react and i shouldn't be worried about it? 

https://github.com/user-attachments/assets/544b88ce-aeba-4fcf-9231-77f42a50a7f6

packages/vscode-extension/lib/instrumentation.js
```diff
@@ -0,0 +1,198 @@
...
+    const { width: screenWidth, height: screenHeight } = Dimensions.get("screen");
+    const statusBarHeight = StatusBar.currentHeight || 0;
+    return {
+      x: 0,
+      y: statusBarHeight,
+      width: screenWidth,
+      height: screenHeight,
+    };
+  } else {
+    const { width: windowWidth, height: windowHeight } = Dimensions.get("window");
```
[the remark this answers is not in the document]
jwajgelt: To correct for the status bar on Android.
Dimensions returned from `measureInWindow` are relative to the `window`, so we should be using that, but because the canvas we use later to draw the overlays covers the entire screen, we need to translate the measurements from `window` coordinate space to `screen`.
On iOS it seems to be the same, but on Android you need to translate everything by the Status Bar height.

packages/vscode-extension/lib/instrumentation.js
```diff
@@ -0,0 +1,198 @@
...
+    boundingRect.y /= windowHeight;
+    boundingRect.width /= windowWidth;
+    boundingRect.height /= windowHeight;
+
+    if (!blueprint) {
+      blueprintMap.set(getFiberId(fiber), {
+        name,
+        count: 1,
+        boundingRect,
+        didCommit: didCommit ? 1 : 0,
```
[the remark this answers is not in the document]
jwajgelt: The renderer expects it in this format, so we have to do the conversion _somewhere_. It might as well be here.

packages/vscode-extension/lib/instrumentation.js
jwajgelt: There was an extra param there at some point, this was lost in refactoring.

packages/vscode-extension/lib/wrapper.js
jwajgelt: Out of the two, I like `setOptions` better. It's not supposed to be "a setter for a field `options`", it's supposed to be a function which sets (some) options.
`updateInstrumentationOptions` might be better in that sense, it doesn't imply a "setter".

packages/vscode-extension/lib/instrumentation.js
```diff
@@ -0,0 +1,198 @@
...
+  const stripped = stripAnimatedFromComponentName(name);
+  return (
+    options.componentBlocklist.has(name) || (stripped && options.componentBlocklist.has(stripped))
+  );
+}
+
+function getWindowRect() {
+  if (Platform.OS === "android") {
+    const { width: screenWidth, height: screenHeight } = Dimensions.get("screen");
+    const statusBarHeight = StatusBar.currentHeight || 0;
```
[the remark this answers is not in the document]
jwajgelt: It seems to work fine Bluesky which, as far as I can tell, uses screens (through native-stack). The app is rendered below the status bar though, so this might be why.
IMO it's better to have this solution, which seems to work for the examples I've checked, than ignore the status bar altogether and have the outlines always offset.

filip131311 (approved):
Thank you

packages/vscode-extension/package.json
```diff
@@ -648,7 +649,9 @@
     "vscode-js-debug": "file:../vscode-js-debug",
     "vscode-test": "^1.5.0",
     "ws": "^8.14.2",
-    "xml2js": "^0.6.2",
-    "tar": "^7.4.3"
+    "xml2js": "^0.6.2"
+  },
+  "dependencies": {
+    "react-scan": "https://github.com/software-mansion-labs/react-scan/releases/download/0.1.3-radon/react-scan-0.1.3-radon.tgz"
```
kmagiera, line 655: we use dev dependencies for everything that is required in build-time.

kmagiera (changes requested):
I'd prefer that we actually use the name `react-scan` instead of "render outlines" given we rely on the tool. 

We also want to have a bit more details about this setup with the extra repo. Specifically, we should put in the description how these different repositoreis are structured. What are the changes that we made. Why is bippy source pulled into the repo etc

packages/vscode-extension/package.json
jwajgelt: I'm not fully familiar with how the extension is built and how it accesses its dependencies yet.
Since the `react-scan` package is used _by the extension_ at runtime, it made more sense to me to include it as a regular dependency.

jwajgelt:
> I'd prefer that we actually use the name `react-scan` instead of "render outlines" given we rely on the tool.

I'm not convinced that's a good idea.
- The name is not very descriptive. That way the user needs to already be familiar with `react-scan` to know what our tool is supposed to do. Granted, "Outline Renders" might not be the best name in that metric either, something like "Outline Components when Rendered" might have been better.
- I think it gives the wrong expectation to users. What we currently implement is a _very small_ part of what the "real" `react-scan` does. If someone who uses `react-scan` on web tries to use our tool, they would probably be disappointed with how little they can do compared to it.

packages/vscode-extension/src/webview/components/RenderOutlinesOverlay.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  useEffect(() => {
+    const host = hostRef.current;
+    const canvasEl = canvasRef.current;
+    if (!host || !canvasEl) {
+      return;
+    }
+
+    const dpr = getDpr();
+    let size = { width: hostRef.current.clientWidth, height: host.clientHeight };
```
kmagiera, line 43: use `host` in both places here or `hostRef` in both places?

packages/vscode-extension/src/webview/components/RenderOutlinesOverlay.css
```diff
@@ -0,0 +1,16 @@
+.render-outlines-overlay {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  pointer-events: none;
+  z-index: 2147483646;
```
kmagiera, line 8: would rather not have it here. Last time we had z-index it'd show stuff over modals etc. Can this be avoided?

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -492,6 +493,7 @@ function Preview({
                 }}
                 className="phone-screen"
               />
+              <RenderOutlinesOverlay />
```
kmagiera, line 496: do we want to conditionally render this? does it make sense?

packages/vscode-extension/lib/wrapper.js
```diff
@@ -407,6 +410,28 @@ export function AppWrapper({ children, initialProps, fabric }) {
...
+  useEffect(() => {
+    if (!devtoolsAgent) {
+      return;
+    }
+    InternalImports.updateInstrumentationOptions({
+      reportRenders: (blueprintOutlines) => {
+        devtoolsAgent._bridge.send("RNIDE_rendersReported", { blueprintOutlines });
+      },
+    });
+    devtoolsAgent._bridge.send("RNIDE_queryInstrumentationOptions");
```
kmagiera, line 422: I don't think this call is necessary. This seems to follow similar scheme as the network plugin. In that case we wait for devtools to connect before we send enable command. This way we can be sure that we always get the enabled state event even when it was sent before reload occured.

packages/vscode-extension/src/webview/components/RenderOutlinesOverlay.tsx
jwajgelt: How did that even happen 🤔

packages/vscode-extension/src/webview/components/RenderOutlinesOverlay.css
jwajgelt: This was taken directly from `react-scan`. It's probably fine without it, since we draw the canvas over the app preview (and we don't need to make sure we draw over the entire window, like the "real" `react-scan` does).

packages/vscode-extension/src/webview/components/Preview.tsx
kmagiera: sure, we can have that logic inside the component impl

packages/vscode-extension/lib/wrapper.js
jwajgelt: fixed by emitting `RNIDE_updateInstrumentationOptions` on `RNIDE_appReady` (extension side)

packages/vscode-extension/src/webview/components/Preview.tsx
jwajgelt: Done.

kmagiera (approved):
Looks good. As final thought I think we should add the license to ale be included in the notices file (one that we generate in buiild step).
I think that we can use submodules-NOTICES.json for this reason despite dippy not being in submodule. If we do that, we don't need LICENSE file in the source tree

PR #971, opened 2025-02-17
Add node version check

filip131311:
This PR adds a node version check to the start up flow that checks node version found in the PATH to the minimum version specified either by the package.json or react-native dependency package.json. 

### How Has This Been Tested: 

-run `react-native-78` test app and change minimum versions in package.json and in react-native dependency, check if the warning appears in "run diagnostics screen".

### DIFF
--- a/packages/docs/docs/guides/troubleshooting.md
+++ b/packages/docs/docs/guides/troubleshooting.md
@@ -112,4 +112,4 @@ If you're using a theme that doesn't look good with Radon IDE, please open an is
 
 ### -sec-num- Node version is not supported
 
-Radon IDE uses node version inherited from a shell run in the application root, and checks if it satisfies react-native minimum requirements, if you have a specific setup that uses node version from outside this range or you want to narrow down the scope of compatible versions for your internal needs you can do so, by specifying `engines` field of your applications `package.json`. If you have a problem installing a compatible node version, we recommend usage of a tool like [nvm](https://github.com/nvm-sh/nvm).
+Radon IDE uses node version inherited from a shell run in the application root, and checks if it satisfies `react-native` minimum requirements. If you have a specific setup that uses node version from outside this range or you want to narrow down the scope of compatible versions for your internal needs, you can do so by specifying `engines` field of your applications `package.json`. If you have a problem installing a compatible node version, we recommend using a tool like [nvm](https://github.com/nvm-sh/nvm).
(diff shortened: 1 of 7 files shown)


packages/docs/docs/guides/troubleshooting.md
```diff
@@ -109,3 +109,7 @@ Radon IDE supports a wide range of themes, but it's not possible to support all
 If you're using a theme that doesn't look good with Radon IDE, please open an issue on the [Radon IDE GitHub repository](https://github.com/software-mansion/radon-ide/issues/new/choose) and we'll try to fix it. Menawhile you can use built-in themes. Radon IDE comes with dedicated light and dark themes.
 
 <img width="698" alt="change-theme-type" src="/img/docs/change_theme_type.png" className="shadow-image"/>
+
+### -sec-num- Node version is not supported
+
+Radon IDE uses node version taken from the PATH of the application root, and checks if it satisfies react-native minimum requirements, if you have a specific setup that uses node version from outside this range or you want to narrow down the scope of compatible versions for your internal needs you can do so, by specifying `engines` field of your applications `package.json`. If you have a problem installing a compatible node version, we recommend usage of a tool like [nvm](https://github.com/nvm-sh/nvm).
```
jwajgelt, line 115:
```suggestion
Radon IDE uses node version taken from the PATH of the application root, and checks if it satisfies react-native minimum requirements, if you have a specific setup that uses node version from outside this range or you want to narrow down the scope of compatible versions for your internal needs you can do so, by specifying `engines` field of your applications `package.json`. If you have a problem installing a compatible node version, we recommend using a tool like [nvm](https://github.com/nvm-sh/nvm).
```

jwajgelt, line 115: > taken from the PATH of the application root

I don't think I understand this sentence. `PATH` is an environmental variable, it's not a property of "the application root".

packages/vscode-extension/src/utilities/minimumSuportedNodeVersion.ts
```diff
@@ -0,0 +1,32 @@
+import fs from "fs";
+import { Logger } from "../Logger";
+
+/* 
+This function return a minimum supported node version for given appRoot assuming node_modules are installed,
+if node modules are not installed the function will return a default value, but it may be incorrect in future versions
+of react native, so it should be always used after node_modules installation.
+*/
+export function minimumSupportedNodeVersion(appRoot: string): string {
```
jwajgelt, line 9: NIT:
Functions should be verbs, not nouns.
```suggestion
export function getMinimumSupportedNodeVersion(appRoot: string): string {
```

jwajgelt (approved):

PR #973, opened 2025-02-17
Add first-party react-query devtools support

maciekstosio:
This PR adds first-party react-query dev tools support. It works without any additional configuration and does not conflict with the Expo Dev plugin, so the old implementation, which is based on Expo Dev Plugins (#878), is being removed. 

**The solution works as follows:** 
- While transforming the react-query library, we inject a code that calls `window.__RNIDE_REACT_QUERY_CLIENT_INIT__` with `queryClient` on mount;
- `__RNIDE_REACT_QUERY_CLIENT_INIT__`  broadcasts the apps react-query state using RN dev tools;
- Radon IDE extension is used as a proxy - receives the events from RN dev-tools and sends them to React Query DevTools webview;
- Thanks to the above, webview has synced react-query state and displays it using [original react-query devtools](https://github.com/TanStack/query/tree/main/packages/query-devtools).

**WebUI**

WebUI needs to be bundled so it can be inserted into webview.  For that, I created a separate repo https://github.com/software-mansion-labs/radon-ide-react-query-webui - which consists of https://github.com/TanStack/query/tree/main/packages/query-devtools with additional logic required for syncing the state between app and dev tools. Built artifacts of webui are committed to the webui repo and are copied while building the Radon IDE extension. 

**Communication**

The reason is to sync both states (apps with dev tools one). The sync is based on https://github.com/TanStack/query/blob/main/packages/query-broadcast-client-experimental/src/index.ts, but instead of `BroadcastChannel`, we communicate through RN dev-tools similarly to redux dev tools #931.

### How Has This Been Tested: 

- run `eas-52-prebuild-with-plugins`
- check if `React Query DevTools`, are available in tools list 
- make sure `React Query DevTools DevPlugin` is not visible
- open `React Query DevTools`
- go to plugins page
- press React Query counter 
- confirm that the events are visible


packages/vscode-extension/src/plugins/react-query-devtools-plugin/react-query-devtools-plugin.ts
```diff
@@ -0,0 +1,78 @@
...
+  );
+
+  return webviewProvider;
+}
+
+export const createReactQueryDevtools = (toolsManager: ToolsManager): ToolPlugin => {
+  const webViewProvider = initializeReactQueryDevPlugin();
+
+  let proxyDevtoolsListener: null | ((event: string, payload: any) => void) = null;
+  webViewProvider?.setListener((webview) => {
```
kmagiera, line 32: wouldn't it make more sense to access toolsManager from webview using `Ide` singleton instead of introducing more singletons?

kmagiera (changes requested):
It'd be good if you could work on the PR description a little bit more. Specifically, why is there a separate repo for UI code, is it based on some other repo, what are the changes? There are no docs about that and the repo has empty README too.

packages/vscode-extension/src/plugins/react-query-devtools-plugin/react-query-devtools-plugin.ts
maciekstosio: Something like this? https://github.com/software-mansion/radon-ide/pull/973/commits/5e6e790de6147a647c9ce86e36159dd42b6cbcc5 it looks much simpler, but we never clear the devtools listener

packages/vscode-extension/src/plugins/react-query-devtools-plugin/ReactQueryDevToolsPluginWebviewProvider.ts
```diff
@@ -0,0 +1,139 @@
...
+): string {
+  const nonce = getNonce();
+  const baseUri = getUri(webview, extensionUri, [PATH]);
+  const files = fs.readdirSync(baseUri.path, { recursive: true });
+  const cssFiles = files
+    .filter((file) => typeof file === "string")
+    .filter((file) => file.endsWith(".css"));
+  const jsFiles = files
+    .filter((file) => typeof file === "string")
+    .filter((file) => file.endsWith(".js"));
```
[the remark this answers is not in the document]
maciekstosio: Nah, there is actually one of each. The vite produces files with hash in the file name, I assumed to avoid caching, thus for convinience I added getting the names automatically. But seems it might not be needed - I tested and it seemes to be ok with explicite name. Changed in: https://github.com/software-mansion/radon-ide/pull/973/commits/05aabf4552cda8596b6af271836cf6db4a4ae544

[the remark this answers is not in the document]
maciekstosio: As for 2. point: static JS is `@tanstack/query-devtools` bundled. The static CSS can be actually skipped - it was generated when creating blank project - changed in: https://github.com/software-mansion/radon-ide/pull/973/commits/47d28a319420a8ea8bd4cc1fcbbe3aacf07af780. CSS used by `react-query-devtools` are added by JS.

packages/vscode-extension/lib/babel_transformer.js
```diff
@@ -124,6 +122,8 @@ function transformWrapper({ filename, src, ...rest }) {
     if (version.startsWith("0.78")) {
       src = `module.exports = require("__RNIDE_lib__/JSXRuntime/react-native-78/${jsxRuntimeFileName}");`;
     }
+  } else if (isTransforming("node_modules/@tanstack/react-query/src/index.ts")) {
+    src = `require("__RNIDE_lib__/plugins/react-query-devtools.js");${src}`;
```
kmagiera, line 126: can you verify how this works exacly, specifically how is this getting resolved in the produced bundle? Was hoping that we can set it up in a way that we don't create require cycle but this seems to make a direct cycle as react-query lib requires the plugin and the plugin requires react-query

packages/vscode-extension/lib/plugins/react-query-devtools.js
```diff
@@ -0,0 +1,91 @@
...
+      const query = queryCache.get(queryHash);
+
+      if (query) {
+        queryCache.remove(query);
+      }
+    });
+  });
+}
+
+register('RNIDE-react-query-devtools');
```
kmagiera, line 84: I think we could just use `react-query` here. I don't see the reason why to prefix the modules. We were doing that for modules form expo-plugins in anticipation of having built-in ones

kmagiera (approved):
Looks good. Thanks for addressing all the comments. Also happy that you found a way for the webview to imports css and js without listing the directory. 

Can you just verify how the require cycle is resolved when loading react query from the lib code? I made an inline comment about that in the relevant place.

packages/vscode-extension/lib/babel_transformer.js
maciekstosio: I think that should help with requrie cycle: https://github.com/software-mansion/radon-ide/pull/973/commits/343c5c9d4c5244681b5aba844d07d4a82c742fcb

PR #974, opened 2025-02-17
Report telemetry events related to tool usage

jwajgelt:
Reports telemetry events related to tool usage:
- enable/disable tool
- opening from tools dropdown
- toggling webview visibility
### How Has This Been Tested: 
By placing breakpoints on `getTelemetryReporter` calls and toggling/opening/hiding the tools

### DIFF
--- a/packages/vscode-extension/src/plugins/expo-dev-plugins/ExpoDevPluginWebviewProvider.ts
+++ b/packages/vscode-extension/src/plugins/expo-dev-plugins/ExpoDevPluginWebviewProvider.ts
@@ -9,6 +9,7 @@ import {
 import { IDE } from "../../project/ide";
 import { ExpoDevPluginToolName } from "./expo-dev-plugins";
 import { Logger } from "../../Logger";
+import { reportToolVisibilityChanged } from "../../project/tools";
 
 function generateWebviewContent(pluginName: ExpoDevPluginToolName, metroPort: number): string {
   const iframeURL = `http://localhost:${metroPort}/_expo/plugins/${pluginName}`;
@@ -37,6 +38,10 @@ export class ExpoDevPluginWebviewProvider implements WebviewViewProvider {
       enableScripts: true,
       localResourceRoots: [Uri.joinPath(this.context.extensionUri, "dist")],
     };
+
+    webviewView.onDidChangeVisibility(() =>
+      reportToolVisibilityChanged(this.pluginName, webviewView.visible)
+    );
 
     const metroPort = IDE.getInstanceIfExists()?.project.metro.port;
     if (!metroPort) {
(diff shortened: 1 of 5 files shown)


kmagiera (changes requested):
Just a few high level comments for those changes:
1) Have you considered passing tool details as parameters of the events rather than part of the event name? One benefit is that under single name the events will be aggregated, such that we can get a sense of how tools are used in general and then narrow it down to specific tool
2) I'm not so sure about using `onDidChangeVisibility` – this is triggered for example when you switch between different panels. Not sure if this gives us any valuable input or allows to draw any conclusions as for the usage of particular tool. We'd only know that people switch between panels which on its own seem like a natural thing to do and doesn't really mean much IMO
3) For disposables you can use `disposeAll` from `src/utilities/disposables` instead of loop (which implements a re-entrant safe way of disposing – not that it matters in this particular place but would be better to spread this pattern)
4) For "opening" tools I'd rather rely on some of the webview controller stuff rather than `openTool` call – this one is only used when you open tool from the tools dropdown. When you have the tool enabled and you restart the editor, you can also open the tool by just selecting its tab in the panel. I think we'd rather have that

jwajgelt:
> Just a few high level comments for those changes:
> 
>     1. Have you considered passing tool details as parameters of the events rather than part of the event name? One benefit is that under single name the events will be aggregated, such that we can get a sense of how tools are used in general and then narrow it down to specific tool
> 
I have, in conversation with @kacperkapusciak he mentioned that the way that we use the dashboard makes it hard to use these parameters, and he'd rather have all the info in the event name and then use it to filter what happened.
>
>     2. I'm not so sure about using onDidChangeVisibility – this is triggered for example when you switch between different panels. Not sure if this gives us any valuable input or allows to draw any conclusions as for the usage of particular tool. We'd only know that people switch between panels which on its own seem like a natural thing to do and doesn't really mean much IMO
>
I've thought that at first, but again, Kacper convinced me that this is useful as well.
Tracking just `enable/disable` doesn't actually let you know if the user _used_ the tool. For example you can have a user enable the Redux Devtools to check them out, decide they don't work well for them and leave them enabled, but never open the panel again. If we only track `enable/disable` events, we'd think the user was using the tool, even though in reality they don't.
The total quantity is meaningless, but something like %(visible was triggered in a session / session count) is (in my opinion, at least) a pretty good indicator how often a tool is actually used.
>
>     3. For disposables you can use disposeAll from src/utilities/disposables instead of loop (which implements a re-entrant safe way of disposing – not that it matters in this particular place but would be better to spread this pattern)
>
:+1:
>
>     4. For "opening" tools I'd rather rely on some of the webview controller stuff rather than openTool call – this one is only used when you open tool from the tools dropdown. When you have the tool enabled and you restart the editor, you can also open the tool by just selecting its tab in the panel. I think we'd rather have that
>
This is precisely what the visibility check is for.

kmagiera:
ok, if you want this to be reported like that.

But given that still having the open event tied to the button doesn't make a lot of sense. Especially given you already have visibility that will be triggered immediately after. In addition, people would still be able to "open" the tool w/o using that button which will still trigger visibility but not open event

jwajgelt:
> ok, if you want this to be reported like that.
> 
> But given that still having the open event tied to the button doesn't make a lot of sense. Especially given you already have visibility that will be triggered immediately after. In addition, people would still be able to "open" the tool w/o using that button which will still trigger visibility but not open event

Yeah, I wasn't sure if I want to leave the `open` event in.
Maybe you're right it doesn't make sense at this point. I'll remove it.

kmagiera (approved):

PR #975, opened 2025-02-18
Fix Path Set Up Process

filip131311:
This PR fixes a problem with extension not being able to refresh node version after it was changed by nvm. Even when the extension was deactivated and activated again. This would in turn lead to user frustration when the node version they ware using was in initially incorrect and applied user solution would not take effect. 

The issue was triggered when the following conditions were met: 
- process vscode was either not restarted completely, or was opened from terminal that would hold the old configuration
- the `$HOME/.nvm/versions/node/{NODE_VERSION}/bin` path was the first one in vscodes path

Because we would open a new shell with vscode processes path passed as an env variable it wouldn't start completely fresh and nvm scripts located in `.zshrc` files  responsible for aliasing seem to have an optimization to not  change the version path if it was already the newest one, causing as receiving the old path to node still living in vscode process.

The solution is to strip the env of a new shell process and allow it to boot fresh. 

### How Has This Been Tested: 

- run `react-native-78` and set a brakepoint after path was recived, run it in following configurations:
    - run with a node 21 
    - reload after switching the node version in seprate terminal 
    - run from scratch

### DIFF
--- a/packages/vscode-extension/src/utilities/subprocess.ts
+++ b/packages/vscode-extension/src/utilities/subprocess.ts
@@ -22,7 +22,10 @@ async function getPathEnv(appRoot: string) {
     "-i",
     "-c",
     `cd "${appRoot}" && echo "${RNIDE_PATH_DELIMITER}$PATH${RNIDE_PATH_DELIMITER}"`,
-  ]);
+  ],{
+    extendEnv:false,
+    env:{}
+  });
   const path = stdout.split(RNIDE_PATH_DELIMITER)[1].trim();
   Logger.debug("Obtained PATH environment variable:", path);
   return path;


packages/vscode-extension/src/utilities/subprocess.ts
```diff
@@ -18,11 +18,14 @@ async function getPathEnv(appRoot: string) {
...
-  const { stdout } = await execa(shellPath, [
-    "-i",
-    "-c",
-    `cd "${appRoot}" && echo "${RNIDE_PATH_DELIMITER}$PATH${RNIDE_PATH_DELIMITER}"`,
-  ]);
+  const { stdout } = await execa(
+    shellPath,
+    ["-i", "-c", `cd "${appRoot}" && echo "${RNIDE_PATH_DELIMITER}$PATH${RNIDE_PATH_DELIMITER}"`],
+    {
+      extendEnv: false,
```
kmagiera, line 25: can you add comment here explaining this option as it is rarely used and not seen in other places in our codebase, therefor may be good to be able to understand it in case someone needs to copy this code for some other purpose

kmagiera (approved):

PR #978, opened 2025-02-19
improve url experience - get routes based on file based routing

kewinzaq1:
This PR improves how expo router integration works.
These changes add initial routes based on the file based routing that allows to use this integration immediately without the need to visit them at first.

### How Has This Been Tested: 

The basic routes (excluding dynamic because currently this integration does not allow to input dynamic values, so it does not make sense including them) should be visible and clickable (first start, after reload, after restart, ...all the time)

https://github.com/user-attachments/assets/b8e2a63b-6808-4d9b-a9b1-4b5b1685b4a9

### DIFF
--- a/packages/vscode-extension/lib/wrapper.js
+++ b/packages/vscode-extension/lib/wrapper.js
@@ -360,6 +360,20 @@ export function AppWrapper({ children, initialProps, fabric }) {
     []
   );
 
+  useAgentListener(devtoolsAgent, "RNIDE_loadFileBasedRoutes", (payload) => {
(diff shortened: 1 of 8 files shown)


filip131311 (commented):
Hey @kewinzaq1, thanks a lot for the input! We took some time to think it over and decided to try a different route for a couple of reasons:
1) We're a bit concerned that your getRoutes utility might not hold up if Expo makes any internal changes down the line. Ideally, we want to source such info directly from Expo.
2) We're leaning towards a more "web-like" experience soon, where typing in the URL—possibly with the help of something like your getRoutes for auto-suggestions—might make more sense.

PR #979, opened 2025-02-19
Don't show full-screen notice when fast refresh happens

kmagiera:
When fast refresh is performed, we'd show a full screen overlay. This was very distracting especially given other tools doesn't do that. This PR introduces some changes to the way we inform about fast refresh:

1) We no longer show the full screen overlay.
2) When fast refresh is ongoing and user tries to interact with the app, only then we block the interaction and show full screen notice. This is to prevent accidentally hitting the in-between state when they may expect the code to already be updated while it is still reloading.
3) We show a tiny notification in bottom left corner when fast refresh is ongoing, but only do it when it's running for at least half a second. Most of reloads are much faster and there's no point in showing this information
4) After successful reload we display a small notice in the same place that says the reload was successful.

Loading popover:
image
image

Screen overlay when attempting to click on it:
image

### How Has This Been Tested: 
1. Run test app
2. Try editin app's code to force fast refresh -> see the notice show up for a bit over a second after successful refresh
4. Add some very large import (`import 'typescript/lib/typingsInstaller';` is almost 2MB file that takes several seconds to fast refresh). Make sure the "loading" notice is show and also see the overlay when interacting with the screen while it is happening.


maciekstosio (approved):
I wonder if the tooltip part shouldn't be something separate that we could use also for other notifications. On the other hand, if we want to post more messages, maybe we should use build-in vscode notifications. Other then that it looks ok.

PR #981, opened 2025-02-20
Additional expo detection method

filip131311:
This PR adds additional method for detecting if expo is used in the user application. The current method was assuming that if expo is used, users have defined scripts that use, expo CLI in their package json. This assumption unfortunately does not hold true for every user especially those using exclusively eas builds. To solve that issue we also check if a user has `"expo"`, configuration in their `app.json`. 

### How Has This Been Tested: 

Run `expo-52-prebuild-with-plugins` test app and remove all scripts form `package.json` before the change  rudon would crash, after it works.

### DIFF
--- a/packages/vscode-extension/src/utilities/expoCli.ts
+++ b/packages/vscode-extension/src/utilities/expoCli.ts
@@ -20,12 +20,18 @@ export function shouldUseExpoCLI() {
 
   const appRootFolder = getAppRootFolder();
   let hasExpoCLIInstalled = false,
-    hasExpoCommandsInScripts = false;
+    hasExpoCommandsInScripts = false,
+    hasExpoConfigInAppJson = false;
   try {
     hasExpoCLIInstalled =
       require.resolve("@expo/cli/build/src/start/index", {
         paths: [appRootFolder],
       }) !== undefined;
+  } catch (e) {}
+
+  try {
+    const appJson = require(path.join(appRootFolder, "app.json"));
+    hasExpoConfigInAppJson = Object.keys(appJson).includes("expo");
   } catch (e) {}
 
   try {
@@ -35,5 +41,5 @@ export function shouldUseExpoCLI() {
     });
   } catch (e) {}
 
-  return hasExpoCLIInstalled && hasExpoCommandsInScripts;
+  return hasExpoCLIInstalled && (hasExpoCommandsInScripts || hasExpoConfigInAppJson);
 }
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/utilities/expoCli.ts
```diff
@@ -20,20 +20,26 @@ export function shouldUseExpoCLI() {
...
+    hasExpoConfigInAppJson = false;
   try {
     hasExpoCLIInstalled =
       require.resolve("@expo/cli/build/src/start/index", {
         paths: [appRootFolder],
       }) !== undefined;
   } catch (e) {}
 
+  try {
+    const appJson = require(path.join(appRootFolder, "app.json"));
```
jwajgelt, line 33: Might be worth handling `app.config.js`, since at least [expo supports it](https://docs.expo.dev/versions/latest/config/app/) -- unless we already assume `app.json` is used elsewhere?

filip131311: @jwajgelt  done, thank you for this suggestion <3

jwajgelt (approved):

packages/vscode-extension/src/utilities/expoCli.ts
```diff
@@ -20,20 +20,35 @@ export function shouldUseExpoCLI() {
 
   const appRootFolder = getAppRootFolder();
   let hasExpoCLIInstalled = false,
-    hasExpoCommandsInScripts = false;
+    hasExpoCommandsInScripts = false,
+    hasExpoConfigInAppJson = false,
+    hasExpoConfigInAppConfigJs = false;
```
jwajgelt, line 25: These are very nitty:
1. this is a personal opinion, but I dislike multiple variable declarations in a single statement, I feel like splitting this would be more readable.
```suggestion
  let hasExpoCommandsInScripts = false;
  let hasExpoConfigInAppJson = false;
  let hasExpoConfigInAppConfigJs = false;
```
2. Personally, I'd rewrite this function with early returns, but that's a stylistic choice.
```
try {
  const hasExpoCliInstalled = ...
  if (!hasExpoCliInstalled) { return false }
} catch {}

try {
  const appJson = require(...)
  if ((appJson).includes("expo")) return true;
} catch {}

...

return false;
```

jwajgelt: And honestly, for 2. in this case, I think your approach is more readable -- you have the explicit condition in the return statement, while in mine you have to read the whole function to understand it.

packages/vscode-extension/src/utilities/expoCli.ts
```diff
@@ -20,20 +20,35 @@ export function shouldUseExpoCLI() {
...
   try {
     hasExpoCLIInstalled =
       require.resolve("@expo/cli/build/src/start/index", {
         paths: [appRootFolder],
       }) !== undefined;
   } catch (e) {}
 
+  try {
+    const appJson = require(path.join(appRootFolder, "app.json"));
```
kmagiera, line 34: should we use `requireNoCache` here? seems like we just want to read the contents of the file

PR #983, opened 2025-02-21
Add command icons setup and use custom off icon for closing IDE

kmagiera:
When IDE is used in secondary side-panel, we display icon that allows for closing the IDE that sits next to the icon for closing the whole sidebar. This has been very misleading and made it unclear what does what:
image

This PR changes the icon to a custom "shutdown" icon. Apparently while VSCode allows menu icons to be provided as SVG, they will not properly inherit colors from the theme as they are loaded using `background-image: url` scheme. For this purpose we need to bundle custom icon as font and use `$(icon-name)` syntax to reference it.

This PR adds scrips and instructions for bundling icons as fonts and uses this new workflow to generate woff2 file with the new "off" icon.

1) For font generation we use `fantasticon` package.
2) We add new `build:icons` script that takes the SVG icons and creates woff file.
3) We add config file for fantasticon where we list each icon with assigned codepoint. Codepoint is assigned explicitely in the config as the same codepoint needs to be referenced in package.json. This is to avoid the codepoint changes when more icons are added.
4) We add README inside font-icons directory explaining how to add more icons in the future
5) We use the new off icon for the "closePanel" action:

image

### How Has This Been Tested: 
1. Run `build:icons` script. Make sure it generates proper woff file
2. Run IDE and test new shutdown button in secondary sidebar.


packages/vscode-extension/package.json
```diff
@@ -620,6 +629,7 @@
     "format": "prettier --write --list-different src",
     "build:tests": "tsc --project tsconfig.test.json",
     "build:plugins": "bash ./scripts/install-dev-plugins.sh",
+    "build:icons": "npm exec fantasticon -- -c ./font-icons/fantasticon.config.json",
```
jwajgelt, line 632: Since we have `fantasticon` as dev dependency, shouldn't
```suggestion
    "build:icons": "fantasticon -c ./font-icons/fantasticon.config.json",
```
be sufficient?

kmagiera: You're right. This works. I just blindly copied command I was testing from command line

jwajgelt (approved):
I don't see any issues.

PR #984, opened 2025-02-21
Add Radon IDE version info to settings dropdown

kmagiera:
This PR adds version information to settings banner. This will make it easier for the users to tell which version they are running as sometimes even though the extension is updated, users may forget to reload extension in which case the new version despite being listed as installed is not running/

image
image

### How Has This Been Tested: 
1. Open device settings dropdown
2. See the current version there. Check it under different vscode themes

### DIFF
--- a/packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.css
+++ b/packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.css
@@ -38,6 +38,12 @@
   margin-bottom: 20px;
 }
 
+.device-settings-version-text {
+  font-size: 80%;
+  justify-content: center;
+  color: var(--swm-disabled-text);
+}
+
 .icons-container {
   display: flex;
   align-items: center;
--- a/packages/vscode-extension/src/webview/components/SettingsDropdown.tsx
+++ b/packages/vscode-extension/src/webview/components/SettingsDropdown.tsx
@@ -28,6 +28,10 @@ function SettingsDropdown({ project, isDeviceRunning, children, disabled }: Sett
   const { movePanelToNewWindow, reportIssue } = useUtils();
   const { telemetryEnabled } = useTelemetry();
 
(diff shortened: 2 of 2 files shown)


jwajgelt (approved):
:+1:

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.css
```diff
@@ -38,6 +38,12 @@
   margin-bottom: 20px;
 }
 
+.device-settings-version-text {
+  font-size: 80%;
```
maciekstosio, line 42: Any specififc reason why %, we stick to px everywhere.

packages/vscode-extension/src/webview/components/SettingsDropdown.tsx
```diff
@@ -28,6 +28,10 @@ function SettingsDropdown({ project, isDeviceRunning, children, disabled }: Sett
   const { movePanelToNewWindow, reportIssue } = useUtils();
   const { telemetryEnabled } = useTelemetry();
 
+  const extensionVersion = document.querySelector<HTMLMetaElement>(
```
maciekstosio, line 31: It seems a bit hacky, what about using just `import {version} from '../../../package.json';`?

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.css
kmagiera: yes, just wanted this to be slightly smaller than text. We don't have variables for font size so wanted to avoid introducing one.

packages/vscode-extension/src/webview/components/SettingsDropdown.tsx
kmagiera: this could be ok too. Haven't tried that but just copied the way we used to show the version during beta

PR #988, opened 2025-02-26
Add an option to build and upload extension artifact with GitHub Actions

balins:
This PR makes it possible to only build the extension and store it as an artifact instead of always publishing.
It also updates `package-lock.json`, bumping our custom `react-scan` tarball (had to fix it upstream to properly build with Vite).

### DIFF
--- a/packages/vscode-extension/vite.config.js
+++ b/packages/vscode-extension/vite.config.js
@@ -59,6 +59,7 @@ export default defineConfig({
           return "assets/[name]-[hash][extname]";
         },
       },
+      external: ["react-scan"],
     },
   },
   server: {
--- a/packages/vscode-extension/vite.config.js
+++ b/packages/vscode-extension/vite.config.js
@@ -1,6 +1,6 @@
+import path from "path";
 import { defineConfig } from "vite";
 import react from "@vitejs/plugin-react";
-import path from "path";
 import license from "rollup-plugin-license";
 
 export default defineConfig({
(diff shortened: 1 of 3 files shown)


kmagiera (commented):
weird you had to add react-scan as external. Maybe you forgot to `npm install` after that dependency got added?

packages/vscode-extension/vite.config.js
```diff
@@ -59,6 +59,7 @@ export default defineConfig({
           return "assets/[name]-[hash][extname]";
         },
       },
+      external: ["react-scan"],
```
jwajgelt, line 62: I don't know much about `vite`/`rollup`, but won't this cause the `react-scan` code to not be included in the final bundle, breaking the features that try to import and use it?

balins:
@kmagiera @jwajgelt
Seems like it had to be fixed upstream https://github.com/software-mansion-labs/react-scan/pull/2, when using this patch + local dependency to `react-scan/packages/scan` I'm again able to build the package.

kmagiera (approved):

kmagiera (approved):
Can you verify the changes in package-lock – they look very suspicious. I wouldn't expect such significant change there. After taking a look there seem to be some binary versions of esbuild getting added that we likely don't want which may cause much longer install times and take significantly more disk space. For example package like `node_modules/@esbuild/freebsd-x64` is listed there while we never had that version listed before. There are other binary versions of esllint including different builds for android which we don't need for sure.

balins:
> Can you verify the changes in package-lock

Discussed in PMs with @kmagiera, the size of `node_modules` didn't change so it's alright. It seems like `esbuild` started to do some shenanigans with package-lock to cover other architectures.

PR #990, opened 2025-02-27
fix "Outline Renders" initialization when tool is already enabled

jwajgelt:
Fixes # (issue)
"Outline Renders" tool not working when the application is reloaded while the tool is enabled.

### How Has This Been Tested: 
- open the Radon IDE panel
- enable the "Outline Renders" tool
- select "Restart app process" from the reload dropdown
- verify the tool still works


filip131311 (approved):
I tested it and it works

PR #991, opened 2025-02-27
Add configuration option to specify arguments passed to expo start

kmagiera:
In one of the projects we heard the feedback from, they use `--scheme` option provided to expo start command. This is necessary for legacy reasons that they keep using expo this way.

The `--scheme` option makes it so that expo bundler uses the provided scheme when generating app URLs that we use for launching the app when expo dev client is detected. We eventually want to avoid relying on launching dev clients via URL, but the necessary changes only landed in SDK 52 (https://github.com/expo/expo/pull/33182), so for compatibility reasons we need to have a way for the schema to be customized on older versions (and until we add a way to launch dev client with extra arguments like proposed here https://github.com/expo/expo/pull/33182)

For the time being, the easiest workaround is to extend the launch configuration and give option to pass additional arguments to expo start. It doesn't appear like any other args really make sense to be added, yet giving more genering "args" option seem to require less explanation.

### How Has This Been Tested: 
1. Launch expo dev client project – make sure nothing breaks
2. Add `--scheme something' as extra args in launch configuration. See that it wouldn't launch and in debug logs you'll see that the app tries to launch using `something://` scheme

### DIFF
--- a/packages/vscode-extension/src/project/metro.ts
+++ b/packages/vscode-extension/src/project/metro.ts
@@ -139,9 +139,18 @@ export class Metro implements Disposable {
     appRootFolder: string,
     libPath: string,
     resetCache: boolean,
+    expoStartExtraArgs: string[] | undefined,
     metroEnv: typeof process.env
   ) {
-    return exec("node", [path.join(libPath, "expo_start.js"), ...(resetCache ? ["--clear"] : [])], {
+    const args = [path.join(libPath, "expo_start.js")];
+    if (resetCache) {
+      args.push("--clear");
+    }
+    if (expoStartExtraArgs) {
+      args.push(...expoStartExtraArgs);
+    }
+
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/project/metro.ts
```diff
@@ -139,9 +139,18 @@ export class Metro implements Disposable {
     appRootFolder: string,
     libPath: string,
     resetCache: boolean,
+    expoStartExtraArgs: string[] | undefined,
     metroEnv: typeof process.env
```
jwajgelt, line 143: Nit:
5 is a lot of arguments, it might be better to pack them in a struct.
```suggestion
    {
    appRootFolder,
    libPath,
    resetCache,
    expoStartExtraArgs,
    metroEnv
    }: LaunchExpoMetroArgs
```

PR #993, opened 2025-02-28
Fix redux plugin after reloads or device switch

### LINKED ISSUE #992: Redux devtools doesn't connect to app after IDE restart or after switching device
**Description**

Steps to reproduce:
1) Launch app with redux plugin
2) Launch the plugin panel and see it works ok
3) Use "Reboot IDE" option from restart menu
4) Open redux plugin again – it no longer shows events

Investigated this a bit and seem like it broke after https://github.com/software-mansion/radon-ide/pull/974

kmagiera:
This PR fixes the issue with redux devtools webview not connecting properly after restarts/reloads or device changes reported in #992.

There were two separate reasons for the described issues:
1) The first issue stems from the timings between redux sents its init messages and when communication channel is established. Uless devtools agent is sent, we silently swallow all messages generated by the devtools. This is mostly ok, as we also need to support connecting the devtools at some later point, so we can't rely on the devtools being ready as soon as the store is created. However, if the devtools frontend was already opened at that time, which happens when we switch between devices, we can't ignore such messages as this would cause the frontend to never properly communicate. The fix for this issue is to make sure the frontent reloads when the communication channel is established. We use "appReady" event for this purpose.
2) The second issue was due to the fact that the plugin was architected in a way that stores stale instances of webview and devtools. As a consequence, after Rebooting IDE instead of routing the messages to the correct instances, we'd attempt to hook to old webview messages or to the stale devtools. The solution is to adapt the workflow the network plugin uses. In that workflow we use webview provider resolve method to get the currenct devtools instance. We then establish the communication between devtools and webview instances from within the resolve method. 

Fixes #992

### How Has This Been Tested: 
1. Open app with redux plugin
2. Make sure plugin works
3. Try different reload options: reload JS, restart metro, reboot IDE
4. Switch to Android and try the same options
5. Before this change, the redux devtools would stop reporting events after reboot IDE and after switching to Android

### DIFF
--- a/packages/vscode-extension/src/plugins/redux-devtools-plugin/ReduxDevToolsPluginWebviewProvider.ts
+++ b/packages/vscode-extension/src/plugins/redux-devtools-plugin/ReduxDevToolsPluginWebviewProvider.ts
@@ -128,6 +128,9 @@ export class ReduxDevToolsPluginWebviewProvider implements WebviewViewProvider {
     }
 
     reduxDevtoolsPlugin.connectDevtoolsWebview(webview);
+    webviewView.onDidDispose(() => {
+      reduxDevtoolsPlugin.disconnectDevtoolsWebview(webview);
+    });
 
     webviewView.onDidChangeVisibility(() => {
       reportToolVisibilityChanged(REDUX_PLUGIN_ID, webviewView.visible);
--- a/packages/vscode-extension/src/plugins/redux-devtools-plugin/ReduxDevToolsPluginWebviewProvider.ts
+++ b/packages/vscode-extension/src/plugins/redux-devtools-plugin/ReduxDevToolsPluginWebviewProvider.ts
@@ -11,8 +11,9 @@ import {
 import { extensionContext } from "../../utilities/extensionContext";
 import { getUri } from "../../utilities/getUri";
 import { getNonce } from "../../utilities/getNonce";
-import { REDUX_PLUGIN_ID } from "./redux-devtools-plugin";
-import { reportToolOpened } from "../../project/tools";
+import { REDUX_PLUGIN_ID, ReduxDevtoolsPlugin } from "./redux-devtools-plugin";
+import { reportToolOpened, reportToolVisibilityChanged } from "../../project/tools";
+import { IDE } from "../../project/ide";
 
 const PATH = "dist/redux-devtools/";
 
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/plugins/redux-devtools-plugin/ReduxDevToolsPluginWebviewProvider.ts
```diff
@@ -106,29 +107,36 @@ function generateWebviewContent(
...
-    webviewView.webview.html = generateWebviewContent(
+    const project = IDE.getInstanceIfExists()?.project;
+    const reduxDevtoolsPlugin = project?.toolsManager.getPlugin(
+      REDUX_PLUGIN_ID
+    ) as ReduxDevtoolsPlugin;
+    if (!reduxDevtoolsPlugin) {
+      throw new Error("Couldn't find redux devtools plugin");
+    }
+
+    reduxDevtoolsPlugin.connectDevtoolsWebview(webview);
```
maciekstosio, line 130: Shouldn't we disconnect on `webviewView.onDidDispose`?

packages/vscode-extension/src/plugins/redux-devtools-plugin/redux-devtools-plugin.ts
```diff
@@ -1,91 +1,84 @@
...
     }
+  };
+
+  connectDevtoolsWebview(webview: Webview) {
+    this.connectedWebview = webview;
+    webview.onDidReceiveMessage((message) => {
+      const { scope, ...data } = message;
+      this.devtools.send(scope, data);
+    });
   }
```
jwajgelt, line 65: What's wrong with getting the `webview` through `webViewProvider.setListener`?

kmagiera: I don't think there's anything inherently wrong. The previous approach had some timing related issue when it comes to that and expected certain callbacks to run in particular order. So I'd assume this was fixable but since we already had a very similar pattern implemented in network that I knew worked, I decided to follow the same approach.

packages/vscode-extension/src/plugins/redux-devtools-plugin/ReduxDevToolsPluginWebviewProvider.ts
kmagiera: I could add that. It wasn't a problem with the current implementation because the `connect` implementation only sets listeners on the webview. So when the webview goes down so do the listeners. But semantically it makes sense to add that and not having it would be more error prone for future updates.

PR #994, opened 2025-02-28
Expo go detaching debugger   alternative fix

maciekstosio:
This PR fixes issue where debugger is detached after restart, when runing Expo Go app. This is due to the fact that when reloading metro, the new instance of debugger with new websocket adress is created. This is alternate solution to: https://github.com/software-mansion/radon-ide/pull/985

This PR solves it as follows:
- after running `metro.reload()` we wait for app to be ready
- then we get list of available runtimes and theire WS addresses 
- we check if current vscSession is still on list from above
- if not we close the connection, and start again 

### How Has This Been Tested: 

**For debugger that detaches during restart i.e. expo-go test app:**
- Lunch expo-go app 
- Test if debugger works (for example log or break point)
- Restart with `Reload JS`
- Wait until it restarts, and make sure the debugger is still attached (there should be also nice ui with "launching", "attaching to debugger") 

**For apps that debugger do not detach i.e. react-native-76**
- Lunch expo-go app 
- Test if debugger works (for example log or break point)
- Add breakpoint in https://github.com/software-mansion/radon-ide/blob/59372cc4e83d28edb2ae6f183b2f0d0a2e48eeb9/packages/vscode-extension/src/debugging/DebugSession.ts#L107
- Restart with `Reload JS`
- Wait until it restarts (**it should not stop at the breakpoint**)
- Make sure the debugger is still attached (the UI should be similar as above, but the "attaching debugger" will pass in the blink of an eye)


packages/vscode-extension/src/project/metro.ts
```diff
@@ -419,10 +408,34 @@ export class Metro implements Disposable {
...
+      retryCount++;
+
+      const list = await fetch(`http://localhost:${this._port}/json/list`);
+      const listJson = await list.json();
+
+      if (listJson.length > 0) {
+        return listJson;
+      }
+
+      await sleep(progressiveRetryTimeout(retryCount));
```
filip131311, line 448: I don't understand this code, it seems to only be reachable when the fetch above fails, but that would throw uncaught  rejection anyway preventing any future retries, wouldn't it? does this condition: `listJson.length` happen often? If so could you add a comment explaining it? 

That being said it seems the fetch above should be wrapped in try catch block

maciekstosio: Not only, basically we're pooling the variables. The list can be empty at the beginning.

Previously we did it in a way that we we're retraing  entrie `fetchDebuggerURL`, but my understanding is that it could fail only when fetch was empty. I needed to have method that fetches the list separatly, so we don't open the websocket to runtime, as it closes previous connection, thus I moved the retry logic to other place, but I belive the entrie process will work as before.

Nevertheless, I agree we should  wrap it in try catch block, good catch!

filip131311 (approved):
Looks good now

packages/vscode-extension/src/project/metro.ts
```diff
@@ -440,10 +429,39 @@ export class Metro implements Disposable {
     return undefined;
   }
 
-  private async fetchDebuggerURL() {
-    // query list from http://localhost:${metroPort}/json/list
-    const list = await fetch(`http://localhost:${this._port}/json/list`);
-    const listJson = await list.json();
+  public async fetchRuntimeList(): Promise<CDPTargetDescription[] | undefined> {
```
kmagiera, line 432: It doesn't list runtimes here but websocket targets. This is a more abstract concept as a single target actually may cover many runtimes. Specifically this happens when you do "reload JS" in react native – new runtime is created but it is listed under the same WS target.

I'd rename to `fetchCDPWebsocketTargets` or something simpler by removing "CDP" or "Websocket" fragments

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -40,6 +40,20 @@ export class DebugSession implements Disposable {
...
+    const availableAddresses =
+      possibleRuntimes?.map((runtime) => runtime.webSocketDebuggerUrl) ?? [];
+
+    if (!availableAddresses.includes(this.vscSession?.configuration?.websocketAddress)) {
+      debug.stopDebugging(this.vscSession);
+      this.vscSession = undefined;
+      return this.start();
+    }
+
+    return false;
```
kmagiera, line 54: I commented about this in the other PR, but I don't like that this method returns `false` in two following scenarios:
 - when there's no need to reconnect
 - when there's need to reconnect and we fail to reconnect properly

Then, this method returns `true` only when we need to reconnect and that was successful.

I'd feel more natural to return `true` always when there's a proper connection established regardless of whether we needed to reconnect or not, and then return `false` in case we couldn't establish the connection

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -40,6 +40,20 @@ export class DebugSession implements Disposable {
     this.debugEventsListener.dispose();
   }
 
+  public async reconnectIfNeeded() {
+    const possibleRuntimes = await this.metro.fetchRuntimeList();
+    const availableAddresses =
+      possibleRuntimes?.map((runtime) => runtime.webSocketDebuggerUrl) ?? [];
+
+    if (!availableAddresses.includes(this.vscSession?.configuration?.websocketAddress)) {
```
kmagiera, line 48: this seems overly complicated I think you just want to find runtime with the given ws URL, so this should be something like:
```
const currentWsAddress = this.vscSession?.configuration?.websocketAddress;
const hasCurrentWsAddress = possibleRuntimes?.some(runtime => runtime.webSocketDebuggerUrl === currentWsAddress)
```

kmagiera (approved):
I left a couple inline comments outside of this review. Please address them prior to merging.

I also think it'd be appropriate for the test plan to cover scenarios w/o expo-go. I understand this PR changes behavior in those cases too. Specifically we should verify that we don't reconnect in those setups. We will also be showing different states in the UI when refresh happens from what I understand?

packages/vscode-extension/src/debugging/DebugSession.ts
maciekstosio: Yeah, I changed it there, but checkout this branch earlier.

maciekstosio:
> I left a couple inline comments outside of this review. Please address them prior to merging.
> 
> I also think it'd be appropriate for the test plan to cover scenarios w/o expo-go. I understand this PR changes behavior in those cases too. Specifically we should verify that we don't reconnect in those setups. We will also be showing different states in the UI when refresh happens from what I understand?

Yes, I'll add information about checking for regression, but the flow will be more or less the same. By more or less I mean that for other setups overlay with "attaching debugger" will pass in the blink of an eye, but will be checked anyway.

filip131311:
The other solution in #985 won

PR #995, opened 2025-02-28
Add a fallback path to the end of the path variable

filip131311:
This PR fixes the issue introduced by  #975, that was causing for the path variable to be incomplete.
The problem could be surfaced  when trying to use `sim-server fingerprint` functionality as it relies on the PATH variable. 

The solution is to in addition to the new path finding method, append the old path at the end of the new one to fill in any missing paths. 

### How Has This Been Tested: 

- Run `react-native-78` test app
- add license

### DIFF
--- a/packages/vscode-extension/src/utilities/subprocess.ts
+++ b/packages/vscode-extension/src/utilities/subprocess.ts
@@ -31,7 +31,14 @@ async function getPathEnv(appRoot: string) {
       env: {},
     }
   );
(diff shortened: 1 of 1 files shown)


kacperkapusciak (approved):
This change fixes an issue I had with license activation on main with error message `Error: Empty result machine ID`

### Before

unnamed (3)



### After

Screenshot 2025-02-28 at 12 34 44

So I'm happy with this change

PR #997, opened 2025-02-28
Sanitize ref name before uploading artifact

balins:
While I didn't find any restrictions in the docs, it seems that artifact names have [some restrictions](https://github.com/actions/upload-artifact/issues/22#issuecomment-568561966). This PR sanitizes the ref name before using it in artifact's name to avoid errors when packaging the extensions from refs containing '\', '/', "', ':', '<', '>', '|', '*', or '?'.


kmagiera (commented):
I wonder if we could just use the name that the Vsix package has? The package scripts outputs a file that gets the version number in it which should be enough

balins:
@kmagiera We will have possibly multiple builds per one version number, so I'd rather keep artifact names unique

filip131311 (approved):
Thank you!

PR #998, opened 2025-02-28
pass a concrete port number to Metro

jwajgelt:
Passes a concrete port number (picked by getting an ephemeral port briefly assigned by the OS and then releasing it) to Metro instead of "0", which causes issues with Metro's internal implementation.

### How Has This Been Tested: 
Open the various test apps in the `test-apps` repo and verify the packager connects.


kmagiera (approved):
Looks good 👍

PR #999, opened 2025-02-28
Fix WebView disconnecting in debug mode when laptop goes to sleep 

maciekstosio:
This PR fixes an issue when webview is disconnecting, when laptop goes sleep. It happens only on development setup. The solution is based on this thread: https://stackoverflow.com/a/73560133

### How Has This Been Tested: 

- Run Radon IDE 
- Close MacBook lid 
- Open it back
- The Radon IDE webview should be working as usuall

### DIFF
--- a/packages/vscode-extension/vite.config.js
+++ b/packages/vscode-extension/vite.config.js
@@ -64,7 +64,7 @@ export default defineConfig({
   server: {
     port: 2137,
     hmr: {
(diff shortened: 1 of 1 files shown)


kmagiera (approved):
Nice find!

Can you just make it clear from the description that this addresses the development setup only. This issue never happens for release versions of the IDE

PR #1000, opened 2025-03-03
Add Radon AI chat

kacperkapusciak:
This PR adds a Radon AI Chat assistant that can answer up-to-date questions about React Native right in VSCode.

You also need to have `GitHub Copilot Chat` extension installed for this to work.

- Added `Radon IDE: Chat with Radon AI` command
  - you can also open the chat via `Chat: Open Chat` and start your question with `@radon` 
- Added check to only allow to use the Chat when you have an active license
- Implemented chat history
- Added telementry
- Fetches data from production backend instance

image

## First usage

When Radon AI is used for the first time user needs allow access to a language model to start chatting.

Screenshot 2025-04-03 at 12 37 43

Screenshot 2025-04-03 at 12 41 57

### Happy path

1. Run command `Chat: Open chat` or `Radon IDE: Chat with Radon AI` to open the GitHub Copilot Chat. 
2. Send react native/expo related question to the chat. Make sure your question is prepended with `@radon` e.g.
`@radon how to setup react native executorch in my project?`
3. Radon AI should respond to your query
image

### Simulate broken backend

1. In `src/chat/api.ts` change the `BASE_RADON_AI_URL` to some other string e.g. "incorrect-url"
2. Rebuild the extension
3. Send react native/expo related question to the chat. Make sure your question is prepended with `@radon` e.g.
`@radon how to setup react native executorch in my project?`
4. Radon AI should respond with "Couldn't connect to Radon AI"
image

### Missing license

1. In `src/chat/index.ts` inside `registerChat` function add `await removeLicense();` to remove the jwt stored in the vscode secret storage
2. Send a question to the Radon AI chat
3. Radon AI will refuse to respond
image


packages/vscode-extension/src/chat/chat.ts
```diff
@@ -0,0 +1,160 @@
...
+          if ("prompt" in response) {
+            history += `USER: ${response.prompt}\n\n`;
+          }
+          if ("response" in response) {
+            response.response.forEach((r) => {
+              if (r instanceof vscode.ChatResponseMarkdownPart) {
+                history += `ASSISTANT: ${r.value.value}\n\n`;
+              }
+            });
+          }
```
jwajgelt, line 102: NIT: If `response` is meant for assistant's messages, maybe we should rename
- `previousResponses` => `chatMessageHistory`
- `response` => `chatMessage`?

packages/vscode-extension/src/chat/chat.ts
```diff
@@ -0,0 +1,160 @@
...
+        } else {
+          Logger.error(String(error));
+        }
+      }
+
+      let systemPrompt = "";
+
+      if (json) {
+        systemPrompt += START_OF_DOCUMENTATION + json.docs + END_OF_DOCUMENTATION;
+      }
```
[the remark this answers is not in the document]
kacperkapusciak: Fixed in 3a3f554b8532af2b02ba46a5620827970db1c2c2

packages/vscode-extension/src/chat/chat.ts
kacperkapusciak: Fixed in c42e402c

packages/vscode-extension/src/chat/history.ts
```diff
@@ -0,0 +1,47 @@
...
+const END_OF_PREVIOUS_RESPONSES = "\n# END OF PREVIOUS RESPONSES.\n\n";
+
+interface ChatMessage {
+  role: "user" | "assistant";
+  content: string;
+}
+
+export function getChatHistory(context: vscode.ChatContext): ChatMessage[] {
+  const chatMessageHistory = context.history.filter(
+    (chatTurn) => chatTurn.participant === CHAT_PARTICIPANT_ID
```
latekvo, line 14: I think we don't want to filter by `chatTurn.participant === CHAT_PARTICIPANT_ID`.

There could be a situation, where the user talks with the regular Copilot, and then decides to use `@radonai`. In such case, i think we should allow Radon to see past messages.

packages/vscode-extension/src/chat/index.ts
```diff
@@ -0,0 +1,160 @@
...
+import { getChatHistory } from "./history";
+
+export const CHAT_PARTICIPANT_ID = "chat.radon-ai";
+const TOOLS_INTERACTION_LIMIT = 3;
+
+interface IChatResult extends vscode.ChatResult {
+  metadata: {
+    command: string;
+  };
+}
```
latekvo, line 15: I think `IChatResult` is redundant. It's only used in the `handler` (line `47`), where the type can be automatically inferred. Removing `IChatResult` does not introduce any warnings as far as I see.

packages/vscode-extension/src/chat/index.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  return [];
+}
+
+export function registerChat(context: vscode.ExtensionContext) {
+  const handler: vscode.ChatRequestHandler = async (
+    request: vscode.ChatRequest,
+    chatContext: vscode.ChatContext,
+    stream: vscode.ChatResponseStream,
+    token: vscode.CancellationToken
```
latekvo, line 57: Same redundancy here. These types should be automatically inferred from the `vscode.ChatRequestHandler` type assignment one line above.

packages/vscode-extension/src/chat/index.ts
latekvo: Correction: the type cannot be automatically inferred, but the usage of `IChatResult` may be replaced with the built-in `vscode.ChatResult`.

packages/vscode-extension/src/chat/index.ts
```diff
@@ -0,0 +1,160 @@
...
+        vscode.LanguageModelChatMessage.User("All requested tool calls have been executed."),
+      ];
+    }
+  }
+
+  return [];
+}
+
+export function registerChat(context: vscode.ExtensionContext) {
+  const handler: vscode.ChatRequestHandler = async (
```
latekvo, line 53: I think the entire `handler` may be brought outside of the `registerChat` function. It doesn't ever use the `context` argument of `registerChat`, so I'm not sure if there's a benefit to having it inside of it.

packages/vscode-extension/src/chat/history.ts
kacperkapusciak: I'm not sure if this is a good idea. Taking messages from Copilot is probably okay but what about other chatbots (Supabase e.g.). I believe it could lead to some unexpected results

packages/vscode-extension/src/chat/index.ts
kacperkapusciak: Fixed in [501d374](https://github.com/software-mansion/radon-ide/pull/1000/commits/501d3746539ac7312e60286f7a3edb100e220265)

packages/vscode-extension/src/chat/index.ts
kacperkapusciak: Fixed in [501d374](https://github.com/software-mansion/radon-ide/pull/1000/commits/501d3746539ac7312e60286f7a3edb100e220265)

packages/vscode-extension/src/chat/index.ts
kacperkapusciak: Fixed in [501d374](https://github.com/software-mansion/radon-ide/pull/1000/commits/501d3746539ac7312e60286f7a3edb100e220265)

filip131311 (dismissed):
If you are using the radon Ai for the first time the system will ask you for permission, and after you choose cancel (as opposed to accept) the chat will stay in "thinking" state with no handling for the rejection

packages/vscode-extension/src/chat/api.ts
```diff
@@ -0,0 +1,86 @@
...
+      Logger.error(`Failed to fetch response from Radon AI with status: ${response.status}`);
+      getTelemetryReporter().sendTelemetryEvent("chat:error", {
+        error: `Failed to fetch with status: ${response.status}`,
+      });
+      return;
+    }
+    return await response.json();
+  } catch (error) {
+    if (error instanceof Error) {
+      Logger.error(error.message);
```
kmagiera, line 30: would be better to print some additional context for the error log and not just the error message here. This way it may be easier to locate where the error occurred based on where it gets printed.

packages/vscode-extension/src/chat/api.ts
```diff
@@ -0,0 +1,86 @@
...
+      getTelemetryReporter().sendTelemetryEvent("chat:error", { error: error.message });
+    } else {
+      Logger.error(String(error));
```
kmagiera, line 33: same here

packages/vscode-extension/src/chat/api.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const results: ToolResult = await response.json();
+    const toolResults = results.tool_results.map((result) => ({
+      callId: result.tool_call_id,
+      content: [result.content],
+    }));
+    return toolResults;
+  } catch (error) {
+    if (error instanceof Error) {
+      Logger.error(error.message);
```
kmagiera, line 79: same comment as above

packages/vscode-extension/src/chat/api.ts
```diff
@@ -0,0 +1,86 @@
...
+      getTelemetryReporter().sendTelemetryEvent("chat:error", { error: error.message });
+    } else {
+      Logger.error(String(error));
```
kmagiera, line 82: again

packages/vscode-extension/src/chat/packages.ts
```diff
@@ -0,0 +1,83 @@
...
+import { findAppRootFolder } from "../utilities/extensionContext";
+import { calculateMD5 } from "../utilities/common";
+import { CHAT_LOG } from ".";
+
+interface Package {
+  name: string;
+  version: string;
+}
+
+const appRootFolder = findAppRootFolder() ?? "";
```
kmagiera, line 13: we recently refactored some bits to no longer assume appRootFolder is a global constant. It'd be better to get it from `Project` instead and only when it is needed rather than storing the copy

packages/vscode-extension/src/chat/packages.ts
```diff
@@ -0,0 +1,83 @@
...
+const packages: Package[] = [];
+let packageJsonHash: string;
+
+async function getPackageJsonHash(): Promise<string> {
+  const rootPackageJson = await vscode.workspace.findFiles(
```
kmagiera, line 18: wonder whether this will work for monorepo projects where you can have most of the dependencies in subproject.

packages/vscode-extension/src/extension.ts
```diff
@@ -331,6 +337,17 @@ async function captureScreenshot() {
   IDE.getInstanceIfExists()?.project.captureScreenshot();
 }
 
+async function openChat() {
+  let prompt = undefined;
+
+  if (!extensionContext.globalState.get(CHAT_OPENED)) {
```
kmagiera, line 343: I'd rename to something  `CHAT_ONBOARDING_COMPLETED`
`CHAT_OPENED` is a bit misleading as it isn't clear that it is persisted across reboots. I understand you want to print this text once the user opens the chat for the very first time?

kmagiera (changes requested):
Added a couple inline comments

packages/vscode-extension/src/chat/api.ts
kacperkapusciak: Yessir 🫡 

Fixed in ac25c354b4b56dca74f4ed8eefd0f0da79d8566b

packages/vscode-extension/src/chat/packages.ts
latekvo: I think it'd make sense to retrieve all packages from all node_modules, and provide them to the LLM grouped by the paths they were extracted from, such that:

```
"main_repo": {
  "react-native": "0.76.0"
  ...
},
"src/package_one": {
  "react-native-reanimated": "4.0.1"
  ...
}
"src/package_two": {
  "react-native-reanimated": "3.13.0"
  "react-native-gesture-handler": "2.22.0"
  ...
}
```

latekvo: Opened a PR adding this feature: https://github.com/software-mansion/radon-ide/pull/1067

packages/vscode-extension/src/chat/index.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+function handleError(err: unknown, stream: vscode.ChatResponseStream): void {
+  // making the chat request might fail because
+  // - model does not exist
+  // - user consent not given
+  // - quote limits exceeded
+
+  if (err instanceof vscode.LanguageModelError) {
+    console.log(err.message, err.code, err.cause);
```
kmagiera, line 147: delete or use Logger

kmagiera (approved):
Looks good. Apart from some small inline comments I'd like for this PR to include a more comprahensive test plan. Specifically testing scenarios where backend returns errors. How do we handle missing / expired license key?

packages/vscode-extension/src/chat/history.ts
latekvo: @kacperkapusciak, what's the final verdict on this? I think we found that other chat participants, including the default one, do have access to all messages, including the ones sent by other participants. I don't recall if there was a reason why we didn't want to do that though.

PR #1001, opened 2025-03-03
Use vscode-js-debug as the debugger implementation

jwajgelt:
Uses `js-debug`'s implementation of the JS debugger for targets which support it (Hermes with RN version >= 0.76).
This allows us to use the features already implemented there (exception breakpoints, conditional breakpoints) without having to reimplement them, and possibly allows us to remove our CDP debugger implementation in the future, once we drop support for the older RN versions.

### How Has This Been Tested: 
- launch various apps from `test-apps` repo
- verify that in each app, the debugger still works, and:
  - the debugger toolbar doesn't appear (we provide our own controls when a breakpoint is hit)
  - you can set "caught/uncaught exception" breakpoints and they work
  - setting regular breakpoints work correctly
  - setting conditional breakpoints work
  - logs are routed to the debug console
  - pausing inside RN code works correctly (i.e. when handling caught exceptions)
  - ???

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -296,6 +296,15 @@
         ]
       },
       {
+        "type": "radon-pwa-node",
+        "label": "Radon IDE internal debugger for React Native",
+        "hiddenWhen": "true",
+        "languages": [
+          "javascript",
+          "typescript"
+        ]
+      },
+      {
         "type": "radon-ide",
         "label": "Radon IDE",
         "languages": [
@@ -724,5 +733,8 @@
     "vscode-test": "^1.5.0",
     "ws": "^8.14.2",
     "xml2js": "^0.6.2"
-  }
+  },
+  "enabledApiProposals": [
+    "portsAttributes"
+  ]
 }
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -639,11 +639,12 @@
     "copy:redux-devtools-plugin": "rm -rf lib/plugins/third-party/redux-devtools-expo-dev-plugin.js && cp ../redux-devtools-expo-dev-plugin/artifacts/RadonIDE/redux-devtools-expo-dev-plugin.js lib/plugins/third-party/redux-devtools-expo-dev-plugin.js",
     "build:redux-devtools-ui": "rm -rf dist/redux-devtools && cp -R ../redux-devtools-expo-dev-plugin/artifacts/RadonIDE/redux-devtools dist/redux-devtools",
     "build:chrome-devtools-ui": "rm -rf dist/network-devtools-frontend && cp -R ../chrome-devtools-ui/build_artifacts/RadonIDE/gen/front_end/ dist/network-devtools-frontend",
-    "build:third-party": "npm run build:chrome-devtools-ui && npm run build:redux-devtools-ui",
+    "build:third-party": "npm run build:chrome-devtools-ui && npm run build:redux-devtools-ui && npm run build:vscode-js-debug",
     "build:extension-debug": "npm run fetch:third-party && node ./scripts/buildExtensionCode.mjs debug && npm run build:third-party",
     "build:extension": "node ./scripts/buildExtensionCode.mjs production && npm run build:third-party",
     "build:webview": "vite build --mode production && cp ./node_modules/@vscode/codicons/dist/codicon.* dist/.",
     "build:sim-server-debug": "bash ./scripts/build-sim-server-debug.sh dist",
+    "build:vscode-js-debug": "cd ../vscode-js-debug && npm run compile && cp dist/src/chromehash_bg.wasm ../vscode-extension/dist/",
     "build:debug": "npm run build:extension-debug && npm run build:sim-server-debug",
     "build:dist": "npm run fetch:third-party && npm run fetch:sim-server-assets && npm run build:extension && npm run build:webview && npm run build:third-party-notice",
     "build:third-party-notice": "node ./scripts/mergeThirdPartyNotices.mjs ./submodules-NOTICES.json ./dist/simulator-server-NOTICES.json ./dist/webview-NOTICES.json ./dist/extension-NOTICES.json ./ThirdPartyNotices.json",
@@ -742,8 +743,5 @@
     "vscode-test": "^1.5.0",
     "ws": "^8.14.2",
     "xml2js": "^0.6.2"
-  },
-  "enabledApiProposals": [
-    "portsAttributes"
-  ]
+  }
 }
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -640,7 +640,8 @@
     "test": "npm run build:extension && npm run build:webview && npm run build:tests && vscode-test",
     "preinstall": "bash ./scripts/init-submodules-if-not-present.sh ../vscode-js-debug && cd ../vscode-js-debug && npm i && npm exec tsc",
     "fetch:sim-server-assets": "bash ./scripts/download-sim-server-release-assets.sh dist",
-    "fetch:third-party": "bash ./scripts/init-submodules-if-not-present.sh ../chrome-devtools-ui ../redux-devtools-expo-dev-plugin"
+    "fetch:third-party": "bash ./scripts/init-submodules-if-not-present.sh ../chrome-devtools-ui ../redux-devtools-expo-dev-plugin",
+    "postinstall": "patch-package"
   },
   "devDependencies": {
     "@babel/preset-react": "^7.23.3",
@@ -699,6 +700,7 @@
     "minimatch": "^9.0.4",
     "mocha": "^10.4.0",
     "node-fetch": "^2.7.0",
+    "patch-package": "^8.0.0",
     "plist": "^3.1.0",
     "prettier": "^2.2.1",
     "re-resizable": "^6.9.16",
@@ -717,6 +719,7 @@
     "url-loader": "^4.1.1",
     "uuid": "^9.0.1",
     "vite": "^5.0.8",
+    "vscode-cdp-proxy": "^0.2.1",
     "vscode-js-debug": "file:../vscode-js-debug",
     "vscode-test": "^1.5.0",
     "ws": "^8.14.2",
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -296,6 +296,15 @@
         ]
       },
       {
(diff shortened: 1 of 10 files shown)


filip131311 (dismissed):
I found three problems with this implementation: 
1) RNIDE_log is never trigered 
2) uncaught exceptions are disabled, by default in this implementation which is a change compared to the previous one and might cause users to think that they stoped working, also this is a departure from "unhandled exception overlay"
3) the stack of unhandled exception seems to be lacking some parts in `react-native-78` test app when you click on throw error button 

image


https://github.com/user-attachments/assets/89a485c1-9e9f-487c-9d63-9d93b78b0118

jwajgelt:
> I found three problems with this implementation:
> 
>     1. RNIDE_log is never trigered
You mean "RNIDE_consoleLog"?
You're right, I didn't notice we do anything other than print them to the debug console (which is handled by `js-debug`) with this event. It should be fine to emit it.

>     2. uncaught exceptions are disabled, by default in this implementation which is a change compared to the previous one and might cause users to think that they stoped working, also this is a departure from "unhandled exception overlay"
It is, but it's also by design. Since we've got the toggle now, I think we should respect it instead of always handling the thrown exceptions.
This does result in a difference in behavior between the old and new debugger now, which may be confusing the user, but that's a wider issue with this feature that I'm not sure there's much we can do about (maybe other than slapping a warning when the old debugger is running that not all features are supported.)

>     3. the stack of unhandled exception seems to be lacking some parts in `react-native-78` test app when you click on throw error button
Yep.
That's because RN actually catches and rethrows the errors in its implementation (several times). When you check "uncaught exceptions", what you actually get is only the last rethrow, coming from the innards of RN.
You can get the debugger to stop on your exception by checking "caught exceptions" instead.
On one hand, this is not as nice as the system we had before, where you would be pointed to the place the exception was thrown.
On the other, this way you can (provided you check "Caught exceptions") actually inspect the state of the application at the throw site, which I'd say is more useful.
This is also consistent with how the React Native DevTools handle exception breakpoints, though that doesn't mean it's the right way to do that.

In this case, I feel that changing the behavior in this way provides a better experience.
Keeping the old overlay would provide an experience where, depending on whether you're in a "breakpoint" or "exception", you get wildly different behavior of the debugger with the same UI, which I'd say is pretty confusing.
Maybe we can improve the messaging (get rid of "Uncaught exceptions" as they are mostly useless, have just "Exceptions", maybe even filter ones which are not thrown from user code or at least those thrown from RN internals), but I'm not convinced that should go into the initial implementation of this feature.

packages/vscode-extension/package.json
```diff
@@ -304,6 +304,24 @@
...
+        "type": "com.swmansion.proxy-debugger",
+        "label": "Radon IDE internal debugger for React Native",
+        "hiddenWhen": "true",
+        "languages": [
+          "javascript",
+          "typescript"
+        ]
+      },
+      {
+        "type": "radon-pwa-node",
```
kmagiera, line 317: not sure I like this name. Seems like we can call it anything.

packages/vscode-extension/src/debugging/DebugAdapter.ts
```diff
@@ -31,22 +33,37 @@ export class DebugAdapter extends DebugAdapterSession {
...
+      if (session.type === debuggerType) {
         this.cdpDebugSession = session;
         didStartHandler?.dispose();
         didStartHandler = null;
       }
     });
 
+    const sourceMapPathOverrides: Record<string, string> = {};
+    if (isUsingNewDebugger && metroWatchFolders.length > 0) {
+      sourceMapPathOverrides["/[metro-project]/*"] = `${metroWatchFolders[0]}${path.sep}*`;
```
kmagiera, line 50: also move the comment for this line from DebugSession.ts

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
```diff
@@ -0,0 +1,299 @@
...
+    return [key, value];
+  });
+}
+
+const CHILD_SESSION_TYPE = "radon-pwa-node";
+
+export class ProxyDebugAdapter extends DebugSession {
+  private cdpProxy: CDPProxy;
+  private disposables: Disposable[] = [];
+  private nodeDebugSession: vscode.DebugSession | null = null;
```
kmagiera, line 45: we're not debugging node here right?

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
```diff
@@ -0,0 +1,299 @@
...
+        this.sendErrorResponse(
+          response,
+          { format: "Failed to attach debugger session", id: 1 },
+          undefined,
+          undefined,
+          ErrorDestination.User
+        );
+      }
+
+      assert(this.nodeDebugSession !== null);
```
kmagiera, line 190: if this asser fails we won't send neither a response nor the error response. Maybe add this check to the `if` block above?

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
```diff
@@ -0,0 +1,299 @@
...
+    });
+  }
+
+  protected async disconnectRequest(
+    response: DebugProtocol.DisconnectResponse,
+    args: DebugProtocol.DisconnectArguments,
+    request?: DebugProtocol.Request
+  ) {
+    this.terminate();
+    this.sendEvent(new Event("RNIDE_continued"));
```
kmagiera, line 229: why we send "continued" even here? explain in a comment

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
```diff
@@ -0,0 +1,299 @@
...
+const CHILD_SESSION_TYPE = "radon-pwa-node";
+
+export class ProxyDebugAdapter extends DebugSession {
+  private cdpProxy: CDPProxy;
+  private disposables: Disposable[] = [];
+  private nodeDebugSession: vscode.DebugSession | null = null;
+  private terminated: boolean = false;
+  private sourceMapRegistry: SourceMapsRegistry;
+
+  constructor(private session: vscode.DebugSession) {
```
kmagiera, line 49: can we call this `parentSession` (we also reference other sessions here so that it'd be clearer when you access `this.parentSession`)

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
```diff
@@ -0,0 +1,299 @@
...
+      if ("value" in result && result.value === "ping") {
+        this.sendEvent(new Event("RNIDE_pong"));
+      }
+    } catch (_) {
+      /** debugSession is waiting for an event, if it won't get any it will fail after timeout, so we don't need to do anything here */
+    }
+  }
+
+  private async startProfiling() {
+    await this.cdpProxy.injectDebuggerCommand({ method: "Profiler.start", params: {} });
```
kmagiera, line 262: I wonder if we could use profiling command from vscode-js-debug now that we use it directly: https://github.com/microsoft/vscode-js-debug/blob/main/src/ui/profiling/uiProfileManager.ts#L113

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -0,0 +1,244 @@
...
+    // some logs may baypass that, especially when printed in initialization phase, so we
+    // need to detect whether the wrapper has added the stack info or not
+    // We check if there are more than 3 arguments, and if the last one is a number
+    // We filter out logs that start with __RNIDE_INTERNAL as those are messages
+    // used by IDE for tracking the app state and should not appear in the VSCode
+    // console.
+    if (args.length > 0 && args[0].value === "__RNIDE_INTERNAL") {
+      // We return here to avoid passing internal logs to the user debug console,
+      // but they will still be visible in metro log feed.
+      return { id: command.id!, result: {} };
```
kmagiera, line 219: will this result in an empty log being appended to the vscode debug console?

packages/vscode-extension/src/debugging/SourceMapsRegistry.ts
```diff
@@ -163,7 +163,7 @@ export class SourceMapsRegistry {
     if (this.sourceMapAliases) {
       // we return the first alias from the list
       for (const [alias, absoluteFilePath] of this.sourceMapAliases) {
-        if (absoluteFilePath.startsWith(absoluteFilePath)) {
+        if (sourceAbsoluteFilePath.startsWith(absoluteFilePath)) {
```
kmagiera, line 166: is this a mistake? `sourceAbsoluteFilePath` is not defined

kmagiera (approved):

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
jwajgelt: We're using the `"node"` debug session from `js-debug`, so I thought the name was appropriate.

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
jwajgelt: No, this sends a response to the application, it doesn't forward the command to the debugger.

packages/vscode-extension/src/debugging/SourceMapsRegistry.ts
jwajgelt: It's the argument to `toSourceMapAliasedFilePath`, the thing we actually want to check in this function.
The previous implementation would check `x.startsWith(x)` and always return path relative to the first alias, which doesn't seem correct to me.

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
jwajgelt: I looked into it briefly, and it seemed from the CDP messages sent between the debugger and the app that js-debug doesn't think the debuggee supports profiling. I didn't spend any time beyond that on it though, so there might be a way to make it work.
I'd keep it out of this PR though.

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
jwajgelt: I changed the assert to a warning, it doesn't seem too broken if its not set correctly. I think the only thing that will break in that case is the graceful shutdown of the proxy session if the JS session terminates unexpectedly.

PR #1003, opened 2025-03-03
Add telemetry to react-query

maciekstosio:
This PR adds telemetry to react-query devtools plugin. Similarly like for other plugins in https://github.com/software-mansion/radon-ide/pull/974.

### How Has This Been Tested: 

- Run expo-52-prebuild-with-plugins 
- Make sure react-query works by enabling it and inreasing counter on seccond tab 
- Restart the app and retest counter
- Restart the pluign and retest counter
- In all cases react-query devtools should work


filip131311 (approved):
LGTM

PR #1004, opened 2025-03-03
Pass app name to gradle

filip131311:
This PR fixes an issue with some setups that contain complicated native structures. 
Some apps contain native android code that is failing gradle builds, but is not application dependency, to avoid that we try to determine  app name and add it as a prefix to build task name.

### How Has This Been Tested: 

- run clean build in `react-native-78` test app

### DIFF
--- a/packages/vscode-extension/src/builders/buildAndroid.ts
+++ b/packages/vscode-extension/src/builders/buildAndroid.ts
@@ -71,10 +71,12 @@ function makeBuildTaskName(productFlavor: string, buildType: string, appName?: s
   // are the names of the productFlavor and buildType that each start with a capital letter.
   // By default, Android Gradle Plugin always creates staging and release buildTypes and does not define any productFlavor.
   // and appName is the name of the application as seen by the gradle build system, omitting appName leads to
-  // whole android file being build.
-  const flavor = productFlavor.charAt(0).toUpperCase() + productFlavor.slice(1);
+  // whole android project being build.
+  const flavorUppercase =
+    productFlavor && productFlavor.charAt(0).toUpperCase() + productFlavor.slice(1);
+  const buildTypeUppercase = buildType && buildType.charAt(0).toUpperCase() + buildType.slice(1);
   const prefix = appName ? `:${appName}:` : "";
-  return prefix + "assemble" + flavor + buildType.charAt(0).toUpperCase() + buildType.slice(1);
+  return `${prefix}assemble${flavorUppercase}${buildTypeUppercase}`;
 }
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -65,12 +66,15 @@ export async function getAndroidBuildPaths(
...
 }
 
-function makeBuildTaskName(productFlavor: string, buildType: string) {
-  // task name is in the format of assemble<ProductFlavor><BuildType> where productFlavor and buildType
+function makeBuildTaskName(productFlavor: string, buildType: string, appName?: string) {
+  // task name is in the format of :<appName>:assemble<ProductFlavor><BuildType> where productFlavor and buildType
   // are the names of the productFlavor and buildType that each start with a capital letter.
   // By default, Android Gradle Plugin always creates staging and release buildTypes and does not define any productFlavor.
+  // and appName is the name of the application as seen by the gradle build system, omitting appName leads to
+  // whole android file being build.
```
jwajgelt, line 74: Pedantic:
```suggestion
  // whole android project being build.
```
?
I don't know what the correct nomenclature here is, but "file" doesn't feel right.

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -65,12 +66,15 @@ export async function getAndroidBuildPaths(
...
   const flavor = productFlavor.charAt(0).toUpperCase() + productFlavor.slice(1);
-  return "assemble" + flavor + buildType.charAt(0).toUpperCase() + buildType.slice(1);
+  const prefix = appName ? `:${appName}:` : "";
+  return prefix + "assemble" + flavor + buildType.charAt(0).toUpperCase() + buildType.slice(1);
```
jwajgelt, line 77:
```suggestion
  const flavorUppercase = productFlavor && productFlavor.charAt(0).toUpperCase() + productFlavor.slice(1);
  const buildTypeUppercase = buildType && buildType.charAt(0).toUpperCase() + buildType.slice(1);
  const prefix = appName ? `:${appName}:` : "";
  return `${prefix}assemble${flavorUppercase}${buildTypeUppercase}`;
```
1. check for emptiness, or assert non-emptiness elsewhere
1. store the build type in an intermediate variable to make the return statement clearer
1. I find format strings nicer to read, but that's a stylistic choice

jwajgelt (changes requested):

jwajgelt (approved):

PR #1006, opened 2025-03-04
fix webview callable objects breaking on clean rebuild

jwajgelt:
Fixes an issue where the objects obtained through `makeProxy` in the webview
couldn't access the newly created objects, breaking:
- `"LaunchConfig"` after changing `appRoot`
- `"RenderOutlines"` after "Clean rebuild" or "Reload IDE" action, 
- possibly others

### How Has This Been Tested: 
- "Reload IDE" or "Clean rebuild" and check that Outline Renderer still works
- change the App Root and verify the `LaunchConfigController.update` method is called on the newly created `LaunchConfigController` object

### DIFF
--- a/packages/vscode-extension/src/panels/WebviewController.ts
+++ b/packages/vscode-extension/src/panels/WebviewController.ts
@@ -27,7 +27,7 @@ export class WebviewController implements Disposable {
     });
   });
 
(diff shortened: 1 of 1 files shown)


filip131311 (approved):
Thank you, looks clean

PR #1007, opened 2025-03-04
fix render outline duration changing with display refresh rate

jwajgelt:
In `react-scan`'s outline renderer implementation, the duration of the outlines is fixed to 45 frames. This causes issues on high refresh rate displays, with frames disappearing too quickly.
This PR updates the version of `react-scan` to our fork which includes a fix for that.
The actual fix was made in https://github.com/software-mansion-labs/react-scan/commit/406f48e31cbd3ff0d16f6568597cbc980626350f


filip131311 (approved):
Looks good

PR #1009, opened 2025-03-05
Fix circular dependency in CDPBreakpoints

filip131311:
This PR fixes an issue with CDPBreakpoints owning CDPSession causing circular dependency after #964.  

We solve the problem by: CDPBreakpoints no longer owns CDPSession and only takes a callback that alows it to send CDP messages

### How Has This Been Tested: 

Run `expo-go` test app and set some breakpoints


jwajgelt (approved):
It's fine, I guess.

PR #1010, opened 2025-03-05
fix stack overflow when serializing breakpoints

jwajgelt:
When setting breakpoints in our Debug Adapter, we sent in the response the list of our internal breakpoint objects. Because they referenced other parts of our implementation, when VSCode attempted to serialize them, it ran into reference cycles which resulted in stack overflows. This PR returns the breakpoint objects from `BreakpointController` in a form which should be safe for serialization.

### How Has This Been Tested: 
- set breakpoints
- don't get stack overflows in logs


kacperkapusciak (approved):
beautiful

PR #1017, opened 2025-03-07
Final touches to the publishing CI

filip131311:
This PR fixes the following issues with the publishing CI: 
1) the target of the publishing was not specified 
2) we tried to specify multiple targets in the `vscode:package` command
3) we used deprecated method of assigning outputs 

### How Has This Been Tested: 

run CI (without publishing)


.github/workflows/package-extension.yml
```diff
@@ -112,3 +113,4 @@ jobs:
         with:
           extensionFile: ${{ steps.package_extension.outputs.vsix_path }}
           pat: ${{ secrets.OPENVSX_MARKETPLACE_TOKEN }}
+          target: ${{ steps.set_targets.outputs.targets }}
```
kacperkapusciak, line 116:
```suggestion
          target: ${{ steps.set_targets.outputs.targets }}
          
```

kacperkapusciak (approved):
🫡

PR #1022, opened 2025-03-10
Fix new debugger detection

filip131311:
This PR changes how we check if the new JS debugger (aka. Fusebox) is enabled by inspecting the debugged Page's `capabilities` for the `prefersFuseboxFrontend` flag, which is set by the RN application when the new debugger is enabled.

### DIFF
--- a/packages/vscode-extension/src/project/metro.ts
+++ b/packages/vscode-extension/src/project/metro.ts
@@ -377,7 +377,7 @@ export class Metro implements Disposable {
         page.reactNative &&
         (page.title.startsWith("React Native Bridge") ||
           page.description.endsWith("[C++ connection]") ||
-          page?.reactNative?.capabilities?.prefersFuseboxFrontend)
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/project/metro.ts
```diff
@@ -376,7 +376,8 @@ export class Metro implements Disposable {
       (page) =>
         page.reactNative &&
         (page.title.startsWith("React Native Bridge") ||
-          page.description.endsWith("[C++ connection]"))
+          page.description.endsWith("[C++ connection]") ||
+          page?.reactNative?.capabilities?.prefersFuseboxFrontend)
```
jwajgelt, line 380: These shouldn't be necessary, we've already checked for `page.reactNative` being truthy.
```suggestion
          page.reactNative.capabilities?.prefersFuseboxFrontend)
```

jwajgelt (approved):

PR #1030, opened 2025-03-13
Fix android app name detection

filip131311:
This PR fixes a regression introduces in #1004. It seems that in some configuration loadConfig is returning an object but gives no details for `platform.android` that caused Radon to crash. As we already expect that config might not exist the fix is to read `projectConfig` conditionally.

### How Has This Been Tested: 

The problem surfaced itself for me in `expo-52-prebuild-with.-plugins` test app. It build correctly after this fix


jwajgelt (approved):
Imagine writing in a type safe language

PR #1035, opened 2025-03-14
Add an option to test deeplinks with app terminated

filip131311:
This PR adds an option to terminate an app before sending deep link to the device.

### How Has This Been Tested: 

- run `expo-52-prebuild-with-plugins` test app 
- test the termination flow with `myapp://` deeplink 
- test flow without termination (ass app is opened nothing should happen) 

https://github.com/user-attachments/assets/b6925895-4d71-4c3f-b9c7-43a74608aee4

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -175,7 +175,7 @@ export interface ProjectInterface {
   resetAppPermissions(permissionType: AppPermissionType): Promise<void>;
 
   getDeepLinksHistory(): Promise<string[]>;
-  openDeepLink(link: string): Promise<void>;
+  openDeepLink(link: string, terminateApp: boolean): Promise<void>;
 
   startRecording(): void;
   captureAndStopRecording(): void;
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -290,6 +290,21 @@ export class IosSimulatorDevice extends DeviceBase {
     }
   }
 
+  terminateApp = async (bundleID: string) => {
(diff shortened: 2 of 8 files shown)


packages/vscode-extension/src/common/Project.ts
```diff
@@ -175,7 +175,7 @@ export interface ProjectInterface {
   resetAppPermissions(permissionType: AppPermissionType): Promise<void>;
 
   getDeepLinksHistory(): Promise<string[]>;
-  openDeepLink(link: string): Promise<void>;
+  openDeepLink(link: string, terminateApp: boolean): Promise<void>;
```
jwajgelt, line 178: I think I'd rather have separate `terminateApp` and `openDeepLink` methods, and call
```ts
if (condition)
  await terminateApp();
await openDeepLink(link);
```
at the call site. 
This way:
1. we can reuse the termination elsewhere without having to refactor this.
2. it reduces the complexity of `openDeepLink`/`sendDeepLink`

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -290,6 +290,21 @@ export class IosSimulatorDevice extends DeviceBase {
     }
   }
 
+  terminateApp = async (bundleID: string) => {
```
jwajgelt, line 293: NIT:
I personally dislike arrow function class members, I find them hard to parse as a method at a glance when reading the source.
I'd prefer if this was a normal method and later in the `map` you wrap it in an arrow function or `.bind(this)` it.

jwajgelt (changes requested):

packages/vscode-extension/src/common/Project.ts
filip131311: I moved the logic to device session but alomost what you asked for

jwajgelt (approved):
Looks good.

PR #1040, opened 2025-03-17
Use LogBox for displaying exceptions

jwajgelt:
Restore the default RN behavior of opening the fullscreen `LogBox` with error details on an uncaught exception.
### How Has This Been Tested: 
- use the "Check uncaught exception" test in the test apps to verify:
  - the `LogBox` appears
  - the stack frame inside is symbolicated
  - pressing on as stack frame opens an editor in the correct place in the codebase
  - pressing "Minimise" on the `LogBox` and triggering the error again should display the `LogBox` again
- verify no other logs trigger the `LogBox`


kmagiera (approved):

kmagiera (approved):
I'd perhaps move the source `atom` script to a different location or at least change it name to sth like `react-fake-editor-executable` given we copy it to dist anyways

kmagiera (approved):

PR #1044, opened 2025-03-20
remove exception handling related code

jwajgelt:
Removes some leftover code after we stopped handling exceptions ourselves in #1040 

### How Has This Been Tested: 
- Open an app with RN version <0.76 (in order to use the old debugger)
- trigger an uncaught exception
- don't crash and burn


kmagiera (approved):
Yollov2

PR #1045, opened 2025-03-20
fix terminating just started app

jwajgelt:
Fixes an issue where the iOS app is sometimes immediately terminated, causing Radon to hang on "Waiting for app" or "Attaching debugger".

The issue was caused by not awaiting the promises for app termination, which caused a race condition between terminating the app and starting it again.


filip131311 (approved):
LGTM

PR #1048, opened 2025-03-20
Refactor DebugSession and DebugAdapter to support only spawning JS debug session.

kmagiera:
In #964 we changed the debugger workflow and started using the debugger to also wrap the metro session to allow us reporting bundle errors.

Further #985 and #1001 contributed some changes to the debugger workflows that impacted the way we handle debug commands with assumption there is a separate debug session for metro and a new one for the js debugger.

This PR clears up that assumption and allows for the DebugSession class to work with the parent metro session but also without it. While the latter mode is not used, this refactor is needed to build the new "connect" functionality that allows Radon to work with the externally controlled metro instances. In such scenario we don't get bundle errors and hence can't use the debug session to report these. In addition, we need to handle disconnect events.

In this PR we are introducing a concept of master and JS debug sessions. Previously we'd use a different naming that I find confusing. The master session is the one that is spawned immediately along metro and is used to push errors from the bundler. The js debug session is started as a child session of the master session and is the main one that controls the JS runtime. The important part is that the JS debug session can now also launch without the parent session.

Below are highlights of what we are changing as part of this refactor:
 - DebugAdapter is no longer the owner of the js debug session. Instead, DebugSession is the main class that controlls both parent and child session. This removes some complexity from DebugAdapter and allows to avoid the adapter from acting as a command proxy
 - We update ping implementation introduced in #985 to use response instead of events for communication. This simplifies the implementation and event handling.
 - We extract a logic for starting debug session that also captures the session reference. This shared logic is now used to launch the parent metro session and the JS debugger session.
 - We rename the session that we start for the names to better indicate those sessions are spawned and controlled by Radon IDE
 - We add a new onTerminate event that is triggered when js debug session is disconnected.
 - We move call to `DebugSession.start` from DeviceSession constructor into `start` method. This allow us to await the start of the parent session before we continue with the normal flow. Before we'd rely on the parent session being started before we report first errors but we'd never await it.

Fixes # (issue)

### How Has This Been Tested: 
1. Run RN 78 and Expo 52 to test new debugger setups. Test reload JS functionality
2. Run RN 75 and Expo 51 to test old debuger setup
3. Run Expo Go and test JS reloads to make sure changes from #985 are still in place (see test scenario from #985)

### DIFF
--- a/packages/vscode-extension/src/debugging/DebugSession.ts
+++ b/packages/vscode-extension/src/debugging/DebugSession.ts
@@ -1,173 +1,219 @@
 import assert from "assert";
-import {
-  commands,
-  debug,
-  DebugSessionCustomEvent,
-  Disposable,
-  DebugSession as VscDebugSession,
-} from "vscode";
+import { commands, debug, DebugConsoleMode, DebugSessionCustomEvent, Disposable } from "vscode";
+import * as vscode from "vscode";
 import { Metro } from "../project/metro";
 import { Logger } from "../Logger";
 import { CDPConfiguration } from "./DebugAdapter";
+import { disposeAll } from "../utilities/disposables";
+import path from "path";
 
 const PING_TIMEOUT = 1000;
+
+const MASTER_DEBUGGER_TYPE = "com.swmansion.react-native-debugger";
+const OLD_JS_DEBUGGER_TYPE = "com.swmansion.js-debugger";
+const PROXY_JS_DEBUGGER_TYPE = "com.swmansion.proxy-debugger";
 
 export const DEBUG_CONSOLE_LOG = "RNIDE_consoleLog";
 export const DEBUG_PAUSED = "RNIDE_paused";
 export const DEBUG_RESUMED = "RNIDE_continued";
 
 export type DebugSessionDelegate = {
-  onConsoleLog(event: DebugSessionCustomEvent): void;
-  onDebuggerPaused(event: DebugSessionCustomEvent): void;
-  onDebuggerResumed(event: DebugSessionCustomEvent): void;
-  onProfilingCPUStarted(event: DebugSessionCustomEvent): void;
-  onProfilingCPUStopped(event: DebugSessionCustomEvent): void;
+  onConsoleLog?(event: DebugSessionCustomEvent): void;
+  onDebuggerPaused?(event: DebugSessionCustomEvent): void;
+  onDebuggerResumed?(event: DebugSessionCustomEvent): void;
+  onProfilingCPUStarted?(event: DebugSessionCustomEvent): void;
+  onProfilingCPUStopped?(event: DebugSessionCustomEvent): void;
+  onDebugSessionTerminated?(): void;
 };
 
 export type DebugSource = { filename?: string; line1based?: number; column0based?: number };
 
-const REACT_NATIVE_SESSION_TYPE = "com.swmansion.react-native-debugger";
+/**
+ * Helpr function that starts a debug session and returns the session object upon sucesfull start
+ */
+async function startDebugging(
+  folder: vscode.WorkspaceFolder | undefined,
+  nameOrConfiguration: string | vscode.DebugConfiguration,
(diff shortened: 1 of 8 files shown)


packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -22,134 +22,148 @@ export type DebugSessionDelegate = {
...
-  /**
-  This method is async to allow for awaiting it during restarts, please keep in mind tho that
-  build in vscode dispose system ignores async keyword and works synchronously.
-  */
-  public async dispose() {
-    this.vscSession && (await debug.stopDebugging(this.vscSession));
-    this.debugEventsListener.dispose();
+  public dispose() {
+    this.stop();
+    disposeAll(this.disposables);
```
jwajgelt, line 158: NIT: Does the order here matter?
In particular, the way it is written now, it's not clear to me if `delegate.onDebugSessionTerminated` here will be called:
- on one hand, `stop` will terminate the session, and we call it before disposing the `terminate` event subscription.
- on the other, `stop` is async, so I'd assume the actual termination will happen in the future, after the subscription is already disposed.

I think I'd prefer if we either called `disposeAll` first, or after the promise from `stop` resolved.

```suggestion
    this.stop().then(() => disposeAll(this.disposables));
```

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -22,134 +22,148 @@ export type DebugSessionDelegate = {
...
-          break;
-        case "RNIDE_profilingCPUStopped":
-          this.delegate.onProfilingCPUStopped(event);
-          break;
-        default:
-          // ignore other events
-          break;
-      }
-    });
+  constructor(private useParentDebugSession: boolean, private delegate: DebugSessionDelegate) {
```
jwajgelt: NIT: I'd prefer if we passed some `options` object instead of naked booleans as arguments, it improves the readability of the code that calls this.
Currently:
```ts
const session = new DebugSession(true, ...);
```
With options:
```ts
const session = new DebugSession(..., { useParentDebugSession: true });
```


```suggestion
  constructor(private delegate: DebugSessionDelegate, private sessionOptions: {
  useParentDebugSession: boolean
  }) {
```

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -116,6 +117,7 @@ export class DebugSession implements Disposable {
   }
 
   public async startParentDebugSession() {
+    this.useParentDebugSession = true;
```
jwajgelt, line 120: I'm worried this might break if at some point we start the JS debugger before Parent, though it probably shouldn't break anything other than having two separate debug sessions visible in the debug panel.
Still, I think I'd prefer the approach from a previous commit, where this was an option passed to the ctor.
The way it is written now, calling `startParentDebugSession` changes the behaviour of all future JS sessions as well, even after `stop`ping this one, which I feel can be surprising to the caller.

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -22,134 +22,148 @@ export type DebugSessionDelegate = {
...
-      didStartHandler?.dispose();
-    }
+    );
   }
 
   public static start(debugEventDelegate: DebugSessionDelegate) {
-    const debugSession = new DebugSession(debugEventDelegate);
-    debugSession.startInternal();
+    const debugSession = new DebugSession(true, debugEventDelegate);
+    debugSession.startParentDebugSession();
```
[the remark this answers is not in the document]
kmagiera: I actually changed this in the meantime. The reason is that we want to await the start method. Constructor cannot be async.

packages/vscode-extension/src/debugging/DebugSession.ts
kmagiera: I changed this already

packages/vscode-extension/src/debugging/DebugSession.ts
kmagiera: This should never be the case. Will add assert here to make sure we guard against it.

PR #1051, opened 2025-03-21
Improve user experience when pods fail to install

jwajgelt:
Stops iOS build if pods need to be installed, but the `pod install` command doesn't install them successfully.
Adds a descriptive message to the build error alert when pods are not installed.

### How Has This Been Tested: 
React Native 78 app in test-apps has an incorrect pods lockfile, which allows testing pod install failure
- try to run `react-native-78` in a freshly cloned repo on iOS -- the pods failed alert should appear
- try to run `react-native-78` in a freshly cloned repo on iOS, with pods not installed correctly -- the pods failed alert should appear
- try to run `react-native-78` in a freshly cloned repo on Android -- the app should build and run correctly
- try to run `react-native-78` in a freshly cloned repo on iOS after resolving the pods issues -- the app should build and run correctly

Previously:
image

Currently:
image

### DIFF
--- a/packages/vscode-extension/src/builders/BuildManager.ts
+++ b/packages/vscode-extension/src/builders/BuildManager.ts
@@ -122,6 +122,10 @@ export class BuildManager {
           if (installPods) {
             getTelemetryReporter().sendTelemetryEvent("build:install-pods", { platform });
             await this.dependencyManager.installPods(iOSBuildOutputChannel, cancelToken);
+            const installed = await this.dependencyManager.checkPodsInstallationStatus();
+            if (!installed) {
+              throw new Error("Pods could not be installed automatically.");
+            }
             // Installing pods may impact the fingerprint as new pods may be created under the project directory.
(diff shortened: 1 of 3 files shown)


jwajgelt:
> Can you add screenshots showing how the error would show up for people? It looks like we're now throwing a new error while it wasn't the case before so wonder how would the pods installation surface before?

It would give a generic build error, with the output not really hinting at pods. Now, because we stop immediately after pods fail, the logs will be focused on the pods error.

> I also think we may consider asking people to run pods installation from command line as a workaround if this error happens? Perhaps we may want to also add that to the troubleshooting guide?

Since the last thing in the logs is the pods error, I think it should be good enough, they seem to be quite good at leading the user to the solution.
Running pods again manually would only give the same error. We can probably add "run pod install in ios" to the alert, but I think it would be cluttered at that point.

kmagiera (approved):

PR #1053, opened 2025-03-21
Enhance XHRInterceptor with TTFB tracking and Network enable/disable support

fnwk:
This PR introduces the following enhancements to the XHRInterceptor:

1. **TTFB (Time to First Byte) Tracking**:
   - Automatically calculates the TTFB when the `readystatechange` event indicates the `readyState` is 2 (headers received).
   - Includes the `ttfb` value in the `Network.responseReceived` message sent to the devtools agent.

2. **Network Enable/Disable Support**:
   - Adds support for `Network.enable` and `Network.disable` methods via the `RNIDE_networkInspectorCDPRequest` listener.
   - Enables or disables the XHRInterceptor based on the received method.

### DIFF
--- a/packages/vscode-extension/lib/network.js
+++ b/packages/vscode-extension/lib/network.js
@@ -110,6 +110,8 @@ export function enableNetworkInspect(devtoolsAgent, payload) {
     }
   });
 
+  const HEADERS_RECEIVED = 2; // readyState value when headers are received
+
   function sendCallback(data, xhr) {
     const requestId = `${requestIdPrefix}-${requestIdCounter++}`;
     const sendTime = Date.now();
@@ -162,7 +164,7 @@ export function enableNetworkInspect(devtoolsAgent, payload) {
     });
 
     xhr.addEventListener("readystatechange", (event) => {
-      if (xhr.readyState === 2) {
+      if (xhr.readyState === HEADERS_RECEIVED) {
         ttfb = Date.now() - sendTime;
       }
     });
--- a/packages/vscode-extension/lib/network.js
+++ b/packages/vscode-extension/lib/network.js
@@ -53,7 +53,7 @@ const XHR_REF_TIMEOUT_MS = 3 * 60 * 1000; // 3 mins
 class FakeWeakRef {
   constructor(obj) {
     this.obj = obj;
-    this.timeout = setTimeout(() => this.obj = undefined, XHR_REF_TIMEOUT_MS)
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/lib/network.js
```diff
@@ -80,6 +80,14 @@ export function enableNetworkInspect(devtoolsAgent, payload) {
   let requestIdCounter = 0;
 
   devtoolsAgent._bridge.addListener("RNIDE_networkInspectorCDPRequest", (message) => {
+    if (message.method === "Network.disable") {
+      XHRInterceptor.disableInterception();
+      console.log("XHRInterceptor disabled");
```
filip131311, line 85: could you add [Radon IDE] in front of this console.log, as this will be logged in debug console alongside application logs it would be good to inform user's where it is comeing from

packages/vscode-extension/lib/network.js
```diff
@@ -80,6 +80,14 @@ export function enableNetworkInspect(devtoolsAgent, payload) {
...
+    } else if (message.method === "Network.enable") {
+      XHRInterceptor.setSendCallback(sendCallback);
+      XHRInterceptor.enableInterception();
```
filip131311, line 88: if you are sending disable log maybe we should consider loging some information here as well?

packages/vscode-extension/lib/network.js
```diff
@@ -151,18 +163,27 @@ export function enableNetworkInspect(devtoolsAgent, payload) {
...
+        ttfb = Date.now() - sendTime;
+      }
+    });
+
     xhr.addEventListener("load", (event) => {
       sendCDPMessage("Network.responseReceived", {
         requestId: requestId,
         loaderId,
         timestamp: Date.now() / 1000,
+        ttfb,
```
filip131311, line 177: can it ever be undefined? if so does sending it take any efect?

filip131311 (commented):
I left some inline comments

packages/vscode-extension/lib/network.js
jwajgelt: I think I'd rather remove this log, I don't think we want our stuff to appear in application logs.

packages/vscode-extension/lib/network.js
fnwk: It can't be undefined, `readystatechange` event is always triggered before `load`

jwajgelt (approved):
I don't see any issues, good job :+1:.

PR #1055, opened 2025-03-24
Only fetch builds from eas with matching fingerprint

jwajgelt:
When fetching builds from EAS Build automatically (i.e. without providing a specific build id), we filter the available builds by the fingerprint returned by the `eas-cli fingerprint:generate` command.

### How Has This Been Tested: 
- run `expo-52-eas` test app on Android
- open launch configuration and select "EAS Build -> Android Build Profile -> development"
- the build should download from EAS Build successfully
- make a change which invalidates the fingerprint
- clean rebuild the app
- the build should fail, and the logs should indicate no build with matching fingerprint was found

### DIFF
--- a/packages/vscode-extension/src/builders/eas.ts
+++ b/packages/vscode-extension/src/builders/eas.ts
@@ -2,14 +2,15 @@ import path from "path";
 import os from "os";
 import { mkdtemp } from "fs/promises";
 
-import { orderBy } from "lodash";
+import assert from "assert";
+import { maxBy } from "lodash";
 import { DevicePlatform } from "../common/DeviceManager";
 import { EasConfig } from "../common/LaunchConfig";
 import { Logger } from "../Logger";
 import { CancelToken } from "./cancelToken";
 import { downloadBinary } from "../utilities/common";
 import {
-  compareFingerprintWithBuild,
+  fingerprintCompareWithBuild,
   EASBuild,
   isEasCliInstalled,
   listEasBuilds,
@@ -61,7 +62,7 @@ async function fetchBuild(config: EasConfig, platform: DevicePlatform, appRoot: 
     return build;
   }
 
-  const builds = await listEasBuilds(platform, config.profile, appRoot);
+  const builds = await listEasBuilds(platform, config, appRoot);
   if (!builds || builds.length === 0) {
     Logger.error(
       `Failed to find any EAS build artifacts for ${platform} with ${config.profile} profile. If you're building iOS app, make sure you set '"ios.simulator": true' option in eas.json.`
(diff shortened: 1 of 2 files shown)


filip131311 (commented):
This is the error you get when there is no build matching fingerprint, I think we should have custom error for eas when the profile is specified to inform the user whats going on

image

packages/vscode-extension/src/builders/easCommand.ts
```diff
@@ -113,3 +121,37 @@ function parseEasBuildOutput(stdout: string, platform: DevicePlatform): EASBuild
...
     });
 }
+
+export interface FingerprintDetails {
+  hash: string;
+}
+
+function isFingerprintDetails(fingerprint: unknown): fingerprint is FingerprintDetails {
+  return (
+    !!fingerprint &&
```
filip131311, line 131: nit: could you explicitly cast using `Boolean(fingerprint)` I find it easy to read "!!" as "!" while glancing through the code quickly.

packages/vscode-extension/src/builders/eas.ts
```diff
@@ -55,7 +62,13 @@ async function fetchBuild(config: EasConfig, platform: DevicePlatform, appRoot:
     return build;
   }
 
-  const builds = await listEasBuilds(platform, config.profile, appRoot);
+  const localFingerprint = await generateFingerprint(platform, appRoot);
```
filip131311, line 65: as I understand it it is an expected scenario that there is no local fingerprint for `listEasBuilds` shouldn't we then catch the potential `Failed to generate build fingerprint: the output of eas-cli is malformed` here?

jwajgelt: `Failed to generate build fingerprint: the output of eas-cli is malformed` is thrown when the JSON output of the `eas-cli` command doesn't match our expected shape.
`generateFingerprint` calculates the fingerprint of local files, it should never fail. Throwing an exception here is well handled by Radon, it just displays an error message that the build failed.

packages/vscode-extension/src/builders/easCommand.ts
jwajgelt: `Boolean(fingerprint)` doesn't let TS know that `fingerprint` is not nullish, unlike `!!`.

packages/vscode-extension/src/builders/eas.ts
filip131311: why is this code: 
```
  if (fingerprintHash !== undefined) {
    commandArgs.push("--fingerprint-hash", fingerprintHash);
  }
``` 
ever needed then?

jwajgelt: Leftover from previous iteration, where we sometimes didn't pass `fingerprintHash` to that function.

filip131311 (approved):
I approve, please remove dead code before merge

packages/vscode-extension/src/builders/easCommand.ts
rolfb: That's strange. Boolean can't return null.

packages/vscode-extension/src/builders/easCommand.ts
```diff
@@ -113,3 +119,37 @@ function parseEasBuildOutput(stdout: string, platform: DevicePlatform): EASBuild
...
+}
+
+export async function generateFingerprint(
+  platform: DevicePlatform,
+  appRoot: string
+): Promise<FingerprintDetails> {
+  const platformMapping = { [DevicePlatform.Android]: "android", [DevicePlatform.IOS]: "ios" };
+  const { stdout } = await exec(
+    "eas",
+    ["fingerprint:generate", "--json", "--non-interactive", "-p", platformMapping[platform]],
```
rolfb, line 143: This command is missing the --build-profile parameter which generates a bad fingerprint causing Radon IDE to fail to start.

PR #1057, opened 2025-03-24
Replace the current network inspector with a custom one

fnwk:
### Webview Content Generation:

* Modified `generateWebviewContent` to accept additional parameters for better customization of the webview content. (`packages/vscode-extension/src/panels/webviewContentGenerator.ts`) [[1]](diffhunk://#diff-3cd38d332bf586fea3a6171b0ed9753f88faddbfec69f5d9f7d9b1fd7c4f1db6L6-R43) [[2]](diffhunk://#diff-3cd38d332bf586fea3a6171b0ed9753f88faddbfec69f5d9f7d9b1fd7c4f1db6L59-R72)
* Updated `SidePanelViewProvider`, `TabPanel`, and `NetworkDevtoolsWebviewProvider` to use the updated `generateWebviewContent` function. (`packages/vscode-extension/src/panels/SidepanelViewProvider.ts`, `packages/vscode-extension/src/panels/Tabpanel.ts`, `packages/vscode-extension/src/plugins/network/NetworkDevtoolsWebviewProvider.ts`) [[1]](diffhunk://#diff-87bb772581661af27ad4c77f95b46f3ba17bacc482c12f73211fe3b0a6af4c4fL36-R39) [[2]](diffhunk://#diff-87bb772581661af27ad4c77f95b46f3ba17bacc482c12f73211fe3b0a6af4c4fL64-R70) [[3]](diffhunk://#diff-b5f31cd7698fe0816035ef3662953429cef09d66d8c26ca90350cc05eef7e409L48-R51) [[4]](diffhunk://#diff-dd0a160ed6c8ea8ef7553bd40882f550a03e1c76627f52004660b38c0579170eL34-R49)

### Build Script Updates:

* Added new build scripts for the network webview and updated existing scripts to include the new network webview build process. (`packages/vscode-extension/package.json`) [[1]](diffhunk://#diff-41ef0db31db6ae1a96349d902b1f423d33e2b104ad20d0b25f5eae835bc5d5c5R655-R666) [[2]](diffhunk://#diff-41ef0db31db6ae1a96349d902b1f423d33e2b104ad20d0b25f5eae835bc5d5c5L676-R679)

### Network Plugin Enhancements:

* Introduced a function to start the Vite server for the network webview and updated the network plugin initialization to use this function. (`packages/vscode-extension/src/plugins/network/network-plugin.ts`) [[1]](diffhunk://#diff-55c4944b125c1798ce87cc4deb95d81758173dc79ad3ff6f08f1154bce78a772R2-R45) [[2]](diffhunk://#diff-55c4944b125c1798ce87cc4deb95d81758173dc79ad3ff6f08f1154bce78a772R54) [[3]](diffhunk://#diff-55c4944b125c1798ce87cc4deb95d81758173dc79ad3ff6f08f1154bce78a772L40-R75) [[4]](diffhunk://#diff-55c4944b125c1798ce87cc4deb95d81758173dc79ad3ff6f08f1154bce78a772R96-R97)

### UI Changes
* We replaced the Chrome Network Inspector with a custom-built React application featuring a redesigned interface. The new inspector is more closely aligned with the VS Code aesthetic, supporting theme synchronization—changing the VS Code theme automatically adjusts the inspector's appearance.


packages/vscode-extension/src/panels/SidepanelViewProvider.ts
```diff
@@ -33,7 +33,10 @@ export class SidePanelViewProvider implements WebviewViewProvider, Disposable {
     this._view.webview.html = generateWebviewContent(
       this.context,
       this._view.webview,
-      this.context.extensionUri
+      this.context.extensionUri,
+      "localhost:2137",
+      "webview",
+      "/src/webview"
     );
```
jwajgelt, line 40: This `generateWebviewContent` call repeats in a couple of places, with the same constants strings passed to it.
In the future, if we want to, for example, change the `src/webview` path or the address of the vite server, we'll have to hunt down all of the places these strings are hard-coded in and change them manually.
It would be better if we either:
- introduced a new function, `generatePreviewWebviewContent(context, webview)` or something like that, which returns the result of 
```ts
generateWebviewContent(
      context,
      webview,
      context.extensionUri,
      "localhost:2137",
      "webview",
      "/src/webview"
    )
```
or
- simply put these strings in constants
```ts
export const EXTENSION_DEV_SERVER_ADDRESS = "localhost:2137"
export const PREVIEW_WEBVIEW_NAME = "webview"
export const PREVIEW_WEBVIEW_PATH = "src/webview"
```
and use them instead wherever we call `generateWebviewContent`

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -1,26 +1,57 @@
...
 import { Devtools } from "../../project/devtools";
 import { ToolKey, ToolPlugin } from "../../project/tools";
 import { extensionContext } from "../../utilities/extensionContext";
+
+import { Logger } from "../../Logger";
 import { NetworkDevtoolsWebviewProvider } from "./NetworkDevtoolsWebviewProvider";
 
 export const NETWORK_PLUGIN_ID = "network";
 
+function startViteServer(onReady: () => void) {
```
jwajgelt, line 15: NIT: I think it's more natural to return a `Promise` from this function, which resolves once the server is ready, and `await` it wherever we call this.

packages/vscode-extension/src/webview/components/shared/ResizableContainer.tsx
```diff
@@ -0,0 +1,120 @@
...
+    <div
+      ref={parentRef}
+      style={{
+        position: "relative",
+        width: side === "bottom" ? "100%" : `${containerWidth}px`,
+        height: side === "bottom" ? `${containerWidth}px` : "auto",
+        display: "flex",
+        flexDirection: side === "bottom" ? "column" : "row",
+        justifyContent:
+          side === "right" || isColumn ? "flex-start" : side === "bottom" ? "flex-end" : "flex-end",
```
jwajgelt, line 93: I find these inline conditional styles hard to read.
I feel like having the parts that depend on `side` separated into three objects (`leftContainerStyle`, `bottomContainerStyle`, `rightContainerStyle`) or even a mapping (`containerStyleBySide: Record<Side, object>`) would be clearer.

packages/vscode-extension/src/webview/components/NetworkFiltersDropdown.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [checked, setChecked] = React.useState(true);
+  return (
+    <DropdownMenuRoot>
+      <DropdownMenu.Trigger asChild disabled={disabled}>
+        {children}
+      </DropdownMenu.Trigger>
+      <DropdownMenu.Portal>
+        <DropdownMenu.Content className="dropdown-menu-content">
+          <DropdownMenu.CheckboxItem checked={checked} onCheckedChange={setChecked}>
+            <DropdownMenu.ItemIndicator>
```
jwajgelt, line 22: Do all of these dropdowns do nothing?
They're all controlled by the same `checked` variable, and it doesn't seem to be used anywhere else.

packages/vscode-extension/src/webview/components/NetworkRequestLog.tsx
```diff
@@ -0,0 +1,312 @@
...
+        ...acc,
+        [title]: 0,
+      }),
+      {}
+    )
+  );
+  const [lastDetailsWidth, setLastDetailsWidth] = useState(0);
+  const [initialRowWidth, setInitialRowWidth] = useState(0);
+
+  useEffect(() => {
```
jwajgelt, line 36: Shouldn't these be `layoutEffect`s, since you want to set UI element sizes?
https://react.dev/reference/react/useLayoutEffect

packages/vscode-extension/src/webview/hooks/useNetworkTracker.tsx
```diff
@@ -0,0 +1,244 @@
...
+                wallTime: params.wallTime,
+                durationMs: undefined,
+                ttfb: params.ttfb,
+              },
+            });
+          }
+        } catch (error) {
+          console.error("Error parsing WebSocket message:", error);
+        }
+      });
```
jwajgelt, line 164: The callback passed to `forEach` here is pretty large. Could we move it to top level, so this `useEffect` is easier to read?

packages/vscode-extension/src/plugins/network/NetworkDevtoolsWebviewProvider.ts
```diff
@@ -31,35 +32,20 @@ export class NetworkDevtoolsWebviewProvider implements WebviewViewProvider {
...
-    window.__websocketEndpoint = 'localhost:${wsPort}';
-    </script>
-    <script type="module" src="${inspectorJsUri}"></script>
-    <body class="undocked" id="-blink-dev-tools"/>
-    </html>`;
+    webview.html = generateWebviewContent(
+      this.context,
+      webviewView.webview,
+      this.context.extensionUri,
+      "localhost:2136",
```
jwajgelt, line 43: Do we launch a separate vite server for the network panel? Is it necessary? Can't we somehow reuse the one serving the main extension webview?

kacperkapusciak (changes requested):
i couldn't make it through it all in one sitting, to be continued tomorrow

packages/vscode-extension/src/panels/webviewContentGenerator.ts
```diff
@@ -56,15 +61,15 @@ export function generateWebviewContent(
                       img-src vscode-resource: http: https: data:;
                       media-src vscode-resource: http: https:;
                       style-src ${webview.cspSource} 'unsafe-inline';
-                      script-src 'nonce-${nonce}';
+                      script-src http: 'nonce-${nonce}';
```
kacperkapusciak, line 64: why this change is needed? I'm not saying it's wrong I just don't know

if its needed then maybe we should also make this change in react query devtools and redux devtools code

packages/vscode-extension/src/panels/webviewContentGenerator.ts
```diff
@@ -3,39 +3,44 @@ import { getUri } from "../utilities/getUri";
...
-`;
-
 export function generateWebviewContent(
   context: ExtensionContext,
   webview: Webview,
-  extensionUri: Uri
+  extensionUri: Uri,
+  viteDevHost: string,
+  webviewName: string,
+  webviewPath: string
```
kacperkapusciak, line 12: if i'm seeing this right i believe `generateWebviewContent` from `webviewContentGenerator.ts` is always used in the same way

redux devtools, expo devplugins,  react query devtools have their own implementations of `generateWebviewContent`

if so, then we don't have to pass `viteDevHost`, `webviewName` and `webviewPath` as arguments to this function as they are constants and can be inlined

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -1,18 +1,54 @@
...
+      if (code !== 0) {
+        reject(new Error(`Process exited with code ${code}`));
+      } else {
+        Logger.debug(`Process exited with code ${code}`);
+      }
+    });
+  });
+}
+
 let initialzed = false;
```
kacperkapusciak, line 44:
```suggestion
let initialized = false;
```
while we are here we can also fix a typo

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -1,18 +1,54 @@
...
 import { extensionContext } from "../../utilities/extensionContext";
+
+import { Logger } from "../../Logger";
 import { NetworkDevtoolsWebviewProvider } from "./NetworkDevtoolsWebviewProvider";
 
 export const NETWORK_PLUGIN_ID = "network";
 
+function startViteServer(): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const process = spawn("npm", ["run", "watch:network-webview"], {
```
kacperkapusciak, line 17: was this added just so we can have both extension Vite instance and network Vite running in parallel?

if so, then we can just use [concurrently](https://www.npmjs.com/package/concurrently)

```
concurrently "npm run watch:extension" "npm run watch:network-webview"
```

we used to use `run-run-all` in react-navigation for this purpose but the project is now abandoned

packages/vscode-extension/src/webview/components/NetworkBar.tsx
```diff
@@ -0,0 +1,51 @@
...
+        onClick={toggleRecording}
+        tooltip={{
+          label: isRecording
+            ? "Stop recording network activity"
+            : "Start recording network activity",
+          side: "bottom",
+        }}>
+        <span
+          style={{ color: isRecording ? "var(--vscode-charts-red)" : "var(--swm-default-text)" }}
+          className={`codicon ${isRecording ? "codicon-record" : "codicon-stop-circle"}`}
```
kacperkapusciak, line 20: might be a bit easier to read using [classnames](https://www.npmjs.com/package/classnames) - we're already using in the project

packages/vscode-extension/src/webview/components/NetworkLogDetails.css
```diff
@@ -0,0 +1,149 @@
...
+  height: 100%;
+}
+
+.request-sent-bar {
+  background-color: #add8e6;
+}
+
+.response-receive-bar {
+  background-color: #90ee90;
+}
```
kacperkapusciak, line 143: hardcoded colors, might be better to use some vscode theme ones

packages/vscode-extension/src/webview/components/NetworkLogDetails.tsx
```diff
@@ -0,0 +1,207 @@
...
+const Section = ({ title, data }: SectionProps) => {
+  const [isExpanded, setIsExpanded] = useState(true);
+
+  return (
+    <div className="section">
+      <div className="section-header">
+        <span
+          className={`codicon ${
+            isExpanded ? "codicon-triangle-down" : "codicon-triangle-right"
+          } gray-icon`}
```
kacperkapusciak, line 56: here will be definitely more readable using `classnames`

packages/vscode-extension/src/webview/components/NetworkLogDetails.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const urlObj = new URL(url);
+  const params: Record<string, string> = {};
+  urlObj.searchParams.forEach((value, key) => {
+    params[key] = value;
+  });
+  return params;
+};
+
+const Section = ({ title, data }: SectionProps) => {
+  const [isExpanded, setIsExpanded] = useState(true);
```
kacperkapusciak, line 48:
```suggestion
  const [isExpanded, toggleExpanded] = useReducer((s) => !s, true);
```

try this instead

packages/vscode-extension/src/webview/components/NetworkLogDetails.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  return (
+    <div className="section">
+      <div className="section-header">
+        <span
+          className={`codicon ${
+            isExpanded ? "codicon-triangle-down" : "codicon-triangle-right"
+          } gray-icon`}
+          onClick={() => setIsExpanded((prev) => !prev)}
```
kacperkapusciak, line 57:
```suggestion
          onClick={toggleExpanded}
```

packages/vscode-extension/src/webview/components/NetworkTimeline.tsx
kacperkapusciak, line 1: would it be possible to write this logic in JSX instead?

https://d3js.org/getting-started#d3-in-react

```jsx
<svg width={containerWidth} height={chartHeight} style={{background: "var(--swm-preview-background)" }}>
  <g ref={gx} transform={``translate(0,${chartHeight - TIMELINE_LEGEND_HEIGHT})``} />
  {...}
```

instead of 

```jsx
const svg = container
      .append("svg")
      .attr("width", containerWidth)
      .attr("height", chartHeight)
      .style("background", "var(--swm-preview-background)");

    svg
      .append("g")
      .attr("transform", `translate(0,${chartHeight - TIMELINE_LEGEND_HEIGHT})`)
      .call(xAxis);
```

🙏

kacperkapusciak (changes requested):

packages/vscode-extension/src/panels/webviewContentGenerator.ts
fnwk: This change is not needed. It was supposed to solve the CORS issue, but the actual problem turned out to be the missing nonce in the `/@vite/client` script.

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -37,7 +73,11 @@ class NetworkCDPWebsocketBackend implements Disposable {
       ws.on("message", (message) => {
         try {
           const payload = JSON.parse(message.toString());
-          if (payload.method === "Network.getResponseBody") {
+          if (
+            ["Network.getResponseBody", "Network.enable", "Network.disable"].includes(
```
kmagiera, line 77: I guess we could just forward anything that starts with `Network.` really

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -1,18 +1,54 @@
...
 import { Devtools } from "../../project/devtools";
 import { ToolKey, ToolPlugin } from "../../project/tools";
 import { extensionContext } from "../../utilities/extensionContext";
+
+import { Logger } from "../../Logger";
 import { NetworkDevtoolsWebviewProvider } from "./NetworkDevtoolsWebviewProvider";
 
 export const NETWORK_PLUGIN_ID = "network";
 
+function startViteServer(): Promise<void> {
```
kmagiera, line 15: not sure why is this here. I don't think we should be using server unless in development mode. Similarly how we do this for the main panel: in dev mode we use vite such that we get hot reload but when bundling the extension we just load the packaged bundle.

packages/vscode-extension/src/panels/webviewContentGenerator.ts
kmagiera, line 64: revert this

packages/vscode-extension/lib/network.js
```diff
@@ -53,7 +53,7 @@ const XHR_REF_TIMEOUT_MS = 3 * 60 * 1000; // 3 mins
 class FakeWeakRef {
   constructor(obj) {
     this.obj = obj;
-    this.timeout = setTimeout(() => this.obj = undefined, XHR_REF_TIMEOUT_MS)
+    this.timeout = setTimeout(() => (this.obj = undefined), XHR_REF_TIMEOUT_MS);
```
kmagiera, line 56: changes here already landed AFAIK. Rebase/merge these such that this file is not included in the PR

packages/vscode-extension/src/webview/components/NetworkBar.tsx
```diff
@@ -0,0 +1,51 @@
+import IconButton from "./shared/IconButton";
+import { useNetwork } from "../providers/NetworkProvider";
+
+const NetworkBar = () => {
```
kmagiera, line 4: I think we typically use `function NetworkBar` type of syntax for components. Would prefer to have this consistent across the repo

packages/vscode-extension/src/panels/webviewContentGenerator.ts
```diff
@@ -56,15 +61,15 @@ export function generateWebviewContent(
                       img-src vscode-resource: http: https: data:;
                       media-src vscode-resource: http: https:;
                       style-src ${webview.cspSource} 'unsafe-inline';
-                      script-src 'nonce-${nonce}';
+                      script-src http: 'nonce-${nonce}';
                       font-src vscode-resource: https:;" />
-        <link rel="stylesheet" type="text/css" href="webview.css" />`
+        <link rel="stylesheet" type="text/css" href="${webviewName}.css" />`
```
kmagiera, line 66: given this variable is only used for the css I don't think it should be called "webviewName" – this creates implicit coupling between "name" of the webview and the name of the stylesheet. Can't this variable just be for the main stylesheet file name and not for "webview name" which seem a bit abstract?

packages/vscode-extension/src/panels/webviewContentGenerator.ts
```diff
@@ -3,39 +3,44 @@ import { getUri } from "../utilities/getUri";
...
+  webviewPath: string
 ) {
   const config = workspace.getConfiguration("RadonIDE");
   const useCodeTheme = config.get("themeType") === "vscode";
   const IS_DEV = context.extensionMode === ExtensionMode.Development;
 
   // The JS file from the React build output
-  const scriptUri = IS_DEV ? `/src/webview/index.jsx` : "webview.js";
+  const scriptUri = IS_DEV
+    ? webviewPath
```
kmagiera, line 20: the name should indicate this is only used in dev

kmagiera (commented):
Added a couple of comments inline

packages/vscode-extension/src/panels/webviewContentGenerator.ts
fnwk: It is also used to generate the main webview, and there are different values that we pass to this function.

packages/vscode-extension/src/panels/webviewContentGenerator.ts
fnwk: It's also used for the js file

packages/vscode-extension/src/plugins/network/NetworkDevtoolsWebviewProvider.ts
```diff
@@ -31,35 +37,20 @@ export class NetworkDevtoolsWebviewProvider implements WebviewViewProvider {
...
+    webview.html = generateWebviewContent(
+      this.context,
+      webviewView.webview,
+      this.context.extensionUri,
+      NETWORK_DEV_SERVER_HOST,
+      PREVIEW_NETWORK_NAME,
+      PREVIEW_NETWORK_PATH
+    ).replace(
+      "</body>",
+      `<script>window.__websocketEndpoint = 'localhost:${wsPort}';</script></body>`
```
kmagiera, line 53: i'd be cleaner to pass this as argument to `generateWebviewContent`. For example add `extraBodyHTML` parameter and embed that into the generated html.

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -1,19 +1,56 @@
...
+
+let initialized = false;
+async function initialize() {
+  if (initialized) {
     return;
   }
-  initialzed = true;
+  Logger.debug("Initilizing Network tool");
+  if (extensionContext.extensionMode === ExtensionMode.Development) {
+    await startViteServer();
```
kmagiera, line 51: why do we start the server here? I see that it only starts in dev, but this is a runtime check so we need to bundle all the dev server related things into the extension which seems unnecessary.

Is there any reason why this just can't follow the same flow as the main vite server we use for hosting the main webview panel? Or event better: perhaps we could make that same vite server host both network and main panel sites?

kmagiera (changes requested):
Good progress. As noted in the comments I'm still concerned by the way the dev server is handled. It seems like we could just use the same vite config with multiple entry points though. I checked and only needed the following changes in rollup configuration in the main vite file for this to work:


```js
rollupOptions: {
      input: { panel: "src/webview/index.jsx", network: "src/network/index.jsx" },
      output: {
        // Fixed name for the JavaScript entry file
        entryFileNames: (chunkInfo) => {
          // Return different filenames based on the entry point
          switch (chunkInfo.name) {
            case "panel":
              return "webview.js";
            case "network":
              return "network.js";
            default:
              return "[name].js";
          }
        },
        // Fixed name for the CSS file
        assetFileNames: (assetInfo) => {
          if (assetInfo.name?.endsWith(".css")) {
            // Handle CSS files based on the entry point
            switch (path.basename(assetInfo.name)) {
              case "panel":
                return "webview.css";
              case "network":
                return "network.css";
              default:
                return "[name].css";
            }
          }
          return "assets/[name]-[hash][extname]";
        },
      },
    },
```

packages/vscode-extension/src/panels/SidepanelViewProvider.ts
```diff
@@ -33,7 +33,10 @@ export class SidePanelViewProvider implements WebviewViewProvider, Disposable {
     this._view.webview.html = generateWebviewContent(
       this.context,
       this._view.webview,
-      this.context.extensionUri
+      this.context.extensionUri,
+      PREVIEW_WEBVIEW_NAME,
+      PREVIEW_WEBVIEW_PATH,
+      ""
```
kmagiera, line 39: last parameter is optional so IMO it'd be better to skip it entirely rather than pass empty string

packages/vscode-extension/package.json
```diff
@@ -652,16 +652,17 @@
     "build:react-query-devtools-ui": "rm -rf dist/react-query-devtools && cp -R ../radon-ide-react-query-webui/artifacts dist/react-query-devtools",
     "build:chrome-devtools-ui": "rm -rf dist/network-devtools-frontend && cp -R ../chrome-devtools-ui/build_artifacts/RadonIDE/gen/front_end/ dist/network-devtools-frontend",
     "build:third-party": "npm run build:chrome-devtools-ui && npm run build:redux-devtools-ui && npm run build:react-query-devtools-ui",
+    "build:webview": "vite build --mode production && cp ./node_modules/@vscode/codicons/dist/codicon.* dist/. && npm run build:network-webview",
```
kmagiera, line 655: we shouldn't need to build these two separately as they are all listed in the single config file that the main build command uses.

packages/vscode-extension/package.json
```diff
@@ -652,16 +652,17 @@
...
     "build:extension": "npm run prepare:fake_editor && node ./scripts/buildExtensionCode.mjs production && npm run build:third-party",
-    "build:webview": "vite build --mode production && cp ./node_modules/@vscode/codicons/dist/codicon.* dist/.",
     "build:sim-server-debug": "bash ./scripts/build-sim-server-debug.sh dist",
     "build:vscode-js-debug": "cd ../vscode-js-debug && npm i && npm exec tsc && npm run compile && cp dist/src/chromehash_bg.wasm ../vscode-extension/dist/",
     "build:debug": "npm run build:extension-debug && npm run build:sim-server-debug",
     "build:dist": "npm run fetch:third-party && npm run fetch:sim-server-assets && npm run build:extension && npm run build:webview && npm run build:third-party-notice",
     "build:third-party-notice": "node ./scripts/mergeThirdPartyNotices.mjs ./submodules-NOTICES.json ./dist/simulator-server-NOTICES.json ./dist/webview-NOTICES.json ./dist/extension-NOTICES.json ./ThirdPartyNotices.json",
     "prepare:fake_editor": "mkdir -p dist && cp react-fake-editor-executable dist/atom",
-    "watch:extension": "vite",
+    "watch:extension": "vite && cd ./src/webview/network && vite",
```
kmagiera, line 665: this shouldn't not be needed

packages/vscode-extension/src/panels/webviewContentGenerator.ts
```diff
@@ -2,40 +2,48 @@ import { ExtensionContext, ExtensionMode, Webview, Uri, workspace } from "vscode
...
-<script type="module" src="/@vite/client"></script>
-`;
+import { EXTENSION_DEV_SERVER_HOST } from "../webview/utilities/constants";
 
 export function generateWebviewContent(
   context: ExtensionContext,
   webview: Webview,
-  extensionUri: Uri
+  extensionUri: Uri,
+  webviewName: string,
```
kmagiera, line 11: I think it'd be cleaner if we place `network` code under `src/network` and not `src/webview` which was meant to be a folder that's dedicated to the main webview panel (I know the name isn't ideal).

I'd recommend that we move the network code to `src/network` and that we pass a single parameter here `moduleName` from which we will derive the dev path `src/${moduleName}/index.jsx`, the release bundle path `dist/${moduleName}.js` and the CSS entrypoint `dist/${moduleName}.css`

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -4,16 +4,19 @@ import { WebSocketServer, WebSocket } from "ws";
...
+import { Logger } from "../../Logger";
 import { NetworkDevtoolsWebviewProvider } from "./NetworkDevtoolsWebviewProvider";
 
 export const NETWORK_PLUGIN_ID = "network";
 
-let initialzed = false;
-function initialize() {
-  if (initialzed) {
+let initialized = false;
+async function initialize() {
```
kmagiera, line 14: why is this changed to async? There are no awaits added in this PR for this method

kacperkapusciak (approved):

PR #1058, opened 2025-03-24
Fix CDP Proxy port assignment strategy

jwajgelt:
Makes CDP Proxy use an ephemeral port assigned by the OS instead of picking a random port and hoping it's open.

### How Has This Been Tested: 
- open an application
- verify the debugger connects correctly


kmagiera (approved):
Just get rid of `this.port` call from `initializeServer`. Otherwise looks goof

PR #1060, opened 2025-03-24
Improve Bundle error experience when auto save is enabled

filip131311:
This PR does two things:
1) disables triggering on source of bundle error when auto focus is enabled, as users give as feedback that this functionality disrupts their flow
2) replaces focus on IDE logs with focus on debug console in Bundle error actions as the bundle error logs are now visible in the console.

### How Has This Been Tested: 

- turn on auto save and trigger bundle errors you can now write without stoping to move the pointer


jwajgelt:
Definitely works better than before :+1:

jwajgelt (approved):

PR #1061, opened 2025-03-24
Improve user interface in case of old node version

filip131311:
This PR adds information to the build error baner that informs the user about missing node installation, or that it is too old to suport react native applications.

### How Has This Been Tested: 

- use command `nvm alias default 17` and open vscode from new shell 
- test `react-native-79` test app 
- the new description shows app correctly


jwajgelt (approved):
I think it might be useful to distinguish between the two cases (`node` not found vs. found but unsupported) and display the information about the found-but-old version (as in, the path and version + min. version we require) so that the user has more info to troubleshoot.

filip131311:
This is extremely rare situation because there is no way that anything react-native related work without correct version of node and would require some development, so I will  do it in some future PR if we evere rework dependency detection system

PR #1066, opened 2025-03-25
Decouple metro and debug session classes

kmagiera:
This PR cleans up the code preparing for changes that'd allow Radon to use existing metro.

In this PR we are decoupling debug session to no longer take metro as argument. This is needed such that we have better control over debug session w/o need of maintaining up-to-date instances of metro class. Eventually, we may want to refactor metro class entirely such that it has a cleaner interface as now it mutates it states when it searches for debugger URL which can be a bit misleading. Regardless – this is out of the scope for this change as we only extract debug session such that it can work based on the session configuration as opposed to fetching details via metro instance.

What's changed:
1. We simplify and move reconnect logic to device session class. Instead of doing promise race we just list the endpoints first. It doesn't seem like the complexity of that code is justified given listing metro endpoints only takes a few milliseconds. The logic is now much simpler as we fetch the endpoints first, and only when our old endpoint is listed we perform ping to check if it is alive. If reconnection is needed we just start a new JS debugger session and debugSession handles that correctly.
2. We extract startDebugging helper method to separate file such that it can be used by ProxyDebugAdapter – this also simplifies the logic as the adapter class no longer needs to have try/finally block and get the session instance via the callback – this is something that the helper method we added recently takes care of.

The main goal of this change was to be able to delete metro import from debugSession file.

### How Has This Been Tested: 
1. Test new debugger connection work as expected including JS reloads (tested on RN78 example)
2. Test reloads on expo go app – make sure debugger stays connected after reloads.

### DIFF
--- a/packages/vscode-extension/src/debugging/DebugSession.ts
+++ b/packages/vscode-extension/src/debugging/DebugSession.ts
@@ -120,20 +120,6 @@ export class DebugSession implements Disposable {
     return this.currentWsTarget;
   }
 
-  public async reconnectJSDebuggerIfNeeded(configuration: JSDebugConfiguration) {
-    if (configuration.websocketAddress === this.currentWsTarget) {
-      // if the websocket address is the same, it means that the old endpoint is still listed
-      // however, the record might be stale as Metro often lists endpoints for some time after
-      // they've been terminated. We need to check if the endpoint is responding before we decide
-      // whether we need to reconnect.
-    }
-    const isAlive = await this.isJsDebugSessionAlive(metro);
-    if (!isAlive) {
-      return this.startJSDebugSession(metro);
-    }
-    return true;
-  }
-
   public async startParentDebugSession() {
     assert(
       !this.jsDebugSession,
--- a/packages/vscode-extension/src/debugging/DebugSession.ts
+++ b/packages/vscode-extension/src/debugging/DebugSession.ts
@@ -1,5 +1,4 @@
 import assert from "assert";
-import path from "path";
 import { commands, debug, DebugConsoleMode, DebugSessionCustomEvent, Disposable } from "vscode";
 import * as vscode from "vscode";
 import { disposeAll } from "../utilities/disposables";
--- a/packages/vscode-extension/src/debugging/DebugSession.ts
+++ b/packages/vscode-extension/src/debugging/DebugSession.ts
@@ -25,9 +25,12 @@ export type DebugSessionDelegate = {
   onDebugSessionTerminated(): void;
 };
 
-export interface DebugExtraConfiguration {
-  websocketAddress?: string;
-  displayDebuggerOverlay?: boolean;
+export interface JSDebugConfiguration {
+  websocketAddress: string;
+  watchFolders: string[];
+  displayDebuggerOverlay: boolean;
+  isUsingNewDebugger: boolean;
+  expoPreludeLineCount: number;
 }
 
 export type DebugSource = { filename?: string; line1based?: number; column0based?: number };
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/debugging/startDebugging.ts
```diff
@@ -0,0 +1,41 @@
...
+import { debug, Disposable } from "vscode";
+
+/**
+ * Helper function that starts a debug session and returns the session object upon sucesfull start
+ */
+export async function startDebugging(
+  folder: vscode.WorkspaceFolder | undefined,
+  nameOrConfiguration: string | vscode.DebugConfiguration,
+  parentSessionOrOptions?: vscode.DebugSession | vscode.DebugSessionOptions
+) {
```
jwajgelt, line 12: NIT:
```suggestion
): Promise<vscode.DebugSession> {
```
I know TS can infer this, but when I was reading this PR it wasn't clear to me:
1. if this function returns something, or is just used for side-effects, due to the `return` statement being buried three indentations deep.
2. if it always returns the session, or can return `undefined` in some cases.

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
```diff
@@ -148,67 +148,48 @@ export class ProxyDebugAdapter extends DebugSession {
...
+        parentSession: this.session,
+        consoleMode: vscode.DebugConsoleMode.MergeWithParent,
+        lifecycleManagedByParent: true,
+        compact: true,
       }
     );
 
-    try {
-      const childSessionStarted = await debug.startDebugging(
+    if (!debugSession) {
```
jwajgelt, line 182: If I read `startDebugging` correctly, it always returns the session, so `debugSession` is always truthy here.
You should wrap the `startDebugging` call in a `try/catch` instead.

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -170,21 +131,16 @@ export class DebugSession implements Disposable {
...
-      return false;
-    }
-
-    const isUsingNewDebugger = metro.isUsingNewDebugger;
+    const isUsingNewDebugger = configuration.isUsingNewDebugger;
     const debuggerType = isUsingNewDebugger ? PROXY_JS_DEBUGGER_TYPE : OLD_JS_DEBUGGER_TYPE;
 
     const sourceMapPathOverrides: Record<string, string> = {};
-    const metroWatchFolders = metro.watchFolders;
+    const metroWatchFolders = configuration.watchFolders;
```
jwajgelt, line 143: NIT: _In my opinion_, we should pass `sourceMapPathOverrides` in the `configuration` here, and calculate them from `watchFolders` either in `Metro` or wherever we call `startJSDebugSession` (i.e. `deviceSession`), but that's not very important.

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -200,9 +156,9 @@ export class DebugSession implements Disposable {
         request: "attach",
         breakpointsAreRemovedOnContextCleared: isUsingNewDebugger ? false : true, // new debugger properly keeps all breakpoints in between JS reloads
```
jwajgelt, line 157: NIT: While we're here, we can probably remove this (or just set to `true`), since we only use the old debug adapter with the old debugger, and the new debug adapter doesn't check this flag.

jwajgelt (approved):
Only one real issue in `ProxyDebugAdapter`,  other than that looks great.
Thank you for your service 🙇

PR #1068, opened 2025-03-27
Pass build errors from Extension to Webview

jwajgelt:
Refactors how build errors are reported to the user.
Previously, the error message shown in the error dialog would be generated in the webview code, based on the state of the workspace and launch configuration.
With this PR, the error is passed by the extension code to the webview for display, along with some extra context (platform and build "type") that the webview needs to adjust the actions available to the user.
This way, we can generate the error message close to where the error actually occurs instead of essentially "guessing" the cause at the time we want to display it.

### How Has This Been Tested: 
Ran `expo-52-eas` and `react-native-78` test apps, breaking configs:
- `app.json` in `expo-52-eas`
- remove `ios` and `android` directories in `react-native-78`
- uninstall `node` and relaunch IDE
To check that relevant errors appear.

### DIFF
--- a/packages/vscode-extension/src/webview/hooks/useBuildErrorAlert.tsx
+++ b/packages/vscode-extension/src/webview/hooks/useBuildErrorAlert.tsx
@@ -62,14 +62,16 @@ export function useBuildErrorAlert(shouldDisplayAlert: boolean) {
   let description = "Open extension logs to find out what went wrong.";
 
   if (projectState.status === "buildError") {
-    description = projectState.buildError.message;
-    if (projectState.buildError.buildType === BuildType.Eas) {
+    const { buildType, message } = projectState.buildError;
+    description = message;
+    if ([BuildType.Local, BuildType.Custom].includes(buildType)) {
+      logsButtonDestination = "build";
+    } else {
       logsButtonDestination = "extension";
-    } else if (projectState.buildError.buildType === BuildType.Unknown) {
-      logsButtonDestination = "extension";
-      if (!ios?.scheme && xcodeSchemes.length > 1) {
-        description = `Your project uses multiple build schemas. Currently used scheme: '${xcodeSchemes[0]}'. You can change it in the launch configuration.`;
-      }
+    }
+
+    if (buildType === BuildType.Unknown && !ios?.scheme && xcodeSchemes.length > 1) {
+      description = `Your project uses multiple build schemas. Currently used scheme: '${xcodeSchemes[0]}'. You can change it in the launch configuration.`;
     }
   }
 


packages/vscode-extension/src/builders/buildIOS.ts
```diff
@@ -79,80 +82,131 @@ export async function buildIos(
...
+      const appPath = await runExternalBuild(
+        cancelToken,
+        customBuild.ios.buildCommand,
+        env,
+        DevicePlatform.IOS,
+        appRoot
+      );
+      if (!appPath) {
+        throw new Error(
+          "Failed to build iOS app using custom script. See the build logs for details."
```
filip131311, line 111: are build logs initiated in that scenario?

packages/vscode-extension/src/project/project.ts
```diff
@@ -838,19 +842,24 @@ export class Project
...
-    if (stage !== this.projectState.startupMessage) {
+    if (this.projectState.status !== "starting" || stage !== this.projectState.startupMessage) {
       return;
     }
     this.updateProjectState({ stageProgress });
   }
 
   private updateProjectState(newState: Partial<ProjectState>) {
+    // NOTE: this is unsafe, but I'm not sure there's a way to enforce the type of `newState` correctly
+    const mergedState: any = { ...this.projectState, ...newState };
```
filip131311, line 853:
```suggestion
   const acceptableKeys = [...list of all posible keys]
    newState = Object.fromEntries(Object.entries(newState).filter((entry) => acceptableKeys.includes(entry[0])))
    const mergedState: any = { ...this.projectState, ...newState };
```
I think something like that would do the trik, wouldn't it?

packages/vscode-extension/src/webview/hooks/useBuildErrorAlert.tsx
```diff
@@ -52,70 +51,26 @@ function BuildErrorActions({
...
-        "Your project uses EAS build, but eas-cli is not installed. Install it and reload the app.";
-    } else {
-      description = "Your project EAS build has failed, see extension logs to see what went wrong.";
+  let description = "Open extension logs to find out what went wrong.";
+
+  if (projectState.status === "buildError") {
+    description = projectState.buildError.message;
+    if (projectState.buildError.buildType === BuildType.Eas) {
+      logsButtonDestination = "extension";
+    } else if (projectState.buildError.buildType === BuildType.Unknown) {
```
filip131311, line 68: i don't know if it is posible to land here with expo-go and custom builds but if so `logsButtonDestination` should be set to "extension" for them as well

filip131311 (changes requested):
I really like this approach, left some inline comments and It would be great if you recorded the effects and put pictures of the ui in the PR description.

packages/vscode-extension/src/project/project.ts
jwajgelt: This only helps in that we skip copying entries from a "larger" type into a "smaller" one, but I don't think that's really an issue (since TS prevents us from accessing these properties anyway). The actual problem is that the typesystem doesn't correctly verify `newState` correctly updates the current state.
Currently, it's impossible to write this function's signature in a safe way, simply due to the fact that you want the `newState` to either:
- only include the fields that the resulting `ProjectState` might have
- include every field that the resulting `ProjectState` must have, if the `newState.status` is different from `this.projectState.status`
But the method type cannot know the type of `this.projectState`.
There are ways to implement this well, but they would require further changes in this file (everywhere that `this.updateProjectState` is called) that I didn't feel were worth the effort right now, since the behaviour is exactly the same as before.

jwajgelt:
> I really like this approach, left some inline comments and It would be great if you recorded the effects and put pictures of the ui in the PR description.

There should be no changes to the UI, maybe other than the wording of the errors changing slightly -- I mostly copied the error string from webview to the extension where we throw them, and then made some cosmetic changes where I thought the messaging could be improved.

packages/vscode-extension/src/webview/hooks/useBuildErrorAlert.tsx
jwajgelt: Possibly, though we didn't do that before, and I wanted to leave the behaviour the same as much as possible.

packages/vscode-extension/src/builders/buildIOS.ts
jwajgelt: I didn't check, since it felt obvious to me that the external build would log to the build output channel, but you're right. It doesn't.
It shouldn't be hard to forward the logs there, let's do it in this PR.

filip131311 (approved):

PR #1069, opened 2025-03-27
fix sourcemap handling for http sourcemap urls

jwajgelt:
Adds support for source map URLs with `http:` schema in the new debugger.
This is needed for the CPU Profiler's source mapping to work in React Native 79 applications, which (at least in the test-app) returns `http:` source map URLs.

### How Has This Been Tested: 
- run `react-native-79` test app
- run the CPU Profiler
- verify the functions in the profile are correctly source mapped

### DIFF
--- a/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
+++ b/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
@@ -213,30 +213,40 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
       .catch(_.noop);
   }
 
-  private handleScriptParsed(command: IProtocolCommand): IProtocolCommand {
+  private async getSourceMapData(sourceMapURL: string) {
+    if (sourceMapURL.startsWith("data:")) {
+      const base64Data = sourceMapURL.split(",")[1];
+      const decodedData = Buffer.from(base64Data, "base64").toString("utf-8");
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -213,30 +213,40 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
...
+    if (sourceMapURL.startsWith("data:")) {
+      const base64Data = sourceMapURL.split(",")[1];
+      const decodedData = Buffer.from(base64Data, "base64").toString("utf-8");
+      const sourceMapData = JSON.parse(decodedData);
+      return sourceMapData;
+    }
+
+    if (sourceMapURL.startsWith("http")) {
+      const data = await fetch(sourceMapURL);
+      return data.json();
```
filip131311, line 226:
```suggestion
      return await data.json();
```

I know this change does not do anything and it is a matter of opinion if it is even good, but in a similar situation some time ago we agreed to indicate that the method is async with explicit await.

filip131311 (approved):
LGTM

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
jwajgelt: Didn't event notice this is async, honestly.
Sure, let's do it.

PR #1070, opened 2025-03-28
Fix android builds not finding metro server 

filip131311:
This PR solves the android application not finding metro when build cache was hit and the application is small enough. 

## What caused the problem? 
If we booting device and building the application was fast enough, the emulators network internals would not be loaded before the start of the application.  This in turn would cause [PackagerStatusCheck](https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/devsupport/PackagerStatusCheck.kt) to fail and the application would think that there is no metro server. 

image

## Solution 
To solve this problem we do not allow the IDE process to think that the emulator is ready until network is operational. 

### How Has This Been Tested: 

- run `react-native-78` test app and switch android devices multiple times to try and caouse the error
- change the behavior of the final condition to expose whats going on like so: 
```
process = exec(ADB_PATH, [
      "-s",
      serial,
      "shell",
      `while ! ping -c 1 10.0.2.2; do sleep 0.5; done;`,
    ]);

    const { stdout, stderr } = await process;
    Logger.debug("wait  for metro", stdout, stderr)

```

- observe multiple failed connections before one successful one

### DIFF
--- a/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
@@ -849,6 +849,10 @@ async function waitForEmulatorOnline(serial: string, timeoutMs: number) {
 
     await process;
 
+    // If booting device and building the application was fast enough, the emulators network internals
+    // would not be loaded before the start of the application. This in turn would cause PackagerStatusCheck
+    // (https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/devsupport/PackagerStatusCheck.kt)
+    // to fail and the application would think that there is no metro server.
     process = exec(ADB_PATH, [
       "-s",
       serial,
--- a/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
@@ -832,13 +832,35 @@ async function parseAvdIniFile(filePath: string) {
 }
 
 async function waitForEmulatorOnline(serial: string, timeoutMs: number) {
-  await exec(ADB_PATH, [
-    "-s",
-    serial,
-    "wait-for-device",
-    "shell",
-    "while [[ -z $(getprop sys.boot_completed) ]]; do sleep 0.5; done; input keyevent 82",
-  ]);
+  return new Promise<void>(async (resolve, reject) => {
+    let process: ChildProcess | undefined;
+    const timeout = setTimeout(() => {
+      process?.kill(9);
+      reject(new Error("Timeout waiting for emulator to boot"));
+    }, timeoutMs);
+
+    process = exec(ADB_PATH, [
+      "-s",
+      serial,
+      "wait-for-device",
+      "shell",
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -832,13 +832,35 @@ async function parseAvdIniFile(filePath: string) {
...
 
 async function waitForEmulatorOnline(serial: string, timeoutMs: number) {
-  await exec(ADB_PATH, [
-    "-s",
-    serial,
-    "wait-for-device",
-    "shell",
-    "while [[ -z $(getprop sys.boot_completed) ]]; do sleep 0.5; done; input keyevent 82",
-  ]);
+  return new Promise<void>(async (resolve, reject) => {
```
kmagiera, line 835: you can use `Promise.withResolvers` here which IMO is cleaner as it avoids additional indent and arrow function

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -832,13 +832,35 @@ async function parseAvdIniFile(filePath: string) {
...
+      "-s",
+      serial,
+      "wait-for-device",
+      "shell",
+      "while [[ -z $(getprop sys.boot_completed) ]]; do sleep 0.5; done; input keyevent 82",
+    ]);
+
+    await process;
+
+    process = exec(ADB_PATH, [
```
kmagiera, line 852: add some comment in code about what is this here for

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -832,13 +832,35 @@ async function parseAvdIniFile(filePath: string) {
...
+      "-s",
+      serial,
+      "shell",
+      `while ! ping -c 1 10.0.2.2>/dev/null 2>&1; do sleep 0.5; done;`,
```
kmagiera, line 856: copilot says you could go `ping -o 10.0.2.2` here which repeats ping unless the first successful one, so you wouldn't need a shell loop at all.

kmagiera (approved):
Left some minor comments inline. Works ok otherwise

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
filip131311: I don't think it is a good case for this method as i await processes in the promise definition If i were to use `withResolvers` I would need to await before returning

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
filip131311: i think it hallucinated as -o is not even an option 
```
Usage: ping [-aAbBdDfhLnOqrRUvV] [-c count] [-i interval] [-I interface]
            [-m mark] [-M pmtudisc_option] [-l preload] [-p pattern] [-Q tos]
            [-s packetsize] [-S sndbuf] [-t ttl] [-T timestamp_option]
            [-w deadline] [-W timeout] [hop1 ...] destination 
```

(output from `adb shell -c ping -h`)

PR #1071, opened 2025-03-31
Adapt updated sourceMapOverrides format in old debug adapter

kmagiera:
This PR updates the original debug adapter implementation to support the updated format of source map aliasing.

For the new react native debugger, we use the option `sourceMapOverrides` in order to translate from "/[metro-project]/" alias that the RN devtools introduced. With the old debug adapter, we've been expecting a different format for this, namely, a list of pairs with alias and absolute file location. Given we may want to use the old debug adapter to support the new debugger while keeping an option to rely on vscode-js-debug, we want to unify these formats, such that we don't need to format the parameters differently depending on the debug adapter we use.

This PR adapts ths old debugger implementation to accept the updated format. Now the overrides are provided as `Record<string,string>` and the override patterns may include wildcards. The basic implementation only works with wildcards at the very end of the path, but given we currently generate the patterns on our own and they meet these criteria this seems like a good tradeoff given the complexity of different pattern schemas we'd have to support for full compatibility.

### How Has This Been Tested: 
1. Open app with new debugger (RN 78 for example)
2. In DebugSession change the line: `const debuggerType = OLD_JS_DEBUGGER_TYPE;` to force the old proxy adapter to be used
3. Test breakpoints, logs etc

### DIFF
--- a/packages/vscode-extension/src/debugging/SourceMapsRegistry.ts
+++ b/packages/vscode-extension/src/debugging/SourceMapsRegistry.ts
@@ -148,42 +148,41 @@ export class SourceMapsRegistry {
   }
 
   public toAbsoluteFilePathFromSourceMapAlias(sourceMapPath: string) {
-    if (this.sourceMapPathOverrides) {
-      for (const [overridePattern, absoluteFilePathPattern] of this.sourceMapPathOverrides) {
-        const absoluteFilePath = replacePathWithOverride(
-          sourceMapPath,
-          overridePattern,
-          absoluteFilePathPattern
-        );
-        if (absoluteFilePath) {
-          return absoluteFilePath;
-        }
+    for (const [aliasedPattern, absoluteFilePathPattern] of this.sourceMapPathOverrides) {
+      const absoluteFilePath = replacePrefixPattern(
+        sourceMapPath,
+        aliasedPattern,
+        absoluteFilePathPattern
+      );
+      if (absoluteFilePath) {
+        return absoluteFilePath;
       }
     }
     return sourceMapPath;
   }
 
   private toSourceMapAliasedFilePath(sourceAbsoluteFilePath: string) {
-    if (this.sourceMapPathOverrides) {
-      // we return the first alias from the list
-      for (const [overridePattern, absoluteFilePathPattern] of this.sourceMapPathOverrides) {
-        const aliasedPath = replacePathWithOverride(
-          sourceAbsoluteFilePath,
-          absoluteFilePathPattern,
-          overridePattern
-        );
-        if (aliasedPath) {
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/debugging/SourceMapsRegistry.ts
```diff
@@ -148,26 +148,48 @@ export class SourceMapsRegistry {
...
+          return aliasedPath;
         }
       }
     }
     return sourceAbsoluteFilePath;
   }
 }
+
+function replacePathWithOverride(filePath: string, fromPattern: string, toPattern: string) {
+  // This method isn't generic enough. It only supports * at the end of the pattern
```
filip131311, line 185: nit: I think this should be a multiline moment above the function declaration as this is a detail that is relavent for the person using the function (and it would be displayed as a context for it after that change.

packages/vscode-extension/src/debugging/SourceMapsRegistry.ts
```diff
@@ -148,26 +148,48 @@ export class SourceMapsRegistry {
...
+        if (aliasedPath) {
+          return aliasedPath;
         }
       }
     }
     return sourceAbsoluteFilePath;
   }
 }
+
+function replacePathWithOverride(filePath: string, fromPattern: string, toPattern: string) {
```
filip131311, line 184: I think it would be better to name this function something like `replacePathPrefix` as this would describe this function's limitations

filip131311 (changes requested):
I left some inline comments

packages/vscode-extension/src/debugging/SourceMapsRegistry.ts
```diff
@@ -148,26 +148,48 @@ export class SourceMapsRegistry {
...
 
   private toSourceMapAliasedFilePath(sourceAbsoluteFilePath: string) {
-    if (this.sourceMapAliases) {
+    if (this.sourceMapPathOverrides) {
       // we return the first alias from the list
-      for (const [alias, absoluteFilePath] of this.sourceMapAliases) {
-        if (sourceAbsoluteFilePath.startsWith(absoluteFilePath)) {
-          return path.join(alias, path.relative(absoluteFilePath, sourceAbsoluteFilePath));
+      for (const [overridePattern, absoluteFilePathPattern] of this.sourceMapPathOverrides) {
+        const aliasedPath = replacePathWithOverride(
```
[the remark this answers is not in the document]
kmagiera: good point, but imo to make it consistent it'd be better to rename overridePattern to aliasedPattern as the method names already have "toAlias" or "toAbsolutePath"

filip131311 (approved):
LGTM

PR #1072, opened 2025-04-01
Handle call frames when logs don't use radon's runtime

kmagiera:
When connecting to the new RN debugger without Radon's runtime, the logs wouldn't appear in the debug console.

The reason was that we improperly transform the output event because we had a special path for creating events that only works when extra parameters are provided via our `wrapConsole` override from `runtime.js`.

However, when we connect to existing metro instance, the override doesn't work, yet we still need to handle the console events. There were a few problems with that:
1) In #846 we changed the way we handle variable references. This was updated in the main path that relied on the call frame being provided from the runtime but not in the `else` branch.
2) When source frame isn't provided via runtime, we never attempt to read the call frame from CDP call which now is properly populated in the new RN debugger.
3) Finally, when connecting to the new debugger, upon connection is sends a number of console API calls from before the debugger was connected. Apparently, before we have the source maps processed we are unable to properly decipher the stack frame.

We do the following to address these problems:
1) We update the `else` branch in CDPDebugSession where no stack information is provided in the console api call. We basically just need to delete the manual handling of the variable reference as it was done for the main branch in #846 
2) We adapt `skipFiles` option and when processing `message.params.callFrame` from CDP message we filter the first frame that doesn't match the skipFiles list.
3) To addess the last issue, we delay handling of console API calls until we get "debug ready" event.

### How Has This Been Tested: 
1. This can only be change with the new Radon connection functionality that will publish a diff off as a follow up
2. The test relies on connecting to existing metro instance and seeing the logs with proper stack frame links

### DIFF
--- a/packages/vscode-extension/src/debugging/CDPSession.ts
+++ b/packages/vscode-extension/src/debugging/CDPSession.ts
@@ -17,6 +17,8 @@ import { CDPCallFrame, CDPDebuggerScope, CDPRemoteObject } from "./cdp";
 import { typeToCategory } from "./DebugAdapter";
 import { annotateLocations } from "./cpuProfiler";
 import { EventEmitter } from "vscode";
+import { CDPConfiguration } from "./CDPDebugAdapter";
+import { minimatch } from "minimatch";
 
 type ResolveType<T = unknown> = (result: T) => void;
 type RejectType = (error: unknown) => void;
@@ -60,24 +62,19 @@ export class CDPSession {
   private debugSessionReady = false;
   private debugSessionReadyEmitter = new EventEmitter<void>();
 
-  constructor(
-    private delegate: CDPSessionDelegate,
-    websocketAddress: string,
-    sourceMapConfiguration: { expoPreludeLineCount: number; sourceMapAliases: [string, string][] },
-    breakpointsConfiguration: { breakpointsAreRemovedOnContextCleared: boolean }
-  ) {
+  constructor(private delegate: CDPSessionDelegate, private configuration: CDPConfiguration) {
     this.sourceMapRegistry = new SourceMapsRegistry(
-      sourceMapConfiguration.expoPreludeLineCount,
-      sourceMapConfiguration.sourceMapAliases
+      configuration.expoPreludeLineCount,
+      configuration.sourceMapAliases
     );
 
     this.breakpointsController = new BreakpointsController(
       this.sourceMapRegistry,
       this,
-      breakpointsConfiguration.breakpointsAreRemovedOnContextCleared
+      configuration.breakpointsAreRemovedOnContextCleared
     );
 
-    this.connection = new WebSocket(websocketAddress);
+    this.connection = new WebSocket(configuration.websocketAddress);
 
     this.connection.on("open", this.setUpDebugger);
 
@@ -228,7 +225,7 @@ export class CDPSession {
       if (message.params.stackTrace) {
         const stackTrace = message.params.stackTrace;
         const { sourceURL, lineNumber1Based, columnNumber0Based } =
-          this.findFirstCallFramePositionFromApp(stackTrace.callFrames);
+          this.findFirstNonSkippedCallFramePosition(stackTrace.callFrames);
         Object.assign(output.body, {
           //@ts-ignore source, line, column and group are valid fields
           source: new Source(sourceURL, sourceURL),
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/debugging/CDPSession.ts
```diff
@@ -13,9 +13,10 @@ import { Logger } from "../Logger";
 import { SourceMapsRegistry } from "./SourceMapsRegistry";
 import { BreakpointsController } from "./BreakpointsController";
 import { VariableStore } from "./variableStore";
-import { CDPDebuggerScope, CDPRemoteObject } from "./cdp";
+import { CDPCallFrame, CDPDebuggerScope, CDPRemoteObject } from "./cdp";
 import { typeToCategory } from "./DebugAdapter";
 import { annotateLocations } from "./cpuProfiler";
+import { EventEmitter } from "vscode";
```
filip131311, line 19: nit: eslint throws a warning here because of improper import order

packages/vscode-extension/src/debugging/CDPSession.ts
```diff
@@ -198,29 +210,57 @@ export class CDPSession {
...
+    let firstPosition;
+    for (const frame of callFrames) {
+      const originalPosition = this.sourceMapRegistry!.findOriginalPosition(
+        frame.scriptId,
+        frame.lineNumber + 1, // cdp line and column numbers are 0-based
+        frame.columnNumber
+      );
+      if (
+        originalPosition.sourceURL !== "__source__" &&
+        !originalPosition.sourceURL.includes("node_modules")
```
filip131311, line 253: In [this PR](https://github.com/software-mansion/radon-ide/commit/1e5b0d60b6c1d2a501931289ae9f0cdbd59d76d4) we introduced "inspectorExcludePattern" mechanism it seems to me that if users don't want some file to show up as an inspection mechanism they wouldn't want it as log source as well. We might need to change the name of this setting, but I think it should be used here as well.

filip131311, line 253: In [this PR](https://github.com/software-mansion/radon-ide/commit/1e5b0d60b6c1d2a501931289ae9f0cdbd59d76d4) we introduced "inspectorExcludePattern" mechanism it seems to me that if users don't want some file to show up as an inspection mechanism they wouldn't want it as log source as well. We might need to change the name of this setting, but I think it should be used here as well.

packages/vscode-extension/src/debugging/CDPSession.ts
```diff
@@ -136,6 +140,8 @@ export class CDPSession {
         );
 
         if (isMainBundle) {
+          this.debugSessionReady = true;
+          this.debugSessionReadyEmitter.fire();
```
filip131311, line 144: Is it possible for this event to fire more then once? When debugger reconnects in expo go for example if so this pattern would cause the logs to appear multiple times.

filip131311 (commented):
Left some inline comments

packages/vscode-extension/src/debugging/CDPSession.ts
kmagiera: this is something a bit different. It is more close to `skipFiles` list we have for the vscode-js-debug mode: https://github.com/software-mansion/radon-ide/blob/main/packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts#L163 and aims to mimic the logic that we have in `wrapConsole`. I guess I could adapt the skip files stuff here given this is a simple pattern matching scheme

packages/vscode-extension/src/debugging/CDPSession.ts
kmagiera: yes, it is, but I'm not changing it here and we already call delegate method here.

filip131311: the delegate method you mention just sends `InitializedEvent` like so: 
```
public onDebugSessionReady = () => {
    this.sendEvent(new InitializedEvent());
  };
```

and i don't think it does any harm if done twice (It might require the change as well if i'm misstaken)

but if this new event is fired multiple times the user will receive old logs a second time, because of this code:
```
this.debugSessionReadyEmitter.event(() => {
            this.handleConsoleAPICall(message);
          });  
```
I think the listener registered here should remove itself after being called to avoid that issue

filip131311 (approved):
Looks good

PR #1074, opened 2025-04-02
Add iPhone 16 Pro to device list

kmagiera:
This PR adds iPhone 16 Pro to device list. We simply copy the configuration from iPhone 15 as all the dimensions and bazels are the same.

### How Has This Been Tested: 
1. Open device creation flow, select iPhone 16
2. Run the newly created device on one project, test text/dark mode settings work ok.


kacperkapusciak (approved):
tested, works

image

PR #1077, opened 2025-04-03
Use new iPhone 16 mas w/o dynamic island

kmagiera:
In #1074 we added iPhone 16 to list of devices by copying device assets from iPhone 15. This turned out not to work super well, as there are some small discrepancies with the location of the dynamic island in these two device images.

After verifying, the iPhone 16's mask does no longer render opaque dyamic island, but instead the simulator image already has the island overlaid on top of the screen's image. 

In this PR, we're updating iPhone 16 assets to accomodate for that.

### How Has This Been Tested: 
1. Run iPhone 16 simulator
2. See it no longer has weird offset around the dynamic island

Before:
image

After:
image


filip131311 (approved):
Looks good

PR #1078, opened 2025-04-03
Force use of Radon's original debug adapter implementation

kmagiera:
In #1001 we introduced a separate implementaton of DebugAdapter that was based on vscode-js-debug project.

While it is better in many aspects from our previous implementation, we discovered that it perform much worse on bigger projects. Specifically the initialization times were very slow as well as JS reloads. For example https://github.com/bluesky-social/social-app project would take as long as 30 seconds on my M3 mac to initialize while both the app and the editor were hanging and waiting for the initialization to complete.

The initialization issue is likely tied to processing source maps as deleting source map information mitigates this problem (although it breaks all other functions like breakpoints, logs, etc). While it requires some additional investigation this PR disables the new vscode-js-debug based implementation and forces the original one for the time being.

Fixes # (issue)

### How Has This Been Tested: 
1. Run project with new debugger (i.e. react-native-78)
2. Test breakpoints, breakpoints after fast refresh, console logging big objects, uncaught exceptions


filip131311 (approved):
😢

PR #1080, opened 2025-04-04
Use vscode elements in network inspector

kacperkapusciak:
This PR refactors our custom network inspector frontend to use VSCode Elements package thus removing a lot of custom components, styles and logic. Now it works and behaves more vscode'y 

## Before

image

## After

image


kmagiera (commented):
I think the headers table should avoid wrapping the header name. On my typical window width I the header names are wrapped so much they are barely readable. On the screenshot below Radon windows I show chrome devtools window that presents the header details for comparison:
image

kmagiera:
I also think, it'd be good to add "General" section at the very top of the headers tab with just the most important information about the request (URL, request type and status) much like it is done in chrome:
image

Also note that all the header names are capitalized and ordered alphabetically which I think we should also be doing.

kmagiera:
Finally, I don't think Domain should be placed as the first column. On Chrome, Domain column is hidden by default and I don't think it is particularily useful if we show the full URL expanded in the details. By default Chrome shows "Name" column which presents the last section of the URL with parameters. Seems to me like a much more sensible default.

kmagiera (approved):
LGTM

bendera:
I love it! <3

PR #1081, opened 2025-04-04
Add options to trigger home button and appswitcher buttons

kmagiera:
This PR adds options in device settings menu that can trigger home and appSwitch button presses on active device.

We utilize the recently updated sim-server for this purpose that now supports new `button` command.

While it may not be the optimal place to add such buttons, they aren't used as frequenly, so we can alter this location once we get more feedback from the users. On top of that we're setting a shortcut that is compatible with iOS simulator for the home button press: Cmd+Shift+H

This change refactors the Project interface and removes method for biometric auth in favor of more generic method that allows us to run a command. This way we can hook into the commands API easily w/o expanding this interface further.

Here's how the new options look like in the device settings menu:
image

### How Has This Been Tested: 
1. Run Android, test home and app switch
2. Do the same on iOS

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -81,6 +81,18 @@
         "enablement": "RNIDE.extensionIsActive"
       },
       {
+        "command": "RNIDE.deviceHomeButtonPress",
+        "title": "Trigger home button press on the current device",
+        "category": "Radon IDE",
+        "enablement": "RNIDE.extensionIsActive"
+      },
+      {
+        "command": "RNIDE.deviceAppSwitchButtonPress",
+        "title": "Open app switcher on the current device",
+        "category": "Radon IDE",
+        "enablement": "RNIDE.extensionIsActive"
+      },
+      {
         "command": "RNIDE.captureReplay",
         "title": "Capture Replay",
         "category": "Radon IDE",
@@ -120,6 +132,11 @@
         "command": "RNIDE.performFailedBiometricAuthorization",
         "key": "ctrl+shift+N",
         "mac": "cmd+shift+N"
+      },
+      {
+        "command": "RNIDE.deviceHomeButtonPress",
+        "key": "ctrl+shift+H",
+        "mac": "cmd+shift+H"
       },
       {
         "command": "RNIDE.captureReplay",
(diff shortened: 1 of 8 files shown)


packages/vscode-extension/package.json
```diff
@@ -80,6 +80,18 @@
...
         "enablement": "RNIDE.extensionIsActive"
       },
+      {
+        "command": "RNIDE.deviceHomeButtonPress",
+        "title": "Trigger home button press on the current device",
+        "category": "Radon IDE",
+        "enablement": "RNIDE.extensionIsActive"
+      },
+      {
+        "command": "RNIDE.deviceAppSwitchButtonPress",
```
filip131311, line 90: nit: I would replace the names of those commands with their results, because there is no such thing as `AppSwitchButton` on iOS 

I would suggest something like: 
`openAppSwitcher`

packages/vscode-extension/src/extension.ts
```diff
@@ -319,6 +325,18 @@ async function performFailedBiometricAuthorization() {
...
+async function deviceHomeButtonPress() {
+  const project = IDE.getInstanceIfExists()?.project;
+  project?.dispatchButton("home", "Down");
+  project?.dispatchButton("home", "Up");
+}
+
+async function deviceAppSwitchButtonPress() {
+  const project = IDE.getInstanceIfExists()?.project;
+  project?.dispatchButton("appSwitch", "Down");
+  project?.dispatchButton("appSwitch", "Up");
```
filip131311, line 337: what appSwitchButton up mean on iOS? is it just ignored?

filip131311 (approved):
I left some small comments inline

packages/vscode-extension/package.json
kmagiera: Command names are not directly visible. What people will see is the description. The "appSwitch" button doesn't exist, but this was a common abstraction that works for Android and iOS that we decided to add in sim-server. It was easier to add one "button" command that'd take appswitch/home/etc as argument since that's how it gets triggered and Android. On iOS we just trigger the double press when appswitch is pressed down.

I didn't want to call this "openAppSwitcher" as this isn't exacly what it does. For example, it wouldn't open the switcher if it is already opened (it would actually close it in that case).

PR #1084, opened 2025-04-07
Prepare DeviceSession management for multidevice capability 

filip131311:
This PR is an initial ground work that will allow as to keep devices alive even when another one is selected. 
Keep in mind that this functionality is never actually used as the current debugSession and tools structure wouldn't work with new setup, but adding those changes a part of this PR would make it even bigger and would slow down review process significantly.

### What has been done:
1) the `metro` and `devtools` objects are now part of `deviceSession` class, so we can create as many of them as we need
2) `DeviceSessionsManager` class was created and it is responsible for selecting and managing running devices. 
3) restart/reload system was collapsed to a single `reload`  method with additional options, it was necessary as we no longer can rely on `selectDevice` as a mechanism for reseting `deviceSession`
4) deviceSession start and restart systems are now updated to use new `cancelToken`  implementation that allows as to cancel in any async methods and not only childProcesses. 

### What more needs to be done to add true multi device functionality?
- multiple debugSession management
- multiple tools management 
- UI for keeping devices alive

### How Has This Been Tested: 

- run `react-native-78` test app and switch between devices 
- use every reload option and do it even if the old reload haven't finished 
- run `documents` folder as a root of the workspace and switch between some application by selecting `appRoot` in launch configuration 
- run `react-native-77` test app and remove android directory to check if native changes detected functionality works as it too was moved from the project, which made it easier to decouple device session from project.


packages/vscode-extension/src/project/project.ts
```diff
@@ -182,26 +149,55 @@ export class Project
...
   private setupAppRoot() {
     const newAppRoot = findAndSetupNewAppRootFolder();
+
     const oldApplicationContext = this.applicationContext;
     this.applicationContext = new ApplicationContext(newAppRoot);
     oldApplicationContext.dispose();
 
-    this.reload("reboot");
+    const oldDeviceSessionsManager = this.deviceSessionsManager;
+    this.deviceSessionsManager = this.deviceSessionsManager = new DeviceSessionsManager(
```
jwajgelt, line 158: I think one assignment should be sufficient
```suggestion
    this.deviceSessionsManager = new DeviceSessionsManager(
```

packages/vscode-extension/src/common/Project.ts
```diff
@@ -61,6 +61,10 @@ type ProjectStateBuildError = {
   };
 } & ProjectStateCommon;
 
+export type SelectDeviceOptions = {
+  preservePreviousDevice?: boolean;
```
kmagiera, line 65: I think it's be more logical to me if the information whether device should be preserved was a feature of a given device rather than an option that we use when switching devices.

packages/vscode-extension/src/common/Project.ts
```diff
@@ -98,12 +102,20 @@ export const StartupStageWeight = [
...
   | "reboot" // reboots device, launch app
   | "reinstall" // force reinstall app
   | "restartProcess" // relaunch app
   | "reloadJs"; // refetch JS scripts from metro
 
+export enum ReloadResult {
+  succeeded,
+  failed,
+  canceled,
+}
```
kmagiera, line 117: Can't we just rely on the exceptions to handle non-successful cases? Specifically if cancel uses a custom error type this should be cleaner IMO

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -211,6 +212,28 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+    super.reboot();
+    const { promise, resolve } = Promise.withResolvers<void>();
+
+    const timeout = setTimeout(async () => {
+      this.emulatorProcess?.off("exit", exitListener);
+      await this.forcefullyResetDevice();
+      resolve();
+    }, REBOOT_TIMEOUT);
+
+    const exitListener = async () => {
```
kmagiera, line 225: this deserves some comments. What are we trying to do here and why?

packages/vscode-extension/src/devices/DeviceBase.ts
```diff
@@ -20,6 +22,10 @@ export abstract class DeviceBase implements Disposable {
 
   abstract get lockFilePath(): string;
 
+  async reboot(): Promise<void> {
+    this.preview?.dispose();
```
kmagiera, line 26: we should do `this.preview = undefined` here – disposed instances shouldn't be reused

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -211,6 +212,28 @@ export class AndroidEmulatorDevice extends DeviceBase {
     await this.internalBootDevice();
   }
 
+  public async reboot() {
```
kmagiera, line 215: there's no mention about the reboot being added to iOS/Android device interfaces in the description. I understand we didn't have this before so why is this needed now?

packages/vscode-extension/src/plugins/react-query-devtools-plugin/ReactQueryDevToolsPluginWebviewProvider.ts
```diff
@@ -87,7 +87,7 @@ export class ReactQueryDevToolsPluginWebviewProvider implements WebviewViewProvi
 
     webview.html = generateWebviewContent(extensionContext, webview, extensionContext.extensionUri);
 
-    const devTools = IDE.getInstanceIfExists()?.project?.toolsManager.devtools;
+    const devTools = IDE.getInstanceIfExists()?.project?.deviceSession?.toolsManager.devtools;
```
kmagiera, line 90: this change may be a bit tricky. I think we rely on the fact there was a single devtools instance here even when you switch between instances. If this changed, we may no longer be able to use a tool once it is open and after switching the device

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -63,29 +95,53 @@ export class DeviceSession implements Disposable {
...
     readonly buildCache: BuildCache,
     debugEventDelegate: DebugSessionDelegate,
-    private readonly eventDelegate: EventDelegate
+    private readonly deviceSessionDelegate: DeviceSessionDelegate,
+    private readonly metroDelegate: MetroDelegate,
+    private readonly toolsDelegate: ToolsDelegate
   ) {
-    this.buildManager = new BuildManager(dependencyManager, buildCache);
+    this.deviceSettings = extensionContext.workspaceState.get(DEVICE_SETTINGS_KEY) ?? {
+      appearance: "dark",
```
kmagiera, line 108: the default settings are not duplicated here and in project, is this intentional?

kmagiera (commented):
I left a couple of comment inline but at a high level I'm not convinced we should actually be changing the structure in a way to allow for device sesstions to have separate devtools/metro instances. This part specifically has been designed with multiple device support in mind and I think it is a logical choice actually. When you don't use Radon you'd only run one metro instance. As for devtools, it makes a bit more sense to have separate instances, however, our devtools class only represents a server that binds to a specific port in which case I think it'd be better to only have one instance of it. Long term, I'd like to restructure devtools in a way where we don't actually use the server at all. I think all the communication that goes over devtools channel should run via the debugger connection.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -211,6 +212,28 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+
+    const timeout = setTimeout(async () => {
+      this.emulatorProcess?.off("exit", exitListener);
+      await this.forcefullyResetDevice();
+      resolve();
+    }, REBOOT_TIMEOUT);
+
+    const exitListener = async () => {
+      await this.internalBootDevice();
+      clearTimeout(timeout);
```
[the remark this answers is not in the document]
filip131311: nice catch

packages/vscode-extension/src/plugins/react-query-devtools-plugin/ReactQueryDevToolsPluginWebviewProvider.ts
filip131311: solved in `fc06354b0acf5a5b6e2f0c164c0ed4a6defe7566`

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
filip131311: I updated the description.

packages/vscode-extension/src/common/Project.ts
```diff
@@ -171,9 +183,8 @@ export type MultimediaData = {
 export interface ProjectInterface {
   getProjectState(): Promise<ProjectState>;
   reload(type: ReloadAction): Promise<boolean>;
-  restart(clean: "all" | "metro" | false): Promise<void>;
   goHome(homeUrl: string): Promise<void>;
-  selectDevice(deviceInfo: DeviceInfo): Promise<boolean>;
+  selectDevice(deviceInfo: DeviceInfo, selectDeviceOptions?: SelectDeviceOptions): Promise<boolean>;
```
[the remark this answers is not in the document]
filip131311: @jwajgelt I added an interface for stoping an unselected device, I didn't add a `start` method as i don't see how it is ever useful, if you have a use case for starting but not selecting a device please let me know

packages/vscode-extension/src/common/Project.ts
filip131311: I though about it and I don't agree, from users point of view the decision to keep, kill the device, should happen at the moment of switching the device, not at the moment of configuring the device, we might add an option to the device `always keep alive` and use it here, but only optionally. 

This opinion has three reasons: 
1) I think it is rather rare that the user would like to keep the second device alive, so more permanent setting does not seem suitable
2) If this is a permanent option there is a risk of users keeping a runaway number of devices alive
3) If we don't ask if you want to keep the old device alive at the point of selecting a new one I think it would be harder to discover this functionality

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -167,45 +169,41 @@ export class DeviceSession implements Disposable {
...
   }
 
   public async deactivate() {
     this.isActive = false;
     this.buildManager.deactivate();
+    this.toolsManager.deactivate();
   }
 
-  public async perform(type: ReloadAction): Promise<ReloadResult> {
+  public async perform(type: ReloadAction): Promise<boolean> {
```
jwajgelt, line 181: I think I prefered the `ReloadResult` enum (though with the failure reported via exception, like you do now), but I understand that might be controversial 😆.

You basically need to read this method's implementation to know that `true` means success (ok, that you could probably guess, seeing that we return a boolean from a function which only has side effects), and `false` means "cancelled" (this you have no chance of guessing without looking here)

jwajgelt: Ideally, I'd throw the `CancelError` from here and rethrow it in the webview (as `CancelError` as well, so that you can tell them apart at call site), but I understand that requires changes to our rpc layer.

packages/vscode-extension/src/common/DeviceSessionsManager.ts
```diff
@@ -0,0 +1,27 @@
...
+  | "rebuild" // clean build, boot device, install app
+  | "reboot" // reboots device, launch app
+  | "reinstall" // force reinstall app
+  | "restartProcess" // relaunch app
+  | "reloadJs"; // refetch JS scripts from metro
+
+export interface DeviceSessionsManagerInterface {
+  reload(type: ReloadAction): Promise<boolean>;
+  stopDevice(deviceId: string): Promise<boolean>;
+  selectDevice(deviceInfo: DeviceInfo, selectDeviceOptions?: SelectDeviceOptions): Promise<boolean>;
```
jwajgelt, line 26: IMO, we should get rid of this method.
I don't see a reason why we can only have a single device "selected" on the extension side.
I feel like "selecting" should only change which device is currently being displayed in the preview panel, which should only concern the panel itself, not the extension.
I'd change it so that
- this method is called `startDevice` or something along these lines
- add a method to `DeviceSessionManagerInterface` to read the state of currently "live" `DeviceSession`s 
- have the Webview "select" which of the running `DeviceSession`s should be displayed at any time, and start a new one if none are.
- for any methods in the Extension <-> Webview interface which operate on the "selected" device, pass a device session id to indicate to which session they are related, instead of assuming some single "selected" device session.

jwajgelt (changes requested):

packages/vscode-extension/src/builders/cancelToken.ts
```diff
@@ -8,24 +14,38 @@ export class CancelToken {
...
+        cancelError = new CancelError("The process was canceled");
+        input.kill(9);
+      });
+
+      const wrappedInput = new Proxy(input, {
+        get(target, prop, receiver) {
+          if (prop === "then") {
+            return (resolve: any, reject: any) => {
+              if (cancelError) {
+                return Promise.reject(cancelError).then(resolve, reject);
```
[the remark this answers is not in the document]
filip131311: I'm confused you wrote the same thing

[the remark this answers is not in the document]
filip131311: ooooo I see you wrote `catch` instead of `then` please refer to this link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch

packages/vscode-extension/src/builders/cancelToken.ts
```diff
@@ -8,24 +14,38 @@ export class CancelToken {
...
+              }
```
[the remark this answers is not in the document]
filip131311: `cancelError` is outside this scope and can be changed after than was called

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -51,21 +55,48 @@ export class DeviceSessionsManager implements Disposable {
...
     }
     return false;
   }
 
+  public async stopDevice(deviceId: string) {
+    if (deviceId === this.selectedDevice) {
+      window.showWarningMessage(
+        "You cannot stop the selected device. Please select another device first.",
+        "Dismiss"
+      );
```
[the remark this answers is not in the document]
filip131311: We don't have front end for this state and I don't think adding it is in this PR's scope

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -51,21 +55,48 @@ export class DeviceSessionsManager implements Disposable {
...
+      return false;
+    }
+    const deviceSession = this.deviceSessions.get(deviceId);
+    if (!deviceSession) {
+      Logger.warn("Failed to stop device, device wasn't running.", "Dismiss");
```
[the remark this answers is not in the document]
filip131311: 👍

packages/vscode-extension/src/common/DeviceSessionsManager.ts
filip131311: I will consider those coments in the PR that implements user interface and probably change it there, I don like doing to many changes releted to how frontend will consume this logic in this quite large PR selecting device is an old concept so lets keep it for now

packages/vscode-extension/src/common/DeviceSessionsManager.ts
```diff
@@ -0,0 +1,27 @@
...
+import { DeviceInfo } from "./DeviceManager";
+
+export type DeviceSessionsManagerDelegate = DeviceSessionDelegate & {
+  onDeviceSelected: (deviceInfo: DeviceInfo) => void;
+  onReloadRequested: (type: ReloadAction) => void;
+};
+
+export type SelectDeviceOptions = {
+  preservePreviousDevice?: boolean;
+};
```
jwajgelt, line 12: I still dislike this.
I'd rather have the callsite call
```ts
if (!preserve) {
manager.stopDevice(oldDevice);
}
manager.selectDevice(newDevice);
```
instead of
```ts
manager.selectDevice(newDevice, { preservePreviousDevice })
```

packages/vscode-extension/src/builders/cancelToken.ts
```diff
@@ -8,24 +14,38 @@ export class CancelToken {
...
+              return Promise.resolve(target).then(resolve, reject);
+            };
+          }
+          return Reflect.get(target, prop, receiver);
+        },
+      });
+
+      return wrappedInput as ReturnType<typeof exec>;
```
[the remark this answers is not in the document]
filip131311: it does not matter when you call `.then` it only matters when the promiseLike resolves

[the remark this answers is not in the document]
filip131311: because
> cancelError is outside this scope and can be changed after than was called

jwajgelt (approved):
I think this is as far as we can take it without also introducing many changes on the frontend side.
I think this refactor still improves a lot on what we've had before. Thank you for your service 🫡

PR #1085, opened 2025-04-08
Bump sim-server to include fixes for keyboard events on iOS

### LINKED ISSUE #835: `Shift` keyboard button isn't working
**Description**

If I type into input fields using `Shift` it isn't working, instead of `U` it returns `u`, same goes for symbols like `!`, etc. I have to use keyboard from the simulator.

**Environment**

- Version of currently used Radon IDE: 1.0.0
- Version of React Native/Expo used in the project: 51 (Expo) and 0.74 (RN)

kmagiera:
This PR updates sim-server version to include fix for handling hardware key presses on iOS.

Fixes #835

### How Has This Been Tested: 
1. Run example with TextInput
2. Start typing immediately after focusing. Use Shift to test capital letters or '@' sign.
3. Before this change, you'd often land in a state where you wouldn't be able to input letters that require modifier key (Shift)


filip131311 (approved):
Shouldn't we also bump the sim-server version to v0.13.0 and publish new before doing that? 

Approving anyway to unblock

kmagiera:
We don't have to cut releases in sim-server unless we are sure the given commit will go as a part of the extension release. The release script checks for tags that match the version from submodule, so now the release process will fail. Once we are sure we're not adding more changes that are going to be included in next radon update, we can publish release of sim-server

PR #1088, opened 2025-04-10
Rework how bundle errors are triggered 

filip131311:
This PR introduces the following changes: 
1) bundle error over lay is removed and we instead rely on build in error message box 
2) We no longer focus on debug console and terminal when bundlingError happens, unless during the startup process, then we focus on the error source 
3) we remove redundant hasBundleBuildFailedError, it was always triggered in pair with  bundlingError and bundlingError was enough to throw ide into error  (display error baner and overlay) so we remove it.
4) because the build in error message box contains clickable elements we list the block on sending inputs to the device while in bundlingError state

image

### How Has This Been Tested: 

- run `react-native-79` test app and force bundle errors in startup and non startup scenario

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -32,7 +32,6 @@ export type ProjectState =
         | "starting"
         | "running"
         | "bootError"
-        | "bundleBuildFailedError"
         | "bundlingError"
         | "debuggerPaused"
         | "refreshing";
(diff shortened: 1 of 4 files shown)


kmagiera (changes requested):
I think we should make the device screen ineractable as now it displays some buttons that people may want to click. The current implementation blocks touches when the error is shown

kmagiera (approved):
Code looks good ✅

PR #1089, opened 2025-04-11
Fix license modal sizing issue

kmagiera:
This PR fixes styles for modals and for the activate license modal in particular.

This addresses the issue with scroll bars appearing on modals when IDE panel is narrow.

This PR makes the following changes to the styles:
 - we use percentage to set width and height of the modal (there's a px max limit that we keep). This way they don't reach the very edges of the view which makes it hard to notice the modal edges.
 - we allow license key input to scale on its own instead of using 100% which cased it to always expand past the boundaries of the modal (because it also has borders and paddings).

### How Has This Been Tested: 
1. Open activate license modal, manage device modal, diagnostics modal and launch config modals
2. Resize the window to see it looks ok with both wide and narrow settings

Before:
image
image

After:
image
image

This change doesn't seem to impact other modals significantly. They looked ok apart from reaching the very edge of the window on narrow screens, now they have some paddings:

Before:
image

After:
image

### DIFF
--- a/packages/vscode-extension/src/webview/components/shared/Modal.css
+++ b/packages/vscode-extension/src/webview/components/shared/Modal.css
@@ -20,9 +20,9 @@ input {
   top: 50%;
   left: 50%;
   transform: translate(-50%, -50%);
-  width: 90vw;
+  width: 75%;
   max-width: 450px;
-  max-height: 85vh;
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
Looks good

PR #1090, opened 2025-04-11
Fix order in which theme.css is applied

kmagiera:
In #1057 a change was added that cause CSS rollup to start splitting CSS into [input].css and theme.css for webview and network inputs. Because of that, in the generated HTML we added theme.css as an extra source of our CSS. However, this change resulted in some UI glitches that started appearing on the release build that were due to the fact the order of CSS changed.

In CSS, when multiple class names are set for an item, and the two classes specify the same property. The class that is defined "later" takes precedence for this particular property. 

The fact that theme.css was listed after the main css bundle, impacted the way some styles were applied in release mode.

In particular, styles for the CPU profiler button were impacted where we use `input-button` and `button-recording-on` classes that both define `width`. We normally expected `button-recording-on` to override `width` property but since `input-button` have width specified in `theme.css`, it was taking the precedence in release builds.

This PR fixes the issue by reordering the CSS imports such that theme.css will always come first.
We also move import of `theme.css` to main network panel entry file, such that we are sure it is always imported first in the dev builds.

### How Has This Been Tested: 
1. Run app, test CPU Profiler button
2. Run `build:webview` script and set IS_DEV to `false` in `webviewContentGenerator` to test how it will look like in release mode.

Before (`IS_DEV = false`):
image

After (`IS_DEV = false`):
image

### DIFF
--- a/packages/vscode-extension/src/network/App.tsx
+++ b/packages/vscode-extension/src/network/App.tsx
@@ -1,5 +1,4 @@
 import "./App.css";
-import "../webview/styles/theme.css";
 import { useMemo, useState } from "react";
 import NetworkBar from "./components/NetworkBar";
 import NetworkRequestLog from "./components/NetworkRequestLog";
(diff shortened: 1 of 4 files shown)


filip131311 (approved):
LGTM

PR #1091, opened 2025-04-11
Delete chrome-devtools-ui from dependencies

kmagiera:
After #1057 we no longer need chrom-devtools-ui dependency in the project.

This PR deletes the submodule and also removes the build steps that were copying the data from that projects which results in significant reduction of the VSIX artifact.

### How Has This Been Tested: 
1. Build VSIX with `vscode:package`
2. Observe it is 10MB lighter than before
3. Install VSIX and test network panel


filip131311 (approved):
LGTM

PR #1095, opened 2025-04-14
Hide internal error popup from production version 

filip131311:
This PR hides internal error popup form production version of the IDE, as not every unhandled exception is necessary fatal. 

### How Has This Been Tested: 

- modify the `activate` function to look like this:
```js
  handleUncaughtErrors(context);
  setImmediate(()=>{throw new Error("test")});
...
}
```
- run it in development mode and observe the popup 
- build the package and open the vscode with the new package installed, popup does not appear

### DIFF
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -37,7 +37,7 @@ import { Connector } from "./connect/Connector";
 const OPEN_PANEL_ON_ACTIVATION = "open_panel_on_activation";
 const CHAT_ONBOARDING_COMPLETED = "chat_onboarding_completed";
 
-function handleUncaughtErrors() {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/extension.ts
```diff
@@ -37,7 +37,7 @@ import { Connector } from "./connect/Connector";
 const OPEN_PANEL_ON_ACTIVATION = "open_panel_on_activation";
 const CHAT_ONBOARDING_COMPLETED = "chat_onboarding_completed";
 
-function handleUncaughtErrors() {
+function handleUncaughtErrors(context: ExtensionContext) {
```
jwajgelt, line 40: VERY NIT: I'd just pass a boolean here, but this is fine.

jwajgelt (approved):

PR #1100, opened 2025-04-15
init mcp configurator

latekvo:
### Work in progress, not ready for review

---

This PR adds MCP support by creating the `mcp.json` file, or adding an entry to said file if it already exists.

To test this feature, ask any chatbot in Cursor for details about any specific react native library. 
A tool call will be made by the chatbot, which will be visible in the chat window.

### DIFF
--- a/packages/vscode-extension/src/mcp/index.ts
+++ b/packages/vscode-extension/src/mcp/index.ts
@@ -1,0 +1,79 @@
+// This module configures the MCP files for Cursor and VSCode
+
+import { Logger } from "../Logger";
+
+enum EditorType {
+  CURSOR = "cursor",
+  VSCODE = "vscode",
+}
+
+interface CursorMcpConfig {
+  mcpServers: Record<string, Object | string>;
+}
+
+interface VscodeMcpConfig {
+  servers: Record<string, Object | string>;
+}
+
+type McpConfig = CursorMcpConfig | VscodeMcpConfig;
+
+const VSCODE_FILE_PATH = ".vscode/mcp.json";
+const CURSOR_FILE_PATH = ".cursor/mcp.json";
+const MCP_BACKEND_URL = "https://radon-ai-backend.swmansion.com/sse";
+
+function getEditorType(): EditorType {
+  const editorType = EditorType.CURSOR;
+  return editorType;
+}
+
+function readMcpConfig(): McpConfig | null {
+  return null;
+}
+
+function writeMcpConfig(config: McpConfig) {
+  const editorType = getEditorType();
+
+  if (editorType === EditorType.CURSOR) {
+    Logger.info(`Writing to ${CURSOR_FILE_PATH}`);
+  } else if (editorType === EditorType.VSCODE) {
+    Logger.info(`Writing to ${VSCODE_FILE_PATH}`);
+  }
+
+  // Cannot have a default case here, as using invalid schema usually results in red errors being thrown.
+}
+
+function newMcpConfig(jwtToken: string): McpConfig {
+  const editorType = getEditorType();
+
+  const radonMcpEntry = {
+    RadonAi: {
+      url: MCP_BACKEND_URL,
+      type: "sse",
+      headers: {
+        Authorization: `Bearer ${jwtToken}`,
+      },
+    },
+  };
+
+  if (editorType === EditorType.VSCODE) {
+    return {
+      servers: radonMcpEntry,
+    };
+  }
+
+  return {
+    mcpServers: radonMcpEntry,
+  };
+}
+
+export function updateMcpConfig(jwtToken: string) {
+  let mcpConfig = readMcpConfig();
+
+  // todo: keep track of config changes, exit if none are detected
+
+  if (!mcpConfig) {
+    mcpConfig = newMcpConfig(jwtToken);
+  }
+
+  writeMcpConfig(mcpConfig);
+}


packages/vscode-extension/src/mcp/index.ts
```diff
@@ -0,0 +1,176 @@
...
+    Logger.info(`Valid MCP config already present.`);
+    return false;
+  }
+
+  const radonMcpEntry = {
+    RadonAi: {
+      url: MCP_BACKEND_URL,
+      type: "sse",
+      headers: {
+        Authorization: "Bearer ${command:RNIDE.getLicenseToken}",
```
latekvo, line 132: Note that `headers` do not work for now due to a [Cursor bug](https://forum.cursor.com/t/cursor-mcp-requests-do-not-honour-api-key-provision-to-sse-only-stdio-works/76889/2). The bug should be fixed with the next Cursor release (0.49.x).

packages/vscode-extension/src/mcp/index.ts
```diff
@@ -0,0 +1,176 @@
...
+    };
+  }
+
+  return {
+    mcpServers: {},
+  };
+}
+
+export async function updateMcpConfig() {
+  let mcpConfig = {};
```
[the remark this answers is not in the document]
latekvo: Marking as resolved. This is invalid, we don't want to do this.

packages/vscode-extension/src/extension.ts
```diff
@@ -299,6 +302,9 @@ export async function activate(context: ExtensionContext) {
   // You can configure the chat in package.json under the `chatParticipants` key
   registerChat(context);
 
+  // Enables Radon AI tooling on editors utilizing MCP configs (e.g. Cursor, VSC Preview).
+  updateMcpConfig();
+
```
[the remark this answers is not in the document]
latekvo: Marking as resolved. All errors are already handled asynchronously, awaiting `updateMcpConfig` is not necessary - Cursor listens to all mcp.json updates.

packages/vscode-extension/src/mcp/server.ts
```diff
@@ -0,0 +1,81 @@
...
+            },
+          ],
+        },
+      };
+    },
+  });
+
+  // todo: find a port programatically
+  const port = 21337;
+
```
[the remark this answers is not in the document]
latekvo: Resolved in code.

packages/vscode-extension/src/mcp/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      reject();
+    }
+
+    const folder = vscode.workspace.workspaceFolders?.[0];
+
+    if (!folder) {
+      Logger.error(`Couldn't read MCP config - no workspace folder open.`);
+      reject();
+    }
+
```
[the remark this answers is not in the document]
latekvo: Good catch.

packages/vscode-extension/src/mcp/index.ts
```diff
@@ -0,0 +1,161 @@
...
+      .then((data) => {
+        const config = JSON.parse(data);
+        Logger.info(`Found valid MCP config - updating.`);
+        resolve(config);
+      })
+      .catch(() => {
+        Logger.info(`Couldn't read MCP config - MCP config not found.`);
+        reject();
+      });
+  });
```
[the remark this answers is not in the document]
latekvo: Good catch. Rewritten that function.

packages/vscode-extension/src/mcp/server.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+async function screenshotToolDefinition(): ToolResponse {
+  const project = IDE.getInstanceIfExists()?.project;
+
+  if (!project || !project.deviceSession) {
+    return (
+      "Could not capture a screenshot!\n" +
+      "The development viewport device is likely turned off," +
+      "tell the user to turn on Radon IDE emulator before proceeding with a."
```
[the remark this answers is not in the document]
latekvo: Done.

packages/vscode-extension/src/extension.ts
```diff
@@ -296,9 +298,14 @@ export async function activate(context: ExtensionContext) {
     })
   );
 
+  const mcpPort = await startLocalMcpServer();
```
kmagiera: I'd rather have this thing extracted to a separate class / module. We don't need to block the initialization process waiting for the server to start. The only thing that needs to happen after the server is booted is to call the method that updates MCP config

packages/vscode-extension/src/mcp/api.ts
```diff
@@ -0,0 +1,88 @@
...
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        "Authorization": `Bearer ${token}`,
+      },
+      body: JSON.stringify({
+        tool_calls: [
+          {
+            name: toolName,
+            id: "3241", // temporarily necessary placeholder
```
kmagiera:
```suggestion
            id: "3241", // this is needed by the API, but the value doesn't matter
```

packages/vscode-extension/src/mcp/api.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (response.status !== 200) {
+      throw new Error(`Network error with status: ${response.status}`);
+    }
+
+    return response.json();
+  } catch (error) {
+    if (error instanceof Error) {
+      const msg = `Failed fetching tool schema with error: ${error.message}`;
+      Logger.error(MCP_LOG, msg);
+      getTelemetryReporter().sendTelemetryEvent("mcp:error", { error: msg });
```
kmagiera: do we want to measure successful calls to mcp. also, are you sure the error doesn't contain any information that is specific to user or their query? we don't want to send user's data to telemetry

packages/vscode-extension/src/mcp/index.ts
```diff
@@ -0,0 +1,161 @@
...
+const MCP_FILE_NAME = "mcp.json";
+
+function getEditorType(): EditorType {
+  // Cursor features different settings than VSCode
+  const config = vscode.workspace.getConfiguration();
+  if (config.get("cursor") !== undefined) {
+    return EditorType.CURSOR;
+  }
+  return EditorType.VSCODE;
+}
```
kmagiera: We should be using this API to set it up instead of writing to mcp config directly: https://code.visualstudio.com/api/extension-guides/mcp#register-an-mcp-server

if it isn't available in cursor, we can use that fact to detect cursor vs vscode

packages/vscode-extension/src/mcp/api.ts
latekvo: >Do we want to measure successful calls to mcp?

Sure, I can add that.

>Also, are you sure the error doesn't contain any information that is specific to user or their query?

I'm quite sure.

The telemetry call you highlighted only logs network errors (usually caused by invalid JWT).
As far as I see, there isn't any personal information contained within these errors.

We also use telemetry for `/api/tool_calls/`. These telemetry logs also don't contain user data,
as `/api/tool_calls/` only ever returns generic errors, like: `"Tool TOOL_NAME not found"`, `"Tool TOOL_NAME was called with missing required arguments."`, or `"Invalid JWT token Not enough segments"`.

I also checked `mcp/index.ts` - all of its telemetry uses are also entirely generic.

packages/vscode-extension/src/mcp/api.ts
latekvo: Done in [10c3cc4](https://github.com/software-mansion/radon-ide/pull/1100/commits/10c3cc490bdb45963f2a2ff8790eecdb49fc0693)

packages/vscode-extension/src/extension.ts
latekvo: We're not awaiting the server to start. We're awaiting a free port to be found, as well as a single API call. The latter is not intentional, and has been fixed in [2f9e33b](https://github.com/software-mansion/radon-ide/pull/1100/commits/2f9e33b0129659d60e2840884264afb3f00a4b21).

I've extracted all mcp init logic into a single function in [2f9e33b](https://github.com/software-mansion/radon-ide/pull/1100/commits/2f9e33b0129659d60e2840884264afb3f00a4b21).
Said function is not awaited. I also started awaiting the server to start, this will prevent any possible issues with `mcp.json` being read before the server is started.

packages/vscode-extension/src/mcp/index.ts
latekvo: This API doesn't seem to exist in the official release yet (checking on the latest version).
It's probably only available in the Insiders edition, not sure if there's a good reason to try to use it right away.

image
image
image

packages/vscode-extension/src/mcp/api.ts
latekvo: Added more telemetry in [e7ab9d6](https://github.com/software-mansion/radon-ide/pull/1100/commits/e7ab9d68b1f00b0cd847b2adda2a81ad0a498bfd) & [2319a2d](https://github.com/software-mansion/radon-ide/pull/1100/commits/2319a2d8cf561e6c6bb8e013e12e014baff7df54).

Let me know if this is what you've had in mind.
I didn't add logging on tool call success, as we're already logging tool call start, and tool call fails.

packages/vscode-extension/src/mcp/index.ts
kmagiera: it is available in version 1.100: https://github.com/microsoft/vscode/commit/0be40f9403ceb692dd7bc24029f2ced23b0745b7

packages/vscode-extension/src/mcp/index.ts
```diff
@@ -0,0 +1,195 @@
...
+      Logger.info(MCP_LOG, `Found valid MCP config - updating.`);
+      return config;
+    });
+  } catch {
+    // This is an expected warning, don't report via telemetry.
+    throw new Error(`Couldn't read MCP config - MCP config not found.`);
+  }
+}
+
+async function writeMcpConfig(config: McpConfig) {
```
kmagiera: would be nice to write some tests for this method. We already have tests setup and this just looks like an easily testable thing. Specifically we should verify that we're not deleting existing entries.

packages/vscode-extension/src/mcp/server.ts
```diff
@@ -0,0 +1,82 @@
...
+    return (
+      "Could not capture a screenshot!\n" +
+      "The development viewport device is likely turned off.\n" +
+      "Please turn on the Radon IDE emulator before proceeding."
+    );
+  }
+
+  const screenshot = await project.deviceSession.captureScreenshot();
+
+  const contents = readFileSync(screenshot.tempFileLocation, { encoding: "base64" });
```
kmagiera, line 31: I wonder, given the file is already on disk somewhere, maybe we can pass `file://` url to the LLM instead of reading and encoding it?

packages/vscode-extension/src/mcp/server.ts
```diff
@@ -0,0 +1,82 @@
...
+export async function startLocalMcpServer(port: number) {
+  const server = new LiteMCP("RadonAiServer", "1.0.0");
+
+  server.addTool({
+    name: "view_screenshot",
+    description: "Get a screenshot of the app development viewport.",
+    execute: screenshotToolDefinition,
+  });
+
+  const toolSchema = await getToolSchema();
```
kmagiera, line 57: is this ok to only run it once here? If someone activates the license at some later point, or just doesn't have the network connection, we will not register the tools. I wonder if it wouldn't be easier if we just hardcode the tools here and only rely on network when the tool is invoked.

packages/vscode-extension/src/mcp/server.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  for (const tool of toolSchema.tools) {
```
kmagiera, line 59: If the above request fails, we won't even register the screenshot tool as the server will never start

packages/vscode-extension/src/mcp/server.ts
latekvo: It might be better to disable authorization on the `/api/get_tool_schema/`.
Having to duplicate tool definitions both locally and remotely will be both annoying and harder to maintain, any minor changes and fixes will require PR to both the backend and the extension.

Let me know what you think.

packages/vscode-extension/src/mcp/server.ts
latekvo: This case is already handled. On fail, `getToolSchema()` returns `{ tools: [] }`.

packages/vscode-extension/src/mcp/server.ts
latekvo: The MCP standard does not seem to mention it being possible to expose a file as `file://` via a [Tool](https://modelcontextprotocol.io/docs/concepts/tools).
Exposing a file via `file://` as a [Resource](https://modelcontextprotocol.io/docs/concepts/resources) requires that it's static. When it's dynamic, it has to be converted to Base64 before being returned. See [here](https://modelcontextprotocol.io/docs/concepts/resources#reading-resources) for official specification. See [here](https://github.com/wong2/litemcp/tree/main?tab=readme-ov-file#resources) for documentation of the lib we're using.

If you're worried about the base64 conversion being a slow operation, I have to mention that it's being performed regardless of our actions. If we remove this conversion from our code base, it'll take place in the background.

packages/vscode-extension/src/mcp/index.ts
latekvo: Added tests for `newMcpConfig` and `insertRadonEntry` in [e5d9873](https://github.com/software-mansion/radon-ide/pull/1100/commits/e5d98730d5159c7f88f8da9c824b9b7f4b74341f) and [d7b5997](https://github.com/software-mansion/radon-ide/pull/1100/commits/d7b5997b34f27f7779cf94af21c7863a36edecd1).

I'll have to spend a little more time on tests for `writeMcpConfig()`, as I'm encountering errors within the `getTelemetryReporter()` when running tests on `readMcpConfig()` and `writeMcpConfig()`.

packages/vscode-extension/src/mcp/api.ts
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    if (response.status !== 200) {
+      throw new Error(`Network error with status: ${response.status}`);
+    }
+
+    const results: ToolResult = await response.json();
+
+    if (results.tool_results.length === 0) {
+      getTelemetryReporter().sendTelemetryEvent("mcp:error", { error: "Tool response empty." });
```
kmagiera: do we really want to label this as mcp:error, the error, if occurs isn't likely tied to mcp but to the radon AI api call, no?

packages/vscode-extension/src/mcp/server.ts
kmagiera: ok, but still we shouldn't resolve to just relying on this one being called once. I often have my project open in cursor for over a week. Also, when this runs with while not connected to network we won't have even the local tools registered as this blocks starting the server entirely.

Does MCP have any mechanism for updating the tools list? Because vscode and cursor need to deal with this scenario for sure and at least once in a while update the list from the MCP server. If that's the case we should likely hook into that call and fetch the list from our server then.

packages/vscode-extension/src/mcp/index.ts
```diff
@@ -0,0 +1,195 @@
...
+    Logger.error(MCP_LOG, msg);
+    getTelemetryReporter().sendTelemetryEvent("mcp:error", { error: msg });
+    throw new Error(msg);
+  }
+
+  Logger.info(MCP_LOG, `Reading MCP config at ${filePath}`);
+
+  try {
+    return await fs.readFile(filePath, { encoding: "utf8" }).then((data) => {
+      const config = JSON.parse(data);
```
kmagiera: I wonder if there's some built-in mechanism in vscode that allows for editing configuration files. I believe that vscode allows for any configuration to use a so-called [JSON-with-comments](https://code.visualstudio.com/docs/languages/json#_json-with-comments) which will likely throw here when parsing with JSON.parse

packages/vscode-extension/src/mcp/index.ts
```diff
@@ -0,0 +1,195 @@
...
+  try {
+    mcpConfig = await readMcpConfig();
+  } catch (info) {
+    if (info instanceof Error) {
+      Logger.info(MCP_LOG, info.message);
+    } else {
+      Logger.info(MCP_LOG, String(info));
+    }
+
+    mcpConfig = newMcpConfig();
```
kmagiera: I think we should only create new config if the file doesn't exists and not for every possible error. I'd prefer that we don't overwrite the config in case there's a parsing error or for example file locking issue. In such scenario we should just bail here

kmagiera (approved):
Added a couple more comments. Accepting here but please address the comments before merging.

packages/vscode-extension/src/mcp/api.ts
latekvo: I agree, does `radon-ai:server-error` work better?
If so, I'll change all other telemetry calls to match this convention as well.

packages/vscode-extension/src/mcp/index.ts
latekvo: You're right. Fixed in [b9a4220](https://github.com/software-mansion/radon-ide/pull/1100/commits/b9a422032ea8c19aecb9948bf4c64f3aee8a4f50)

packages/vscode-extension/src/mcp/api.ts
latekvo: Rewritten all telemetry labels in [70ae676](https://github.com/software-mansion/radon-ide/pull/1100/commits/70ae6761eaf609852ed2604a97bf3f9b5b739818)

latekvo:
> > ok, but still we shouldn't resolve to just relying on this one being called once. I often have my project open in cursor for over a week. [...] Does MCP have any mechanism for updating the tools list? Because vscode and cursor need to deal with this scenario for sure and at least once in a while update the list from the MCP server. If that's the case we should likely hook into that call and fetch the list from our server then.
>
> I haven't thought of that... I know that both vscode and cursor periodically suggest reloading the tool list (in vscode the "servers" button highlight in bright blue with a "reload" icon), although I'm not sure how often that happens, or what triggers it. It would make sense to add an automatic watcher for new tools being added, I will have to look into how the litemcp server works, if it's possible to either kill or modify it after it's been already launched.

Fixed in [6df1b7a](https://github.com/software-mansion/radon-ide/pull/1100/commits/6df1b7a9f5fa0b47fcd0a01ea37f2a325b403c02). 
On every license token update, the previous MCP server is killed, and a new one is started, which ensures:
1. Tools are reloaded at least once a day.
2. Radon AI starts working as soon as the user opens up the IDE panel or activates their license.

One edge case still persists - the user only uses Radon AI and not the IDE panel. In such case, the license token will never be reloaded.

packages/vscode-extension/src/mcp/server.ts
latekvo: Fixed and answered [here](https://github.com/software-mansion/radon-ide/pull/1100#issuecomment-2962322316)

packages/vscode-extension/src/mcp/index.ts
latekvo: Works, done in [c65bccb](https://github.com/software-mansion/radon-ide/pull/1100/commits/c65bccb4085b1917d7a9ef72fd652fa3ae9f9e58) and [5a5c77b](https://github.com/software-mansion/radon-ide/pull/1100/commits/5a5c77b2761c31d920b3eaa6b3154d584b02238f)

image

noma4i:
Would be extra helpfull to be able to retrieve metro logs to MCP. This is a main loop I face where there is an error in the logs like rule of hooks or typo made by LLM so I need to copy-paste the log to chat. Partially will be solved via sending screenshot but react doesn't show all the errors on the device screen.

PR #1103, opened 2025-04-16
update "Auto" description for eas build id

jwajgelt:
Updates the description of the "Auto" option for EAS build id in the lauch config dialog.

### DIFF
--- a/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
+++ b/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
@@ -522,7 +522,7 @@ function EasBuildConfiguration({
             className="input-configuration"
             type="string"
             defaultValue={buildUUID ?? ""}
-            placeholder="Auto (the latest available build will be used)"
+            placeholder="Auto (newest build with matching fingerprint)"
             onBlur={onBuildUUIDInputBlur}
           />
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
lgtm

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -522,7 +522,7 @@ function EasBuildConfiguration({
             className="input-configuration"
             type="string"
             defaultValue={buildUUID ?? ""}
-            placeholder="Auto (the latest available build will be used)"
+            placeholder="Auto (newest build with matching fingerprint)"
```
kmagiera, line 525: is "newest" here actually accurate? Could there be multiple builds with matching fingerprint? Even if that's the case, I think we should put emphasis on "matching fingerprint" here and write:
> "build with matching fingerprint (most recent one if there are many)"
or just
> "build with matching fingerprint"
as I expect it isn't common scenario there are multiple builds anyways and taking the most recent one in that case seems like a logical choice.

I think it'd be also good to update description in `package.json` as we should document the default state for `buildUUID` when it is not specified (i.e. "uses build with matching fingerprint when no buildUUID is specified")

PR #1104, opened 2025-04-16
Support EAS local builds

jwajgelt:
- Allows configuring EAS apps to build locally instead of downloading from the EAS Build server.
- Allows setting the EAS build `local` flag through UI.
- Updates the docs to include the new flag.

image

## How Has This Been Tested:
- open the `expo-52-eas` test app
- add a `"local": true` entry to `eas.ios` or `eas.android` launch config
- check that the app is built locally for the platform
- open the "Launch Configuration" menu
- verify you can select "Build Locally > Yes/No" under the EAS options and that the setting is respected

### DIFF
--- a/packages/vscode-extension/src/builders/easCommand.ts
+++ b/packages/vscode-extension/src/builders/easCommand.ts
@@ -1,6 +1,7 @@
+import { OutputChannel } from "vscode";
 import { DevicePlatform } from "../common/DeviceManager";
 import { Logger } from "../Logger";
-import { exec } from "../utilities/subprocess";
+import { exec, lineReader } from "../utilities/subprocess";
 
 type UnixTimestamp = number;
 
@@ -62,6 +63,11 @@ export async function isEasCliInstalled(appRoot: string) {
   }
 }
 
+const DEVICE_TO_EAS_PLATFORM_MAPPING = {
+  [DevicePlatform.Android]: "android",
+  [DevicePlatform.IOS]: "ios",
+};
+
 type ListEasBuildsOptions = { profile: string; fingerprintHash: string };
 
 export async function listEasBuilds(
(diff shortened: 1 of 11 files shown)


packages/vscode-extension/src/builders/easCommand.ts
```diff
@@ -161,3 +170,36 @@ export async function generateFingerprint(
...
+    outputPath,
+  ];
+  const buildProcess = exec("eas", commandArgs, { cwd: appRoot, quietErrorsOnExit: true });
+  lineReader(buildProcess).onLineRead((line) => {
+    outputChannel.appendLine(line);
+  });
+  try {
+    await buildProcess;
+  } catch {
+    throw new Error("EAS local build failed. See the build logs for details.");
```
filip131311, line 203: shouldn't this be `BuildError`?

filip131311 (commented):
Left one inline comment, but i would prefer to review it again after #1097 is mereged

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -440,7 +443,9 @@ function EasBuildConfiguration({
   const updateEasConfig = (configUpdate: Partial<EasConfig>) => {
     const currentPlaftormConfig = eas?.[platform] ?? {};
     const newPlatformConfig = Object.fromEntries(
-      Object.entries({ ...currentPlaftormConfig, ...configUpdate }).filter(([_k, v]) => !!v)
+      Object.entries({ ...currentPlaftormConfig, ...configUpdate }).filter(
```
filip131311, line 446: We used to just remove the `eas` configuration for the platform if the user have chosen the disabled option but it seems to me that after this PR it might be frustrating for a user that have chosen the `build locally` option and then chosen `disabled` option that his choice is not save I would suggest to introduce a new `disabled` state that is not represented as just lack of configuration to avoid that problem.

filip131311 (commented):
I left one reasnoble and one not important comment inline but the code looks good

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
filip131311: I'll leave it to your judgment then, this is a niche use case anyway

filip131311 (approved):
👍

PR #1106, opened 2025-04-17
Fix Extension Development Host failing to start again

jwajgelt:
Fixes a problem where, after launching once, trying to start/restart the Extension Development Host would not work.
Fixes the issue where the network inspector's development server never actually launched.


filip131311 (approved):
looks good, thank you <3

PR #1107, opened 2025-04-17
[chore] Update deps that don't cause any issues

kmagiera:
This PR cleans up our package dependencies. We update most of deps that don't cause any issues when bumped. We only skip the following ones:
 - vite and its plugins – which seem to require some additional work to get updated, just bumping it results in the vite server throwing errors in dev mode when loading webview app
 - execa – which we already started using from an old version because of lack of commonjs support in vscode (now it is present). Bumping it now will require updates in several critical code paths so not doing it here.

On top of that we also remove dependencies that are deprecated or not used at all, those are:
 - chrome-remote-interface, glob – not used
 - @vscode/webview-ui-toolkit – is deprecated and we already added @vscode-elements package – in this PR we update a few places where we still used the ui-toolkit

Eslint version bump also require that we migrate the config file from .eslintrc.json to `eslint.config.mjs` – this was done using automated tool `@eslint/migrate-config`

Additionally, update of prettier resulted in format check failing in several files – mostly CSS. Those files are updated in the PR as well.

This change is largely needed for future work on devtools which require React 19 on the frontend.

### How Has This Been Tested: 
1. Full test on expo-51
2. Verify the progress ring displays properly (it is only shown when Radon IDE is loading + I also forced deep link search component in code to be in loading state to see the ring there)


filip131311 (approved):
lgtm

PR #1109, opened 2025-04-17
Remove vite call from src/network

kmagiera:
Calling `vite` from `src/network` is unnecessary. First, because `vite` command goes into watch mode so it will never execute the commands after "&&". We also don't need to have a separate instance of vite for network panel because the main instance hosts the network and webview panels.

### How Has This Been Tested: 
1. Build in dev mode and make sure network panel loads


filip131311 (approved):
👍

PR #1111, opened 2025-04-23
fix test runner, remove outdated tests

jwajgelt:
After #1107 the test runner crashes due to issues with `node-fetch` and `mocha` imports.
Also, the Linux platform check test fails, since rudimentary Linux support has been added a while back.
This PR resolves these issues.

### How Has This Been Tested: 
- run `npm run test`
- the tests should pass and the test runner shouldn't crash


packages/vscode-extension/package.json
```diff
@@ -771,8 +771,8 @@
     "file-loader": "^6.2.0",
     "lodash": "^4.17.21",
     "minimatch": "^10.0.1",
-    "mocha": "^11.1.0",
-    "node-fetch": "^3.3.2",
+    "mocha": "^10.2.0",
```
filip131311, line 774: why is the new version incompatible with our set up? do we have some other dependency that clashes with it?

packages/vscode-extension/tsconfig.test.json
```diff
@@ -8,6 +8,7 @@
     "rootDir": "src",
     "strict": true,
     "esModuleInterop": true,
+    "resolveJsonModule": true,
```
filip131311, line 11: why is it related to the tests crashing, there was a repository state when they worked and this was set to default value

filip131311 (commented):
inline

packages/vscode-extension/package.json
jwajgelt: As for `node-fetch`, they stopped shipping "CommonJS" modules in version 3, which results in `tsc` failing to build anything.
I'm not sure why it works with the main package though, since `tsconfig` is the same there. I wasn't able to resolve it any other way, and didn't want to spend too much time on it.

filip131311 (approved):
I run it and it works <3

PR #1117, opened 2025-04-24
Add react devtools profiler support

kmagiera:
This PR adds a new tool – React Profiler

React Profiler integration allows controlling (starting/stopping) of the profile via React Devtools and also to open the profiles in a standalone editor panel.

https://github.com/user-attachments/assets/b525808b-6eb1-4fa4-95ac-de052121e7de

Below is an overview of the changes made in this PR:
 - We use devtools-hedless package to access react devtools' profiler store to control the profiling
 - When profiling is finished, we use the export method exposed via devtools-headless package to dump the profile into a file and save under '.reactprofile' extension
 - We're implementing a new editor provider that can open '.reactprofile' files. The editor uses the devtools-inline package to present the UI and uses the profile import method from devtools-headless package to process the profile data.
 - We're adding buttons to the tools menu to initiate the profiling and use the CPU profiling style indicator when the profiler is running.
 - We're updating vite configuration – for some reason after adding a new input, vite in release mode started splitting CSS into more chunks. Before it was just main chunk and theme chunk. Instead of adding more dependencies I decided it will be easier to force it to not do chunking at all and use single style.css for all the webviews.
 - We are adding additional options to the webview CSP because devtools use JS's eval and also boot up a worker thread using blob URL.

### How Has This Been Tested: 
1. Open IDE panel and wait for app to load
2. Click "Start React Profiler"
3. Observe a profiler indicator appear (similar as with the CPU profiling)
4. Do something with the app
5. Click stop
6. The profiler will stop (indicator is gone) and a profile will be automatically opened in a new editor tab


filip131311 (commented):
I believe that the react profiler has an option to be displayed in light mode, if it is possible to set it in the hedless package we use, I think it would look much better for users that use light mode in their vscode

image

kmagiera:
forgot about that, just made an update to account for the theme

packages/vscode-extension/src/project/project.ts
```diff
@@ -291,6 +294,18 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
 
+  async startProfilingReact() {
+    await this.deviceSession?.devtools.startProfilingReact();
+  }
+
+  async stopProfilingReact() {
+    const uri = await this.deviceSession?.devtools.stopProfilingReact();
+    if (uri) {
+      // open profile file in vscode using our custom editor
+      commands.executeCommand("vscode.open", uri);
```
jwajgelt, line 305: Does VSCode guess the editor using the file extension? Maybe there's some way to explicitely tell the `vscode.open` command which editor it should open with?

packages/vscode-extension/src/webview/components/ToolsDropdown.tsx
```diff
@@ -96,6 +96,15 @@ function ToolsDropdown({ children, disabled }: { children: React.ReactNode; disa
             <span className="codicon codicon-chip" />
             {isProfilingCPU ? "Stop JS CPU Profiler" : "Start JS CPU Profiler"}
           </DropdownMenu.Item>
+          <DropdownMenu.Item
+            className="dropdown-menu-item"
+            onSelect={() =>
+              isProfilingReact ? project.stopProfilingReact() : project.startProfilingReact()
```
jwajgelt, line 102: What happens if `startProfilingReact()` is called twice, without calling `stopProfilingReact()`?
Can't that happen if the user tries to toggle this quickly, while the extension hasn't sent the `isProfiling` event yet?

jwajgelt (approved):
The code looks ok, but I haven't tested how it works yet.

jwajgelt:
image
The UI crops the settings cog when the panel is not wide enough. The same happens with the CPU profiler. It's not something that should be fixed in this PR, but we should try and solve this somehow.

jwajgelt:
After I took the profile, it took a good while to open the editor tab with it, with no feedback (that I could see) that something's happening in the background. Might be worth adding some sort of spinner while the profile is being opened to indicate it's busy, not broken.

packages/vscode-extension/src/project/project.ts
kmagiera: we register custom editor to open .reactprofile files and this works based on that mechanism. The file is saved with .reactprofile extension. This is also good, because if someone saves the file on disk, they can open it later on using in VSCode using our editor.

The only case where it may not work is if some other extension registers an editor that opens .reacprofile files in which case I don't necessarily have anything agains that because it'd be the user's choice to use a different editor over the one that we provide.

There is a command called `vscode.openWith` that allows you to specify the editor but for the reasons mentioned above I'd rather use just plain `open` method here and rely on VSCode to use the editor that the user prefers or the one that we provide (in most of the cases)

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -107,10 +118,14 @@ export class Devtools implements Disposable {
...
       const bridge = createBridge(wall);
       const store = createStore(bridge);
+      this.store = store;
 
       ws.on("close", () => {
         this.socket = undefined;
         bridge.shutdown();
+        if (this.store === store) {
+          this.store = undefined;
+        }
```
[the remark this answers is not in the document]
kmagiera: the check was just defensive code as we never call `starInternal` more than once on devtools. But it'd be good to have the same check for `this.socket` anyways.

packages/vscode-extension/src/webview/components/ToolsDropdown.tsx
kmagiera: this is safe to be called, the subsequent calls are going to be ignored.

PR #1118, opened 2025-04-24
Allow multiple devices running at the same time

filip131311:
This PR allows for multiple devices running at the same time, with smooth transition between them. To achieve this we implement activation and deactivation system for device sessions

### How Has This Been Tested: 

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.


jwajgelt (approved):
I still have the same issues with the API, but I understand this is more of a continuation of #1084 rather than actual implementation of the feature (since there doesn't seem to be any way for the user to trigger this?), so my problems will have to wait until the PR with the UI.
The code itself looks ok.

PR #1120, opened 2025-05-05
fix debugger always restarting after app reload

jwajgelt:
Currently, the logic for reconnecting the debugger after the application is reloaded always restarts the debug session.
This was uncaught since the behaviour is _technically correct_ (in that in the end, we'll always have a debug session connected to the application after reload), but undesirable (we'd rather reuse the existing debug session since its initialization is relatively slow).
This PR fixes the minor bugs which caused the issue.

Right now, after this fix is applied, the current debugger implementation breaks after JS reload, presumably due to issues with how source maps are cleaned up.
This surfaces by not mapping the sources correctly in the Debug Console (instead always mapping to `__source__`)
This PR also makes the debug session always restart on the old debugger as a workaround for that issue.

### How Has This Been Tested: 
- use the `PROXY_JS_DEBUGGER_TYPE` in `DebugSession.startJSDebugSession` to enable the new `js-debug`-based debugger
- add log points to `DeviceSession.reconnectJSDebuggerIfNeeded`
- check that after "Reload JS" is triggered in Radon, the function returns early (without calling `this.connectJSDebugger`)

### DIFF
--- a/packages/vscode-extension/src/debugging/CDPDebugAdapter.ts
+++ b/packages/vscode-extension/src/debugging/CDPDebugAdapter.ts
@@ -339,6 +339,13 @@ export class CDPDebugAdapter extends DebugSession implements CDPSessionDelegate 
   }
 
   private async ping() {
+    // NOTE: This is a temporary workaround to avoid issues with source maps
+    // breaking after a JS reload happens.
+    // This causes the debugger session to always restart after a JS reload happens.
+    const ALWAYS_FAIL = true;
+    if (ALWAYS_FAIL) {
+      return false;
+    }
     if (!this.cdpSession) {
       Logger.warn("[DebugAdapter] [ping] The CDPSession was not initialized yet");
       return;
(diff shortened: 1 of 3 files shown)


filip131311 (commented):
The code looks good but it exposes other bugs in our debugger implementation, so I would not marge it until they are fixed

packages/vscode-extension/src/debugging/CDPDebugAdapter.ts
```diff
@@ -339,6 +339,13 @@ export class CDPDebugAdapter extends DebugSession implements CDPSessionDelegate
   }
 
   private async ping() {
+    // NOTE: This is a temporary workaround to avoid issues with source maps
+    // breaking after a JS reload happens.
+    // This causes the debugger session to always restart after a JS reload happens.
+    const ALWAYS_FAIL = true;
```
filip131311, line 345: I don't like how deeply buried in the implementation this is, it's really easy to miss in the future, I think this force fail should be implemented in `reconnectJSDebuggerIfNeeded` when we check if reconnection is in fact needed.

filip131311 (changes requested):
I left inline comment

packages/vscode-extension/src/debugging/CDPDebugAdapter.ts
jwajgelt: `reconnectJSDebuggerIfNeeded` isn't (and shouldn't) be aware of which debug implementation is used, it's an implementation detail of `DebugSession`.
An argument could be made that `DebugSession` should control this, but I still disagree. The issue lies in `CDPDebugAdapter`, so it should inform its callers that it always needs to be reconnected, rather than the caller deciding.

filip131311 (approved):
Thank you

PR #1125, opened 2025-05-06
Fix js debug startup delay and reenable js-debug for new debugger targets

jwajgelt:
This PR fixes the issue of _excruciatingly_ long debug startup on larger applications when using the `js-debug`-based debugger implementation and reenables it for React Native apps using the new ("Fusebox") debugger.

It also updates the version of `js-debug` used and applies patches which should improve startup performance in the React Native case, as described in the PR: https://github.com/software-mansion-labs/vscode-js-debug/pull/4

### How Has This Been Tested: 
- launch `social-app` (or another large application)
- set some breakpoints
- restart the application a couple of times in various ways (JS reload, full app restart, IDE restart)
- verify that the delay before the set breakpoints are active (indicated in the UI by the breakpoint dot turning red) takes on the order of seconds, and not tens-of-seconds

### DIFF
--- a/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
+++ b/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
@@ -4,12 +4,13 @@ import _ from "lodash";
 import { CDPProxyDelegate, ProxyTunnel } from "./CDPProxy";
 import { SourceMapsRegistry } from "./SourceMapsRegistry";
 import { Logger } from "../Logger";
+import { Minimatch } from "minimatch";
 
 export class RadonCDPProxyDelegate implements CDPProxyDelegate {
   private debuggerPausedEmitter = new EventEmitter<{ reason: "breakpoint" | "exception" }>();
   private debuggerResumedEmitter = new EventEmitter();
   private consoleAPICalledEmitter = new EventEmitter();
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -18,7 +19,12 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
   public onDebuggerResumed = this.debuggerResumedEmitter.event;
   public onConsoleAPICalled = this.consoleAPICalledEmitter.event;
 
-  constructor(private sourceMapRegistry: SourceMapsRegistry) {}
+  constructor(
+    private sourceMapRegistry: SourceMapsRegistry,
+    patterns: string[]
```
kmagiera, line 24: I'd keep "skipFiles" name here and in the class instance. The "skipFiles" concept have the semantics of regular and negated patterns that needs to be considered in particular order while "patterns" is more a generic term

kmagiera (approved):
the skip file logic is quite confusing. I'd rather avoid using reduce there which adds extra bit of confusion. In `CDPSession.ts` we already have this logic which is writte using `flipNegate` option for Minimatch and IMO is easier to follow, but that's just my preference so not going to request changes here unless you feel the same.

PR #1127, opened 2025-05-07
wait for sources before jumping to breakpoint

jwajgelt:
After a JS Reload is triggered, in the new Hermes debugger implementation the breakpoints remain set, but our CDP implementation discards the source maps. This can cause a breakpoint to be hit before the source maps are loaded by the debugger, resulting in the IDE trying to display the unmapped source (as shown in the screenshot:)

image

This PR waits for the source maps to load before displaying the breakpoint to the user, which avoids the issue

### How Has This Been Tested: 
- launch an app
- set breakpoints such that they will be hit soon after loading the bundle
- trigger a JS Reload
- verify that the IDE displays the position in the source code where the breakpoint was set

### DIFF
--- a/packages/vscode-extension/src/debugging/CDPSession.ts
+++ b/packages/vscode-extension/src/debugging/CDPSession.ts
@@ -18,6 +18,7 @@ import { CDPCallFrame, CDPDebuggerScope, CDPRemoteObject } from "./cdp";
 import { typeToCategory } from "./DebugAdapter";
 import { annotateLocations } from "./cpuProfiler";
 import { CDPConfiguration } from "./CDPDebugAdapter";
+import { Disposable, EventEmitter } from "vscode";
 
 type ResolveType<T = unknown> = (result: T) => void;
 type RejectType = (error: unknown) => void;
@@ -59,6 +60,9 @@ export class CDPSession {
   private breakpointsController: BreakpointsController;
 
   private debugSessionReady = false;
+  private debugSessionReadyEmitter = new EventEmitter();
+  private onDebugSessionReady = this.debugSessionReadyEmitter.event;
+
   private consoleAPICallsQueue: any[] = [];
 
   private resumeEventTimeout: NodeJS.Timeout | undefined;
@@ -148,6 +152,7 @@ export class CDPSession {
 
         if (isMainBundle) {
           this.debugSessionReady = true;
+          this.debugSessionReadyEmitter.fire({});
           this.flushEnqueuedConsoleAPICalls();
           this.delegate.onDebugSessionReady();
         }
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/debugging/CDPSession.ts
```diff
@@ -286,6 +291,26 @@ export class CDPSession {
   }
 
   private async handleDebuggerPaused(message: any) {
+    if (!this.debugSessionReady) {
+      await new Promise((resolve, reject) => {
```
kmagiera, line 295: I prefer `Promise.withResolvers()` approach as it is easier to follow and we are using it more and more often thorough the codebase.

packages/vscode-extension/src/debugging/CDPSession.ts
```diff
@@ -148,6 +152,7 @@ export class CDPSession {
 
         if (isMainBundle) {
           this.debugSessionReady = true;
+          this.debugSessionReadyEmitter.fire({});
```
kmagiera, line 155: given you added emitter maybe you can refactor `flishEnqueuedConsoleAPICalls` to just follow the same approach (instead of adding to queue we create event subscription for every log entry that needs to be scheduled)

kmagiera (approved):

packages/vscode-extension/src/debugging/CDPSession.ts
jwajgelt: I'm not 100% sure the order of listeners is guaranteed, so I'd rather not.
We can still have a listener which flushes queue though, it should be cleaner.

packages/vscode-extension/src/debugging/CDPSession.ts
jwajgelt: I felt like since we `await` it immediately, the constructor form made more sense, but maybe you're right.

packages/vscode-extension/src/debugging/CDPSession.ts
jwajgelt: Extracted the "resolve a promise on event or reject after a timeout" part to a helper function, I think it reads better that way.

packages/vscode-extension/src/debugging/CDPSession.ts
jwajgelt: Extracted the "resolve a promise on event or reject after a timeout" part to a helper function, I think it reads better that way.

packages/vscode-extension/src/debugging/CDPSession.ts
jwajgelt: Moved the body of this `if` statement to a listener on `onDebugSessionReady`

PR #1129, opened 2025-05-07
fix debugger not ignoring injected files correctly

jwajgelt:
Due to the difference in paths between development mode and packaged extensions, the `lib/` directory of the extension was not correctly ignored when the extension was installed from a package (rather than running in the Extension Development Host).
This PR fixes that.

### How was this tested?
- build an install an extension `.vsix` package
- open a test app from `radon-ide-test-apps`
- check "Uncaught exceptions" breakpoint
- click the "Check uncaught exceptions" button
- verify the debugger doesn't pause inside RN renderer

### DIFF
--- a/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
+++ b/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
@@ -23,7 +23,9 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
     private sourceMapRegistry: SourceMapsRegistry,
     skipFiles: string[]
   ) {
-    this.ignoredPatterns = skipFiles.map((pattern) => new Minimatch(pattern, { flipNegate: true }));
+    this.ignoredPatterns = skipFiles.map(
+      (pattern) => new Minimatch(pattern, { flipNegate: true, dot: true })
+    );
   }
 
   public async handleApplicationCommand(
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -23,7 +23,9 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
     private sourceMapRegistry: SourceMapsRegistry,
     skipFiles: string[]
   ) {
-    this.ignoredPatterns = skipFiles.map((pattern) => new Minimatch(pattern, { flipNegate: true }));
+    this.ignoredPatterns = skipFiles.map(
+      (pattern) => new Minimatch(pattern, { flipNegate: true, dot: true })
```
kmagiera, line 27: Is this change on purpose? There's no trace of it in the description

kmagiera (approved):
Looks good, just not sure if the minimatch change is intentional

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
jwajgelt: It's necessary, because the extension directory (potentially) includes a dotted path (`.vscode` in my case), and without this option, `Minimatch` doesn't match it.

PR #1132, opened 2025-05-08
Fix reload after bundle error

filip131311:
This PR fixes an issue with how launch completion was reported that prevented users from using preview functionality in the following scenario:
- cause a bundle error 
- try to reload in that state 
- fix the bundle error 
- run auto-reload 
- now the app is launched but `isLaunching` prop is set to `true`

### How Has This Been Tested: 

run the reproduction and check if it works


jwajgelt:
Closing in favour of the (much handsomer) #1136

PR #1133, opened 2025-05-09
Fix build error dialog not appearing on reload

jwajgelt:
Currently, when a build error happens during a reload, we fail to display the error dialog, instead hanging on "Building...".
The root cause was us not catching the error in `DeviceSessionsManager.reload`, meaning we never set the project state to "build error".
This PR fixes that.

### How Has This Been Tested: 
- have an app which fails to build
- open it in Radon
- wait for the build to fail
- reload the app (either from the error dialog or with the Reload button in the URL bar)
- the build error dialog should appear again <- previously, the build would appear to hang instead


filip131311 (approved):
lgtm

PR #1143, opened 2025-05-12
Improve UrlBar/UrlSelect functionality

HexHyperion:
This PR improves the UrlSelect component responsible for navigating through the app's path history. Previously, the feature functioned like a `<select>`, only allowing traversal to locations that have been visited before.

Now, the user will be able to enter any path, valid or not, which will be passed to the emulator and handled by the application. The old dropdown has been repurposed as a suggestion and history panel, similar to what we see in modern web browsers. In the future, the panel is planned to list all available static paths, too.

Recording of the functionality: https://github.com/user-attachments/assets/21f01e74-50e0-4fc5-9e38-512dfe4419bd

### How Has This Been Tested:
- Open any multi-screen RN project using Expo Router or React Navigation
- The path bar becomes enabled once the app is fully loaded
- Any changes in navigation invoked from the emulator will be instantly visible in the bar
- Click on the path field - it should become active and open the suggestions
- Enter the desired path and press Enter/Return - the app should navigate to the location or display a "404" page if it's invalid
- As you're writing, suggestions from the history should appear below
- When the field is active, press the down arrow on your keyboard - the first option from the suggestion panel should become focused - confirm with Enter/Return, navigate with up/down arrows
- When the first option from the list is highlighted, pressing the up arrow should focus the input field
- Open a component preview from the editor window - the app should display it and a path starting with `preview://` should show up on the path bar
- The preview URL appears in the history and is navigable like any other URL


packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -23,58 +34,255 @@ interface UrlSelectProps {
...
+                            const prevItem = (e.target as HTMLDivElement).previousElementSibling;
+                            if (prevItem && prevItem.classList.contains("url-select-item")) {
+                              (prevItem as HTMLDivElement).focus();
+                            } else {
+                              (textfieldRef.current as HTMLInputElement).focus();
+                            }
+                          }
+                        }}>
+                        {getNameFromId(item.id)}
+                      </PopoverItem>
```
jwajgelt, line 220: Personally, I find large chunks of logic inside the component's `tsx` return value hard to read.
I'd extract this to a separate function, which takes the `item` and returns the `PopoverItem` element with the props set.
Since you want to access some values from `UrlSelect` you might need to define that function inside it, but if I'm not wrong, the only thing you actually need is
```ts
handleValueChange(`filtered#${item.id}`);
setInputValue(getNameFromId(item.id));
setIsDropdownOpen(false);
```
So you could just pass it as a callback (called `onItemSelected` or something) to it, and make this into a separate component outside of `UrlSelect`.

jwajgelt (changes requested):
Left a couple of inline comments, but overall, the code looks very nice to me.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
HexHyperion: Okay, done, but now I wonder if it's better the way it is now, or like in the commit before, inside `UrlSelect`? Because when it's inside, the `<PopoverItem>` in JSX needs only `item` as a prop.
Or have I made the whole keyboard navigation function actually worse than the `onKeyDown`'s at the beginning?

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -22,59 +60,208 @@ interface UrlSelectProps {
...
+      setFilteredItems(filtered);
+    }
+  }, [inputValue, items]);
+
+  useEffect(() => {
+    const filteredOut = items.filter(
+      (item) => !filteredItems.some((filteredItem) => filteredItem.id === item.id)
+    );
+    setFilteredOutItems(filteredOut);
+  }, [items, filteredItems]);
```
jwajgelt, line 141: Since the only place that `filteredItems` are set is in the first effect here, I think we could combine these two `useEffect` into a single one.
Then, you could also get the `filteredOutItems` at the same time as you calculate `filteredItems`, instead of checking for each element if it's in the `filteredItems` list (using either a for loop or `lodash`'s [`partition`](https://lodash.com/docs/#partition))
```suggestion
  useEffect(() => {
    if (disabled) {
      setFilteredItems([]);
      setFilteredOutItems(items);
      return;
    }
    
    const inputValueLowerCase = inputValue.toLowerCase();
    const [filtered, filteredOut] = _.partition(items, (item) =>
      getNameFromId(item.id).toLowerCase().includes(inputValueLowerCase)
    );
    setFilteredItems(filtered);
    setFilteredOutItems(filteredOut);
  }, [inputValue, items]);
```
(actually, I'm not sure about the `if (disabled)` behaviour -- I think what you're doing right now is just `return;`, but I'm not sure that's what we want.)

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -22,59 +60,208 @@ interface UrlSelectProps {
...
+            {filteredItems.length > 0 && filteredOutItems.length > 0 && (
+              <div className="url-select-separator" />
+            )}
+
+            {filteredOutItems.length > 0 && (
+              <div className="url-select-group url-select-group-other">
+                <div className="url-select-label">Other paths:</div>
+                {filteredOutItems.map(
+                  (item) =>
+                    item.name && (
```
jwajgelt, line 249: NIT: You could probably filter out `item`s with `item.name === ""` in the effect where you set `filteredItems` and `filteredOutItems`, but the current approach is also good (since you can see that only non-empty items are rendered). Feel free to do whichever you think is better.

jwajgelt (approved):
Code looks good to me, and seems to work fine as well.
I'm. not a fan of the `document.querySelector` calls to find the next item to focus, but I also don't know a better way to achieve that -- maybe someone more experienced with webdev will have an idea.

HexHyperion:
> not a fan of the `document.querySelector` calls to find the next item to focus, but I also don't know a better way to achieve that

I guess adding the `PopoverItem`s to a set with `item.id` as a key would do the trick, but I'll probably change the dropdown layout anyway when playing with route indexing, so maybe then?

PR #1147, opened 2025-05-13
Improve window management

kmagiera:
This PR improves window/panel position management. Before, when using IDE in editor panel, we'd always get the editor position structure messed up when restarting the editor. Thanks to the made changes, we can also take adventage of the new "compact" mode for new windows.

Here's what we're changing:
1. We're deleting code for handling secondary side bar. This includes the setting option etc. Due to limitation of vscode extension API, we cannot control this way and the value gets out of sync between the UI and our configuration. Going forward we have to rely on the user moving it correctly (will be updating the docs to reflect that as well).
2. The panel location now only tell whether we should use "side-bar" or "tab" mode for the IDE panel. This is the only combination that we can reliably implement. 
3. We're updating the menu option for changing the position to only list "Move to Side Bar", "Move to New Window" and "Move to Editor Tab". We hide the side bar option when we know the panel is in the side bar. The other options stay on at all time as we can still shuffle the view between tabs and windows even when the panel is already in new window or in editor tab.
4. We're removing `OPEN_PANEL_ON_ACTIVATION` in favor of using WebviewSerializer API. The Serializer API allows us to rely on VSCode to restore the window in the right place after the editor is restarted. This is much more reliable than what we had before.
5. We're changing the "showIDEPanel" logic, such that it takes the new location as parameter. Sometimes the parameter will be left out, for example when the panel location changes as a result of the settings change. The parameter can also include a hint whether the "tab" mode is requested with "move to new window" or "move to editor tab".
6. When opening panel in new window, we use the new "compact" mode added to VSCode 1.100

### How Has This Been Tested: 
1. Run any project and play with the panel position
2. Test the following conversions:
  a) side-panel -> editor-tab (should open in "beside" mode)
  b) side-panel -> new-window
  c) editor-tab -> new-window
  d) new->window -> editor-tab (will open in the first editor group)
  e) new-> window -> new-window (will open a "newer" window)
  d) editor-tab -> editor-tab (when the IDE is in the first group, it will move to "beside" group, otherwise it will move to the first group)
3. Make sure changing panelLocation in settings is reflected
4. Restart VSCode with the panel open in all three states. Make sure the window is in the same location after restart
5. Test new-window mode on Cursor where no compact mode command fails
6. Make sure "Show IDE Panel" command focuses the panel (i.e. when it is placed in a hidden side panel)

https://github.com/user-attachments/assets/6c759f8d-c6dd-40f3-a764-afd6cdbc819c

### DIFF
--- a/packages/vscode-extension/src/common/utils.ts
+++ b/packages/vscode-extension/src/common/utils.ts
@@ -9,6 +9,8 @@ export interface UtilsEventMap {
   telemetryEnabledChanged: boolean;
 }
 
+export type IDEPanelMoveTarget = "new-window" | "editor-tab" | "side-panel";
+
 export interface UtilsInterface {
   getCommandsCurrentKeyBinding(commandName: string): Promise<string | undefined>;
 
@@ -18,7 +20,7 @@ export interface UtilsInterface {
 
   saveMultimedia(multimediaData: MultimediaData): Promise<boolean>;
 
-  movePanelTo(location: "new-window" | "editor-tab" | "side-panel"): Promise<void>;
+  movePanelTo(location: IDEPanelMoveTarget): Promise<void>;
 
   showDismissableError(errorMessage: string): Promise<void>;
 
--- a/packages/vscode-extension/src/common/utils.ts
+++ b/packages/vscode-extension/src/common/utils.ts
@@ -18,7 +18,7 @@ export interface UtilsInterface {
 
   saveMultimedia(multimediaData: MultimediaData): Promise<boolean>;
 
-  movePanelToNewWindow(): Promise<void>;
+  movePanelTo(location: "new-window" | "editor-tab" | "side-panel"): Promise<void>;
 
   showDismissableError(errorMessage: string): Promise<void>;
 
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -33,6 +33,7 @@ import { registerChat } from "./chat";
 import { ProxyDebugSessionAdapterDescriptorFactory } from "./debugging/ProxyDebugAdapter";
 import { Connector } from "./connect/Connector";
 import { ReactDevtoolsEditorProvider } from "./react-devtools-profiler/ReactDevtoolsEditorProvider";
+import { IDEPanelMoveTarget } from "./common/utils";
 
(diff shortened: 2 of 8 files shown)


packages/vscode-extension/src/common/utils.ts
```diff
@@ -18,7 +18,7 @@ export interface UtilsInterface {
 
   saveMultimedia(multimediaData: MultimediaData): Promise<boolean>;
 
-  movePanelToNewWindow(): Promise<void>;
+  movePanelTo(location: "new-window" | "editor-tab" | "side-panel"): Promise<void>;
```
jwajgelt, line 21: Since this union type is used in multiple places, I'd prefer
```suggestion
  movePanelTo(location: IDEPanelLocation): Promise<void>;
```
With a `type IDEPanelLocation = "new-window" | "editor-tab" | "side-panel"` alias somewhere else, or changing the existing `PanelLocation` union to include the new `"new-window"` value and using that.

packages/vscode-extension/src/extension.ts
```diff
@@ -85,17 +87,37 @@ export async function activate(context: ExtensionContext) {
 
   commands.executeCommand("setContext", "RNIDE.sidePanelIsClosed", false);
 
-  async function showIDEPanel() {
+  let updatingConfigProgrammatically = false;
+  async function showIDEPanel(newLocation?: "editor-tab" | "new-window" | "side-panel") {
```
jwajgelt, line 91: As suggested above,
```suggestion
  async function showIDEPanel(newLocation?: IDEPanelLocation) {
```

packages/vscode-extension/src/utilities/utils.ts
```diff
@@ -123,8 +123,8 @@ export class Utils implements UtilsInterface {
     return true;
   }
 
-  public async movePanelToNewWindow() {
-    commands.executeCommand("workbench.action.moveEditorToNewWindow");
+  public async movePanelTo(location: "new-window" | "editor-tab" | "side-panel") {
```
jwajgelt, line 126: Similarly here.
```suggestion
  public async movePanelTo(location: IDEPanelLocation) {
```

jwajgelt (approved):
Just one nitpick, other than that it looks ok.

PR #1148, opened 2025-05-14
Kmagiera/unify skip files

kmagiera:
This PR unifies the skip files logic we had in two places. It also adapts the skip files concept to the element inspector that was previously using a custom setting `RadonIDE.inspectorExcludePattern`.

1. We extract the skip files logic to separate class and use it in the two places where we had it before
2. We deprecte `RadonIDE.inspectorExcludePattern` option 
3. We add a new launch configuration option `skipFiles` that will append to the set of default skip files (node modules, etc)

### How Has This Been Tested: 
1. Run project, make sure the debugger works
2. Use right-click inspector to see that the built-in RN items doesn't show up by default.

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -413,6 +413,10 @@
                   }
                 }
               },
+              "skipFiles": {
+                "type": "array",
+                "description": "Files matching this pattern will be filtered out or hidden from debugger stack traces / exception breakpoints in debugger or when using inspector. The patterns support negation (when '!' is place at the start of the pattern) Radon IDE has a default set of patterns and this option will add to that list. If you want to force the default patterns to be ignored, add '!**/*' as the first pattern to your list."
+              },
               "eas": {
                 "type": "object",
                 "description": "Configuration for EAS build service",
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -194,6 +194,7 @@
           "type": "string",
           "scope": "window",
           "default": null,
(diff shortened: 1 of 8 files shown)


filip131311 (changes requested):
I don't like the direction this PR is heading, I think that workspace/user settings is a correct place for such a setting, in the spirit of vscode guidelines. I think there are users that familiar with how vscode works and we should have settings that they can intuit where to find in vscode interface. If the concern is duplication we can move skipFiles to settings. 

The `.vscode` directory contains a file with workspace settings the same as it does `launch.json` so the setting should be easily committable to share with the resto of developers.

jwajgelt:
> I don't like the direction this PR is heading, I think that workspace/user settings is a correct place for such a setting, in the spirit of vscode guidelines. I think there are users that familiar with how vscode works and we should have settings that they can intuit where to find in vscode interface. If the concern is duplication we can move skipFiles to settings.
> 
> The `.vscode` directory contains a file with workspace settings the same as it does `launch.json` so the setting should be easily committable to share with the resto of developers.

I agree with this point.
Having it in settings/preferences makes it more accessible for users who will look there first for configuration.
Having it in the Launch Config makes it so there's no real way for the user to discover this setting, they have to either already know it's there, or go to the docs and read about it there (which many people simply won't do)

kmagiera:
Settings are for things that depend on users' preferences like for example the position of the IDE panel. 

The skip files setting is specific to a given project and unlike the position of the IDE, should be shared with other people working on the given codebase. Moreover, we may want to allow multiple launch configurations for example in case of monorepos with many apps or whitelabel projects. For me this definitely belongs in launch config and not in the settings.

We can make launch configuration features more easily accessible, but this is not a problem specific to skip files settings as we already have a number of project specific configuration options there like whether the project uses expo, or what scheme to use for building etc.

PR #1150, opened 2025-05-15
Fix occasional webview crash in UrlSelect

jwajgelt:
Due to mistyped parameters, sometimes `value` passed to the `UrlSelect` component is `undefined`, causing an error due to `.toLowerCase` being called on it.
This PR fixes the tyle of `urlSelectValue` state and checks for `undefined` where necessary

### How Has This Been Tested: 
- the issue was hard to reproduce reliably, but sometimes the IDE webview would crash due to method call on undefined that is impossible now


filip131311 (approved):
code looks good

PR #1153, opened 2025-05-16
Fix UrlSelect behavior when Expo Router is not used

HexHyperion:
This PR fixes bugs caused mainly by the application path bar being always enabled, even if Expo Router was not present:
- the input field was active, so the user could input a route and send it to a nonexistent handler, resulting in a hidden error and the app not navigating anywhere, but changing the displayed path;
- the bar did not have any routes to present, but the dropdown could be opened, resulting in broken layout;
- there was no `/` route on the list, so users couldn't intuitively navigate to the homepage with the bar;
- paths not starting with `"/"` were displayed with id instead of name, resulting in unreadable preview paths.

### Fixes:
If Expo Router is not used:
- disable the UrlSelect if the path history doesn't exist;
- make the bar's field read-only;
- create a simpler route list without the search result section;
- hard-code a `/` route that uses `project.goHome()` on selection into the dropdown;
- make the input field lose focus after clicking outside the dropdown;
- set a fail-safe `"/"` path bar placeholder instead of `""`

### How Has This Been Tested: 
- open a React Native project without Expo Router - the path bar will stay disabled until the first navigation change
- after navigating, i.e. to a preview, click the UrlSelect -  the input field is not editable, but text can be highlighted, copied, etc.
- in the dropdown, there is only one section with `/` route first, and then the other visited ones sorted from newest
- open a component preview - now both on bare RN and on ER projects the component name is displayed instead of the file path

### DIFF
--- a/packages/vscode-extension/src/webview/components/UrlSelect.tsx
+++ b/packages/vscode-extension/src/webview/components/UrlSelect.tsx
@@ -3,6 +3,7 @@ import * as Popover from "@radix-ui/react-popover";
 import { VscodeTextfield } from "@vscode-elements/react-elements";
 import { partition } from "lodash";
 import "./UrlSelect.css";
+import { useProject } from "../providers/ProjectProvider";
 
 export type UrlItem = { id: string; name: string };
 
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -234,31 +262,70 @@ function UrlSelect({ onValueChange, recentItems, items, value, disabled }: UrlSe
...
+                            onNavigate={navigateBetweenItems}
+                            getNameFromId={getNameFromId}
+                            textfieldRef={textfieldRef as React.RefObject<HTMLInputElement>}
+                          />
+                        )
+                    )}
+                  </div>
                 )}
-              </div>
+              </>
```
jwajgelt, line 328: NIT:
You have a bunch of code that looks the same here. I'd extract this into a separate component which takes a list of `item`s and renders the `<div>items.map(...)</div>`

jwajgelt (approved):
Left some smaller comments, but overall the code looks very good.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -166,8 +172,15 @@ function UrlSelect({ onValueChange, recentItems, items, value, disabled }: UrlSe
...
             disabled={disabled}
+            readonly={dropdownOnly}
             onInput={() => setInputValue(textfieldRef.current?.value ?? "")}
-            onMouseDown={() => setIsDropdownOpen(true)}
+            onMouseDown={() => {
+              if (!dropdownOnly || !isDropdownOpen) {
+                setIsDropdownOpen(true);
+              } else if (dropdownOnly) {
+                setIsDropdownOpen(false);
+              }
```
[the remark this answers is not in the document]
HexHyperion: Right, onFocus is much better, changed it now.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
HexHyperion: Sure, but maybe I'll bring the route-indexing branch up to date first, because I made some bigger changes to the structure there.

PR #1154, opened 2025-05-16
Refactor project and device state to fix issues with device switch and simplify our logic

kmagiera:
This PR refactors a bunch of places in order to straighten up the logic and data flow and to make the project state management more robust and self-contained.

Key objectives:
1) Switching between devices sometimes causes an undesirable state coming from unselected/old device to leak into project settings. As this is undesirable, we update `DeviceSessionManager` to guard against this. Now, all the project state updates are routed via `DeviceSessionManager` which verifies that the device session from which the update comes is the one that is currently selected.
2) Our state is currently distributed between the extension process and webview. Most of the "project state" actually belongs to a particular device session. When we switch between sessions, we don't want to manually keep track of the state updates on the webview site. This was previously happening. When switching we'd need to reset navigation state. But this also applies to other state elements, like whether we're currently profiling or if there's a stale build. Previously, we'd use events to pass that information and then store state in webview. Now, we're moving most of that into the "project state" and update it along all the other variables. This includes things like navigation history which is now kept in the state too.

The biggest change in the PR that affects the biggest number of lines is that we're moving most of the "project state" out of the project class and into device session class. This includes a number of methods and logic responsible for monitoring state updates and responding to them.

This PR also aims to fix some issues and design flaws in project / device-session-manager / device session classes. Below, we're listing notable changes:

1. device settings management moved from device session into device base. Now, device base class is responsible for storing the updates for device settings and also managing replays and show touches settings.
2. we're restructuring the startup flow for the device session class. Before it'd control all the aspects of starting the session. Since we recently added "activate" / "deactivate" methods that aim to control the session "visibility" when switching, we move control over the debug session there. We also branch out from the startup flow to make sure that debugger is only started when in active state.

### How Has This Been Tested: 
1. Run app (test on several versions)
3. Switch devices during boot
4. Test restart options
5. Switch during restart
6. After switching verify there are no multiple debug sessions in the debug console dropdown
7. Test Url bar on expo-router project
8. Verify changing device settings works and persist across device switching
9. Test replays and "show toucches" settings in particular

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -33,7 +33,12 @@ export type BuildErrorDescriptor = {
   buildType: BuildType | null;
 };
 
-export type ProfilingState = "stopped" | "running" | "saving";
+export type ProfilingState = "stopped" | "profiling" | "saving";
+
+export type NavigationHistoryItem = {
+  displayName: string;
+  id: string;
+};
 
 export type DeviceSessionState = {
   status:
@@ -52,10 +57,10 @@ export type DeviceSessionState = {
   fastRefreshOngoing: boolean;
   profilingReactState: ProfilingState;
   profilingCPUState: ProfilingState;
-  navigationHistory: { displayName: string; id: string }[];
-  toolsState: ToolsState | undefined;
+  navigationHistory: NavigationHistoryItem[];
+  toolsState: ToolsState;
   isDebuggerPaused: boolean;
-  logCount: number;
+  logCounter: number;
   hasStaleBuildCache: boolean;
 };
 
@@ -70,9 +75,9 @@ export const DeviceSessionInitialState: DeviceSessionState = {
   profilingReactState: "stopped",
   profilingCPUState: "stopped",
   navigationHistory: [],
-  toolsState: undefined,
+  toolsState: {},
   isDebuggerPaused: false,
-  logCount: 0,
+  logCounter: 0,
   hasStaleBuildCache: false,
 };
 
@@ -153,13 +158,9 @@ export enum ActivateDeviceResult {
 }
 
 export interface ProjectEventMap {
-  log: { type: string };
   projectStateChanged: ProjectState;
   deviceSettingsChanged: DeviceSettings;
-  toolsStateChanged: ToolsState;
   licenseActivationChanged: boolean;
-  navigationChanged: { displayName: string; id: string };
-  needsNativeRebuild: void;
   replayDataCreated: MultimediaData;
   isRecording: boolean;
 }
@@ -193,6 +194,7 @@ export interface ProjectInterface {
   focusExtensionLogsOutput(): Promise<void>;
   focusDebugConsole(): Promise<void>;
   openNavigation(navigationItemID: string): Promise<void>;
+  navigateBack(): Promise<void>;
   openDevMenu(): Promise<void>;
 
   activateLicense(activationKey: string): Promise<ActivateDeviceResult>;
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -56,7 +56,6 @@ export type DeviceSessionState = {
   buildError: BuildErrorDescriptor | undefined;
   selectedDevice: DeviceInfo | undefined;
   previewURL: string | undefined;
-  fastRefreshOngoing: boolean;
   profilingReactState: ProfilingState;
   profilingCPUState: ProfilingState;
   navigationHistory: NavigationHistoryItem[];
@@ -73,7 +72,6 @@ export const DeviceSessionInitialState: DeviceSessionState = {
   buildError: undefined,
   selectedDevice: undefined,
   previewURL: undefined,
-  fastRefreshOngoing: false,
   profilingReactState: "stopped",
   profilingCPUState: "stopped",
   navigationHistory: [],
@@ -179,7 +177,6 @@ export type MultimediaData = {
 
 export interface ProjectInterface {
   getProjectState(): Promise<ProjectState>;
-  goHome(homeUrl: string): Promise<void>;
   renameDevice(deviceInfo: DeviceInfo, newDisplayName: string): Promise<void>;
   updatePreviewZoomLevel(zoom: ZoomLevelType): Promise<void>;
 
@@ -197,6 +194,7 @@ export interface ProjectInterface {
   focusDebugConsole(): Promise<void>;
   openNavigation(navigationItemID: string): Promise<void>;
   navigateBack(): Promise<void>;
+  navigateHome(): Promise<void>;
   openDevMenu(): Promise<void>;
 
   activateLicense(activationKey: string): Promise<ActivateDeviceResult>;
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -36,6 +36,17 @@ export type BuildErrorDescriptor = {
 export type ProfilingState = "stopped" | "running" | "saving";
 
 export type DeviceSessionState = {
+  status:
+    | "starting"
+    | "running"
+    | "bootError"
+    | "bundlingError"
+    | "debuggerPaused"
+    | "refreshing"
+    | "buildError";
+  startupMessage: StartupMessage | undefined;
+  stageProgress: number | undefined;
+  buildError: BuildErrorDescriptor | undefined;
   selectedDevice: DeviceInfo | undefined;
   previewURL: string | undefined;
   fastRefreshOngoing: boolean;
@@ -46,24 +57,24 @@ export type DeviceSessionState = {
   isDebuggerPaused: boolean;
   logCount: number;
   hasStaleBuildCache: boolean;
-} & (
-  | {
-      status:
-        | "starting"
-        | "running"
-        | "bootError"
-        | "bundlingError"
-        | "debuggerPaused"
-        | "refreshing";
-      startupMessage: StartupMessage | undefined;
-      stageProgress: number | undefined;
-    }
-  | {
-      status: "buildError";
-      selectedDevice: DeviceInfo | undefined;
-      buildError: BuildErrorDescriptor;
-    }
-);
+};
+
+export const DeviceSessionInitialState: DeviceSessionState = {
+  status: "starting",
+  startupMessage: undefined,
+  stageProgress: undefined,
+  buildError: undefined,
+  selectedDevice: undefined,
+  previewURL: undefined,
+  fastRefreshOngoing: false,
+  profilingReactState: "stopped",
+  profilingCPUState: "stopped",
+  navigationHistory: [],
+  toolsState: undefined,
+  isDebuggerPaused: false,
+  logCount: 0,
+  hasStaleBuildCache: false,
+};
 
 export type ProjectState = {
   initialized: boolean;
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -40,15 +40,17 @@ export type NavigationHistoryItem = {
   id: string;
 };
 
+export type DeviceSessionStatus =
+  | "starting"
+  | "running"
+  | "bootError"
+  | "bundlingError"
+  | "debuggerPaused"
+  | "refreshing"
+  | "buildError";
+
 export type DeviceSessionState = {
-  status:
-    | "starting"
-    | "running"
-    | "bootError"
-    | "bundlingError"
-    | "debuggerPaused"
-    | "refreshing"
-    | "buildError";
+  status: DeviceSessionStatus;
   startupMessage: StartupMessage | undefined;
   stageProgress: number | undefined;
   buildError: BuildErrorDescriptor | undefined;
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -27,8 +27,27 @@ export type ToolsState = {
   [key: string]: ToolState;
 };
 
-export type ProjectState =
-  | ({
+export type BuildErrorDescriptor = {
+  message: string;
+  platform: DevicePlatform;
+  buildType: BuildType | null;
+};
+
+export type ProfilingState = "stopped" | "running" | "saving";
+
+export type DeviceSessionState = {
+  selectedDevice: DeviceInfo | undefined;
+  previewURL: string | undefined;
+  fastRefreshOngoing: boolean;
+  profilingReactState: ProfilingState;
+  profilingCPUState: ProfilingState;
+  navigationHistory: { displayName: string; id: string }[];
+  toolsState: ToolsState | undefined;
+  isDebuggerPaused: boolean;
+  logCount: number;
+  hasStaleBuildCache: boolean;
+} & (
+  | {
       status:
         | "starting"
         | "running"
@@ -36,26 +55,20 @@ export type ProjectState =
         | "bundlingError"
         | "debuggerPaused"
         | "refreshing";
-    } & ProjectStateCommon)
-  | ProjectStateBuildError;
+      startupMessage: StartupMessage | undefined;
+      stageProgress: number | undefined;
+    }
+  | {
+      status: "buildError";
+      selectedDevice: DeviceInfo | undefined;
+      buildError: BuildErrorDescriptor;
+    }
+);
 
-type ProjectStateCommon = {
-  previewURL: string | undefined;
-  selectedDevice: DeviceInfo | undefined;
+export type ProjectState = {
   initialized: boolean;
   previewZoom: ZoomLevelType | undefined; // Preview specific. Consider extracting to different location if we store more preview state
-  startupMessage: StartupMessage | undefined;
-  stageProgress: number | undefined;
-};
-
-type ProjectStateBuildError = {
-  status: "buildError";
-  buildError: {
-    message: string;
-    platform: DevicePlatform;
-    buildType: BuildType | null;
-  };
-} & ProjectStateCommon;
+} & DeviceSessionState;
 
 export type ZoomLevelType = number | "Fit";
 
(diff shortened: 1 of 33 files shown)


packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -1,298 +1,163 @@
...
+import { DeviceSessionInitialState } from "../common/Project";
 
 const LAST_SELECTED_DEVICE_KEY = "last_selected_device";
 
-export class DeviceSessionsManager implements DeviceSessionsManagerInterface, Disposable {
-  // selected device id
-  private selectedDevice: string | undefined;
-  private deviceSessions: Map<string, DeviceSession> = new Map();
+export class DeviceSessionsManager implements Disposable, DeviceSessionsManagerInterface {
+  private deviceSessions: Set<DeviceSession> = new Set();
```
jwajgelt, line 22: I'd change this to a map `DeviceId` -> `DeviceSession`, since we probably only want to either
- lookup by device id
- walk over all sessions
But that's not very important when there are probably < 3 sessions at any time.

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -1,298 +1,163 @@
...
-      this.deviceSessionManagerDelegate.onDeviceSelected(
-        deviceInfo,
-        this.selectedDeviceSession?.previewURL
-      );
-      return true;
-    }
-
-    const device = await this.selectDeviceOnly(deviceInfo);
+    // if there's an existing session for the device, we use it instead of starting a new one
+    const existingDeviceSession = this.deviceSessions
```
filip131311, line 56: if deviceSessions was a map it would be much cleaner and i don't see a reason for it not to be

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -1,298 +1,163 @@
...
+      .values()
+      .find((session) => session.deviceInfo.id === deviceInfo.id);
+    if (existingDeviceSession) {
+      this.updateSelectedSession(existingDeviceSession);
+      const otherSessions = this.deviceSessions
+        .values()
+        .filter((session) => session !== existingDeviceSession);
+      await Promise.all(otherSessions.map((session) => this.terminateSession(session)));
```
filip131311, line 64: this seems like something that should be conditioned by `killPreviousDeviceSession`

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -1,298 +1,163 @@
...
-      }
-    }
-
-    this.updateProjectState({
-      selectedDevice: deviceInfo,
-      initialized: true,
-      status: "starting",
-      startupMessage: StartupMessage.InitializingDevice,
-      previewURL: undefined,
+    const newDeviceSession = new DeviceSession(this.applicationContext, deviceInfo, device, {
```
filip131311, line 75: `deviceInfo` seems to be a new constructor argument, Why is it needed? doesn't `device` already hold all the important information?

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -1,298 +1,163 @@
...
-        this.deviceSessionManagerDelegate
-      );
-      this.deviceSessions.set(newDeviceId, newDeviceSession);
-      this.selectedDevice = newDeviceId;
+    const previousSessions = Array.from(this.deviceSessions.values());
+    this.deviceSessions.add(newDeviceSession);
+    this.updateSelectedSession(newDeviceSession);
+
+    if (killPreviousDeviceSession) {
+      await Promise.all(previousSessions.map((session) => this.terminateSession(session)));
```
filip131311, line 89: I have two remarks here:
1) I don't think this is what `killPreviousDeviceSession` option should mean, If wee kill all previous sessions we should rename it to `killPreviousDeviceSessions` or even better for clarity `killAllPreviousDeviceSessions`, but I don't think an option like that would ever be necessary. 
2) `await Promise.all(previousSessions.map((session) => this.terminateSession(session)));` is now used twice and i belive something similar will be useful for dispose method so it might be better to make it into a separate utility.

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -1,298 +1,163 @@
...
-    if (anyActiveDeviceSessionId) {
-      const selectedActiveSession = await this.trySelectingActiveDeviceSession(
-        anyActiveDeviceSessionId,
-        true
-      );
-      if (selectedActiveSession) {
-        return true;
-      }
+  private findInitialDeviceAndStartSession = async () => {
+    if (this.selectedDeviceSession && !this.findingDevice) {
```
filip131311, line 100: the name of this method does not indicate this condition (i left a comment above) i would create a second method like this: 
```js
private handleDevicesChanged(){
  if (this.selectedDeviceSession && !this.findingDevice) {
    return 
  }
  this. findInitialDeviceAndStartSession();
}

```

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -1,298 +1,163 @@
...
 
   // used in callbacks, needs to be an arrow function
   private removeDeviceListener = async (device: DeviceInfo) => {
-    if (this.selectedDevice === device.id) {
-      this.updateProjectState({ status: "starting" });
-      await this.killAndRemoveDevice(device.id);
-      await this.trySelectingDevice();
-    }
+    // if the deleted device was running an active session, we need to terminate that session
+    this.deviceSessions.forEach((session) => {
```
filip131311, line 133: this method coupled with `findInitialDeviceAndStartSession` implicitly assumes that "deviceRemoved" event happens before "deviceChanged" event, I think this is a bit dangerous as someone implementing the `deviceManager` would have to know that. I would prefer to the old way of removeDeviceListener triggering `trySelectingDevice` (or current alternative) to avoid this problem.

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -482,50 +651,85 @@ export class DeviceSession implements Disposable {
...
+          platform: this.device.platform,
+        };
+      }
+      throw e;
+    } finally {
+      this.emitStateChange();
+    }
+  }
+
+  private emitStateChange() {
```
filip131311, line 731: I don't like this approach as it requires two lines of code every time we update `sessionState` and can lead to mistakes made by not calling it. I would prefer one object `SessionState` that looks something like: 
```
 status: this.status,
      startupMessage: this.startupMessage,
      stageProgress: this.stageProgress,
      buildError: this.buildError,
      profilingCPUState: this.profilingCPUState,
      profilingReactState: this.profilingReactState,
      navigationHistory: this.navigationHistory,
      selectedDevice: this.deviceInfo,
      previewURL: this.previewURL,
      toolsState: this.toolsManager.getToolsState(),
      isDebuggerPaused: this.isDebuggerPaused,
      logCounter: this.logCounter,
      hasStaleBuildCache: this.hasStaleBuildCache,
```

and then an utility `updateSessionState` that updates the value and calls `onStateChange`

packages/vscode-extension/src/project/project.ts
```diff
@@ -466,14 +343,14 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
   }
 
   private async reloadMetro() {
-    if (await this.deviceSession?.perform("reloadJs")) {
+    if (await this.deviceSession?.performReloadAction("reloadJs")) {
       this.updateProjectState({ status: "running" });
```
filip131311, line 347: this seems to be left out from previous implementation, as now status should be updated by `deviceSession`

filip131311: btw why not `this.deviceSessionsManager.reloadCurrentSession`?

packages/vscode-extension/src/project/project.ts
```diff
@@ -591,13 +455,18 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
   }
 
   public async focusDebugConsole() {
+    this.deviceSession?.resetLogCounter();
```
filip131311, line 458: I think we should also call `resetLogCounter` before selecting a new device, so when the user comes back is not welcomed with the old counter (might not be necessary or in scope of this PR)

packages/vscode-extension/src/project/project.ts
```diff
@@ -691,34 +556,12 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
-  }
-
   private updateProjectState(newState: Partial<ProjectState>) {
-    // NOTE: this is unsafe, but I'm not sure there's a way to enforce the type of `newState` correctly
-    const mergedState: any = { ...this.projectState, ...newState };
-    // stageProgress is tied to a startup stage, so when there is a change of status or startupMessage,
-    // we always want to reset the progress.
-    if (
-      newState.status !== undefined ||
-      ("startupMessage" in newState && newState.startupMessage !== undefined)
```
filip131311, line 708: is there a similar logic in the new implementation? Or is it not necessary anymore?

filip131311 (commented):
Left inline comments

packages/vscode-extension/src/common/Project.ts
```diff
@@ -129,18 +158,11 @@ export enum ActivateDeviceResult {
...
 export interface ProjectEventMap {
-  log: { type: string };
   projectStateChanged: ProjectState;
   deviceSettingsChanged: DeviceSettings;
-  toolsStateChanged: ToolsState;
   licenseActivationChanged: boolean;
-  navigationChanged: { displayName: string; id: string };
-  needsNativeRebuild: void;
   replayDataCreated: MultimediaData;
   isRecording: boolean;
```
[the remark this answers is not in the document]
kmagiera: yes

packages/vscode-extension/src/project/DeviceSessionsManager.ts
kmagiera: no, it is not coupled, the logic here is very simple:
1) when device list changes we try to select initial device. If the device was already selected we do nothing. This is why the findInitialDevice... method is hooked to that event. I added comment in code to explain it. I think the name is self explanatory as it has "initial" in it which suggests we only do it if it isn't already initialized.
2) this code here is not coupled, it only guards agains the current device session being selected after the device is removed. There's no expectations on the order of the events

packages/vscode-extension/src/project/deviceSession.ts
kmagiera: It will be replaced once we add a state management solution so I don't think it makes sense to invest more in it. The state here collects some data from other objects like the toolsManager + we also sometimes want to batch updates meaning we want to update a couple of state items and only then trigger update. I'd rather rely on whichever mechanism the state management library provide here rather than building my own here.

packages/vscode-extension/src/project/DeviceSessionsManager.ts
kmagiera: I followed the other sugestion and added separate method for handling device change events

packages/vscode-extension/src/project/project.ts
kmagiera: this change was just because I was refactoring the method name. But project is already quite messy, sometimes we call deviceSession methods and sometimes we call methods on the manager. They actually have slightly different behavior, for example the device session one would throw if there's no device selected. I'd rather not change this as I don't understand what could break

packages/vscode-extension/src/project/project.ts
filip131311: 👍

packages/vscode-extension/src/project/project.ts
filip131311: 👍

packages/vscode-extension/src/common/Project.ts
```diff
@@ -27,35 +27,66 @@ export type ToolsState = {
...
+  selectedDevice: DeviceInfo | undefined;
+  previewURL: string | undefined;
+  profilingReactState: ProfilingState;
+  profilingCPUState: ProfilingState;
+  navigationHistory: NavigationHistoryItem[];
+  toolsState: ToolsState;
+  isDebuggerPaused: boolean;
+  logCounter: number;
+  hasStaleBuildCache: boolean;
+  isRecordingScreen: boolean;
```
filip131311, line 66: why replay data is kept in a separate communication flow?

packages/vscode-extension/src/project/DeviceSessionsManager.ts
filip131311: I don't think I quite understand: 

When we changeDevices (when the device is added or removed but well be only interested in removing)  we trigger: 
```js
private devicesChangedListener = async () => {
    // this method is triggered when new devices are added, we don't want to
    // run the device selection process, if an existing session is already running
    // or if we're already in the process of finding a device (either because of a
    // previous event or becuase we only just booted up the manager).
    if (this.selectedDeviceSession && !this.findingDevice) {
      return;
    }
    this.findInitialDeviceAndStartSession();
  };
```

this method will do noting if `this.selectedDeviceSession` is anything else then `undefind`, additionally when `deviceRemoved` is triggered we run: 

```js
private removeDeviceListener = async (device: DeviceInfo) => {
    // if the deleted device was running an active session, we need to terminate that session
    await this.terminateSession(device.id);
  };
```

the `terminateSession` method is setting `this.selectedDeviceSession` to `undefined` which allows `devicesChangedListener` to pass `if (this.selectedDeviceSession && !this.findingDevice) ` guard and select a new device, but this whole mechanism assumes that `deviceRemoved` is happening before `devicesChanged` this is in fakt true because `await Promise.all([emulators, simulators]);` is allowing `deviceRemoved` to be emitted before `devicesChanged` but this is an assumption that the maintainer of `deviceManager` shouldn't have to know.

packages/vscode-extension/src/common/Project.ts
kmagiera: Replays are not part of the state as capturing a replay is just an asynchronous function. Whether replays are enabled or not is kept in the settings because that's where it is placed in the UI and settings are shared for all the devices atm

packages/vscode-extension/src/project/project.ts
kmagiera: ok, I see now. Makes sense, will add it here

kmagiera: added resetting log counter in deactivate right before we disconnect the JS debugger as that's where the logs are being lost

packages/vscode-extension/src/common/Project.ts
```diff
@@ -27,35 +27,66 @@ export type ToolsState = {
...
 
-type ProjectStateBuildError = {
-  status: "buildError";
-  buildError: {
-    message: string;
-    platform: DevicePlatform;
-    buildType: BuildType | null;
-  };
-} & ProjectStateCommon;
+export const DeviceSessionInitialState: DeviceSessionState = {
```
jwajgelt, line 69: NIT: I'd prefer we'd reserve `PascalCase` for type names
```suggestion
export const DEVICE_SESSION_INITIAL_STATE: DeviceSessionState = {
```

jwajgelt (changes requested):

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -1,298 +1,171 @@
...
     private readonly applicationContext: ApplicationContext,
     private readonly deviceManager: DeviceManager,
-    private readonly deviceSessionManagerDelegate: DeviceSessionsManagerDelegate,
-    private readonly updateProjectState: (newState: Partial<ProjectState>) => void
+    private readonly deviceSessionManagerDelegate: DeviceSessionsManagerDelegate
   ) {
-    this.trySelectingDevice();
+    this.findInitialDeviceAndStartSession();
     this.deviceManager.addListener("deviceRemoved", this.removeDeviceListener);
+    this.deviceManager.addListener("devicesChanged", this.devicesChangedListener);
```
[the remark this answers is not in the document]
kmagiera: you're right but this is how it was before and I didn't want to change more things which would result in the PR getting even bigger. The only thing is that we previously would only register for the change event in `selectInitialDevice` and IMO it is cleaner if we just do it in one place especially that we also remove the listeners in dispose.

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -1,298 +1,171 @@
...
-        this.selectedDeviceSession?.previewURL
-      );
-      return true;
+    // if there's an existing session for the device, we use it instead of starting a new one
+    const existingDeviceSession = this.deviceSessions.get(deviceInfo.id);
+    if (existingDeviceSession) {
+      this.updateSelectedSession(existingDeviceSession);
+      if (killPreviousDeviceSession) {
+        await this.terminatePreviousSessions();
+      }
```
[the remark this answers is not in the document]
kmagiera: This code is actually not used at the moment. We never request an existing session to be selected. I changed it here such that this method works consistently regardless of whether the device is running or not. Before we'd ignore the `preservePreviousDevice` when the selected device was already running and only consider it when it was not. Now, we handle it consistently.

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -1,298 +1,171 @@
...
+  };
 
   // used in callbacks, needs to be an arrow function
   private removeDeviceListener = async (device: DeviceInfo) => {
-    if (this.selectedDevice === device.id) {
-      this.updateProjectState({ status: "starting" });
-      await this.killAndRemoveDevice(device.id);
-      await this.trySelectingDevice();
+    // if the deleted device was running an active session, we need to terminate that session
+    await this.terminateSession(device.id);
```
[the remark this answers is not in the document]
kmagiera: you're right. This is how it worked before. Will update

jwajgelt (approved):

PR #1158, opened 2025-05-19
Safeguard agains debug sessions started in parallel

kmagiera:
This PR extracts some code changes from #1154 that can be landed separately and improves the codebase regardless.

We've recently struggled with an error that was caused because of the session management is very complex right now. As a consequence, we'd end up getting a multiple debug sessions opened and sticking around in the debug console.

This PR updates the DebugSession code, such that we always guard against the sessions being started in parallel. What would happen before, is that from two different places, we'd request the debug session to start. We'd typically store the debug session in a member variable, that the assignment is only done after await. So, when another call happens before the session is assigned to variable, we don't detect that and the newly created session would override that member variable resulting in one of the debug session leaking.

What we do here is simply store the awaited session in a local variable and check if the session was set in  the meantime after awaiting again. If that happens we print a warning, as it could be helpful to track down the root cause of this behavior regardless.

### How Has This Been Tested: 
1. Test whether the app loads properly and that we get metro errors logged on the startup time.

### DIFF
--- a/packages/vscode-extension/src/debugging/DebugSession.ts
+++ b/packages/vscode-extension/src/debugging/DebugSession.ts
@@ -5,6 +5,7 @@ import { disposeAll } from "../utilities/disposables";
 import { sleep } from "../utilities/retry";
 import { startDebugging } from "./startDebugging";
 import { extensionContext } from "../utilities/extensionContext";
+import { Logger } from "../Logger";
 
 const PING_TIMEOUT = 1000;
 
@@ -35,16 +36,22 @@ export interface JSDebugConfiguration {
 
 export type DebugSource = { filename?: string; line1based?: number; column0based?: number };
 
+type DebugSessionOptions = {
+  useParentDebugSession?: boolean;
+};
+
 export class DebugSession implements Disposable {
   private parentDebugSession: vscode.DebugSession | undefined;
   private jsDebugSession: vscode.DebugSession | undefined;
 
-  private useParentDebugSession = false;
   private disposables: Disposable[] = [];
 
   private currentWsTarget: string | undefined;
 
-  constructor(private delegate: DebugSessionDelegate) {
+  constructor(
+    private delegate: DebugSessionDelegate,
+    private options: DebugSessionOptions = {}
+  ) {
     this.disposables.push(
       debug.onDidTerminateDebugSession((session) => {
         if (session.id === this.jsDebugSession?.id) {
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -135,13 +146,16 @@ export class DebugSession implements Disposable {
     if (this.jsDebugSession) {
       await this.restart();
     }
+    if (this.options.useParentDebugSession && !this.parentDebugSession) {
+      await this.startParentDebugSession();
+    }
```
jwajgelt, line 151: Can this `if` condition ever be satisfied? We've just started the `parentDebugSession` in `restart`

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -102,27 +108,32 @@ export class DebugSession implements Disposable {
         suppressSaveBeforeStart: true,
       }
     );
+    if (this.parentDebugSession) {
```
filip131311, line 111: nit : shouldn't this code evaluate before starting the new session and do we actually need the new constant instead of something like: 

```js 
 if (this.parentDebugSession) {
      Logger.warn("Parent debugger session has spawned concurrently, dropping the earlier session");
      debug.stopDebugging(this.parentDebugSession);
    }
  this.parentDebugSession =  await startDebugging(...;
```

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -169,7 +183,11 @@ export class DebugSession implements Disposable {
         compact: true,
       }
     );
-
+    if (this.jsDebugSession) {
```
filip131311, line 186: same question as above

filip131311 (commented):
left inline comments

packages/vscode-extension/src/debugging/DebugSession.ts
kmagiera: this is exactly the key of this PR. We need this after, because we use `await` there before the assignment. So if we for example call `restart` twice in a row before the first one resolves, it will not stop the first session as it wouldn't be assigned to the variable before the second call. I fixed one problem where this was happening already in #1138 and while doing my refactor I encountered it again, so decided to this code here here such that we avoid similar problems in the future.

filip131311 (approved):
LGTM

PR #1160, opened 2025-05-19
Extract route information from Expo Router and display it in UrlSelect

HexHyperion:
This PR further enhances the extension's integration with Expo Router by automatically extracting all indexed paths and displaying them in the existing URL suggestion panel. Now, the user doesn't need to navigate to a path for it to become listed, which significantly improves the usability of the navigation bar.

Paths visited recently, including previews and invalid URLs, are displayed above them for easy access. The indexed routes are sorted by depth, and then alphabetically, ensuring the paths closer to the app root will always stay on top.

As in big apps the list can get really long, the user can type into the bar to filter the routes and navigate to them either with mouse or keyboard. The suggested paths now highlight the query substring to improve readability.

The component identifies and highlights dynamic route segments, and if a route with one or more such segments is selected, the user will be able to fill them in the input field before navigating.

### How Has This Been Tested: 
- open a React Native project with Expo Router
- wait until the app loads and click on the URL field - the list now contains all URLs available in the app
- type something into the text field - paths will be suggested and highlighted based on the query
- if the app has dynamic routes, the parts in `[]` are colored, and the path is editable in the text field upon click - the first dynamic segment will be auto-selected, and the user can use (Shift+)Tab to jump between them
- try to make the bar content overflow - now the list becomes scrollable without breaking the layout

https://github.com/user-attachments/assets/283fee9a-56d3-4f12-9cf3-774a546c50d7


packages/vscode-extension/lib/expo_router_v2_plugin.js
```diff
@@ -20,6 +21,18 @@ function useRouterPluginMainHook({ onNavigationChange }) {
...
+    if (!store.routeNode) {
+      return;
+    }
+    const routeList = extractNestedRouteList(store.routeNode);
+    if (global.__REACT_DEVTOOLS_GLOBAL_HOOK__?.reactDevtoolsAgent?._bridge) {
+      global.__REACT_DEVTOOLS_GLOBAL_HOOK__.reactDevtoolsAgent._bridge.send("RNIDE_routeListRetrieved", {
+        routes: routeList,
+      });
+    }
+  }, [store.routeNode]);
```
kmagiera, line 34: it looks like we should extend the navigation plugin abstraction for this instead of acquiring the bridge directly here. Would it make more sense to add another method along `onNavigationChange` that we'd pass to the main navigation plugin hook and then use this method here

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -167,6 +169,9 @@ export class DeviceSession implements Disposable {
     devtools.onEvent("RNIDE_isProfilingReact", (isProfiling) => {
       this.deviceSessionDelegate.onAppEvent("isProfilingReact", isProfiling);
     });
+    devtools.onEvent("RNIDE_routeListRetrieved", (payload) => {
```
kmagiera, line 172: I'd collocate this with `RNIDE_navigationChanged` event from above because they are related. We could also rename this such that it starts with "navigation", for example `RNIDE_navigationRouteListUpdated`

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -63,6 +64,7 @@ export type AppEvent = {
   fastRefreshComplete: undefined;
   isProfilingReact: boolean;
   isSavingReactProfile: boolean;
+  routeListRetrieved: Route[];
```
kmagiera, line 67: "retrieved" is not the best term I think as from what I understand we may be getting updates later on if this list changes. Better call it "updated" as it'd make it easier to understand when it gets called

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -74,12 +30,16 @@ function UrlSelect({
...
   const [filteredItems, setFilteredItems] = React.useState<UrlItem[]>([]);
   const [filteredOutItems, setFilteredOutItems] = React.useState<UrlItem[]>([]);
-
   const [textfieldWidth, setTextfieldWidth] = React.useState<number>(0);
   const textfieldRef = useRef<HTMLInputElement>(null);
 
+  const routes = useRoutes();
+  const routeItems = useRoutesAsItems();
+
+  const itemRefs = useRef<Array<React.RefObject<HTMLDivElement>>>([]);
```
kmagiera, line 42: it's a single ref of many items so should be "itemsRef"

packages/vscode-extension/src/webview/providers/RoutesProvider.tsx
```diff
@@ -0,0 +1,55 @@
+import { createContext, useContext, useEffect, useState } from "react";
+import { useProject } from "./ProjectProvider";
+
+declare global {
+  interface Window {
+    RNIDE_lastRouteList?: Route[];
```
kmagiera, line 6: it is very weird to handle it this way, why?

packages/vscode-extension/src/webview/providers/RoutesProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export default function RoutesProvider({ children }: { children: React.ReactNode }) {
+  const [routes, setRoutes] = useState<Route[] | null>(null);
+  const { project } = useProject();
+
+  useEffect(() => {
+    if (window.RNIDE_lastRouteList) {
+      setRoutes(window.RNIDE_lastRouteList);
+    }
```
kmagiera, line 27: I think it'd be best to wait for changes from #1154 – there we save routes history in state and we should do the same with the available routes list avoiding this type of hacks

packages/vscode-extension/src/project/deviceSession.ts
HexHyperion: From what I figured, the only case where we'd need to update the list would be if Expo somehow hot-reloaded the app after a file was added/removed (which shouldn't happen, it should just fully reload), otherwise I _think_ there's no way for it to suddenly update in the middle of app's operation.

I put it in a separate `useEffect` only because sometimes the store wouldn't contain the list immediately after the app loaded, so with `[]` we got `undefined`, but okay, maybe that counts as updating...

packages/vscode-extension/src/webview/providers/RoutesProvider.tsx
HexHyperion: Right, it's a leftover from before I got the `RoutesProvider` to work, changed the listener now to just pass the routes to the provider as a prop (right now we need the routes only in the UI), but it's still kind of an ugly solution, and without it, it's possible to miss the event...

Probably would make sense to wait for the merge.

packages/vscode-extension/src/common/Project.ts
```diff
@@ -40,6 +40,14 @@ export type NavigationHistoryItem = {
   id: string;
 };
 
+export type Route = {
```
kmagiera, line 43: can we rename to `NavigationRoute`. This is a very central interface file that provides definitions for a lot of different functionalities. It would be better for the future if the names we export here are more specific and less generic

packages/vscode-extension/src/webview/components/UrlSelectItemGroup.tsx
```diff
@@ -0,0 +1,57 @@
...
+  onNavigate: (
+    e: React.KeyboardEvent,
+    prev?: UrlSelectFocusable,
+    next?: UrlSelectFocusable
+  ) => void;
+  getNameFromId: (id: string) => string;
+  noHighlight?: boolean;
+}
+
+const UrlSelectItemGroup: React.FC<UrlSelectItemGroupProps> = ({
```
kmagiera, line 22: we typically use `function()` syntax rather than arrow functions in the codebase.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -1,77 +1,49 @@
...
   const [filteredItems, setFilteredItems] = React.useState<NavigationHistoryItem[]>([]);
   const [filteredOutItems, setFilteredOutItems] = React.useState<NavigationHistoryItem[]>([]);
-
+  const [inputValue, setInputValue] = React.useState("/");
+  const [dynamicSegmentNames, setDynamicSegmentNames] = React.useState<string[]>([]);
+  const [currentDynamicSegment, setCurrentDynamicSegment] = React.useState<number>(0);
   const [textfieldWidth, setTextfieldWidth] = React.useState<number>(0);
   const textfieldRef = useRef<HTMLInputElement>(null);
 
+  const itemsRef = useRef<Array<React.RefObject<HTMLDivElement>>>([]);
```
kmagiera, line 38: It is very difficult to read thought the changes here. It feels like the code it more complicated than it needs to be. Specifically very unusual patterns like this one needs to be explained in comments and in the PR description.

I understand that some of this weirdness is because of the limitations of the components we are using. But if that's the case it requires better explanation nevertheless

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -80,60 +52,131 @@ function UrlSelect({ onValueChange, navigationHistory, disabled, dropdownOnly }:
...
+
+  // FIXME - if no Expo Router, this will be the previous path on reload!
   useEffect(() => {
     setInputValue(navigationHistory[0]?.displayName ?? "");
   }, [navigationHistory[0]?.id]);
 
+  // Update the itemsRef to ensure all items are focused correctly
+  useEffect(() => {
+    itemsRef.current = [
+      ...(dropdownOnly ? [{ id: "/", name: "/" }] : []),
```
kmagiera, line 141: the `itemsRef` type suggests we're storing refs to components but here we assign plain objects to `.current` – how's that typescript doesn't yell here?

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -148,6 +191,17 @@ function UrlSelect({ onValueChange, navigationHistory, disabled, dropdownOnly }:
     }
   }, []);
 
+  // Hacky way to change the cursor style of a readonly input,
+  // since the VscodeTextfield component doesn't provide any parts
+  // or props, and according to the authors, it's not going to.
+  useEffect(() => {
```
kmagiera, line 197: this may need to be a layout effect, no?

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -168,24 +222,32 @@ function UrlSelect({ onValueChange, navigationHistory, disabled, dropdownOnly }:
...
-                textfieldRef.current?.blur();
+                setTimeout(() => textfieldRef.current?.blur(), 0);
               }
               if (e.key === "ArrowDown") {
                 if (isDropdownOpen) {
                   navigateBetweenItems(
                     e,
                     undefined,
-                    document.querySelector(".url-select-item") as HTMLDivElement
+                    itemsRef.current[0]?.current as UrlSelectFocusable
```
kmagiera, line 250: we call `navigateBetweenItems` here but in this context word "navigate" means that we update the navigation state of the app. I understand that this method in that context doesn't actually "navigate". Can we change the name of it in that case as it is misleading.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -80,60 +52,131 @@ function UrlSelect({ onValueChange, navigationHistory, disabled, dropdownOnly }:
...
       textfieldRef.current?.focus();
     }
   };
 
+  // Props for UrlSelectItems and UrlSelectItemGroups to reduce code duplication
+  const commonItemProps = {
+    width: textfieldWidth,
+    itemsRef: itemsRef.current,
+    textfieldRef: textfieldRef as React.RefObject<HTMLInputElement>,
+    onNavigate: navigateBetweenItems,
```
kmagiera, line 129: is this really about navigation or are we mixing some concepts here, I'm confused

kmagiera (changes requested):
Most of the code looks good now. I'm a bit confused by the changes and their complexity in the UrlSelect and related code. Would be best to get clear explanation for all the unusual things that are happening there if they are all necessary

packages/vscode-extension/src/webview/components/UrlSelect.tsx
HexHyperion: Nope, it's for moving the focus between dropdown items, I changed the name now.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
HexHyperion: It's more of a React thing, because this array can't be recreated on every render - we'd lose the refs to the elements this way, so the "big" ref is used to keep the list unaltered between all the state changes, and the only time we're altering it is when the dropdown content changes.

I added a comment to explain it.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
HexHyperion: There's a `.map()` after the closing bracket of the array:
```tsx
itemsRef.current = [
  ...(dropdownOnly ? [{ id: "/", name: "/" }] : []),
  ...filteredItems,
  ...filteredOutItems,
].map((_, i) => itemsRef.current[i] || React.createRef<HTMLDivElement>());
```
I see how that can be confusing though, so I split it into two steps in the newest commit.

HexHyperion:
I think route extraction is finished now, but I found at least 3 bugs related to each other:
- If the user navigates to an incorrect route right after the app refreshes/loads, and then comes back using the back button (or "back" in the screen header), the path in the UrlSelect will not change to "/"
- If the user reloads a non-Expo Router app, the path in the UrlSelect will not get refreshed
- If the user clears Metro cache with the reload button or causes an error in the app, the UrlSelect won't get disabled
The bugs are not related to the topic of the PR though, so I guess they shouldn't be fixed here.

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -1,139 +1,190 @@
...
+function UrlSelect({
+  onValueChange,
+  navigationHistory,
+  routeList,
+  disabled,
+  dropdownOnly,
+}: UrlSelectProps) {
   const [isDropdownOpen, setIsDropdownOpen] = React.useState(false);
-  const [inputValue, setInputValue] = React.useState("/");
+  const [allItems, setAllItems] = React.useState<NavigationHistoryItem[]>([]);
```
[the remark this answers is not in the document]
HexHyperion: That makes sense, done.

packages/vscode-extension/src/webview/components/UrlSelectItem.tsx
```diff
@@ -0,0 +1,159 @@
...
+  onArrowPress: (
+    e: React.KeyboardEvent,
+    prev?: UrlSelectFocusable,
+    next?: UrlSelectFocusable
+  ) => void;
+  getNameFromId: (id: string) => string;
+  noHighlight?: boolean;
+}
+
+const UrlSelectItem = React.forwardRef<HTMLDivElement, PropsWithChildren<UrlSelectItemProps>>(
```
jwajgelt, line 22:
```suggestion
const UrlSelectItem = 
```
This isn't `forwardRef` anymore, right?

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -148,6 +184,17 @@ function UrlSelect({ onValueChange, navigationHistory, disabled, dropdownOnly }:
     }
   }, []);
 
+  // Hacky way to change the cursor style of a readonly input,
+  // since the VscodeTextfield component doesn't provide any parts
+  // or props, and according to the authors, it's not going to.
+  useLayoutEffect(() => {
+    if (textfieldRef.current && textfieldRef.current.shadowRoot) {
+      const style = document.createElement("style");
+      style.textContent = "input[readonly] { cursor: text !important; }";
```
kmagiera, line 193: I wonder if we could just have this in our css with proper selector to make sure this will only apply to input fields under url-select-input container

packages/vscode-extension/src/webview/components/UrlSelectItem.tsx
```diff
@@ -0,0 +1,154 @@
...
+  width,
+  style,
+  itemList,
+  refIndex,
+  textfieldRef,
+  onConfirm,
+  onArrowPress,
+  getNameFromId,
+  noHighlight = false,
+  ...props
```
kmagiera, line 33: this isn't used?

kmagiera (approved):

packages/vscode-extension/src/webview/components/UrlSelect.tsx
HexHyperion: Nope, I tried _everything_ to set this in CSS, and there's no way to access the hidden input - the authors of the lib responded to a similar issue with "it's by design"...

And even if it affects other fields, it's still weird to hit the users with a 🚫 on a focusable input

PR #1161, opened 2025-05-20
Allow disabling element inspector with the button

### LINKED ISSUE #948: Clicking inspector button to turn off inspection does not work
**Description**

When you start inspecting using the inspector button and want to stop by clicking the inspector button again, nothing happens. Which may be confusing. We assume that you click the first time to select, leave the component highlighted (so you know which component's code you see), and click the second time to turn off inspection. But allowing to turn off the inspection by inspector button would make sense too. 

**Environment**
- Repo: expo-52-dev-client (Probably not really related)

<video src="https://github.com/user-attachments/assets/38c0f4ec-5283-46cc-8106-bc7f6939fc14" />

HexHyperion:
This PR fixes #948, allowing the user to close the inspector mode at any moment by clicking on the active inspector button. 

The button now stays active after opening the clicked element's code (when the inspector highlight frame is present, but `isInspecting` is false), ensuring the user intuitively knows where the highlight came from and how to get rid of it.

I also fixed a bug where an appearing editor panel could prevent the inspector from visually selecting the element, which was caused by `setIsInspecting` being called on `mouseUp` instead of `mouseDown`.

### How Has This Been Tested: 
- open any RN project with components
- after the app loads, click on the inspector button - it becomes active, and hovering over components highlights them
- you can turn the inspector off by clicking the button again - the highlight frame will disappear
- with the inspector open, click on a component - the editor opens the code, the element stays highlighted, and the button is still active
- click the button or anywhere on the device screen - the highlight disappears, and the button stops being active

https://github.com/user-attachments/assets/fae82404-4877-4804-a0b9-5c2bffe626b3


packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -90,6 +90,7 @@ function PreviewView() {
   const { showDismissableError } = useUtils();
 
   const [isInspecting, setIsInspecting] = useState(false);
+  const [isInspectorActive, setIsInspectorActive] = useState(false);
```
kmagiera, line 93: it doesn't seem like we need another state for this. Why can't we use `isInspecting` here?

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -41,6 +41,7 @@ type Props = {
   setInspectFrame: (inspectFrame: Frame | null) => void;
   setInspectStackData: (inspectStackData: InspectStackData | null) => void;
   onInspectorItemSelected: (item: InspectDataStackItem) => void;
+  resetInspector: () => void;
```
kmagiera, line 44: There's already setIsInspecting and now resetInspector. We should only have one callback to control inspect state

kmagiera (changes requested):

packages/vscode-extension/src/webview/views/PreviewView.tsx
HexHyperion: `isInspecting` is true only between the inspector activation and element selection - the moment the editor jumps to the code, the state is being set to false (and I can't change that, because the element mouseover logic relies on it), but the inspector overlay is still present, so when the button got clicked, it would just enable element selection again.

From what I figured, the inspector has to have 3 different states for the button to work, but maybe there's a better way?

HexHyperion:
Is it better now? It seems much more readable with the ternary state, but do we even want the 3-state behavior?

packages/vscode-extension/src/common/Project.ts
```diff
@@ -35,6 +35,8 @@ export type BuildErrorDescriptor = {
 
 export type ProfilingState = "stopped" | "profiling" | "saving";
 
+export type InspectorState = "disabled" | "inspecting" | "selected";
```
jwajgelt, line 38: NIT: I'd prefer an enum

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -35,8 +36,8 @@ function TouchPointIndicator({ isPressing }: { isPressing: boolean }) {
 }
 
 type Props = {
-  isInspecting: boolean;
-  setIsInspecting: (isInspecting: boolean) => void;
+  inspectorState: string;
```
jwajgelt, line 39: Shouldn't this be
```suggestion
  inspectorState: InspectorState;
```

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -197,19 +198,15 @@ function Preview({
...
   const shouldPreventInputEvents =
     debugPaused || projectStatus === "refreshing" || !showDevicePreview || !!replayData;
 
-  const shouldPreventFromSendingTouch = isInspecting || !!inspectFrame;
+  const shouldPreventFromSendingTouch = inspectorState === "inspecting" || !!inspectFrame;
 
   function onMouseMove(e: MouseEvent<HTMLDivElement>) {
     e.preventDefault();
-    if (isInspecting) {
+    if (inspectorState === "inspecting") {
```
jwajgelt, line 210: I'd rather have `const isInspecting = inspectorState === "inspecting"` instead of repeating this check in multiple places.

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -84,7 +85,11 @@ function PreviewView() {
     useProject();
   const { showDismissableError } = useUtils();
 
-  const [isInspecting, setIsInspecting] = useState(false);
+  // The inspector can be in one of the following states:
+  // "disabled" - inspector is not active, device is interactable
+  // "inspecting" - inspector is active and inspecting elements on hover
+  // "selected" - no more hover, but highlight is active on an element
```
jwajgelt, line 91: If we change `InspectorState` to an enum, we can have this as a doc comment on the enum variants.

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -295,17 +307,31 @@ function PreviewView() {
...
+              case "disabled":
+                setInspectorState("inspecting");
+                break;
+              case "inspecting":
+                setInspectorState("disabled");
+                break;
+              case "selected":
+                setInspectorState("disabled");
+                break;
+            }
```
jwajgelt, line 333: I'd move this to a function (like `toggleInspecting`) to better communicate the intent of what this is supposed to do.

packages/vscode-js-debug
jwajgelt, line 1: Why is there a change to `vscode-js-debug`?

jwajgelt (changes requested):

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -238,30 +235,31 @@ function Preview({
   function onMouseDown(e: MouseEvent<HTMLDivElement>) {
```
filip131311, line 236: I am not a fan of moving effects of clicks into `mouseDown` this is expected behaviorv that moving out of focus after pressing something will cancel the action 


https://github.com/user-attachments/assets/cecf9053-8ad1-40d3-8146-8feae5340901

HexHyperion:
Too many things broke here for such a small functionality, probably better to start over than fix it in rush...

PR #1162, opened 2025-05-20
Add approot select to Preview

jwajgelt:
Adds a dropdown to the main Preview screen which allows the user to quickly switch between applications in a monorepo as well as see at a glance which app root is selected.

### Known issues:
- sometimes after switching the AppRoot, a "Couldn't start selected device" error appears. This doesn't prevent the app to build and run correctly (and after it does, Radon will stop showing the error and work normally), so it seems to be an issue with error handling in device startup.

### TODO:
- [x] ~cleanup css classes in webview code~ I gave up
- [x] improve how application's name is derived (support `app.config.js`, use `package.json` as a fallback)

Screenshot 2025-05-20 at 14 26 06
Screenshot 2025-05-20 at 14 27 26

### How this was tested:
- open a monorepo or a folder containing multiple RN apps
- the half-pill button next to the device picker should display the name of the current application
- press the button
- the dropdown should contain all the RN apps in the workspace
- selecting an entry should switch Radon to that app

### DIFF
--- a/packages/vscode-extension/src/panels/LaunchConfigController.ts
+++ b/packages/vscode-extension/src/panels/LaunchConfigController.ts
@@ -2,6 +2,7 @@ import path from "path";
 import { EventEmitter } from "stream";
 import { ConfigurationChangeEvent, workspace, Disposable } from "vscode";
 import {
+  ApplicationRoot,
   LaunchConfig,
   LaunchConfigEventListener,
   LaunchConfigEventMap,
@@ -14,9 +15,44 @@ import { getIosSourceDir } from "../builders/buildIOS";
 import { readEasConfig } from "../utilities/eas";
 import { EasBuildConfig } from "../common/EasConfig";
 import { getLaunchConfiguration } from "../utilities/launchConfiguration";
-import { promises } from "fs";
 
 const CUSTOM_APPLICATION_ROOTS_KEY = "custom_application_roots_key";
+
+function readApplicationRoot(appRootPath: string): ApplicationRoot {
+  const appRootAbsolutePath = path.resolve(workspace.workspaceFolders![0].uri.fsPath, appRootPath);
+  try {
+    const appRootConfig = require(appRootAbsolutePath + "/app.json");
+    if (appRootConfig) {
+      return {
+        path: appRootPath,
+        name: appRootConfig.name ?? appRootConfig.expo?.name ?? path.basename(appRootPath),
+        displayName: appRootConfig.displayName,
+      };
+    }
+  } catch {}
+  try {
+    const configProvider = require(appRootAbsolutePath + "/app.config.js");
+    const appRootConfig = configProvider({ config: {} });
+    if (appRootConfig) {
+      return {
+        path: appRootPath,
+        name: appRootConfig.name ?? appRootConfig.expo?.name ?? path.basename(appRootPath),
+        displayName: appRootConfig.displayName,
+      };
+    }
+  } catch {}
+  try {
+    const appPackageJson = require(appRootAbsolutePath + "/package.json");
+    return {
+      path: appRootPath,
+      name: appPackageJson.name ?? path.basename(appRootPath),
+    };
+  } catch {}
+  return {
+    path: appRootPath,
+    name: path.basename(appRootPath),
+  };
+}
 
 export class LaunchConfigController implements Disposable, LaunchConfig {
   private config: LaunchConfigurationOptions;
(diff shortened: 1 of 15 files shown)


packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -54,7 +54,7 @@ function LaunchConfigurationView() {
       <AppRootConfiguration
         appRoot={appRoot}
         update={update}
-        applicationRoots={applicationRoots}
+        applicationRoots={applicationRoots.map((root) => root.path)}
```
kmagiera, line 57: any reason we can't just pass the whole objects here? given its called applcationRoots and we now have a structure that is called ApplicationRoot I'd expect this prop to take array of ApplicationRoot objects and not just the paths

packages/vscode-extension/src/project/project.ts
```diff
@@ -98,6 +90,18 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
       }
     );
 
+    this.projectState = {
```
kmagiera, line 93: this will likely conflict with #1154

packages/vscode-extension/src/panels/LaunchConfigController.ts
```diff
@@ -17,6 +18,42 @@ import { getLaunchConfiguration } from "../utilities/launchConfiguration";
 
 const CUSTOM_APPLICATION_ROOTS_KEY = "custom_application_roots_key";
 
+function readApplicationRoot(appRootPath: string): ApplicationRoot {
+  const appRootAbsolutePath = path.resolve(workspace.workspaceFolders![0].uri.fsPath, appRootPath);
+  try {
+    const appRootConfig = require(appRootAbsolutePath + "/app.json");
```
kmagiera, line 24: assuming we just want to read a json file here we should use requireNoCache here

packages/vscode-extension/src/panels/LaunchConfigController.ts
```diff
@@ -17,6 +18,42 @@ import { getLaunchConfiguration } from "../utilities/launchConfiguration";
...
+    if (appRootConfig) {
+      return {
+        path: appRootPath,
+        name: appRootConfig.name ?? appRootConfig.expo?.name ?? path.basename(appRootPath),
+        displayName: appRootConfig.displayName,
+      };
+    }
+  } catch {}
+  try {
+    const configProvider = require(appRootAbsolutePath + "/app.config.js");
```
kmagiera, line 34: I'd be hesitant to load user's js files like that. They may execute some code and load extra bunch of dependencies along which may result in increased memory usage and pollute the node environment that the extension uses. Maybe there's some better way or in the worst case we should start a node subprorcess to read and return the data we want

packages/vscode-extension/src/panels/LaunchConfigController.ts
```diff
@@ -17,6 +18,42 @@ import { getLaunchConfiguration } from "../utilities/launchConfiguration";
...
+    if (appRootConfig) {
+      return {
+        path: appRootPath,
+        name: appRootConfig.name ?? appRootConfig.expo?.name ?? path.basename(appRootPath),
+        displayName: appRootConfig.displayName,
+      };
+    }
+  } catch {}
+  try {
+    const appPackageJson = require(appRootAbsolutePath + "/package.json");
```
kmagiera, line 45: use requireNoCache here?

kmagiera (changes requested):

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
jwajgelt: We can, I just didn't want to refactor that component.
It's no issue.

packages/vscode-extension/src/project/project.ts
jwajgelt: Yes, but we have to pass it somehow, and this seems to be the most sensible place to me.
It shouldn't be hard to adapt either PR to this change.

packages/vscode-extension/src/panels/LaunchConfigController.ts
jwajgelt: For sure :+1:
I read it using `fs` initially, and later found some place where we used `require` for it, so I just copied the approach here.

packages/vscode-extension/src/panels/LaunchConfigController.ts
jwajgelt: I've thought about it.
I felt the risk of people doing weird things in a `config.js` file should be low enough to not worry about it, but maybe that _is_ overly optimistic.
I think I'll play around with it for a bit and just ignore this file if I don't find a solution that I like -- the `package.json` fallback should be fine, since most templates fill the name there anyway.

kmagiera (approved):
Looks good 👍

PR #1165, opened 2025-05-21
Fix build stage state updates

kmagiera:
This PR fixes stage progress updates after #1154 

1) We're resetting stageProgress to undefined as it is only expected to be set for stages that can actually report progress (build and bundle)
2) We're resetting stale build cache flag when restarting the device
3) We're only updating stale build cache flag when in running state to prevent accidental reports when build is ongoing and modifies files inside the build folder.

### How Has This Been Tested: 
1. Run app
2. Use "clean build" option
3. See the progress is only reported in build and bundle phases and there's no "0%" progress reported elsewhere
4. Make sure that the stale build dialog doesn't pop up when in build process
5. Add some dummy line to podfile and see the stale build dialog appear
6. Use restart button from the dialog
7. It should be gone immediately and not show up after the device reboots. Further updates to Podfile should make the alert re-appear

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -141,10 +141,17 @@ export class DeviceSession
     this.startupMessage = startupMessage;
     this.stageProgress = undefined;
     this.buildError = undefined;
+    this.hasStaleBuildCache = false;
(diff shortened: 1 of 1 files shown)


jwajgelt (approved):
Other than the one question, I see no issues.

PR #1166, opened 2025-05-21
Use skip files for logs in radon connect

kmagiera:
This PR updates the Proxy Debug Adapter such that it considers the skipFiles setting when filtering the stack frame for console logs.

This fixes an issue when using Radon Connect which doesn't override the built-in console log method in a way that'd provide accurate stack frames. For the main IDE panel experience, we append the extra frame data to the log calls and the extract them in the code that handles console logging. Then we use that to strip out the frame from the framework code that supports logging.

With Radon connect we don't have that, and all logs just appear with lines from withing the framework code which isn't super helpful. To fix this, we use skip files to filter the log call frame. This code is universal enough that we should use it in all the configuration hence no Radon Connect checks are necessary.

### How Has This Been Tested: 
1. Run app with Radon Connect
2. Print some logs
3. See that the logs no longer point to console.js file in framework internals

### DIFF
--- a/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
+++ b/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
@@ -301,8 +301,8 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
       stackTrace?.callFrames.splice(0, originalCallFrameIndex);
     }
 
-    if (stackTrace?.callFrames) {
-      stackTrace.callFrames = stackTrace?.callFrames.filter((frame) => {
+    if (stackTrace) {
+      const filteredCallFrames = stackTrace?.callFrames.filter((frame) => {
         const { scriptId, lineNumber, columnNumber } = frame;
         const { sourceURL } = this.sourceMapRegistry.findOriginalPosition(
           scriptId,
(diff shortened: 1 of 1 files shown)


jwajgelt (commented):
I'm not sure this is what we want to do.
On one hand, this is actually an improvement over what we had in cases where the user has some library which wraps logging (such as sentry).
On the other, if you call some imported function which logs something internally, since we skip `node_modules`, you'll see your code which calls the function in the logs, not the actual dependency which logged.
Maybe that's not an issue?

filip131311 (approved):
Tested on RN78, works

filip131311:
@jwajgelt 
> On the other, if you call some imported function which logs something internally, since we skip node_modules, you'll see > > your code which calls the function in the logs, not the actual dependency which logged.

I think this is expected by the user, but if they really want to they can always use negate pattern to look into `node_modules` in fact I expect that the set of users who would like to look into `node_modules` closely matches the set of people who would think about negation pattern.

jwajgelt:
> @jwajgelt
> 
> > On the other, if you call some imported function which logs something internally, since we skip node_modules, you'll see > > your code which calls the function in the logs, not the actual dependency which logged.
> 
> I think this is expected by the user, but if they really want to they can always use negate pattern to look into `node_modules` in fact I expect that the set of users who would like to look into `node_modules` closely matches the set of people who would think about negation pattern.

I disagree.
I think it will be confusing when you see unexpected logs (since you didn't log them), and when you click the link to check where the log happened, you get sent to a line of code which doesn't log anything. Maybe the users will figure out that it means the logs come from the external function they call there, but I'd still find that weird.

jwajgelt:
That said, I would also not expect other libraries to flood my logs without my permission, so I wouldn't expect this to come up often.

kmagiera:
This is our best shot to show user's code in the debug console. This is also the approach that the stock devtools are relying on

PR #1171, opened 2025-05-23
Fix replay overlay styles

kacperkapusciak:
The background overlay for Replay feature was hardcoded and didn't adjust to the currently selected vscode theme.

This PR makes the background color aligned to the current vscode theme. Also, fixed the length selector background color. 

## Before
https://github.com/user-attachments/assets/d799b5d6-85d3-44ad-9c23-ecaa979a0c58

## After
https://github.com/user-attachments/assets/24b6b4cc-cf0a-46e9-a899-068cb1b2450c


filip131311 (approved):
Looks much better now Thank you

PR #1173, opened 2025-05-23
Fix Radon not starting when no devices are found

filip131311:
This PR fixes an issue with radon not loading when no devices are available. 

The problem was caused by `initialized` in `projectState` not being updated as after #1154 `initialized` is only updated when `updateProjectState` and it does not happen if no devices are found. 

To solve this problem we introduce a new `deviceSessionsManagerDelegate` method and update `initialized` even when no device can be selected 

### How Has This Been Tested: 

- Run radon with no devices


jwajgelt (approved):
I feel like I'd prefer an `async` method called `initialize` which is awaited in project and sets the state then, but the current approach is fine as well.

PR #1175, opened 2025-05-23
Fix bundle error popup after fast refresh

filip131311:
This PR fixes an issue with bundle error popup either flashing or completely not showing when bundle error is caused be "fast refresh". The problem was happening because even if fast refresh is unsuccessful it is reported and coupled which means that `RNIDE_fastRefreshComplete` event does not mean that the app is running. 

To avoid this issue we update device status, only if the current status is "refreshing" which means that the refresh process has been uninterrupted.

### How Has This Been Tested: 

- run `react-native-77` test app and cause bundle error with fast refresh

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -265,6 +265,9 @@ export class DeviceSession
       this.emitStateChange();
     });
     devtools.onEvent("RNIDE_fastRefreshComplete", () => {
+      if (this.status !== "refreshing") {
+        return;
+      }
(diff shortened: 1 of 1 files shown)


kmagiera (approved):
nice, while there's perhaps not much of a difference I'd feel safer if we restored the previous logic here, that is check for "starting" and "bundleError" states:
https://github.com/software-mansion/radon-ide/pull/1154/files#diff-8d5143a1f3455de6dbfaa820c4b22f0f3af065a16dd2cf0a27a5e4f3b59d52d6L226-L230
image

PR #1179, opened 2025-05-25
Restore logic for debugger paused state to version before #1154

kmagiera:
In #1168 I broke the debugger overlay by introducing additional state variable that wasn't hooked into the UI properly.

While this was fixed in #1168 I think it'd be safer to restore the previous logic from before changes in #1168.

Before that change, we've been using "debuggerPaused" as one of the statuses for the project. We've been also using that status when processing debugger resumed events. This PR restores this behavior.

### How Has This Been Tested: 
1. Run project, make sure the debugger overlay appears


jwajgelt:
Why do you think having a separate state is better in this case?

PR #1180, opened 2025-05-25
Extract generic interface for communicating between app process and the extension

kmagiera:
Previously, we've been relying on the react devtools websocket connection as communication channel for all Radon specific communication including, sending inspect commands, notifying about refresh events, or communicating between the plugins (network, redux, etc).

Apparently, this approach is not ideal when considering Radon Connect that doesn't control the metro or devtools servers.

For this reason, we're abstracting away the communication related code to a separate interface that can be implemented separately by the react devtools server and over CDP for Radon Connect (not a part of this PR).

This PR is purely focus on extracting the interface and some cleanup around what type of messages we use when communicating over react devtools:
1) Previously, we'd have separate event names for different events. We're generalizing this now and only use a single Radon specific event name This makes it easier for us to use any names we want without worrying about name collisions with what devtools is sending. It also simplifies the logic somewhat as we no longer need to assign listeners to all event types that we want to support.
2) On the `lib/` side, all the code will now communicate via `inspector_bridge` module which works on top of an "agent" interface which is designed based on the webworker API (`onmessage` and `sendMessage` methods).
3) We're providing a react devtools backed implementation for the agent that relies on our single custom message `"RNIDE_message"`
4) Since we want the agent to be swappable and possibly available from different places, we store it under a global `__radon_agent` property.

In addition:
1) We're introducing some renamings such that the class/method names better reflect what they do and are less generic.
2) We're restructuring render-overlays code to work similarily to other plugins. It's code is now moved to a separate file and it uses "pluginMessage" command for communication.

### How Has This Been Tested: 
1. Run example app with plugins
2. Test react scan, network and react query plugin (communicating both ways)
3. Test redux plugin both ways via expo (set up in the expo-52-prebuild-with-plugins) project
4. Comment out devtoolEnhancer in the app code to test the built-in redux integration as well.


packages/vscode-extension/lib/react_devtools_agent.js
```diff
@@ -0,0 +1,39 @@
...
+};
+
+const hook = window.__REACT_DEVTOOLS_GLOBAL_HOOK__;
+if (hook.reactDevtoolsAgent) {
+  setDevtoolsAgent(hook.devtoolsAgent);
+} else {
+  hook.on("react-devtools", setDevtoolsAgent);
+}
+
+globalThis.__radon_agent = agent;
```
jwajgelt, line 39: Because it's named `agent`, I initially confused this global to be storing the `reactDevtoolsAgent` when reading the other files. I'd change this to something that's less likely to be confused with anything else.

This class is used as a way for the extension and application to communicate. Maybe something like `RadonExtensionChannel` or `RadonExtensionConnection` would be more fitting?

packages/vscode-extension/lib/inspector_bridge.js
```diff
@@ -0,0 +1,25 @@
+const agent = globalThis.__radon_agent;
+
+if (!agent) {
+  throw new Error("Radon inspector bridge agent is not installed");
+}
+
+const messageListeners = [];
+
+const inspectorBridge = {
```
jwajgelt, line 9: For me, the term "Inspector" has a specific meaning, relating to the Element Inspector feature. I'd avoid reusing that name for other things -- it's surprising to me that other tools use an "Inspector Bridge", and not just the "Inspector".
Since this is meant to be used for communicating with the extension, I'd rename this to something like `ExtensionBridge`

packages/vscode-extension/src/project/bridge.ts
```diff
@@ -0,0 +1,104 @@
...
+  fastRefreshStarted: [];
+  fastRefreshComplete: [];
+  openPreviewResult: [{ previewId: string; error?: string }];
+  inspectData: [{ id: number }];
+  devtoolPluginsChanged: [{ plugins: string[] }];
+  pluginMessage: [{ pluginId: string; type: string; data: any }];
+  isProfilingReact: [boolean];
+}
+
+export interface RadonInspectorBridge {
```
jwajgelt, line 17: Like mentioned before, I'd rename this to something like `RadonApplicationBridge` to avoid confusing this as something Inspector-specific.

PR #1186, opened 2025-05-26
Add label to approot select dropdown

kmagiera:
Adds a label to select dropdown to match the look from device select dropdown. Also, the dropdown w/o any label look very "mysterious" for projects where only one option is available.

### How Has This Been Tested: 
1. Open project select dropdown

**Before**
image

**After**
image

### DIFF
--- a/packages/vscode-extension/src/webview/components/AppRootSelect.tsx
+++ b/packages/vscode-extension/src/webview/components/AppRootSelect.tsx
@@ -16,6 +16,7 @@ function renderAppRoots(
 
   return (
     <Select.Group>
+      <Select.Label className="approot-select-label">Select project</Select.Label>
       {applicationRoots.map(({ path, displayName, name }) => (
         <RichSelectItem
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/webview/components/AppRootSelect.tsx
```diff
@@ -16,6 +16,7 @@ function renderAppRoots(
 
   return (
     <Select.Group>
+      <Select.Label className="approot-select-label">Select project</Select.Label>
```
jwajgelt, line 19: I think I'd prefer
```suggestion
      <Select.Label className="approot-select-label">Select application</Select.Label>
```
But :shrug:

I think both XCode and Visual Studio (the fat one) use the terminology of "Project" being a part of "Workspace", so your version makes sense in that context.
I don't have a strong opinion on that.

jwajgelt (approved):

PR #1191, opened 2025-05-27
Fix broken UI styles

HexHyperion:
This PR adjusts the styles of the extension's webview, specifically of device/approot selectors and the device management menu:
- the focus outline of the selectors was clipped by `overflow: hidden;` - now it's able to escape the container;
Zrzut ekranu 2025-05-27 o 11 33 28

- long letters (j, y, g, etc.) were being clipped in the selectors and in the device management screen - line and container height has been adjusted to prevent clipping:
Zrzut ekranu 2025-05-27 o 11 33 59
Zrzut ekranu 2025-05-27 o 11 34 18
Zrzut ekranu 2025-05-27 o 11 35 16

- sharp corners of selectors' last items visually collided with the dropdown's `border-radius` - now the last items have rounded bottom corners:
Zrzut ekranu 2025-05-27 o 11 35 46
Zrzut ekranu 2025-05-27 o 11 36 09

- the device selection dropdown was aligned to the start of the trigger, which caused it to stick out to the right - the alignment has been changed to `center`, making it feel more like a single two-part selector rather than two separate ones:
Zrzut ekranu 2025-05-27 o 12 04 04

- \+ a few other minor fixes.

### How Has This Been Tested: 
1. Start the extension and wait for webview to load
2. Click between the inspector button and the selector and press Tab - the approot selector should become focused properly
3. Open the selector and hover over the last item - now the corners match those of `<Select.Content>`
4. Open the device selector, select a device with long letters in the name - the text isn't clipped anywhere
5. Hover over the "Manage devices..." button - it now has rounded corners at the bottom
6. Click the button to open the device management window - the devices' names on the list aren't clipped anymore

### DIFF
--- a/packages/vscode-extension/src/webview/components/UrlSelect.css
+++ b/packages/vscode-extension/src/webview/components/UrlSelect.css
@@ -124,3 +124,11 @@
     max-height: 50vh;
   }
 }
+@media (width <= 350px) {
+  .url-select-content {
+    display: none;
+  }
+  .url-select-input {
+    border-radius: 18px !important;
+  }
+}
--- a/packages/vscode-extension/src/webview/views/ManageDevicesView.css
+++ b/packages/vscode-extension/src/webview/views/ManageDevicesView.css
@@ -23,7 +23,7 @@
   align-items: center;
   gap: 10px;
   line-height: 1.2;
-  height: 48px;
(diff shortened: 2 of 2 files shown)


kmagiera (approved):
Thanks 🙌

PR #1192, opened 2025-05-28
fallback to app name "app" on Android

jwajgelt:
When building android, the task invoked by Android Studio and Expo cli to build the apk is `{appName}:assemble{Variant}`. Currently, if Radon cannot read an app name from the application's config, it falls back to executing `assemble{Variant}` which at least on one project tested failed due to `.aar` dependencies (when using `react-native-executorch`).
This PR changes the behaviour to fall back to `appName = "app"` when none is specified in the config, which is the default app name in React Native projects and the app name assumed by expo cli (https://github.com/expo/expo-cli/blob/54997b9b1aa66329f91e33e913f98155bcbb2464/packages/expo-cli/src/commands/run/android/runAndroid.ts#L107)

### How Has This Been Tested: 
I cannot reproduce the problem in a fresh app, but this fixed it for me on the internal appjs app.

### DIFF
--- a/packages/vscode-extension/src/builders/buildAndroid.ts
+++ b/packages/vscode-extension/src/builders/buildAndroid.ts
@@ -66,7 +66,7 @@ export async function getAndroidBuildPaths(
   return { apkPath, packageName };
 }
 
-function makeBuildTaskName(productFlavor: string, buildType: string, appName?: string) {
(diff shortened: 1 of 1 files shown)


filip131311 (commented):
I am worried about the situation in which someone has non default app but `loadConfig` still returns no appName For example if someone does not use expo and changes this configuration on the native side

PR #1193, opened 2025-05-28
Fix UrlSelect item focusing and unreadable dynamic segment colors

HexHyperion:
This PR fixes two bugs noticed in the behavior of the `UrlSelect` component:
- the keyboard item focusing functionality sometimes breaking after adding a new path to the dropdown, e.g. after opening a preview - caused by duplicated entries in the array of item references
- ~the suggestions extracted from Expo Router not showing after the app started, especially on Android - caused by incorrect logic of the DevTools agent listener, which led to the route list not being sent to the webview~
fixed in #1180

Additionally, I made the dynamic segments in the dropdown more visible by changing `color: var(--swm-disabled-text)` to a new, custom color.

### How Has This Been Tested: 
1. Open a multi-route Expo Router project with Expo version ≤52 and launch it on an Android device
2. After the app loads, click on the `UrlSelect` - the indexed routes should be displayed properly and with dynamic segments brightly colored
3. Navigate to some route, then open a preview, navigate back to `/`, play with the filtering in the bar while testing the arrow key focusing at each point - it shouldn't get stuck or focus incorrect items anymore


packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -33,7 +33,8 @@ function UrlSelect({
   const [currentDynamicSegment, setCurrentDynamicSegment] = React.useState<number>(0);
   const [textfieldWidth, setTextfieldWidth] = React.useState<number>(0);
 
-  const dropdownItemsRef = React.useRef<Array<HTMLDivElement>>([]);
+  const dropdownItems: UrlSelectFocusable[] = [];
```
jwajgelt, line 36: I don't like the idea of creating a new object to pass as a prop to children on each render.
Maybe it's fine in this case, since we want to always store the rendered elements here.
I'd rather someone else give their opinion on this.
@kmagiera @filip131311

packages/vscode-extension/lib/wrapper.js
```diff
@@ -365,15 +375,13 @@ export function AppWrapper({ children, initialProps, fabric }) {
...
       setDevtoolsAgent(hook.reactDevtoolsAgent);
-    } else {
-      hook.on("react-devtools", setDevtoolsAgent);
-      return () => {
-        hook.off("react-devtools", setDevtoolsAgent);
-      };
     }
+    return () => {
+      hook.off("react-devtools", setDevtoolsAgent);
+    };
```
jwajgelt: Is this change necessary?

packages/vscode-extension/src/webview/components/UrlSelect.tsx
HexHyperion: It may as well stay like it was, but since the children add themselves to the array at specific indexes, it'd need to be cleared every time the items change order (otherwise keyboard focusing breaks), so either on each render (making `useRef` useless) or in the `combinedItems` `useMemo` which is probably slightly more effective, but also a bit less readable.

packages/vscode-extension/lib/wrapper.js
HexHyperion: Now that I look at it, no, it's a leftover from trying to get the routes working...

HexHyperion:
After the bridge refactor the `devtoolsAgent` fix is no longer needed, so merging this just as a fix for focusing, while the navigation bar problems that appeared after the refactor will be fixed in #1211.

PR #1199, opened 2025-06-02
Fix dynamic route params duplicating as query params in route displayNames

HexHyperion:
This PR enhances the readability of displayed dynamic paths by removing all params already present in the route URL from the query (from after the `?` sign). This way, the values won't be passed twice, significantly shortening the path string.

Before:
Zrzut ekranu 2025-06-2 o 16 17 53
After:
Zrzut ekranu 2025-06-2 o 16 18 12

### How Has This Been Tested: 
1. Open an Expo Router project
2. Navigate to a dynamic route - params should now only be present inside the actual path
3. Navigate to a "not-found" route or a static route that accepts query params - normal params are still displayed in the route name

### DIFF
--- a/packages/vscode-extension/lib/expo_router_helpers.js
+++ b/packages/vscode-extension/lib/expo_router_helpers.js
@@ -6,6 +6,22 @@ export function computeRouteIdentifier(pathname, params) {
   return query ? `${pathname}?${query}` : pathname;
 }
 
+// Helper function to prevent duplicating dynamic segments of the route
+// in the params object, as they are already part of the route path.
+export function getParamsWithoutDynamicSegments(routeInfo) {
+  const params = routeInfo?.params || {};
+  const dynamicSegments = routeInfo?.segments.filter((segment) => segment.startsWith("[") && segment.endsWith("]")) || [];
+
+  Object.keys(params).forEach((key) => {
+    if (dynamicSegments.map((segment) => segment.slice(1, -1)).includes(key)) {
+      delete params[key];
+    }
+  });
+
+  return params;
+}
+
+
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/lib/expo_router_helpers.js
```diff
@@ -6,6 +6,22 @@ export function computeRouteIdentifier(pathname, params) {
...
 }
 
+// Helper function to prevent duplicating dynamic segments of the route
+// in the params object, as they are already part of the route path.
+export function getParamsWithoutDynamicSegments(routeInfo) {
+  const params = routeInfo?.params || {};
+  const dynamicSegments = routeInfo?.segments.filter((segment) => segment.startsWith("[") && segment.endsWith("]")) || [];
+
+  Object.keys(params).forEach((key) => {
+    if (dynamicSegments.map((segment) => segment.slice(1, -1)).includes(key)) {
```
kmagiera: you can move the `map` call above as it can be calulcated before the `forEach` loop

packages/vscode-extension/lib/expo_router_plugin.js
```diff
@@ -14,7 +14,7 @@ function useRouterPluginMainHook({ onNavigationChange, onRouteListChange }) {
   const pathname = routeInfo?.pathname;
   const params = routeInfo?.params;
 
-  const filteredParams = params ?? {};
+  const filteredParams = getParamsWithoutDynamicSegments(routeInfo);
```
kmagiera, line 17: would it make sense to also handle deleting of the __EXPO_ROUTER_key property in this newly added method given we call it from all the versions of the router plugin?

kmagiera (approved):
Accepting, but please address the inline comments first

PR #1200, opened 2025-06-03
Use a login shell to evaluate the PATH variable

jwajgelt:
This PR makes the shell process used for determining the `PATH` variable a "login" shell. This allows us to read the configuration from `.profile`/`.zprofile` and similar.

### How Has This Been Tested: 
- set up some important PATH variable (`node` via `nvm` or `ruby` via `rbenv`) in `.zprofile` (if using `zsh`)
- start Radon
- verify the correct version is used

### DIFF
--- a/packages/vscode-extension/src/utilities/subprocess.ts
+++ b/packages/vscode-extension/src/utilities/subprocess.ts
@@ -25,7 +25,12 @@ async function getPathEnv(appRoot: string) {
   // which would otherwise fill missing variables with those from the currently running process.
   const { stdout } = await execa(
     shellPath,
-    ["-i", "-c", `cd "${appRoot}" && echo "${RNIDE_PATH_DELIMITER}$PATH${RNIDE_PATH_DELIMITER}"`],
+    [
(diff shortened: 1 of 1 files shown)


jwajgelt:
> Looks good, but I wonder now if we need the `-i` option. We added it in #515 such that the rc files are loaded, but I'd assume that the login option does that too and we don't actually need interactive option now given we only print the path variable

From the `zsh` man page:
> If the shell is a login shell, commands are read from /etc/zprofile and then $ZDOTDIR/.zprofile. Then, if the shell is interactive, commands are read from /etc/zshrc and then $ZDOTDIR/.zshrc. Finally, if the shell is a login shell, /etc/zlogin and $ZDOTDIR/.zlogin are read.

Meaning we need to launch a "login" _and_ "interactive" shell if we want to handle both `profile` and `rc` files.
"Login interactive" should also be the closest to what the shell inside a terminal actually is (at least on macos with `iTerm`/`iTerm2` which are the terminals I have tested with).

PR #1202, opened 2025-06-03
Refactor "refreshing" from project status to boolean flag

jwajgelt:
Refactors the "refreshing" project state from a separate project status to a boolean flag in device session state and hides the refreshing indicator when the preview is not displayed.

### How Has This Been Tested: 
- open an app
- start a fast refresh
- verify the fast refresh indicator is shown and hidden as expected
- verify the fast refresh indicator doesn't appear when installing pods/dependencies or installing the app while the app is still loading

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -159,6 +159,7 @@ export class DeviceSession
     this.startupMessage = startupMessage;
     this.stageProgress = undefined;
     this.buildError = undefined;
+    this.isRefreshing = false;
     this.hasStaleBuildCache = false;
     this.profilingCPUState = "stopped";
     this.profilingReactState = "stopped";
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -279,15 +278,11 @@ export class DeviceSession
...
+      this.isRefreshing = true;
       this.emitStateChange();
     });
     devtools.onEvent("fastRefreshComplete", () => {
-      const ignoredEvents = ["starting", "bundlingError"];
-      if (ignoredEvents.includes(this.status)) {
-        return;
-      }
-      this.status = "running";
+      this.isRefreshing = false;
```
kmagiera, line 286: this is bad that we only ever set it to `false` here. If we miss the event (i.e. the app would crash when refreshing, or someone reloads the app while refreshing), it will never reset. We should at least set it to false in `resetStartingState` method and perhaps in some other places too?

kmagiera (approved):
Looks good overall. Accepting to unblock but note that we need to reset the flag in other places too (added inline comment about it)

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: `resetStartingState` seems to be enough, since it's called wherever the app is reloaded.

PR #1211, opened 2025-06-04
Prevent sending early and duplicate navigation events

HexHyperion:
This PR fixes the issues exposed after the switch to a custom bridge in #1180, that is:
- empty or `undefined` `navigationChanged` events being sent before the app started and appearing in path suggestions
- paths from outside the main application (e.g. previews) being reset to the last in-app path in the URL bar after a split second from opening, caused by sending duplicated `navigationChanged`

The `useEffect` responsible for sending the events was already activating in those situations before the bridge change, but the events were usually blocked by the `devtoolsAgent` being `null`, whereas the new `inspectorBridge` is available all the time, so we need to actively prevent sending the problematic updates.

### How Has This Been Tested: 
1. Open an Expo Router app
2. Open the UrlSelect - there should be no `undefined` or `""` on the list
3. Play with the navigation, confirm all paths are displayed correctly
4. Open a preview - the `preview:...` path should stay in the bar properly
5. Open a "404" route - the path should reflect the app's current location and update on `.goBack()`

### DIFF
--- a/packages/vscode-extension/lib/expo_router_helpers.js
+++ b/packages/vscode-extension/lib/expo_router_helpers.js
@@ -6,6 +6,18 @@ export function computeRouteIdentifier(pathname, params) {
   return query ? `${pathname}?${query}` : pathname;
 }
 
+export function compareNavigationDescriptors(a, b) {
+  if (a.pathname !== b.pathname) {
+    return false;
+  }
+  for (const key in a.params) {
+    if (a.params[key] !== b.params[key]) {
+      return false;
+    }
+  }
+  return true;
+}
+
 // Helper function to extract the route list from Expo Router's routeNode, which is a tree-like object
 // returned by the router store and the getRoutes() function, containing all indexed routes.
 // For future reference: https://github.com/expo/expo/blob/main/packages/expo-router/src/getRoutes.ts
--- a/packages/vscode-extension/lib/expo_router_v5_plugin.js
+++ b/packages/vscode-extension/lib/expo_router_v5_plugin.js
@@ -1,11 +1,16 @@
-import { useEffect } from "react";
+import { useEffect, useRef } from "react";
 import { useRouter } from "expo-router";
 import { store, useRouteInfo } from "expo-router/build/global-state/router-store.js";
-import { computeRouteIdentifier, extractNestedRouteList } from "./expo_router_helpers.js";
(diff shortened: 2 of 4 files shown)


packages/vscode-extension/lib/expo_router_helpers.js
```diff
@@ -6,6 +6,18 @@ export function computeRouteIdentifier(pathname, params) {
   return query ? `${pathname}?${query}` : pathname;
 }
 
+export function compareNavigationDescriptors(a, b) {
```
jwajgelt:
```suggestion
export function areNavigationDescriptorsEqual(a, b) {
```
`compare` (to me) suggests a "comparator" (like the callback to [Array.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#parameters)) that orders the arguments, not just checks for equality.

packages/vscode-extension/lib/expo_router_v5_plugin.js
```diff
@@ -23,12 +33,19 @@ function useRouterPluginMainHook({ onNavigationChange, onRouteListChange }) {
   }, [store.routeNode]);
 
   useEffect(() => {
-    onNavigationChange({
-      name: displayName,
-      pathname,
-      params,
-      id: computeRouteIdentifier(pathname, params),
-    });
+    if (
```
filip131311: if this is the same logic for all expo plugins we could extract it into `expo_router_helpers.js`

packages/vscode-extension/lib/expo_router_helpers.js
```diff
@@ -6,6 +6,18 @@ export function computeRouteIdentifier(pathname, params) {
...
+  if (a.pathname !== b.pathname) {
+    return false;
+  }
+  for (const key in a.params) {
+    if (a.params[key] !== b.params[key]) {
+      return false;
+    }
+  }
```
jwajgelt: NIT:
I'd use `Array.some()` or `Array.every()` for this sort of "for loop checking some predicate", since they generally better communicate the intent, like
```suggestion
  return Object.keys(a.params).every(key => a.params[key] === b.params[key])
```
That said, shouldn't we also check that the keys are the same for both objects?
For example, for:
```
a.params = { a: "a" }
b.params = { a: "a", b: "b" }
```
this method will return `true`

jwajgelt (approved):

PR #1216, opened 2025-06-04
docs: add a Contact Us page

kacperkapusciak:
This PR adds a new `/contact` page

#### desktop
image

#### mobile
image

### DIFF
--- a/packages/docs/src/components/Contact/index.tsx
+++ b/packages/docs/src/components/Contact/index.tsx
@@ -1,7 +1,0 @@
-// import styles from "./styles.module.css";
-
-// export default function ContactUsScreen(): JSX.Element {
-//   return (
-
-//   );
-// }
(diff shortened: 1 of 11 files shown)


packages/docs/src/components/Contact/index.tsx
jwajgelt, line 1: You can probably remove this file.

packages/docs/src/pages/contact.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+                <div>
+                  <a
+                    href="mailto:projects@swmansion.com"
+                    target="_top"
+                    className={styles.contactButton}>
+                    Talk to Sales <ArrowRightSmallIcon />
+                  </a>
+                </div>
+              </div>
```
jwajgelt: Maybe move this to a `<ContactCard>` component? there's a lot of repetition here, with only the text, `href` and `target` props changing.

kacperkapusciak: done

PR #1217, opened 2025-06-05
Refactor device session management in project state

jwajgelt:
Refactors project state to store the information about all currently running device sessions in ProjectState. This refactor is necessary to implement running multiple devices simultaneously. 

### How was this tested:
- verify there are no regressions:
  - try switching between devices
  - try adding/removing devices
  - try removing the running device
  - try renaming devices
- to test that multiple device sessions are correctly reported, pass `preservePreviousDevice` to `startOrActivateSessionForDevice` and log the project state to check all sessions are present

### DIFF
--- a/packages/vscode-extension/src/project/DeviceSessionsManager.ts
+++ b/packages/vscode-extension/src/project/DeviceSessionsManager.ts
@@ -86,7 +86,13 @@ export class DeviceSessionsManager implements Disposable, DeviceSessionsManagerI
     Logger.debug("Selected device is ready");
 
     const newDeviceSession = new DeviceSession(this.applicationContext, device, {
-      onStateChange: (state) => this.deviceSessionManagerDelegate.onDeviceSessionChange(state),
+      onStateChange: (state) => {
+        if (!this.deviceSessions.has(state.deviceInfo.id)) {
+          // NOTE: the device is being removed, we shouldn't report state updates
+          return;
+        }
+        this.deviceSessionManagerDelegate.onDeviceSessionChange(state);
+      },
     });
 
     this.deviceSessionManagerDelegate.onDeviceSessionStarted(newDeviceSession.getState());
@@ -168,11 +174,11 @@ export class DeviceSessionsManager implements Disposable, DeviceSessionsManagerI
   private async terminateSession(deviceId: string) {
     const session = this.deviceSessions.get(deviceId);
     if (session) {
-      this.deviceSessionManagerDelegate.onDeviceSessionStopped(session.getState());
-      this.deviceSessions.delete(deviceId);
       if (session === this.activeSession) {
         this.updateSelectedSession(undefined);
       }
+      this.deviceSessionManagerDelegate.onDeviceSessionStopped(session.getState());
+      this.deviceSessions.delete(deviceId);
       await session.dispose();
     }
   }
--- a/packages/vscode-extension/src/project/DeviceSessionsManager.ts
+++ b/packages/vscode-extension/src/project/DeviceSessionsManager.ts
@@ -9,14 +9,21 @@ import { AndroidEmulatorDevice } from "../devices/AndroidEmulatorDevice";
 import { IosSimulatorDevice } from "../devices/IosSimulatorDevice";
 import {
   DeviceSessionsManagerInterface,
-  DeviceSessionsManagerDelegate,
   ReloadAction,
   SelectDeviceOptions,
 } from "../common/DeviceSessionsManager";
 import { disposeAll } from "../utilities/disposables";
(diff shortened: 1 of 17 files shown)


packages/vscode-extension/src/project/project.ts
```diff
@@ -110,6 +111,33 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
+    };
+    this.updateProjectState({ deviceSessions: newDeviceSessions });
+  }
+
+  onDeviceSessionStopped(state: DeviceSessionState): void {
+    const deviceId = state.deviceInfo.id;
+    const newDeviceSessions = { ...this.projectState.deviceSessions };
+    delete newDeviceSessions[deviceId];
+    this.updateProjectState({ deviceSessions: newDeviceSessions });
+  }
```
jwajgelt: Maybe it'd be worth moving this wholesale to `DeviceSessionsManager` and having it send `{ selectedDeviceId, deviceSessions }` through `updateProjectState` instead.

jwajgelt: That approach makes more sense to me. Done.

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -37,6 +41,15 @@ export class DeviceSessionsManager implements Disposable, DeviceSessionsManagerI
     this.deviceManager.addListener("devicesChanged", this.devicesChangedListener);
   }
 
+  private get state(): DeviceSessionsManagerState {
+    return {
+      selectedSessionId: this.activeSession?.getState().deviceInfo.id ?? null,
+      deviceSessions: Object.fromEntries(
```
filip131311, line 47: this is counter intuitive to me that active session is not a getter that uses some internal filed "activeSessionId" or something it causes a wanky situations like that in which we need to understand the structure of deviceSession to get the id

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -313,7 +313,8 @@ export class DeviceSession
       this.isActive = true;
       this.toolsManager.activate();
       if (this.startupMessage === StartupMessage.AttachingDebugger) {
-        await this.reconnectJSDebuggerIfNeeded();
+        this.debugSession = new DebugSession(this, { useParentDebugSession: true });
```
filip131311, line 316: why is it part of of this pr? what does it do?

jwajgelt: Without it, debugger overlay will break when you switch to an already-running device.
This is because `deactivate` will call `dispose` on the `DebugSession`, but later `activate` calls will try to reuse it (which is illegal 👮).

packages/vscode-extension/src/project/DeviceSessionsManager.ts
jwajgelt: Changed it.

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -147,24 +163,28 @@ export class DeviceSessionsManager implements Disposable, DeviceSessionsManagerI
...
-    const previousSession = this.activeSession;
-    this.activeSession = session;
-    if (previousSession !== session) {
-      previousSession?.deactivate();
-      session?.activate();
-      this.deviceSessionManagerDelegate.onActiveSessionStateChanged(
-        session?.getState() ?? DEVICE_SESSION_INITIAL_STATE
-      );
+    const previousSession = this.selectedDeviceSession;
+    this.activeSessionId = session?.getState().deviceInfo.id;
```
filip131311, line 167: after the last change it seems to me that it would be reasonable to chege the argument of this function to id as well for example:

```js
    const existingDeviceSession = this.deviceSessions.get(deviceInfo.id);
    if (existingDeviceSession) {
      this.updateSelectedSession(existingDeviceSession);
```

this code in deviace session manger seems unnecessery

jwajgelt: I prefer it the way it is.
It makes more sense to me to pass the session here instead of having this function get it by id (and somehow handle the case where no session with that id exists).

> this code in deviace session manger seems unnecessery

I don't know which code you're referring to, but if you mean the code block you quoted, we check if the session already exists by retrieving it from the map since we want to handle the cases where a session already exists and the case where it doesn't differently -- so even if we passed just the id, we'd still have to look up if `deviceSessions.has(deviceInfo.id)` (and then check it again inside `updateSelectedSession`)

filip131311 (approved):

PR #1219, opened 2025-06-05
Add an option to remove navigation history entries

HexHyperion:
This PR improves the functionality of the `UrlSelect` by allowing the user to remove unwanted entries (e.g. with typos) from the navigation history, since we don't clear it after reload anymore.

If the removed item is a static route present in the `routeList` extracted from Expo Router, it will be moved down to the alphabetically-sorted "other" section, and all other items will just be removed. The currently selected route can't be removed until another route is opened, as this would break the history.

https://github.com/user-attachments/assets/35c745cb-83af-48dc-8eb1-172c1bb493fc

### How Has This Been Tested: 
1. Open an Expo Router app
2. Play with navigation, open some static, dynamic and incorrect routes
3. Open the UrlSelect, there should be an "X" next to each visited entry
4. Click the "X"-es on the items - they should disappear or move down in the suggestions
5. Refresh the app - the deleted items shouldn't show up until visited again

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -38,6 +38,7 @@ export type ProfilingState = "stopped" | "profiling" | "saving";
 export type NavigationHistoryItem = {
   displayName: string;
   id: string;
+  removable?: boolean;
 };
 
 export type NavigationRoute = {
@@ -207,6 +208,7 @@ export interface ProjectInterface {
   openNavigation(navigationItemID: string): Promise<void>;
   navigateBack(): Promise<void>;
   navigateHome(): Promise<void>;
+  removeNavigationHistoryEntry(id: string): Promise<void>;
   openDevMenu(): Promise<void>;
 
   activateLicense(activationKey: string): Promise<ActivateDeviceResult>;
--- a/packages/vscode-extension/src/webview/components/UrlSelect.tsx
+++ b/packages/vscode-extension/src/webview/components/UrlSelect.tsx
@@ -56,6 +56,10 @@ function UrlSelect({
     return itemForID.displayName;
   };
 
+  const removeHistoryEntry = (id: string) => {
+    project.removeNavigationHistoryEntry(id);
+  };
+
   const findDynamicSegments = (item: NavigationHistoryItem) => {
     const matchingRoute = routeList.find((route) => route.path === item.id);
     if (matchingRoute && matchingRoute.dynamic) {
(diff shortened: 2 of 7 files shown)


packages/vscode-extension/src/common/Project.ts
```diff
@@ -38,6 +38,7 @@ export type ProfilingState = "stopped" | "profiling" | "saving";
 export type NavigationHistoryItem = {
   displayName: string;
   id: string;
+  removable?: boolean;
```
jwajgelt: Does it need to be here? It looks to me like it's only used on frontend side, and you could probably just pass it as a separate prop to `UrlSelectItem`

packages/vscode-extension/src/webview/components/UrlSelectItemGroup.tsx
```diff
@@ -46,6 +48,7 @@ function UrlSelectItemGroup({
               itemList={itemList}
               textfieldRef={textfieldRef}
               noHighlight={noHighlight}
+              onRemove={onRemove}
```
jwajgelt, line 51: Re. the first comment, pass `removable` to `UrlSelectItemGroup` (since one group has removable elements, and one does not, right?), and then here pass
```
removable={removable && index > 0}
```
to `UrlSelectItem`

packages/vscode-extension/src/webview/components/UrlSelect.tsx
```diff
@@ -132,16 +136,23 @@ function UrlSelect({
...
+      ...item,
+      removable: index !== 0,
+    }));
     const routesNotInHistory = differenceBy(
       routeItems,
       navigationHistory,
       (item: NavigationHistoryItem) => item.displayName
     );
-    return [...navigationHistory, ...routesNotInHistory];
+    return [...navigationHistoryWithRemovable, ...routesNotInHistory];
```
jwajgelt, line 159: Alternatively, you could type this as `NavigationHistoryItem & { removable?: boolean }`

packages/vscode-extension/src/webview/components/UrlSelectItemGroup.tsx
HexHyperion: > since one group has removable elements, and one does not, right?

Well, not exactly, because we want the "X"-es only on `navigationHistory` items, while the `UrlSelectItemGroup` relies on a `filtered`/`filteredOut` division, so each of those can have both removable and non-removable items...

I removed the field from the common type and added a local one, but not sure if it's better...

jwajgelt (approved):

PR #1221, opened 2025-06-05
fix Network Inspector crash caused by invalid request URLs

### LINKED ISSUE #1215: Network inspector unusable after clicking on a pending/failed request
**Description**
1. Open the network inspector
2. Run a network request that takes time to load or fails

Result:
1. Failed request don't seem to be updated and are left in pending status
Image
2. Clicking on one of the pending requests results in a blank panel, only fixable by disabling and enabling network inspector
Image

**Environment**

- Radon IDE v1.7.1
- Expo 52 / React native 0.76.6
- Axios 1.8.2

jwajgelt:
Currently, if a request with a malformed URL is made, the Network Inspector will crash when you try to open the request details.
This PR safeguards against that.

Fixes #1215

### How Has This Been Tested: 
- open the network inspector
- make a request to an invalid URL
- open the details of the request
- don't crash :)

### DIFF
--- a/packages/vscode-extension/src/network/components/PayloadTab.tsx
+++ b/packages/vscode-extension/src/network/components/PayloadTab.tsx
@@ -5,12 +5,16 @@ interface PayloadTabProps {
 }
 
 const getParams = (url: string): Record<string, string> => {
-  const urlObj = new URL(url);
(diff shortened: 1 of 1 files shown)


filip131311 (approved):
👍

PR #1223, opened 2025-06-05
Add a reliable Expo Router availability check for the frontend and fix navbar for multi-app projects

HexHyperion:
This PR introduces a new `isExpoRouterProject` field in the `ProjectProvider` context value, which indicates if the currently running app uses Expo Router for navigation, and updates each time the `appRoot` changes.

This allows the webview components to reliably depend on Expo Router availability, especially in monorepo projects, where some apps can use different navigation solutions.

The change was motivated mainly by the navigation bar breaking if the `appRoot` was switched from an Expo Router app to a non-Expo one and vice versa, as the dependency check was performed only on component mount.

### How Has This Been Tested: 
1. Open a folder with more than one React Native app, e.g. [radon-ide-test-apps](https://github.com/software-mansion-labs/radon-ide-test-apps)
2. Launch the IDE panel and open an app with Expo Router - the navigation bar has full functionality
3. Change the app to a bare RN one with the `AppRootSelect` - the navbar switches to dropdown mode and disables the back button
4. Switch back to the ER app - appropriate functions are enabled again

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -204,11 +204,13 @@ export interface ProjectInterface {
   focusBuildOutput(): Promise<void>;
   focusExtensionLogsOutput(): Promise<void>;
   focusDebugConsole(): Promise<void>;
+  
+  usesExpoRouter(): Promise<boolean>;
   openNavigation(navigationItemID: string): Promise<void>;
   navigateBack(): Promise<void>;
   navigateHome(): Promise<void>;
   openDevMenu(): Promise<void>;
-
+  
   activateLicense(activationKey: string): Promise<ActivateDeviceResult>;
   hasActiveLicense(): Promise<boolean>;
 
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -204,13 +204,13 @@ export interface ProjectInterface {
   focusBuildOutput(): Promise<void>;
   focusExtensionLogsOutput(): Promise<void>;
   focusDebugConsole(): Promise<void>;
-  
+
   usesExpoRouter(): Promise<boolean>;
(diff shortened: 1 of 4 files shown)


jwajgelt (commented):
I think it'd be better to refresh the dependencies when the appRoot instead, so that the `useDependencies` hook reruns and the `UrlBar` component rerenders with new `dependencies` when the appRoot changes.

jwajgelt:
> I think it'd be better to refresh the dependencies when the appRoot instead, so that the `useDependencies` hook reruns and the `UrlBar` component rerenders with new `dependencies` when the appRoot changes.

Right now, the `DepencencyManager` that the webview calls is bound in `WebviewController` once, when the webview is first created, but whenever the appRoot changes, `Project` creates a new `DependencyManager` with a new appRoot, which is why the webview doesn't see these changes to dependencies.
This means that wherever we may want to use the `DependencyManager` proxy in the webview we'll have the same issue in monorepo setups.
That's why I think that we should solve the expo router issue this way, and instead try to fix how the `dependencies` are passed to the UI.

I think there are two approaches to fix this problem:
- don't create a new `DependencyManager` when the appRoot changes, and instead change its `appRootFolder` property and rerun checks. I feel like there may be some unexpected consequences to that approach, but maybe it would work just fine. (I also kind of disagree with this on principle, since having an "immutable" `DependencyManager` makes the code easier to reason about, but I digress.)
- wait for @filip131311's state management refactor. If we move all of `dependencies` into the shared state, we should get this for free since we'll be able to remove the proxied objects wholesale (or maybe not, depends on how that refactor handles function calls on proxies -- in that case we'd still need to fix them so that they're called on the new `DependencyManager` object after the appRoot changes.)

HexHyperion:
Is [this](https://github.com/software-mansion/radon-ide/pull/1226) approach better?
Or we can wait for the state refactor as you say.

HexHyperion:
Done better in #1226

PR #1229, opened 2025-06-06
Move the home and deeplink buttons inside the UrlSelect, unify the dropdown layout for non/ER apps and fix navigationChanged events

HexHyperion:
This PR improves the layout of the navigation bar by moving the "home" and "open deeplink" buttons to the bottom of the `UrlSelect` dropdown, giving more space to the icons on the top toolbar, and matching the label styles to those of other selectors in the UI (not very visible in this theme, but they _are_ different):

For Expo Router:
Zrzut ekranu 2025-06-6 o 12 24 44

Without Expo Router:
Zrzut ekranu 2025-06-6 o 12 26 30

It also unifies the layout and behavior of the dropdown to avoid unnecessary condition checking for non-Expo apps and fixes a small bug in the Expo Router plugin where the navigation descriptors were compared shallowly, allowing some events to slip through the equality check.

### How Has This Been Tested: 
1. Open an app
2. Click on the URL bar - the buttons are at the bottom, and stay visible even with the list overflowing
3. Try to focus the buttons using arrow keys - they should get focused properly and be selectable with Enter
4. Test the functionality of the home button, deeplink button and the navigation history - everything should still work properly
5. Open a preview and confirm if the path in the bar is correct
6. Test this for both Expo and non-Expo apps


packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -171,12 +170,6 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
               </DropdownMenu.SubContent>
             </DropdownMenu.Portal>
           </DropdownMenu.Sub>
-          <DropdownMenu.Item
-            className="dropdown-menu-item"
-            onSelect={() => openModal("Open Deep Link", <OpenDeepLinkView />)}>
-            <span className="codicon codicon-link" />
-            Open Deep Link
-          </DropdownMenu.Item>
```
jwajgelt: NIT:
I think we can keep it here, for people who already use it and are used to it being here.
That said, removing it is also fine.

PR #1231, opened 2025-06-09
Allow running devices in the background

jwajgelt:
Allow the user to leave devices running in the background, to enable switching quickly between them without having to wait for the device and application to load.

This PR adds UI elements to allow for managing running devices:
- "Select device" dropdown now displays a separate "Running devices" section when >1 device is running.
image

- an indicator next to the "Select device" dropdown showing the number of devices running in the background (if any)
image

- In the "Manage Devices" modal, a "Shut down devices when switching" toggle was added to restore the previous behaviour if a user doesn't need devices to run in the background
- In the "Manage Devices" modal running devices have a "stop" button to shut down a single device.
image

- In the "Manage Devices" modal you can now press the device entry to select it (which saves us from adding too many buttons for devices running in the background, after adding the "stop" button).

### How Has This Been Tested: 
- open an app in Radon
- go to "Manage Devices" and uncheck "Shut down devices when switching"
- select a different device
- verify the device switches correctly and the app starts
- select the previous device
- verify both running devices are shown in the "Running devices" section in the device selector
- verify the device switches and the app appears immediately, without having to boot/build/install/etc.
- check that stopping the device from "Manage devices" works correctly
- check that checking "Shut down devices when switching" restores the previous behaviour

### DIFF
--- a/packages/vscode-extension/src/webview/components/DeviceSelect.css
+++ b/packages/vscode-extension/src/webview/components/DeviceSelect.css
@@ -110,7 +110,6 @@
   border: 1px solid var(--swm-button-counter-border);
   padding: 2px 4px;
   font-size: 10px;
-  transform: translate(-50%, -75%);
   transition:
     left 200ms ease-in-out,
     opacity 200ms ease-in-out;
--- a/packages/vscode-extension/src/webview/components/DeviceSelect.tsx
+++ b/packages/vscode-extension/src/webview/components/DeviceSelect.tsx
@@ -128,12 +128,12 @@ function DeviceSelect() {
           <div className="device-select-value">
             <span className="codicon codicon-device-mobile" />
             <span className="device-select-value-text">{text}</span>
+            {backgroundDeviceCounter > 0 && (
+              <span className="device-select-counter">+{backgroundDeviceCounter}</span>
+            )}
           </div>
         </Select.Value>
       </Select.Trigger>
-      {backgroundDeviceCounter > 0 && (
-        <span className="device-select-counter">+{backgroundDeviceCounter}</span>
-      )}
 
       <Select.Portal>
         <Select.Content


packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -41,6 +41,10 @@ export class DeviceSessionsManager implements Disposable, DeviceSessionsManagerI
     return this.activeSessionId ? this.deviceSessions.get(this.activeSessionId) : undefined;
   }
 
+  public async stopSession(deviceId: DeviceId): Promise<void> {
```
filip131311: I wonder if we need to add a new method instead of just exposing `terminateSession` in the interface, if what you mean by stop is terminate then you should just use it. 

More over the consumer of this interface could think that "stop"refers to something like "freeze" tho i don't think it is likely

packages/vscode-extension/package.json
```diff
@@ -195,6 +195,12 @@
           "scope": "window",
           "default": null,
           "description": "Files matching this pattern fill be excluded for the inspector jump to file functionality. This can be used if your codebase has some design system primitives that are used everywhere in your codebase and that you don't want to always get opened when using inspect functionality."
+        },
+        "RadonIDE.preservePreviousDevice": {
+          "type": "string",
+          "scope": "window",
+          "default": true,
+          "description": "When set, switching between devices will keep the previous device running in the background instead of shutting it down."
```
kmagiera: it is counterintuitive that we have a reversed option in the UI where "setting" means the device will be terminated

jwajgelt:
> haven't reviewed the code, only left one inline comment, but wanted to also comment on the UI bits.
> 
>     1. I think the switch in "manage devices" should be on top. it doesn't feel right where it is placed right now.
> 

It didn't feel right to me at the top. This is meant to be the same as what we already have in the "Open Deep Link", where the toggle controlling the behavior is below the list. I don't think it's very important, someone with better UI/UX sensibilities can revisit it in the future if this is problematic for some reason.

>     2. We should have some limit (configurable) for the number of running devices. Especially for Android, where devices are much more resource intensive. By default we should limit to 1 Android and maybe 2/3 iOS devices?

We can do that as a separate PR if we see it being a problem. 
I wouldn't expect people to create a lot of devices in the first place, so this shouldn't be an issue.
Historically we've had issues with discoverability of various settings we provide. I think it'd be confusing for the user if they could start 2/3 iOS devices, but starting a second Android device would kill the previous one automatically, and I doubt they'd go looking through the settings to see if there's some default limit. I'd keep it as is.

>     3. Would be good to add a command / keybinding that'd switch between running devices.

Probably, but it's a nontrivial amount of work I'd keep out of this PR. We can add it separately without holding this feature back.
There's also the issue that we don't have a specific order that devices are listed in, so just adding a "Next/Previous Device" shortcut might cause the switching to happen in a confusing way, unless we change the existing code to make sure the order in the UI is the same as in the extension (or handle the command in the UI code, which has some different issues). 
I doubt the usefulness of such command though, I'd this "Next/Previous _Running_ Device" would probably be more useful (and in that case, we can probably just use the order returned by `Object.keys`, since it should be the same on both sides).
I could also see a "Select Device" command which would let you pick the device in the command palette, but that doesn't lend itself to a shortcut either way.

packages/vscode-extension/package.json
jwajgelt: I kept it that way because that's what the existing function on `DeviceSessionsManager` accepted, but you're right. I flipped the meaning of the flag to match the UI.

jwajgelt:
>     2. Add some visual indicator in the device switcher (not just when it is open but also the closed one) that'd indicate there are more devices apart from the current one.
> 

I added an indicator to the device switcher when there are devices running in the background
image

We can add a warning when some threshold of devices is crossed, but I don't want to add any more things to this PR.

Issue #1203, opened 2025-06-03
Localization setting doesn't work on android emulator


anna-malina:
**Description**
While testing the sample app to display location and localization data I have observed that in Radon IDE after opening Device Settings and changing Localization value - this action reflects on application only for iOS devices, for android this doesn't work. It is described in docs, but this is not a user friendly path.

**Preconditions:**

1. Radon IDE extension is available and launched for the expo application project, which displays  location and localization data
2. Permissions are granted 

**Steps to Reproduce:**

1. Run app on android device
2. Click on device settings / Localization
3. Select some option and confirm it
4. Wait for app rebuild and load

**Actual Behavior:**
After loading the app - still the old (default) localization is displayed.
It works fine for iOS devices.

<img width="819" alt="Image" src="https://github.com/user-attachments/assets/72aceff1-7c1b-496a-a465-020b21da86c4" />

**Expected Behavior:**
If some option is not yet supported on android device - it should be hidden from device settings options and displayed only when using iOS devices.
If there is a option for quick fix for android - please also update a limitation (remove it) from docs.

**Environment**
Version of used Radon IDE: 1.7.1
Version of React Native: 0.76.7
Version of Expo: 53.0.9

xk0fe:
There is a bug where user cannot change localization on new emulators

PR #1243, opened 2025-06-12
add option to select camera output for android emulator

### LINKED ISSUE #1242: Change camera output on Android emulator
**Description**
Using Device Manager in Android Studio it's possible to change Camera output to:
- VirtualScene
- Emulated
- Webcamera

I could not find this feature in Radon's device manager. 

Image
Screenshot from Android Studio Device Manager

**Environment**

- Version of currently used Radon IDE: 1.7.1
- Version of React Native: 0.79

xk0fe:
Adds an option to the device dropdown menu to select camera. Works only for Android

Fixes #1242 

### How Has This Been Tested: 

No testing, could not find clear instructions on how to build. Tried, failed. Requires testing

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -15,6 +15,10 @@ export type DeviceSettings = {
   locale: Locale;
   replaysEnabled: boolean;
   showTouches: boolean;
+  camera: {
+    back: "emulated" | "none" | "webcam0" | "virtualscene";
+    front: "emulated" | "none" | "webcam0";
+  };
 };
 
 export type ToolState = {
--- a/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
@@ -90,77 +90,16 @@ export class AndroidEmulatorDevice extends DeviceBase {
     }, DISPOSE_TIMEOUT);
   }
 
-  private async shouldUpdateLocale(newLocale: Locale) {
-    const locale = newLocale.replace("_", "-");
-    const { stdout } = await exec(ADB_PATH, [
-      "-s",
-      this.serial!,
-      "shell",
-      "settings",
-      "get",
-      "system",
-      "system_locales",
-    ]);
-
-    // if user did not use the device before it might not have system_locales property
-    // as en-US is the default locale, used by the system, when no setting is provided
-    // we assume that no value in stdout is the same as en-US on some devices
-    // stdout is a string "null" instead of undefined so we need to handle it separately
-    if ((stdout ?? "en-US") === locale || stdout === "null") {
-      return false;
-    }
-    return true;
-  }
-
-  /**
-   * This method changes device locale by modifying system_locales system setting.
-   */
-  private async changeLocale(newLocale: Locale): Promise<void> {
-    const locale = newLocale.replace("_", "-");
-
-    await exec(ADB_PATH, [
-      "-s",
-      this.serial!,
-      "shell",
-      "settings",
-      "put",
-      "system",
-      "system_locales",
-      locale,
-    ]);
-
-    // this is needed to make sure that changes will persist
-    await exec(ADB_PATH, ["-s", this.serial!, "shell", "sync"]);
-
-    // TODO:  Find a way to change or remove persist.sys.locale without root access. note: removing the whole
-    // data/property/persistent_properties file would also work as we persist device settings globally in Radon IDE
-    const { stdout } = await exec(ADB_PATH, [
-      "-s",
-      this.serial!,
-      "shell",
-      "getprop",
-      "persist.sys.locale",
-    ]);
-
-    if (stdout) {
-      Logger.warn(
-        "Updating locale will not take effect as the device has altered locale via system settings which always takes precedence over the device setting the IDE uses."
-      );
-    }
-  }
-
   async changeSettings(settings: DeviceSettings) {
-    let shouldRestart = false;
-    if (await this.shouldUpdateLocale(settings.locale)) {
-      shouldRestart = true;
-      await this.changeLocale(settings.locale);
-    }
+    // Create setting appliers for configuration that requires boot to take effect
+    const settingAppliers: SettingApplier[] = [
+      new LocaleSettingApplier(settings.locale, this.serial),
+      new CameraSettingApplier(settings.camera),
+    ];
 
-    if (await this.shouldUpdateCameraSettings(settings.camera)) {
-      shouldRestart = true;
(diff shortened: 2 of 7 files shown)


packages/vscode-extension/src/common/Project.ts
```diff
@@ -15,6 +15,10 @@ export type DeviceSettings = {
   locale: Locale;
   replaysEnabled: boolean;
   showTouches: boolean;
+  camera: {
+    back: "emulated" | "none" | "webcam0" | "virtualscene";
```
filip131311, line 19: I have never used emulated cameras on android, so it might be an obvious question, but what is `virtualscene` and more importantly why is it only available on the frontend camera?

packages/vscode-extension/src/common/Project.ts
```diff
@@ -15,6 +15,10 @@ export type DeviceSettings = {
   locale: Locale;
   replaysEnabled: boolean;
   showTouches: boolean;
+  camera: {
```
filip131311, line 18: If you change settings type and later on assume they are set you should modify the default settings as well

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -206,6 +211,56 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+        }
+      }
+      
+      const updatedContent = lines.join("\n");
+      await fs.promises.writeFile(configIni, updatedContent, "utf-8");
+    } catch (e) {
+      Logger.error("Failed to update AVD camera settings", e);
+      throw new Error("Failed to update camera settings");
+    }
+  }
```
filip131311: This logic conforms to the previous design we used for:
```js
 if (await this.shouldUpdateLocale(settings.locale)) {
      shouldRestart = true;
      await this.changeLocale(settings.locale);
    }
```

and this is nice but it causes a file system operation to be performed twice `fs.promises.readFile(configIni, "utf-8");` 

I belive that we should have `maybeUpdateConfigurationSetting` method (maybe with a better name) that checks if uodate is needed and performs it in one sweep, it would also be best if it is able to update/check multiple settings at once as i believe that it is the way will want to support screen rotation as well

what do you think?

filip131311 (changes requested):
Hello, thank you so much for you contribution!!!!!! 

I left some inline comments/ questions

filip131311:
And here is [documentation](https://ide.swmansion.com/docs/guides/development) on how to build the project. 

Please let us know if you were succesful in building development environment

packages/vscode-extension/src/common/Project.ts
xk0fe: VirtualScene is a camera feed that is used for ARCore, developers can provide their own images through it
Here's link to the [documentation](https://developer.android.com/studio/run/emulator-use-camera#arcore)

There is no clear indication as to why it's not being used on front camera. I've decided to include these camera simply to mirror Virtual Device Configuration options.

image

image

xk0fe: We can only guess why it's like this on AVD, but it's possible that when VirtualScene was initially added there was no support for front camera on ARCore side. Later, when support was added, AVD was left unchanged

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
xk0fe: Does this mean that rotation changes will require reboot? 

I've also noticed that `changeSettings` returns `shouldRestart` which could be confusing. Ideally, it should return `true` if any setting were changed.

xk0fe:
I was able to run extension in debug mode following the documentation. My changed are ready for the review

xk0fe:
Here's video preview for the code provided


https://github.com/user-attachments/assets/cdf6db2c-b011-4979-83bc-30b650fd493e

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -90,72 +90,16 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
-      "get",
-      "system",
-      "system_locales",
-    ]);
-
-    // if user did not use the device before it might not have system_locales property
-    // as en-US is the default locale, used by the system, when no setting is provided
-    // we assume that no value in stdout is the same as en-US on some devices
-    // stdout is a string "null" instead of undefined so we need to handle it separately
-    if ((stdout ?? "en-US") === locale || stdout === "null") {
```
xk0fe, line 109: This logic contains a bug where users with new emulators cannot change their language

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -982,3 +967,148 @@ function getNativeQemuArch() {
...
+      "system",
+      "system_locales",
+    ]);
+
+    // if user did not use the device before it might not have system_locales property
+    // as en-US is the default locale, used by the system, when no setting is provided
+    // we assume that no value in stdout is the same as en-US on some devices
+    // stdout is a string "null" instead of undefined so we need to handle it separately
+    const currentLocale = (stdout === "null" || stdout === undefined) ? "en-US" : stdout;
+    return currentLocale !== newLocale;
```
xk0fe, line 1012: Partially fixes #1203 

There is another issue however. Changes are not applied after first reboot, so I had to reboot my device twice for the language settings to be applied. 
I believe this change is not related to the code changes and deserves it's own PR

Here's video of the bug:
1) Change locale & wait for reboot
2) Notice that locale does not change
3) Change locale again & wait for reboot
5) Locale changes

https://github.com/user-attachments/assets/e199fd19-63d0-4ed3-b040-ee1282a6d56a

packages/vscode-extension/src/webview/views/CameraSettingsView.css
```diff
@@ -0,0 +1,106 @@
...
+  justify-content: flex-end;
+}
+
+/* Confirmation view styles */
+.camera-change-wrapper {
+  padding: 20px;
+  display: flex;
+  flex-direction: column;
+  gap: 16px;
+  min-width: 400px;
```
jwajgelt, line 65: I think `400px` is too much here.
image
The modal of this size could easily fit all of the information we want, but we still get a scroll bar and parts of the text are hidden, because we set a large `min-width`.
Also, I don't think the action buttons should ever be hidden, they should shrink with the size of the modal.

packages/vscode-extension/src/common/Project.ts
jwajgelt, line 18: Since we only use this on Android, I think this should be optional (or otherwise typed so that we can leave this unset on iOS, instead of setting it to some arbitrary value that we never use, like we do now.)

PR #1245, opened 2025-06-12
Fix webview showing crashed preview

filip131311:
This PR solves an issue with with webview showing crashed preview state in the following scenario: 

1) run application 
2) when on "waiting for an app to load" screen click on the device to show preview
3) when app loads use restart option for example "clean metro cache" 
4) now you should see crashed preview 

this was happening because we changed the device status only when the build started and not limitedly after restart this pr fixes that 

### How Has This Been Tested: 

run reproduction steps on `expo-53`


jwajgelt (approved):
I can see no issues that this could potentially cause.

PR #1247, opened 2025-06-12
add keyboard shortcuts for quickly switching running devices

jwajgelt:
Adds vscode commands and keyboard shortcuts for quickly switching between running devices.
- "Cmd + Shift + ]" switches to the next running device
- "Cmd + Shift + [" switches to the previous running device

### Known issues:
- holding the shortcut causes the devices to be switched rapidly, extra debug consoles appear (#1254)

### How Has This Been Tested: 
- open a project in radon
- launch multiple devices at the same time
- check the keyboard shortcuts work

### DIFF
--- a/packages/vscode-extension/src/project/DeviceSessionsManager.ts
+++ b/packages/vscode-extension/src/project/DeviceSessionsManager.ts
@@ -17,7 +17,7 @@ import { disposeAll } from "../utilities/disposables";
 import { DeviceId, DeviceSessionsManagerState } from "../common/Project";
 
 const LAST_SELECTED_DEVICE_KEY = "last_selected_device";
-const SWITCH_DEVICE_THROTTLE_MS = 500;
+const SWITCH_DEVICE_THROTTLE_MS = 300;
 
 export type DeviceSessionsManagerDelegate = {
   onInitialized(): void;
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -16,6 +17,7 @@ import { disposeAll } from "../utilities/disposables";
 import { DeviceId, DeviceSessionsManagerState } from "../common/Project";
 
 const LAST_SELECTED_DEVICE_KEY = "last_selected_device";
+const SWITCH_DEVICE_THROTTLE_MS = 300;
```
filip131311, line 20: This time is either too small or the whole aproch needs a refactor because we still loose connections to debug sessions when holding the shourtcut
Screenshot 2025-06-13 at 09 28 15

filip131311 (changes requested):
I left expolnation inline

packages/vscode-extension/src/project/DeviceSessionsManager.ts
filip131311, line 20:
```suggestion
const SWITCH_DEVICE_THROTTLE_MS = 500;
```

after incriseing the timeout I'm unable to cause the problem using the shortcut

filip131311 (approved):
I aprove this changes as I can see that the problem with "run away" debug sessions will probably not affect many users if any, but please document the problem in some way before marging

PR #1249, opened 2025-06-13
Support for react native 80 

filip131311:
This PR adds a support for react native 80 by doing following things: 
1) Update React version support to 19.1, this requires rebuilding renderers and `jsx-runtime` from [this commit](https://github.com/software-mansion-labs/react-radon-ide/commit/0ab8b16ddd2892b8e8dafc10475900dfb5966a4e)
2) change the source of `AppRegistry` core component 
3) initialize `parseErrorStack` before it is passed into the consoleWrapper to avoid infinite loops 
4) override the `warn-on-deep-imports` babel plugin that was added into the RN 80 to warn users that deep inports from react native are deprecated, as the warnings seem useful we don't want to remove them completely with https://reactnative.dev/blog/2025/06/12/moving-towards-a-stable-javascript-api, so we create our onw plug in that ignores any imports made by a file in our `lib`  directory

### How Has This Been Tested: 
- run and play with a new test app added in [this commit](https://github.com/software-mansion-labs/radon-ide-test-apps/commit/809b17e70cb9c5c5a18b47e08c1a3632eeb5f677)

### DIFF
--- a/packages/vscode-extension/lib/JSXRuntime/react-native-80/react-jsx-dev-runtime.development.js
+++ b/packages/vscode-extension/lib/JSXRuntime/react-native-80/react-jsx-dev-runtime.development.js
@@ -1,0 +1,355 @@
+/**
+ * @license React
+ * react-jsx-dev-runtime.development.js
+ *
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+"use strict";
+"production" !== process.env.NODE_ENV &&
+  (function () {
+    function getComponentNameFromType(type) {
+      if (null == type) return null;
+      if ("function" === typeof type)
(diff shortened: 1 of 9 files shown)


packages/vscode-extension/lib/JSXRuntime/react-native-80/react-jsx-dev-runtime.development.js
jwajgelt, line 1: You made some changes to this file when compared to the upstream version, right? Could you annotate that with some comments, like
```ts
// RADON PATCH BEGIN
... changes ...
// RADON PATCH END
```
so that we can identify these sections easily?

filip131311: This information is available [here](https://github.com/software-mansion-labs/react-radon-ide/commit/0ab8b16ddd2892b8e8dafc10475900dfb5966a4e) and is linked in the readme I belive it is more useful to inspect source code changes

PR #1252, opened 2025-06-16
fix "shut down devices" toggle not responding on first click

jwajgelt:
When toggling the "Shut down devices when switching" switch  for the first time, it would not visually change until a second press.
This was due to a typo, where the config reported by the `"configChange"` event used the value initially loaded by the extension instead of the current one.

### How Has This Been Tested: 
- go to "Manage Devices"
- toggle the switch
- the switch should visually toggle
- the behaviour should match the visual state of the switch


filip131311 (approved):
👍

PR #1253, opened 2025-06-16
warn the user when many devices are running concurrently

jwajgelt:
Adds a warning notification when more than a certain number of emulators (>3 iOS or >1 Android) are running at the same time.

### How Has This Been Tested: 
- open Radon
- launch 2 Android devices
- a warning should appear
- stop one of the devices
- open 4 iOS devices
- a warning should appear

### DIFF
--- a/packages/vscode-extension/src/project/DeviceSessionsManager.ts
+++ b/packages/vscode-extension/src/project/DeviceSessionsManager.ts
@@ -130,6 +130,7 @@ export class DeviceSessionsManager implements Disposable, DeviceSessionsManagerI
 
     this.deviceSessionManagerDelegate.onDeviceSessionsManagerStateChange(this.state);
     this.deviceSessions.set(deviceInfo.id, newDeviceSession);
+    this.maybeWarnAboutRunningDevices();
     this.updateSelectedSession(newDeviceSession);
     this.deviceSessionManagerDelegate.onInitialized();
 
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -22,6 +23,9 @@ export type DeviceSessionsManagerDelegate = {
   onDeviceSessionsManagerStateChange(state: DeviceSessionsManagerState): void;
 };
 
+const MAX_ALLOWED_IOS_DEVICES = 3;
```
filip131311, line 27: I don't like the name `ALLOWED` as it suggests that it is not possible to have more then that

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -130,6 +135,36 @@ export class DeviceSessionsManager implements Disposable, DeviceSessionsManagerI
...
+    const [iosDevices, androidDevices] = _.partition(
+      this.deviceSessions.values().toArray(),
+      (session) => session.getState().deviceInfo.platform === DevicePlatform.IOS
+    );
+
+    if (
+      iosDevices.length > MAX_ALLOWED_IOS_DEVICES ||
+      androidDevices.length > MAX_ALLOWED_ANDROID_DEVICES
+    ) {
+      window
```
filip131311, line 163: Shouldn't we have some king of timeout flag that prevents showing this multiple times in the row? if some one wants to destry their machine it might be quite annoying that we are warning them all the time about it

filip131311 (approved):
I left some inline comments but it seems a good addition <3

PR #1256, opened 2025-06-17
UI adjustments to device select dropdown when multiple devices are running

kmagiera:
This change adjusts the UI of the device select dropdown to provide better visual clues and also better ergonomics for some popular use-cases:

1. We no longer group running devices on top. When there are just a handful devices (2 or 3 devices in total), this doesn't look great.
2. We add a "running" badge on the dropdown, this gives a visual clue associated with the "+1" badge. Otherwise it is unclear what that "+1" badge would mean.
3. We make the badge clickable, such that users can stop devices from the menu directly. This is useful when you just switched the device, but don't want the keep the previous one running and you don't want to change the defaults / haven't discovered that option at all.

On top of implementing the above, we're making the following adjustments:
1. We're introducing new theme variable that is used to display the close button which should normally be red and contrast with the badge color.
2. We're modifying the hover effect for subtitle under device manager modal. Previously hovering would result in the text turning into multiline when long. Instead we use `title` to display full subtitle text as tooltip.

Screenshots:
image
image
image

### How Has This Been Tested: 
1. Start device and observe it displays a "running" badge
3. Start more devices and see "+1" and badges in the dropdown
4. Use the badge to stop devices
5. Check the close button in different themes and in Radon themes in particular.
6. Make sure the app selector look good as the modified RichSelectItem component is reused there


packages/vscode-extension/src/webview/components/shared/RichSelectItem.tsx
```diff
@@ -4,17 +4,39 @@ import "../DeviceSelect.css";
...
+        e.preventDefault();
+        e.stopPropagation();
+      }}
+      onClick={onStopClick}>
+      <Badge variant="activity-bar-counter" className="running-badge-button">
+        <span />
+      </Badge>
+    </div>
+  );
+}
```
jwajgelt: I'd move this to `ManageDevicesView` and pass it as a prop/child to `RichSelectItem` (since it's supposed to be generic)

kmagiera: good suggestion, just updated the code

PR #1257, opened 2025-06-17
only enable device switching shortcut when the Preview is focused

jwajgelt:
- sets the extension context whenever the Radon panel is focused/blurred
- only enables the device switching shortcut when the Radon panel is focused

### How Has This Been Tested: 
- start multiple devices
- focus on the Radon panel
- "Cmd+Shift+]" should switch between running devices
- focus on an editor with multiple open tabs
- "Cmd+Shift+]" should switch between editor tabs (unless unset)

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -121,13 +121,13 @@
         "command": "RNIDE.nextRunningDevice",
         "title": "Switch to next running device",
         "category": "Radon IDE",
-        "enablement": "RNIDE.extensionIsActive && RNIDE.panelIsOpen"
+        "enablement": "RNIDE.extensionIsActive && RNIDE.panelIsOpen && RNIDE.isPreviewFocused"
       },
       {
         "command": "RNIDE.previousRunningDevice",
         "title": "Switch to previous running device",
         "category": "Radon IDE",
-        "enablement": "RNIDE.extensionIsActive && RNIDE.panelIsOpen"
+        "enablement": "RNIDE.extensionIsActive && RNIDE.panelIsOpen && RNIDE.isPreviewFocused"
       }
     ],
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/package.json
```diff
@@ -121,13 +121,13 @@
         "command": "RNIDE.nextRunningDevice",
         "title": "Switch to next running device",
         "category": "Radon IDE",
-        "enablement": "RNIDE.extensionIsActive && RNIDE.panelIsOpen"
+        "enablement": "RNIDE.extensionIsActive && RNIDE.panelIsOpen && RNIDE.isPreviewFocused"
```
filip131311, line 124: is there a way to sdisable cust the keybinding and not the whole command?

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -198,7 +199,18 @@ function PreviewView() {
     .padStart(2, "0")}`;
 
   return (
-    <div className="panel-view">
+    <div
+      className="panel-view"
+      onFocus={(e) => {
+        vscode.postMessage({
```
filip131311, line 205: I don't like introducing communication with vscode without any interface, but i guess will change all of the communication soon anyway and its fine

jwajgelt: I don't like it either, but I thought it was better than the alternative of adding it to an existing class or introducing `WebviewController`-specific channel when it already controls the actual communication.

packages/vscode-extension/package.json
jwajgelt: There's a `"when"` property for the keybindings, but it doesn't seem to do what we want, sadly.

PR #1258, opened 2025-06-17
Fix Radon showing dead preview after bundle error and "Reload IDE"

filip131311:
This PR fixes an issue introduces after abstracting inspector bridge interface in #1180. In this PR we moved on from `addListener` `removeListener` pattern to `disposable` one. The new implementation assumed that the listeners array for one event would not change when they were called, which caused some of the listeners to be ignored if the ones defined before them were removed before the end of the iteration. 

This caused problems with `waitForAppReady` promisees not being resolved which caused debbuger not attaching and the problem with Reloading the ide. 

### How Has This Been Tested: 

I run the following reproduction steps on `expo-53` test app:
- run app 
- cause bundle error 
- click on realod button a few times 
- fix the bundle error 
- reload 
- use "Reboot IDE" option

### DIFF
--- a/packages/vscode-extension/src/project/bridge.ts
+++ b/packages/vscode-extension/src/project/bridge.ts
@@ -1,5 +1,6 @@
 import { Disposable } from "vscode";
 import { NavigationRoute } from "../common/Project";
+import { Logger } from "../Logger";
 
 export interface RadonInspectorBridgeEvents {
   appReady: [];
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/project/bridge.ts
```diff
@@ -38,7 +39,22 @@ export abstract class BaseInspectorBridge implements RadonInspectorBridge {
...
-    this.listeners.get(event)?.forEach((listener) => listener(payload));
+    const listeners = this.listeners.get(event);
+    if (!listeners) {
+      return;
+    }
+
+    // We need to clone the listeners array to avoid issues with concurrent modifications
+    // it is a common pattern to create listeners that dispose themselves which could lead to 
+    // issues if we modify the array while iterating over it.
+    const listenersIterable = [...listeners]
```
jwajgelt: NIT:
```suggestion
    const listenersCopy = [...listeners]
```

"Iterable" has a specific meaning in JS, I'd avoid using it here."
We're making a "copy", so I'd reflect that in the name.

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -68,8 +68,7 @@ export class DeviceBootError extends Error {
 }
 
 export class DeviceSession
-  implements Disposable, MetroDelegate, ToolsDelegate, DebugSessionDelegate, BuildManagerDelegate
-{
+  implements Disposable, MetroDelegate, ToolsDelegate, DebugSessionDelegate, BuildManagerDelegate {
```
jwajgelt: Did the formatting change for some reason?

PR #1261, opened 2025-06-17
[Radon AI] Rewrite to HTTP & merge with radon-chat

latekvo:
As of Cursor 1.0.0 release, support for `SSE` has been silently dropped.

This PR replaces the `SSE` transport with a `Streamable HTTP` transport.
This PR also merges the `mcp` and `chat` implementations, they now share parts of their code.


packages/vscode-extension/src/ai/mcp/toolRegistration.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export async function registerMcpTools(server: McpServer) {
+  server.registerTool(
+    "view_screenshot",
+    {
+      description: "Get a screenshot of the app development viewport.",
+      inputSchema: {},
+    },
+    screenshotToolDef
```
kmagiera: this is the executor of the tool not the def (perhaps "declaration" is the term). The "def" is the whole thing, so maybe that's what you want to move into the `toolDefinitions`.

Maybe it'd be better to have method like `registerWithServer` and then pass the server there and the method will pass all the necessary arguments.

packages/vscode-extension/src/ai/mcp/server.ts
```diff
@@ -0,0 +1,99 @@
...
+  app.use(express.json());
+
+  app.post("/mcp", async (req, res) => {
+    // Check for existing session ID
+    const sessionId = req.headers["mcp-session-id"] as string | undefined;
+    let transport: StreamableHTTPServerTransport;
+
+    if (sessionId && transports[sessionId]) {
+      // Recycle transport
+      transport = transports[sessionId];
```
kmagiera: would be good to verify that vscode and cursor actually send the session id with subsequent calls. Otherwise we could end up storing new sessions for every single request and never deleting them

packages/vscode-extension/src/ai/mcp/server.ts
```diff
@@ -0,0 +1,99 @@
...
+      });
+
+      // Clean up transport when closed
+      transport.onclose = () => {
+        if (transport.sessionId) {
+          delete transports[transport.sessionId];
+        }
+      };
+      const server = new McpServer({
+        name: "example-server",
```
kmagiera: should this be somethig else than example-server?

packages/vscode-extension/src/ai/mcp/server.ts
```diff
@@ -0,0 +1,99 @@
...
+    const transport = transports[sessionId];
+    await transport.handleRequest(req, res);
+  };
+
+  app.get("/mcp", handleSessionRequest);
+  app.delete("/mcp", handleSessionRequest);
+  return app;
+}
+
+export async function startLocalMcpServer(port: number) {
```
kmagiera: now that we have full control over the server, maybe we could just bind on ephemeral port instead. That is: we don't pass port number here, and pass `0` to `listen`. Then, once `listen` resolves we read the port from `server.address().port`

packages/vscode-extension/src/ai/mcp/server.ts
```diff
@@ -0,0 +1,99 @@
...
+  const server = getHttpServer();
+
+  Logger.info(`Starting local MCP server on port: ${port}`);
+
+  await new Promise<void>((resolve, reject) => {
+    try {
+      server.once("error", reject);
+      server.listen(port, () => {
```
kmagiera: we should pass `127.0.0.1` to listed after the port. This is important as otherwise it binds on all interfaces and can be potentially accessed from other devices on the network if someone has firewall disabled

kmagiera (changes requested):

packages/vscode-extension/src/ai/mcp/server.ts
latekvo: Sure, done in [f376268](https://github.com/software-mansion/radon-ide/pull/1261/commits/f376268c698cbc7d8787e16af59801a6ed4e32dd). 
Although I have to mention, `localhost` is already an alias for `127.0.0.1`. As far as I see, there isn't a way it could be accessed via network, because it is explicitly an alias for the loopback interface.

latekvo: Excuse my previous comment, I misunderstood what you said.
Done in [11f1e7b](https://github.com/software-mansion/radon-ide/pull/1261/commits/11f1e7bf66e28da5a52f70ac17d557a4895c0fbf).

packages/vscode-extension/src/ai/mcp/server.ts
latekvo: Fixed in [8803149](https://github.com/software-mansion/radon-ide/pull/1261/commits/8803149a6ea997d046babe35e5d0ef4905d5f95d)

packages/vscode-extension/src/ai/mcp/server.ts
latekvo: Done in [e940c88](https://github.com/software-mansion/radon-ide/pull/1261/commits/e940c887cc0549213a6ea585b4e95a25fb486e08).

packages/vscode-extension/src/ai/mcp/server.ts
latekvo: Limited open sessions to one in [ac96142](https://github.com/software-mansion/radon-ide/pull/1261/commits/ac9614234082d52e0c227e5fbbe3e83823dc1cb4)

PR #1266, opened 2025-06-23
@jwajgelt/select android camera output

### LINKED ISSUE #1242: Change camera output on Android emulator

jwajgelt:
Based on #1243, fixes #1242

Adds an option to the device dropdown menu to select camera. Works only for Android

https://github.com/user-attachments/assets/917553a7-32a9-4498-bdbe-ffa724a4d82c

### How Has This Been Tested: 
- open an application which uses the camera (tested on the appjs app)
- change the values in the device settings dropdown -> Camera Settings
- check that the camera display changes


packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -982,3 +987,157 @@ function getNativeQemuArch() {
...
 }
+
+interface SettingApplier {
+  shouldUpdate(configLines: string[]): Promise<boolean>;
+  /**
+   * Updates the runtime settings of the emulator.
+   * @param configLines - The lines of the config.ini file. We can modify them if needed, but we should not return them.
+   * @returns True if the runtime settings were changed, false otherwise.
+   */
+  updateRuntime(configLines: string[]): Promise<boolean>;
```
kmagiera: Is this even possible for emulator to automatically pick up the changes made in config.ini? I doesn't seem like we are event using it here, the `updateRuntime` implementation always returns false, so meybe there's no point in keeping this method here which also further complicates the implementation of `tryUpdateConfigurationSettings` for no reason?

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -982,3 +987,157 @@ function getNativeQemuArch() {
...
+  /**
+   * Updates the boot settings of the emulator.
+   * @param configLines - The lines of the config.ini file.
+   * @returns True if the boot settings were changed, false otherwise. The updated config lines are returned.
+   */
+  updateBoot(
+    configLines: string[]
+  ): Promise<{ bootChanged: boolean; updatedConfigLines: string[] }>;
```
kmagiera: this doesn't seem to be designed in a way to cover the existing two use cases. The locale settings don't require any updates to config.ini but when they change we'd still udpate the config file. Maybe we should consider a different abstraction that'd more closely match and track the current usecases.

kmagiera (changes requested):

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
jwajgelt: I agree, adding it here seems premature.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
jwajgelt: I think it's fine, though we could probably check if any changes were actually return before writing back to the file.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -982,3 +981,142 @@ function getNativeQemuArch() {
...
+        } else if (key === "hw.camera.front" && this.cameraSettings?.front !== undefined) {
+          configLines[i] = `hw.camera.front=${this.cameraSettings.front}`;
+          hasChanges = true;
+        }
+      }
+
+      // camera changes require boot to take effect
+      return { bootChanged: hasChanges, updatedConfigLines: configLines };
+    } catch (e) {
+      Logger.error("Failed to update AVD camera settings", e);
```
kmagiera: when can this even throw? We seem to be just doing some string concatenation & dict updates. Also, if any of that really throws, I don't think we should still be updating the config lines, this could mean that we'd updated the config partially which may not be ideal

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -203,7 +165,44 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
+          bootSettingsChanged = true;
+        }
+        if (updatedConfigLines !== undefined) {
+          configLines = updatedConfigLines;
+        }
+      }
+    }
+
+    const finalizedConfig = configLines.join("\n");
+    if (bootSettingsChanged && finalizedConfig !== configIniContent) {
```
kmagiera: the semantics of this new abstraction still confuses me. It'd seem more natural that the boot flag indicates whether we need to reboot after change, and the updatedConfig would indicate when we need to update the config. But this doesn't seem to be the case as we need to have both boot flag and the config flag set in order to apply the config updates.

Then, the implementation can actually return boot:false but also the list of updates, which are going to be discarded anyways.

kmagiera (approved):
Left a few more inline comments. I don't personally think we need this setting applier abstraction at this point given the update semantics is completely different (camera settings read and write config file, while the locale doesn). Will leave it up to you to decide whether you want to keep it, but regardless there are some inconsistencies in the update method that I'd like to be address before this gets merged in

kmagiera (approved):

Issue #1267, opened 2025-06-23
Surplus debugger sessions


anna-malina:
**Description**
There's a problem with the surplus number of sessions, when I previously tested a some number of devices, I closed them all and then again started for the project a few.

**Preconditions**
Radon IDE extension is available and launched for the react native application project.

**Steps to Reproduce**

1. Previously I used the app for a few (about 4-5) devices and tested fast switching between them (by kayborad shortcut and manually)
2. But after this I terminated each running device by clicking "Stop" on device list
3. Then I run the project once agin on 2 new devices (android, ios)
4. For a while at first I had 3 debuggers to choose, after a while then it reduced to 2
5. But when on one device I was just testing console logs in the app I had a problem when switching between devices or turning on the console that they were not automatically shown - "Radon IDE debugger 4" was initially selected, but it wasn't proper debugger for running devices
6. I had to manually switch between debuggers myself

https://github.com/user-attachments/assets/551676f6-e972-4c6e-9c93-434a1c1a90a9

**Actual Behavior**
For both running devices I had 2 debuggers: "Radon IDE debugger 4" and "Radon IDE debugger"

**Expected Behavior**
One device should always be 1 dedicated session. Changing the active device should also correctly switch the console / debugger context.

**Environment**
Version of used Radon IDE: 1.8 (from VS Code extension marketplace)
Version of React Native: 0.78

Issue #1254, opened 2025-06-16
Holding the "Next/Previous Running Device" shortcut causes extra debug consoles to be opened


jwajgelt:
**Description**
When the keyboard shortcut for switching between devices is held, the devices switch rapidly, which sometimes causes vscode to display an already stopped debug session in the "Debug Console" view.

**Environment**

- Version of currently used Radon IDE: 1.8
- Version of React Native/Expo used in the project: irrelevant
- Any other versions of tools related to the problem: irrelevant

PR #1269, opened 2025-06-24
Name debug sessions after target devices, improve Debug Session shutdowns

jwajgelt:
- names the debug session after the device it targets. This is supposed to help users switch to the correct debug console when they encounter #1254 #1267 
- fixes an issue where disposing/stopping the debug session while it's starting would leave the master debug session running, causing some cases of #1254 #1267 

### Known issues:
#1254 and #1267 still happen, but the console seems to now switch correctly after changing devices.
It seems that vscode keeps the consoles open after the session has been stopped for whatever reason, so I'm not sure how much further we can improve the current behavior.

### How Has This Been Tested: 
- open an app in Radon
- open the debug console
- start several devices at once
- verify that switching between devices displays the correct console (the name should match the selected device and logs from the application should work)

### DIFF
--- a/packages/vscode-extension/src/debugging/DebugSession.ts
+++ b/packages/vscode-extension/src/debugging/DebugSession.ts
@@ -100,11 +100,11 @@ export class DebugSession implements Disposable {
 
     this.cancelStartingDebugSession();
 
-    const startPromise = startDebugging(
+    this.parentDebugSession = await startDebugging(
       undefined,
       {
         type: MASTER_DEBUGGER_TYPE,
-        name: `${this.options.displayName}`,
+        name: this.options.displayName,
         request: "attach",
       },
       {
@@ -115,9 +115,6 @@ export class DebugSession implements Disposable {
       },
       this.cancelStartDebuggingToken
     );
-
-    const newParentDebugSession = await startPromise;
-    this.parentDebugSession = newParentDebugSession;
   }
 
   public async restart() {
@@ -156,6 +153,10 @@ export class DebugSession implements Disposable {
   public async startJSDebugSession(configuration: JSDebugConfiguration) {
     if (this.jsDebugSession) {
       await this.restart();
+    }
+
+    if (this.options.useParentDebugSession && !this.parentDebugSession) {
+      await this.startParentDebugSession();
     }
 
     const isUsingNewDebugger = configuration.isUsingNewDebugger;


packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -89,16 +91,21 @@ export class DebugSession implements Disposable {
     return this.currentWsTarget;
   }
 
+  private startParentSessionPromise: Promise<vscode.DebugSession> | undefined;
```
kmagiera: wonder why we add this when handling parent session but not for js session, seems like the problem with concurrent starting can happen for both of those sessions

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -89,16 +91,21 @@ export class DebugSession implements Disposable {
...
+
   public async startParentDebugSession() {
     assert(
       !this.jsDebugSession,
       "Cannot start parent debug session when js debug session is already running"
     );
-    const newParentDebugSession = await startDebugging(
+    if (this.startParentSessionPromise) {
+      this.startParentSessionPromise.then((session) => debug.stopDebugging(session));
+    }
```
kmagiera: wouldn't it make more sense to add `cancelToken` as a parameter to startDebugging?

kmagiera (commented):
My main question is whether it'd be easier to follow `cancelToken` pattern rather than add custom handling for the case where session is started concurrently

packages/vscode-extension/src/debugging/DebugSession.ts
jwajgelt: Possibly, I haven't thought of that. I'll look into it.

jwajgelt: It seems to work fine, and I think it encapsulates the idea better. Thanks for the suggestion 🙏

packages/vscode-extension/src/debugging/DebugSession.ts
jwajgelt: Mostly because issues when launching multiple JS sessions _kind of_ resolve themselves, since the debugged application would automatically disconnect the extra sessions.
There are potential issues with race conditions (when we save the session that disconnected instead of the one that succeeded) but they didn't come up in testing.

That said, there isn't a good reason (other than making the code harder to follow) to _not_ do it, so I changed it to use the cancel token as well when I changed the approach according to your other suggestion.

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -94,41 +97,49 @@ export class DebugSession implements Disposable {
...
-    const newParentDebugSession = await startDebugging(
+
+    this.cancelStartingDebugSession();
+
+    const startPromise = startDebugging(
       undefined,
       {
         type: MASTER_DEBUGGER_TYPE,
-        name: "Radon IDE Debugger",
+        name: `${this.options.displayName}`,
```
kmagiera:
```suggestion
        name: this.options.displayName,
```

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -94,41 +97,49 @@ export class DebugSession implements Disposable {
...
-      }
+      },
+      this.cancelStartDebuggingToken
     );
-    if (this.parentDebugSession) {
-      Logger.warn("Parent debugger session has spawned concurrently, dropping the earlier session");
-      debug.stopDebugging(this.parentDebugSession);
-    }
+
+    const newParentDebugSession = await startPromise;
```
kmagiera: we can revert this change and just have `newParentDebugSession` defined above with `await` in a single line instead of having `startPromise`

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -146,20 +157,19 @@ export class DebugSession implements Disposable {
     if (this.jsDebugSession) {
       await this.restart();
     }
-    if (this.options.useParentDebugSession && !this.parentDebugSession) {
```
kmagiera, line 149: why is this removed? Is that because we'd always expect the `parentDebugSession` to be set? If that's the case maybe it'd be good to add an `assert` here. Especially given that we reuse the cancel token that we also use for starting the parent session. If the parent session wasn't set it would not be safe to reuse the cancel token for sure. I actually still feel a bit uncertain whether it is safe to use a single cancel token and would prefer that we have separate ones for parent debug session and for the JS debug one.

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -181,7 +191,8 @@ export class DebugSession implements Disposable {
         suppressSaveBeforeStart: true,
         consoleMode: DebugConsoleMode.MergeWithParent,
         compact: true,
-      }
+      },
+      this.cancelStartDebuggingToken
```
kmagiera, line 196: I'd feel more safe if we had separate tokens. It perhaps doesn't make sense given the current configuration as we'd never want the js debug session to run or be started when the parent one isn't there, but worried that this is still an implicit assumption that may break with some changes in the future

jwajgelt: If we had separate cancel tokens right now, we'd either:
- cancel both anyway -- when starting the parent session, since you don't want to have the "old" JS session start while its parent is being stopped
- check that the parent session already exists (and thus isn't starting)

Having a single cancel token for me better reflects the behavior that we never want the single `DebugSession` to be launching multiple debug sessions at once:
- of the same type, because they conflict with one another
- of different types, because one is dependent on the other and needs to be started after it

packages/vscode-extension/src/debugging/DebugSession.ts
jwajgelt: `restart` starts the `parentDebugSession` if it wasn't already running, and it's the last thing it does, so we can be sure it's running here.

jwajgelt: Ok, but you _can_ enter this function without the `jsDebugSession`, in which case you don't start the parent. Don't know why I missed that.

PR #1272, opened 2025-06-24
Refactor BuildManager

jwajgelt:
Refactors `BuildManager`:
- makes `BuildManager` accept a `BuildConfig` rather than reading the launch config
- gets rid of `DisposableBuild` interface in favor of using `CancelToken` to cancel builds
- moves the "cache stale" event that `DeviceSession` listens to from `BuildManager` to `BuildCache`
- lifts the `BuildManager` out of `DeviceSession` to `ApplicationContext`

Motivation:
- move some responsibilities out of `BuildManager` (watching the workspace for cache staleness, generating build configs from launch config, reading launch configs)
- passing `BuildConfig` to `BuildManager` will make it easier to handle multiple `LaunchConfig`s in the future
- lifting `BuildManager` out of `DeviceSession` will make it possible to share builds-in-progress between compatible devices
- simply cleaning up

### How Has This Been Tested: 
- open app in Radon
- check it builds for Android and iOS
- check that build cache works
- check that changing the native code invalidates the cache and displays the error alert
- check that switching approots works

### DIFF
--- a/packages/vscode-extension/src/builders/BuildManager.ts
+++ b/packages/vscode-extension/src/builders/BuildManager.ts
@@ -12,10 +12,6 @@ import { Logger } from "../Logger";
 import { BuildConfig, BuildType } from "../common/BuildConfig";
 import { isExpoGoProject } from "./expoGo";
 import { LaunchConfigurationOptions } from "../common/LaunchConfig";
-import { throttleAsync } from "../utilities/throttle";
-import { watchProjectFiles } from "../utilities/watchProjectFiles";
-
-const FINGERPRINT_THROTTLE_MS = 10 * 1000; // 10 seconds
 
 export type BuildResult = IOSBuildResult | AndroidBuildResult;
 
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/builders/BuildManager.ts
```diff
@@ -1,35 +1,25 @@
...
-
-const FINGERPRINT_THROTTLE_MS = 10 * 1000; // 10 seconds
 
 export type BuildResult = IOSBuildResult | AndroidBuildResult;
 
-export interface DisposableBuild<R> extends Disposable {
-  readonly build: Promise<R>;
-}
-
 export interface BuildManagerDelegate {
```
kmagiera: can we delete this interface now?

packages/vscode-extension/src/builders/BuildManager.ts
```diff
@@ -102,307 +195,129 @@ export class BuildManager implements Disposable {
...
       platform,
       type: forceCleanBuild ? "clean" : "incremental",
     });
 
-    const buildApp = async () => {
-      const currentFingerprint = await this.buildCache.calculateFingerprint(platform);
-
-      // Native build dependencies when changed, should invalidate cached build (even if the fingerprint is the same)
-      const buildDependenciesChanged = await this.checkBuildDependenciesChanged(deviceInfo);
+    const currentFingerprint = await this.buildCache.calculateFingerprint(platform);
```
kmagiera, line 210: shouldn't we check cancelToken after async calls? when possible even pass the token to async methods in case they call subprocesses.

packages/vscode-extension/src/builders/BuildManager.ts
jwajgelt: We can, forgot to do it here.

packages/vscode-extension/src/builders/BuildManager.ts
jwajgelt: We should, but we didn't before.
I'll take a look at it separately from this PR, since it already has a lot of changes.

PR #1273, opened 2025-06-24
Fix reseting breakpoints on hot module reloading and debugger paused

filip131311:
This PR solves a problem with resetting the breakpoints while the debugger is paused.

What was happening: 
1) Immediately after a user was saving the file vscode sends `setBreakpointByUrl` command it is initially triggered by the vscode frontend and not the `vscode-js-debug` so we can not do anything about it 
2) because the debugger is paused the breakpoint would be set before the changes to the updated module would be evaluated by the application runtime, so additional breakpoint would be set in the line following the line with original breakpoint
3) after debugger was continued it would stop on this phantom breakpoint

What we change to avoid it:
1) we subscribe to metro `/hot` web socket endpoint to listen for fast refresh updates, we additionally need to register the entry point we want to observe and we use the one that comes with first `scriptParsed` event as it is the main bundle. (this part might require changes when we support multiple bundles like in brownfield applications for example)
2) we add `isBreakpointSettingPaused` state to `RadonCDPProxyDelegate` that is set when metro sends out the event `update-start`
3) we store the upcoming `setBreakpointByUrl` in a list that we dispatch when  new `scriptParsed` comes our way (this is when the update completes)

_note_: there exists an `update-end` event that is sent by the `/hot` metro endpoint but it is meaningless!!! All it means is that the update was sent out in full from metro to application and nothing about application state, that's why we need to use `scriptParsed` debugger event to signal update finish instead.

### Test Plan:

- run 'expo-53' test app 
- stop the debugger in such a line that the next line has some code that debugger could stop on as well for example:
```js
 const text = "console.log()"; //stop here
 console.log(text);
```
- add an empty line above the line you stoped on and save the file
- continue the debugger
- before the changes the debugger would stop again seemingly on the same line (but realy on the next line)


packages/vscode-extension/src/debugging/CDPProxy.ts
```diff
@@ -126,6 +151,7 @@ export class ProxyTunnel {
   constructor(
     applicationTarget: Connection,
     debuggerTarget: Connection,
+    public readonly HMRSocket: WebSocket,
```
jwajgelt: This is not CDP-related, it should be a part of the `RadonCDPProxyDelegate` or `ProxyDebugAdapter` (possibly created in the `DebugAdapter` and passed to `ProxyDelegate`)

- `CDPProxy` - implements the proxy itself (sets up a server, handles websocket connections, pumps messages)
- `RadonCDPProxyDelegate` - implements the actual Radon business logic

The proxy shouldn't care about the HMR server.

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -164,12 +164,37 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
...
+  private pendingBreakpointCommands: Array<{
+    command: IProtocolCommand;
+    resolve: (cmd: IProtocolCommand) => void;
+  }> = [];
+
+  private dispatchPendingBreakpointCommands() {
+    for (const pending of this.pendingBreakpointCommands) {
+      pending.resolve(pending.command);
+    }
+  }
```
jwajgelt:
```suggestion
  private dispatchPendingBreakpointCommands() {
    const pendingCommands = this.pendingBreakpointCommands;
    this.pendingBreakpointCommands = [];
    for (const pending of pendingCommands) {
      pending.resolve(pending.command);
    }
  }
```
We need to clear the array to avoid memory leaks.

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -164,12 +164,37 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
           command.id!,
           command.params as Cdp.Debugger.SetBreakpointByUrlParams
         );
-        return command;
+
+        const paused = () => {
+          return this.isBreakpointSettingPaused;
+        };
```
jwajgelt, line 174: Why is it thunked? You should be able to just read `this.isBreakpointSettingPaused` inside the `setImmediate`

filip131311: It is not unfortunately `this` is undefined for some reason

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -246,19 +285,34 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
...
     try {
       const sourceMapData = await this.getSourceMapData(sourceMapURL);
       const isMainBundle = sourceMapData.sources.some((source: string) =>
         source.includes("__prelude__")
       );
 
       this.sourceMapRegistry.registerSourceMap(sourceMapData, url, scriptId, isMainBundle);
+
+      if (isMainBundle) {
+        this.HMRSocket.send(
```
kmagiera, line 309: does it mean we will only be notified of the changes made in the main bundle then? For example, when we reload a module that has already been reloaded we won't be getting that?

Why is the replace statement there? Give example and explain

kmagiera (changes requested):
To me this seems like a lot of work to workaround a relatively minor issue.

For me this seems quite natural that when you put a breakpoint in a code that is about to be evaluated, it will stop there. I understand that after the module is hot-reloaded we also place the breakpoint to the new location such that it ends up being placed in the correct place once everything settles.

For me this approach is quite fragile as it seem to depend on the timing of how exactly the hot reload is processed while the debugger is paused. AFAIK this isn't documented.

I'd be afraid that this change could impact for example the case when we are setting the breakpoint while paused, but we don't intend to modify any code and hence the reload wouldn't happen. We still want to be able to hit that breakpoint (this is a pretty common scenario for me personally as opposed to modifying the code when paused).

I'm not entirely sure as for the timing of HMR/CDP commands, but one approach that I think would be less invasive would be to handle this in the code that handles the debugger pause events. We now automatically continue debugger pauses in skipped files, but we can also do the same for files overridden by HRM reload.

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
filip131311: I know this is confusing, but HMRServer only sends out updates to clients that has registered an entry point [here](https://github.com/facebook/metro/blob/4f2110412e3c4cbe6feb4f4cad27085edf7db1b0/packages/metro/src/HmrServer.js#L211) is the code responsible for that registration on metro side, to receive information about those updates we need to register to the entry point of the application that is currently running and this is the same as the URL of the Main bundle.



As tho the comments from you main message: 
> For me this seems quite natural that when you put a breakpoint in a code that is about to be evaluated, it will stop there. I understand that after the module is hot-reloaded we also place the breakpoint to the new location such that it ends up being placed in the correct place once everything settles.

I am not sure that it is expected by the user in this scenario as users probably don't know when the evaluation happens and might be lost as to that this double braek means

>For me this approach is quite fragile as it seem to depend on the timing of how exactly the hot reload is processed while the debugger is paused. AFAIK this isn't documented.

It's not documented and we plan to add the communication about HMR updates to CDP, but adding that will not solve the problem for current versions

>I'd be afraid that this change could impact for example the case when we are setting the breakpoint while paused, but we don't intend to modify any code and hence the reload wouldn't happen. We still want to be able to hit that breakpoint (this is a pretty common scenario for me personally as opposed to modifying the code when paused).

This case will not be affected by this change 

https://github.com/user-attachments/assets/091fd525-a346-423f-a67d-bc5f0b93b7ab

>I'm not entirely sure as for the timing of HMR/CDP commands, but one approach that I think would be less invasive would be to handle this in the code that handles the debugger pause events. We now automatically continue debugger pauses in skipped files, but we can also do the same for files overridden by HRM reload.

I am not sure it is less invasive the only thing my approach does is delaying the setBreakpoint command, the solution you write about would have similar effect but would need to have some separate logic that understands that some brakpoints should be skipped (once) and use source maps correctly to identify those. I find this more risky 


>To me this seems like a lot of work to workaround a relatively minor issue.

It was, but I did most of that work for unrelated reasons anyway.

kmagiera: No, this isn't confusing that HMR server requires registration. What I was asking there was that we only register for the main bundle and not for other endpoints. I'd expect that we should register for all scripts that gets eventually parsed, as for example the HRM may load additional files that didn't exist in the original main bundle in which case we'd never react to the changes made in those files.

kmagiera: As for my "alternative approach" what I was proposing was to make it so that we skip when debugger pauses on a breakpoint that isn't set in the most recent version of the code. For this purpose, we wouldn't even need to maintain the HMR connection as we could tell this just purely based on the source maps.

Have you checked if this issue you're solving here happens when connecting chome debugger?

PR #1276, opened 2025-06-25
Prevent running the same build concurrently

jwajgelt:
After #1118 we encounter an issue where, if two devices (with the same platform) are launched at the same time, and we need to build the native application, both will start building at the same time. This causes issues (at least on iOS, running `xcodebuild` from the same project concurrently causes errors in one of the builds) as well as is simply wasteful (since the application build can be used for both devices anyway).

This PR introduces a "batching" mechanism, where if a device requests an application build while a compatible build is already running, it will use the result of that in-progress build instead.

### How Has This Been Tested: 
- open an app (which builds the native part locally) in Radon
- choose the "Clean rebuild" option to force a build
- start a second device with the same platform
- check that the build only runs once -- either by inspecting the Radon IDE output or adding breakpoints/logpoints in `BuildManager`
- check that stopping all of the devices also cancels the build

### DIFF
--- a/packages/vscode-extension/src/project/ApplicationContext.ts
+++ b/packages/vscode-extension/src/project/ApplicationContext.ts
@@ -3,13 +3,21 @@ import { BuildCache } from "../builders/BuildCache";
 import { DependencyManager } from "../dependency/DependencyManager";
 import { LaunchConfigController } from "../panels/LaunchConfigController";
 import { disposeAll } from "../utilities/disposables";
-import { BuildManager } from "../builders/BuildManager";
+import { BuildManager, BuildManagerInterface } from "../builders/BuildManager";
+import { BatchingBuildManager } from "../builders/BatchingBuildManager";
+
+function createBuildManager(
+  dependencyManager: DependencyManager,
+  buildCache: BuildCache
+): BuildManagerInterface {
+  return new BatchingBuildManager(new BuildManager(dependencyManager, buildCache));
+}
 
 export class ApplicationContext implements Disposable {
   public dependencyManager: DependencyManager;
   public launchConfig: LaunchConfigController;
   public buildCache: BuildCache;
-  public buildManager: BuildManager;
+  public buildManager: BuildManagerInterface & Disposable;
   private disposables: Disposable[] = [];
 
   constructor(public appRootFolder: string) {
@@ -17,7 +25,7 @@ export class ApplicationContext implements Disposable {
 
     this.launchConfig = new LaunchConfigController(appRootFolder);
     this.buildCache = new BuildCache(appRootFolder);
-    this.buildManager = new BuildManager(this.dependencyManager, this.buildCache);
+    this.buildManager = createBuildManager(this.dependencyManager, this.buildCache);
 
     this.disposables.push(
       this.launchConfig,
@@ -41,7 +49,7 @@ export class ApplicationContext implements Disposable {
 
     this.launchConfig = new LaunchConfigController(newAppRoot);
     this.buildCache = new BuildCache(newAppRoot);
-    this.buildManager = new BuildManager(this.dependencyManager, this.buildCache);
+    this.buildManager = createBuildManager(this.dependencyManager, this.buildCache);
     this.disposables.push(
       this.launchConfig,
       this.dependencyManager,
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/project/ApplicationContext.ts
```diff
@@ -3,21 +3,29 @@ import { BuildCache } from "../builders/BuildCache";
...
+): BuildManagerInterface {
+  return new BatchingBuildManager(new BuildManager(dependencyManager, buildCache));
+}
 
 export class ApplicationContext implements Disposable {
   public dependencyManager: DependencyManager;
   public launchConfig: LaunchConfigController;
   public buildCache: BuildCache;
-  public buildManager: BuildManager;
+  public buildManager: BuildManagerInterface & Disposable;
```
kmagiera: `BuildManagerInterface` already extends `Disposable`, is this needed then?

kmagiera (commented):
I like the idea and see the point of extracing the code that manages multiple builds into a separate class. However, I don't think it makes sense to design an interface level abstraction for this. Especially that we would never use the `BuildManager` directly.

I think the approach that could work better is:
1) AppContext just references the BatchedBuildManager instead of the interface
2) We refactor BuildManager to just be a function (buildApp) instead of being a class. The whole class is almost "stateless" anyways and only stores the output channel, which can equally well be kept in the `BatchedBuildManager` anyways.
3) I'd rename BatchedBuild to perhaps just `BuildManager` and this way it'd effectively replace the current `BuildManager` class that'd became a function

wdyt?

packages/vscode-extension/src/project/ApplicationContext.ts
jwajgelt: At some point while I was working on it it didn't, and I must have missed that.

jwajgelt:
> I like the idea and see the point of extracing the code that manages multiple builds into a separate class. However, I don't think it makes sense to design an interface level abstraction for this. Especially that we would never use the `BuildManager` directly.
> 
> I think the approach that could work better is:
> 
>     1. AppContext just references the BatchedBuildManager instead of the interface
> 
>     2. We refactor BuildManager to just be a function (buildApp) instead of being a class. The whole class is almost "stateless" anyways and only stores the output channel, which can equally well be kept in the `BatchedBuildManager` anyways.
> 
>     3. I'd rename BatchedBuild to perhaps just `BuildManager` and this way it'd effectively replace the current `BuildManager` class that'd became a function
> 
> 
> wdyt?

I had two reasons in mind when I implemented it this way:
- it made it easy to write tests for the batching, since we can actually mock the build logic this way. If we injected the `buildApp` function we'd get the same behaviour though.
- I actually wanted to go further in this direction, to remove the current dependency of `BuildManager` on `BuildCache` and instead introduce another decorator class, `CachingBuildManager` which implements the caching logic.

> However, I don't think it makes sense to design an interface level abstraction for this.

I might be OOP-pilled at this point, but I generally like the idea of depending on abstractions instead of concrete implementations (i.e. the Dependency Inversion Principle). While at the end of the day, we do only use a single implementation (the one that `ApplicationContext` will expose via `buildManager` property) it makes it easier to apply other design patterns (like the decorator pattern here) and test things depending on the `BuildManager` without jumping through too many hoops.

packages/vscode-extension/src/project/ApplicationContext.ts
```diff
@@ -3,7 +3,12 @@ import { BuildCache } from "../builders/BuildCache";
 import { DependencyManager } from "../dependency/DependencyManager";
 import { LaunchConfigController } from "../panels/LaunchConfigController";
 import { disposeAll } from "../utilities/disposables";
-import { BuildManager } from "../builders/BuildManager";
+import { BuildManagerImpl, BuildManager } from "../builders/BuildManager";
+import { BatchingBuildManager } from "../builders/BatchingBuildManager";
+
+function createBuildManager(dependencyManager: DependencyManager, buildCache: BuildCache) {
+  return new BatchingBuildManager(new BuildManagerImpl(dependencyManager, buildCache));
+}
```
kmagiera, line 11: maybe we could extract those few lines of code that creates new managers/config objects in constructor and `updateAppRootFolder` to separate method, then we'd only call this `new BatchingBuildManager` in one place and won't need a helper method for it.

jwajgelt: We can't really do that, since if we move the prop assignments from the `constructor` to a separate method TS cannot figure out that the fields are initialized properly.
I'd rather go back to the previous idea behind `ApplicationContext`, where it's an immutable grouping of `appRoot`-dependent objects that is created from scratch whenever we switch `appRoot`s.
That would require either moving the `DependencyManager` out of it or creating it in `Project` and passing as a dependency to `ApplicationContext`, but I think that's fine.
I'd do that outside of this PR though, since the change isn't really related.

PR #1281, opened 2025-06-27
Setup CDP-based communication mechanism for Radon Connect

kmagiera:
This PR implements the "agent" interface for Radon Connect.

Agent interface is what we use for communication between the extension and the app process. When running metro inside the extension, we implement the agent based on the react devtools websocket connection. This is then used by the dev tools extensions like redux, react query and others in order to exchange messages.

In Radon Connect, we don't have a way to communicate over the devtools protocol, because we don't control metro proces and hence we don't also control the react devtools port.

The new mechanism relies on the CDPs "binding" mechanism. Using this mechanism, we can register a global method (we call it `__radon_binding`), which when called, will trigger `Runtime.bindingCalled` method. We use this mechanism to trigger message calls from the app process to the extension (via debugger).

To send messages from extension to app process, we simply use `evaluate` and call another global method that we need to register beforeahead (`__radon_dispatch`).

Similarly to how we do things in "Radon Classic", we also need to have some runtime code executed. The runtime code is responsible for setting up the dispatch command and defining the agent interface that uses dispatch command and the registered binding. In order to load the runtime code however, we can no longer inject some extra modules into metro. Instead, we again use `evaluate` mechanism. But, because we want to eventually also load a bit more code beyond the agent definition (for example to include the redux dev tools code), we are using a pre-built script that we load into the eval method.

The script is built using esbuild and mostly follows the same pipeline as the main extension code, apart from the fact we need a slightly different configuration, because the bundled code will be loaded into a React Native process rather than a node process.

Here's the breakdown of the changes made:
1) We add a new `connect_runtime.ts` file which is an entry point for the runtime loaded into RN process when Radon Connect is connected
2) We update esbuild configuration to build the runtime bundle
3) We update the debug session code, such that (when requested) it sets up the Radon Connect runtime and installs necessary bindings
4) We refactor Connector.ts by defining new ConnectSession class that now manages the debug session along with metro etc. This simplifies the Connector code a bit while also allowing us to expand it for handling tools in the future.

This change has mostly no effect as now we don't utilize the communication channel (this will be needed once we add tools support). The only effect this change has is on Radon Connect runtime where we load the runtime bundle upon connecting.

### How Has This Been Tested: 
1) Since this change has no effect apart from loading runtime code, the way I tested it is by checking the communication channel. For that you need to add some lines to `connect_runtime.ts`:
2) add `globalThis.__radon_agent.postMessage("Hello");` and place a breakpoint in `ConnectSessionDelegate.onBindingCalled` to see that the event gets delivered there
3) add `globalThis.__radon_agent.onMessage` handler and call `inspectorBidge.send` to see that the mesages are delivered both ways.


packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -246,7 +255,43 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
...
 
-  private async handleScriptParsed(command: IProtocolCommand): Promise<IProtocolCommand> {
+  private async setupRadonConnectRuntime(tunnel: ProxyTunnel) {
+    // load script from lib/connect_runtime.js and evaluate it
+    const runtimeScriptPath = path.join(
+      extensionContext.extensionPath,
+      "dist",
+      "connect_runtime.js"
+    );
+    const runtimeScript = fs.readFileSync(runtimeScriptPath, "utf8");
```
jwajgelt: 1. why not the `fs.promises` API? We're already in an `async` context, I don't think we need to stop the whole extension host for this.
2. I dislike the idea of doing fs operations in the `RadonProxyDelegate`, but I'm not sure there's a better way to do it without changing too many things just for the sake of it.
Maybe instead of passing a `boolean` `installConnectRuntime` flag we could pass a `injectedRuntime: string` to `RadonProxyDelegate` instead? We could also then store it in a static somewhere (lazily, after the first read) so that we don't have to read it from the filesystem on each debug connection.

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -246,7 +255,43 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
...
+    });
+
+    const result = (await tunnel.injectDebuggerCommand({
+      method: "Runtime.evaluate",
+      params: {
+        expression: runtimeScript,
+      },
+    })) as Cdp.Runtime.EvaluateResult;
+    if (result.exceptionDetails) {
+      Logger.error("Failed to setup Radon Connect runtime", result.exceptionDetails);
```
jwajgelt, line 282: Maybe this should be handled by Radon? I'm not sure what the consequences are, but won't it cause the Preview to get stuck in some state?

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -230,6 +235,10 @@ export class DebugSession implements Disposable {
     return Promise.race([resultPromise, timeout]).catch((_e) => false);
   }
 
+  public postMessage(data: any) {
+    this.jsDebugSession?.customRequest("RNIDE_postMessage", data);
+  }
```
jwajgelt: I feel like this method's name is too generic. It's supposed to be used specifically for the inspector bridge connection, right? Maybe something more like
```suggestion
  public dispatchToRadonAgent(data: any) {
    this.jsDebugSession?.customRequest("RNIDE_dispatchToRadonAgent", data);
  }
```
since it calls `radon_dispatch` on the runtime? Or
```suggestion
  public postInspectorBridgeMessage(data: any) {
    this.jsDebugSession?.customRequest("RNIDE_postInspectorBridgeMessage", data);
  }
```
If we wanted to be less specific about how it's implemented, and more about the intent.
 I don't know really.

jwajgelt (changes requested):
Overall looks good to me, but I have left two questions and one nitpick in the inline comments.

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
kmagiera: I changed this to async, but don't think it is worth optimizing the 2nd point. The RadonCDPProxyDelegate is currently coupled with the new communication mechanism which isn't limited to reading and evaluating the script at a given moment but also to registering the binding. If I was to extract the script loading to uncouple it, I'd also need to extract setting up the binding at the correct moment and also abstract the binding call handlers, which would create an unnecessarily complex abstraction. 

We may want to add a caching mechanism, but given this is only called once per app reload, and that the runtime file will be <1MB I don't think that's necessary either.

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
kmagiera: For now, this will have no consequences. Once we start relying on the communication between process and extension (i.e. for the plugins), this will prevent the plugins from working. Not sure how this could be surfaced best if not via logs. Perhaps we want to also add this to telemetry as I'd expect this to not be a common error but once it happens we will likely want to know how and why.

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
jwajgelt: The difference between how this is handled vs. the runtime binding is, the way we handle the binding here is not really Connect-related (from the perspective of ProxyDelegate) -- you install a binding and emit an event whenever it's called, the mechanism itself is generic and can be reused by different code which wants to talk to the runtime.
At the end of the day, I'd say it's fine either way, it just bugs me a little -- but there are bigger transgressions in the debugger code anyway.

jwajgelt (approved):
I don't see any other problems with the code, I haven't tested it though.

PR #1290, opened 2025-07-02
Make BuildCache independent from `LaunchConfig`, batch fingerprint calculations when workspace doesn't change

jwajgelt:
Makes the `BuildCache` class independent from `LaunchConfig`.
- calls to `BuildCache`'s methods now require passing either `appRoot` or `FingerprintOptions` which include the options previously retrieved from `LaunchConfig`.
- the "build cache stale" listener has been moved from `BuildCache` to `DeviceSession`. 
- fingerprint calculations are now batched, so that when multiple callers try to get the fingerprint while no changes to the project where made, they will get the result of the same calculation

Motivation:
- batch and cache fingerprint calculations to save time on unnecessary computations
- make `BuildCache` independent from `launchConfig` to allow for having multiple launch configs in the future

### How Has This Been Tested: 
- launch the app
- verify it grabs the build from cache
- change something that impacts the fingerprint
- verify the "native changes" alert appears
- verify that when custom fingerprint is configured, it is used

### DIFF
--- a/packages/vscode-extension/src/project/ApplicationContext.ts
+++ b/packages/vscode-extension/src/project/ApplicationContext.ts
@@ -5,7 +5,6 @@ import { LaunchConfigController } from "../panels/LaunchConfigController";
 import { disposeAll } from "../utilities/disposables";
 import { BuildManagerImpl, BuildManager } from "../builders/BuildManager";
 import { BatchingBuildManager } from "../builders/BatchingBuildManager";
-import { FingerprintProvider } from "./FingerprintProvider";
 
 function createBuildManager(dependencyManager: DependencyManager, buildCache: BuildCache) {
   return new BatchingBuildManager(new BuildManagerImpl(dependencyManager, buildCache));
@@ -14,16 +13,16 @@ function createBuildManager(dependencyManager: DependencyManager, buildCache: Bu
 export class ApplicationContext implements Disposable {
   public dependencyManager: DependencyManager;
   public launchConfig: LaunchConfigController;
-  public buildCache: BuildCache;
   public buildManager: BuildManager;
-  public fingerprintProvider: FingerprintProvider = new FingerprintProvider();
   private disposables: Disposable[] = [];
 
-  constructor(public appRootFolder: string) {
+  constructor(
+    public appRootFolder: string,
+    public readonly buildCache: BuildCache
+  ) {
     this.dependencyManager = new DependencyManager(appRootFolder);
 
     this.launchConfig = new LaunchConfigController(appRootFolder);
-    this.buildCache = new BuildCache(this.fingerprintProvider);
     const buildManager = createBuildManager(this.dependencyManager, this.buildCache);
     this.buildManager = buildManager;
 
@@ -43,7 +42,6 @@ export class ApplicationContext implements Disposable {
     disposeAll(this.disposables);
 
     this.launchConfig = new LaunchConfigController(newAppRoot);
-    this.buildCache = new BuildCache(this.fingerprintProvider);
     const buildManager = createBuildManager(this.dependencyManager, this.buildCache);
     this.buildManager = buildManager;
 
@@ -52,6 +50,5 @@ export class ApplicationContext implements Disposable {
 
   public dispose() {
     disposeAll(this.disposables);
-    this.fingerprintProvider.dispose();
   }
 }
--- a/packages/vscode-extension/src/project/ApplicationContext.ts
+++ b/packages/vscode-extension/src/project/ApplicationContext.ts
@@ -25,7 +25,7 @@ export class ApplicationContext implements Disposable {
     const buildManager = createBuildManager(this.dependencyManager, this.buildCache);
     this.buildManager = buildManager;
 
-    this.disposables.push(this.launchConfig, this.dependencyManager, buildManager, this.buildCache);
+    this.disposables.push(this.launchConfig, this.dependencyManager, buildManager);
   }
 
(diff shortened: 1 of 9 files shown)


packages/vscode-extension/src/project/FingerprintProvider.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    this.fingerprintCache.set(cacheKey, fingerprintPromise);
+    fingerprintPromise.catch(() => {
+      // NOTE: on error, we remove the promise from the cache so that further calls will retry
+      if (this.fingerprintCache.get(cacheKey) === fingerprintPromise) {
+        this.fingerprintCache.delete(cacheKey);
+      }
+    });
+
+    return await fingerprintPromise;
```
filip131311: this is something different then `return this.fingerprintCache.get(this.makeCacheKey(options))!;` as the awaits are inconsistant tho I don't think it affects the usage in any way.

packages/vscode-extension/src/project/FingerprintProvider.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  constructor() {
+    this.workspaceWatcher = watchProjectFiles(this.onProjectFilesChanged);
+  }
+
+  public async calculateFingerprint(options: FingerprintOptions): Promise<Fingerprint> {
+    const cacheKey = this.makeCacheKey(options);
+    if (this.fingerprintCache.has(cacheKey)) {
+      Logger.debug("Using cached fingerprint");
+      return this.fingerprintCache.get(this.makeCacheKey(options))!;
```
filip131311: why "!" is necessery?

filip131311: I think you can avoid it by replaceing it with ` this.fingerprintCache.get(cacheKey)`

packages/vscode-extension/src/project/ApplicationContext.ts
```diff
@@ -41,14 +43,15 @@ export class ApplicationContext implements Disposable {
...
     const buildManager = createBuildManager(this.dependencyManager, this.buildCache);
     this.buildManager = buildManager;
 
-    this.disposables.push(this.launchConfig, this.dependencyManager, buildManager, this.buildCache);
+    this.disposables.push(this.launchConfig, this.dependencyManager, buildManager);
   }
 
   public dispose() {
     disposeAll(this.disposables);
+    this.fingerprintProvider.dispose();
```
filip131311: why not just ad it to disposables?

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -269,10 +299,8 @@ export class DeviceSession
 
   //#endregion
 
-  //#region Build manager delegate methods
-
-  onCacheStale = (platform: DevicePlatform) => {
-    if (platform === this.device.platform && this.status === "running") {
+  onCacheStale = () => {
```
filip131311, line 302: why do you drop the platform check? won't it cause the reload even if the other platform is stale?

filip131311 (commented):
left inline comments

packages/vscode-extension/src/project/ApplicationContext.ts
jwajgelt: We don't need to dispose it whenever the appRoot changes, unlike the other dependencies.

packages/vscode-extension/src/project/FingerprintProvider.ts
jwajgelt: Pretty sure you can't (ts cannot express these kinds of conditions), but it should still use `cacheKey`.

packages/vscode-extension/src/project/FingerprintProvider.ts
jwajgelt: It cannot be anything else, because there are no `await` points in this function (other than this one in the return statement). That said, we could probably just return `fingerprintPromise` and make this function non-async to highlight this.

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: https://github.com/software-mansion/radon-ide/pull/1290/files/023d1b30160c680f0a4cfdc75974846399234588#diff-67fa1a702549ac04c423443aace83235ede1b8cf967257fd7a03980c1e0879cdR218-R222

We pass the platform when calling `isCacheStale`.

filip131311: I get it but now that i think about it shouldn't we propagate it no matter the platform and then block it in the device session state? When a user has two devices turned on, they should see "native changes" warning when switching back to the other platform and not only after restart?

packages/vscode-extension/src/project/ApplicationContext.ts
filip131311: hmmm we usually tho it other way around and expose the objects that can change out of disposables so to force the pattern of adding everything that is not special to `disposables` as `fingerprintProvider` is here. Maybe we should think about some other approach of updating the disposables and adding everything there no matter what...?

jwajgelt: Or initialize this somewhere else.
The only reason it's initialized here is because it's a dependency of `BuildCache`, which itself is only initialized here because it's a dependency of `BuildManager` (which _should_ be initialized here, that's the end of the chain).
We could probably set this up in Project and inject into `ApplicationContext` instead.

filip131311 (approved):
👍

PR #1296, opened 2025-07-03
Fix local MCP server not being reloaded on license token change

latekvo:
Local MCP was intended to be restarted on every change of the license token.
This behaviour ensures that the paid MCP tools are loaded as soon as a valid license is available. 

With direct-injection mode (vscode `1.101+`), this behaviour was bugged, and the server never restarted 
due to the server-starting promise being only executed once before being cached.

This PR: 
- Fixes said bug
- Adds a missing `Disposable` (`watchLicenseTokenChange`) to the `extensionContext.subscriptions` array
- Prevents AI registration from blocking IDE initialisation
- Applies a more uniform naming scheme in the `ai/mcp/index.ts`.

### How Has This Been Tested: 

- On each token reload, a new `port` is being used, and the server now sends appropriate web requests to the backend.

### DIFF
--- a/packages/vscode-extension/src/ai/mcp/index.ts
+++ b/packages/vscode-extension/src/ai/mcp/index.ts
@@ -72,7 +72,6 @@ function isDirectLoadingAvailable() {
 }
 
 export default function registerRadonAi() {
-  // The `registerRadonAi` is async, but never awaited, to prevent slowing down Radon IDE startup.
   if (isDirectLoadingAvailable()) {
     directLoadRadonAI();
   } else {


packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -65,20 +63,23 @@ async function loadRadonAI() {
   }
 }
 
-export default function registerRadonAi() {
-  if (
-    // @ts-ignore lm.registerMcpServerDefinitionProvider API is only availble in VSCode 1.101+, we do runtime check here as it shouldn't be used in Cursor or Windsurf
+function isDirectLoadingAvailable() {
+  return (
+    // @ts-ignore lm.registerMcpServerDefinitionProvider API is only availble in VSCode 1.101+. Cursor and Windsurf both use <1.101 VSCode versions
```
kacperkapusciak:
```suggestion
    // @ts-ignore lm.registerMcpServerDefinitionProvider API is only available in VSCode 1.101+. Cursor and Windsurf both use <1.101 VSCode versions
```
nit: typo

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -65,20 +63,23 @@ async function loadRadonAI() {
...
-  ) {
-    // We need to register MCP definition before extension's activate resolves.
-    // As this is a sync call, we don't need to await for the server to actually load
-    // and we can simply return here to not delay the activation flow.
-    // We will need the server to start before
-    return loadRadonAIOnVscode1_101();
+  );
+}
+
+export default async function registerRadonAi() {
```
kacperkapusciak:
```suggestion
export default async function registerRadonAI() {
```
nit: inconsistent casing, let's go with RadonAI

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -65,20 +63,23 @@ async function loadRadonAI() {
...
+  // The `registerRadonAi` is async, but never awaited, to prevent slowing down Radon IDE startup.
```
kmagiera: this doesn't seem to make any sense. we should not have a method that is async but doesn't await. Instead, for the methods where we explicitly omit await we should add a comment that we don't await there because this method needs to return synchronously.

packages/vscode-extension/src/ai/mcp/index.ts
latekvo: Fixed in [33ad179](https://github.com/software-mansion/radon-ide/pull/1296/commits/33ad1793a269b6dfa205c91156c711d10fa5f904)

packages/vscode-extension/src/ai/mcp/index.ts
latekvo: Fixed in [de1ff0e](https://github.com/software-mansion/radon-ide/pull/1296/commits/de1ff0e5e6853780c3a1288849d44a2e35868915)

latekvo:
@kmagiera 

>The server should not rely on the license verification as it may be also started while offline in which case we would't restart it once we get back the network connection

The server doesn't require license verification to start, it requires license verification to load the tools that use our API.

We could try reloading the tools when when a license token changes without restarting the MCP, but that'll add complexity without really solving any issues. Reloading tools (removing stale ones, adding new ones) is something the MCP server doesn't support, so we'd probably have to write patches to support it.

This is of course all doable if necessary, but I don't think there's a need to change the current tool loading system.

packages/vscode-extension/src/ai/mcp/index.ts
latekvo: You're right, fixed in [4e21b2c](https://github.com/software-mansion/radon-ide/pull/1296/commits/4e21b2c9a9ec4517ecd066072cccb253a7ae20c6) and [18cbf13](https://github.com/software-mansion/radon-ide/pull/1296/commits/18cbf131f48621a909c1cc9e7c96638f7f04f7c2)

kmagiera (approved):

PR #1297, opened 2025-07-04
Store used launch config in project 

jwajgelt:
- passes launch configuration from `Project` to various components of Radon that need it, instead of reading directly from the `launch.json` file through `getLaunchConfiguration()`
- adds the selected launch configuration and the list of all configured launch configs in the Project State.
- adds a `setLaunchConfiguration` method do `ProjectInterface` to be called by the frontend to allow switching between configurations from the UI in the future.

The goal is to preserve the current behavior while preparing for user-facing changes in a following PR.

### How Has This Been Tested: 
- open a monorepo setup with no launch config
- the current behavior of some application starting and a "multiple application roots detected" warning appearing should still occur
- add a "radon-ide" launch config with `appRoot` configured, e.g.
```
{
  "configurations": [
    {
      "type": "radon-ide",
      "name": "Radon IDE panel",
      "request": "launch",
      "appRoot": "expo-52-eas",
      "eas": {
        "android": {
          "profile": "development"
        }
      }
    }
  ]
}
```
- Radon should switch to the configured `appRoot` and apply other configured options
- check that the app root select and Launch Configuration modal still work
- no user facing changes should be present


packages/vscode-extension/src/common/Project.ts
```diff
@@ -215,6 +218,8 @@ export interface ProjectInterface {
   renameDevice(deviceInfo: DeviceInfo, newDisplayName: string): Promise<void>;
   updatePreviewZoomLevel(zoom: ZoomLevelType): Promise<void>;
 
+  setLaunchConfiguration(launchConfig: LaunchConfigurationOptions): Promise<void>;
```
kmagiera: maybe `selectLaunchConfigurationWithOptions` ? We don't actually send launch config object but options here

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -30,26 +29,25 @@ import { getTelemetryReporter } from "../utilities/telemetry";
...
-    return this._appRootFolder;
+    return this.launchConfiguration.absoluteAppRoot;
   }
 
-  set appRootFolder(newRoot: string) {
-    if (this._appRootFolder !== newRoot) {
-      this._appRootFolder = newRoot;
+  setLaunchConfiguration(newLaunchConfiguration: LaunchConfiguration) {
+    const newRoot = newLaunchConfiguration.absoluteAppRoot;
+    if (this.appRootFolder !== newRoot) {
```
kmagiera: for clarity I'd compare against `this.launchConfiguration.absoluteAppRoot`

packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -30,26 +29,25 @@ import { getTelemetryReporter } from "../utilities/telemetry";
...
   private packageManagerInternal: PackageManagerInfo | undefined;
 
-  constructor(appRootFolder: string) {
-    this._appRootFolder = appRootFolder;
-  }
+  constructor(private launchConfiguration: LaunchConfiguration) {}
 
   get appRootFolder() {
-    return this._appRootFolder;
+    return this.launchConfiguration.absoluteAppRoot;
```
kmagiera: we can likely remove this entirely. Seems to be only used in this class and we can completely switch all the methods that take it to use LaunchConfiguration instead which should be cleaner (i.e. `isExpoGoProject` or `npmPackageVersionCheck`)

packages/vscode-extension/src/utilities/launchConfiguration.ts
```diff
@@ -8,6 +8,13 @@ export function getLaunchConfiguration() {
   return ideConfig;
 }
 
+export function getLaunchConfigurations(): Array<LaunchConfigurationOptions> {
```
kmagiera, line 11: do we still need the above method `getLauchConfiguration` ?? Would be better to remove it as someone may use it assuming there's only one configuration

packages/vscode-extension/src/project/project.ts
```diff
@@ -157,24 +182,7 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
-      this.applicationContext,
-      this.deviceManager,
-      this
-    );
-    oldDeviceSessionsManager.dispose();
-    this.updateProjectState({
-      appRootPath: this.relativeAppRootPath,
-    });
-  }
+  private async setupAppRoot() {}
```
kmagiera: delete?

packages/vscode-extension/src/project/project.ts
```diff
@@ -140,7 +165,7 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
   }
 
   get launchConfig() {
-    return this.applicationContext.launchConfig;
+    return this.applicationContext.launchConfigurationController;
```
kmagiera, line 168: do we want to rename the getter to `launchConfigController` then?

packages/vscode-extension/src/project/project.ts
```diff
@@ -93,29 +99,48 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
+          this.applicationContext.updateLaunchConfig(selectedLaunchConfig);
+          this.updateProjectState({
+            selectedLaunchConfiguration: selectedLaunchConfig,
+          });
         }
       })
     );
   }
 
+  async setLaunchConfiguration(options: LaunchConfigurationOptions): Promise<void> {
```
kmagiera: similar comment re the naming here, we take options as input, so using the setter notation doesn't feel right

packages/vscode-extension/src/project/project.ts
```diff
@@ -93,29 +99,48 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
-              "Dismiss"
-            );
-            return;
-          }
+      this.launchConfigsManager.onLaunchConfigurationsChanged((launchConfigs) => {
+        this.updateProjectState({
+          customLaunchConfigurations: launchConfigs,
+        });
+        const selectedLaunchConfig =
+          launchConfigs[0] ?? launchConfigurationFromOptions({ appRoot: this.relativeAppRootPath });
```
kmagiera, line 108: not sure if that's what I'd expect here. I understand that this triggers when someone adds another launch configuration. I think I'd like to keep the previously selected launch configuration unless that one got removed\, no?

kmagiera (approved):
I like this change. Looks good for the most part. I left a few minor comments and one that discussed how we handle launch configuration changes.

On top of that I wonder why do we need "absoluteAppRoot" now while before it wasn't necessary? Feels like we should be able to derive the absolute root path when it is necessary and we don't need to duplicate it in the configuration.

packages/vscode-extension/src/project/project.ts
jwajgelt: This is meant to preserve the current behaviour, which it does -- currently we always look at the first (`radon-ide`) config in `launch.json`, and switch applications when it changes the configured `appRoot` (no matter if it's a "changed" config or an "added" one).

packages/vscode-extension/src/utilities/launchConfiguration.ts
jwajgelt: It's used in two places: 
- extension context where we don't have the project context, and so we pick any launch config. It's used to find _some_ app root and activate the extension if it's present.
- `LaunchConfigController`, which I plan to either change or remove completely in a follow up PR, since it shouldn't be necessary once we adapt the frontend to the changes here.

I'd leave it as is for now, and remove it once I'm finished with `LaunchConfigController`

packages/vscode-extension/src/common/Project.ts
jwajgelt: `Options` is just `LaunchConfiguration` that you don't have to specify the optional fields for. I think it's fine. I changed the name to `select`.

PR #1301, opened 2025-07-04
Config and init ui-tests to open project and validate webview

anna-malina:
### Description
This PR introduces first steps for automated UI testing for the Radon IDE Visual Studio Code extension. These tests aim to improve the quality of the plugin. For now, this is just the initial phase of a simple smoke test

The main changes include:
- Addition of Selenium-based UI tests with using [vscode-extension-tester framework](https://github.com/redhat-developer/vscode-extension-tester)
- Adding a scripts to build the test environment (install dependencies like vsc and extension) and run ui-tests
- Adding a two test cases to check that 1) whether the Radon IDE webview is correctly enabled for the opened project and correctly displays the project name, 2) whether, depending on the device state, the correct elements / texts are displayed

**Motivation:** The motivation behind this addition is to catch bugs early in the development lifecycle and reduce manual testing efforts.

### How Has This Been Tested:
The newly added UI tests have been run on local development environment. Here's the outline of the testing process:

- **Local Testing:** Ran via script "npm run setup-run-tests" test cases on local machine using the VS Code Extension Tester - both running tests were successful.

**Test Configuration:**
- **OS:** macOS Sequoia 15.5
- **VSCode Version:** 1.101.2 (Universal)
- **Node.js Version:** v22.16.0
- **Extension Version:** 1.8.0

At the moment not documented, in future with development planned to add md file. 

### Setup:
Installation:
- NodeJS [v22.16.0]
- Visual Studio Code [1.101.2]
- ChromeDriver [138.0.7204.49]
- Python [3.13.5]
- selenium-webdriver [4.34.0]
- mocha [11.7.1]
- chai [5.2.0]
- vscode-extension-tester 

Prepare and mock test data:
1. Inside the `data` folder add a vsix radon file (with the version you want to test) named `radon-ide.vsix`. If you want to name it differently, rename the vsix file name in package.json in the script `install-vsix`
2. Inside the 'data' folder add a new folder named `radon-ide-test-apps` which will contain the test applications folders. In my case, `react-native-80` is there. If the folder with apps or the name of the folder with the tested app itself has a different name - change it in `testData.js` inside te `data` folder. This file also contains the expected name of the project (in my case `reactNative80`) - change it if your application is named differently.

Run setup and tests:
1. After preparing the data, run the command `npm run setup-run-tests` which performs the following actions: retrieves Visual Studio Code and ChromeDriver, installs the `./data/radon-ide.vsix` and runs each tests inside the `./ui-tests/*.test.js`

### DIFF
--- a/packages/vscode-extension-tester/ui-tests/smoke.test.js
+++ b/packages/vscode-extension-tester/ui-tests/smoke.test.js
@@ -31,7 +31,6 @@ describe("Smoke tests Radon IDE", () => {
         await browser.waitForWorkbench();
 
         workbench = new Workbench();
-        await new Promise((resolve) => setTimeout(resolve, 500));
 
         view = new WebView();
     });
--- a/packages/vscode-extension-tester/utils/projectLauncher.js
+++ b/packages/vscode-extension-tester/utils/projectLauncher.js
@@ -17,13 +17,13 @@ async function openProject(browser, driver, projectPath, workbench) {
     );
 
     await workbench.executeCommand('RNIDE.openPanel');
-    await driver.sleep(5000);
+    await driver.sleep(1000);
 
-    const webview = await driver.findElement(By.className('webview'));
+    const webview = await driver.findElement(By.css('iframe[class*="webview"]'));
     await waitForElement(driver, webview);
     await driver.switchTo().frame(webview);
 
-    const iframe = await driver.findElement(By.tagName('iframe'));
+    const iframe = await driver.findElement(By.css('iframe[title="Radon IDE"]'));
     await waitForElement(driver, iframe);
     await driver.switchTo().frame(iframe);
 


packages/vscode-extension/test/utils/projectLauncher.js
```diff
@@ -0,0 +1,28 @@
...
+    await driver.sleep(5000);
+
+    const webview = await driver.findElement(By.className('webview'));
+    await driver.switchTo().frame(webview);
+
+    const iframe = await workbench.getDriver().findElement(By.tagName('iframe'));
+    await workbench.getDriver().switchTo().frame(iframe);
+}
+
+export { openProject };
```
filip131311: please, add new line at the end of each file

packages/vscode-extension/test/utils/projectLauncher.js
```diff
@@ -0,0 +1,28 @@
...
+        async () => {
+            console.log('Additional wait after workbench is ready');
+            await new Promise(resolve => setTimeout(resolve, 3000));
+        }
+    );
+
+    await workbench.executeCommand('RNIDE.openPanel');
+    await driver.sleep(5000);
+
+    const webview = await driver.findElement(By.className('webview'));
```
filip131311: this does not seem to be a very distinct className can we add some custom tags to our code so we can be sure that we click on correct elements?

packages/vscode-extension/test/utils/projectLauncher.js
```diff
@@ -0,0 +1,28 @@
...
+    await browser.openResources(
+        path.resolve(__dirname, projectPath),
+
+        async () => {
+            console.log('Additional wait after workbench is ready');
+            await new Promise(resolve => setTimeout(resolve, 3000));
+        }
+    );
+
+    await workbench.executeCommand('RNIDE.openPanel');
```
filip131311: Note (probably ot of scope): this seems like an important functionality to cover with tests, before we start testing other things as this working as expected may affect most of other test cases

packages/vscode-extension/test/ui-tests/example.test.js
```diff
@@ -0,0 +1,84 @@
...
+import { openProject } from "../utils/projectLauncher.js";
+
+describe("Smoke tests Radon IDE", () => {
+    let browser;
+    let driver;
+    let workbench;
+    let view;
+
+    beforeEach(async function () {
+        this.timeout(8000);
```
filip131311: this seems extremely costly do we have any way of telling if Initialization was completed and resumeing after that to avoid unnecessary wait times?

packages/vscode-extension/test/ui-tests/example.test.js
```diff
@@ -0,0 +1,84 @@
...
+    });
+
+    it("should open Radon IDE webview for a specific project", async function () {
+        this.timeout(15000);
+        const title = await driver.getTitle();
+        assert.equal(title, texts.pageTitle, "Page title should be 'Visual Studio Code'");
+
+        await openProject(browser, driver, paths.projectPath, workbench);
+
+        const element = await driver.findElement(By.className('approot-select-value'));
```
filip131311: This is more destinct the "webview" but is there a way of adding custom tags for tests to make sure we test the thing we expect?

packages/vscode-extension/test/utils/projectLauncher.js
anna-malina: The best solution would be to add id value to either the checked elements or the some test attributes, like data-test="input-email". Can this be added to main elements in the views? Indeed, catching by class is not recommended.

packages/vscode-extension/test/utils/projectLauncher.js
anna-malina: added via visual studio code settings (commit a366ddfdc4b6291fccfbc236cce5a46f6b4ea779)

packages/vscode-extension/test/utils/projectLauncher.js
anna-malina: I have added a tags named data-test for used elements in tests (commit a366ddfdc4b6291fccfbc236cce5a46f6b4ea779)

packages/vscode-extension/test/utils/projectLauncher.js
anna-malina: I have separated this into first test, if its fails - other tests will be paused (commit a366ddfdc4b6291fccfbc236cce5a46f6b4ea779) [smoke.test.js 44-51]

packages/vscode-extension/test/ui-tests/example.test.js
anna-malina: Right, I removed the unnecessary waits from before / after functions. Inside the tests now I am using waiting until (await driver.wait(until.elementIsVisible(element))

packages/vscode-extension/test/ui-tests/example.test.js
anna-malina: I have added data-test attributes

packages/vscode-extension-tester/utils/projectLauncher.js
```diff
@@ -0,0 +1,34 @@
...
+        path.resolve(__dirname, projectPath),
+
+        async () => {
+            console.log('Additional wait after workbench is ready');
+            await new Promise(resolve => setTimeout(resolve, 3000));
+        }
+    );
+
+    await workbench.executeCommand('RNIDE.openPanel');
+    await driver.sleep(5000);
```
filip131311: is there a way to check that pannel was opened programmatically to not have a fixed await time each time?

packages/vscode-extension-tester/utils/projectLauncher.js
```diff
@@ -0,0 +1,34 @@
...
+
+    const webview = await driver.findElement(By.className('webview'));
```
filip131311: is it necessary to identify this component by "webview" classname? Is it possible for other webviews to be opened in the enviorment? For example one of our own devtools like network Inspector

packages/vscode-extension-tester/utils/projectLauncher.js
```diff
@@ -0,0 +1,34 @@
...
+    await waitForElement(driver, webview);
+    await driver.switchTo().frame(webview);
+
+    const iframe = await driver.findElement(By.tagName('iframe'));
```
filip131311: same comment as webview

packages/vscode-extension-tester/ui-tests/smoke.test.js
```diff
@@ -0,0 +1,96 @@
...
+        assert.equal(title, 'Radon IDE — ' + texts.pageTitle, `Page title should be: ${texts.pageTitle}`);
+
+        const approot = await driver.findElement(By.css('[data-test="approot-select-value"]'));
+        await waitForElement(driver, approot);
+
+        const text = await approot.getText();
+        assert.equal(text, texts.expectedProjectName, "Text of the element should be a name of the project");
+    });
+
+    it("should the correct context be displayed based on the availability of devices after opening a project", async function () {
```
filip131311, line 65: this should be a few separate tests that mock the scenario that they are testing

filip131311: If it is a lot of work to manipulate the environment in order to have differed devices installed please don't include this in this PR as it is already quite large. We can do it in the followup

packages/vscode-extension-tester/ui-tests/smoke.test.js
```diff
@@ -0,0 +1,96 @@
...
+        driver = browser.driver;
+        if (!driver) {
+            console.error("Failed to obtain driver from VSBrowser.");
+            return;
+        }
+
+        await browser.waitForWorkbench();
+
+        workbench = new Workbench();
+        await new Promise((resolve) => setTimeout(resolve, 500));
```
filip131311: can we observe programmatically if workbench is ready instead?

filip131311 (commented):
Left some inline comments

packages/vscode-extension-tester/utils/projectLauncher.js
anna-malina: Unfortunately, I was not able to improve it, I tried to wait until the webview in several ways is displayed, however, this action ended in failure, because it was performed before the panel fully opened. The same with catching other elements like iframes. I have not found a better way. The current way is also included in the examples vscode-extension-tester-example. At the moment I have just I reduced the unnecessarily long time. [commit c2c39c72595067450b2f461d0a3ffe646c8fce1c]

packages/vscode-extension-tester/utils/projectLauncher.js
anna-malina: From what I have read and tested these tags are outside extenstion, at the level vs code so it is difficult to add attributes (like ids, data-test). In selenium it is necessary to switch first to the parent iframe-webview, and only after that you can to the context of the second iframe (you can not combine it). These actions are necessary to have access to the elements inside the radon extension, I've fixed it a little bit this catching on more precise css.

Unfortunately, the network tab in the extension also displays only an iframe with the webview class so there may be 2 same elements (iframe.webview) - but this view is shown only in the already loaded Radon IDE panel and only after the application is built, at this stage it doesn't cause problems. When writing tests for Network, I will need to catch this network tab more carefully.

packages/vscode-extension-tester/utils/projectLauncher.js
anna-malina: Improved via css title attribute [commit c2c39c72595067450b2f461d0a3ffe646c8fce1c]

packages/vscode-extension-tester/ui-tests/smoke.test.js
anna-malina: The only way I have found is to wait for the already displayed items inside the webview which is done in the next steps in the openProject function. I tested and this waiting was unnecessary here, the test passes without it because the loading of the view elements is checked later. I have removed this line. [commit c2c39c72595067450b2f461d0a3ffe646c8fce1c]

PR #1307, opened 2025-07-07
Move additional reload options trigger to more visible position

filip131311:
This PR moves an arrow that is triggering additional reload options to more visible position in order to make this functionality more visible. 

### Before: 
Screenshot 2025-07-07 at 16 37 59
Screenshot 2025-07-07 at 16 38 08
Screenshot 2025-07-07 at 16 38 16
### After: 

https://github.com/user-attachments/assets/c03d1977-6d36-408c-98ce-acc7077a2215

### How Has This Been Tested: 

Run any test application. 

I didn't do it yet I will upon approval 

If applicable (for UI changes or new features) provide a link to a PR that updates the relevant documentation pages.

### DIFF
--- a/packages/vscode-extension/src/webview/components/IconButtonWithOptions.tsx
+++ b/packages/vscode-extension/src/webview/components/IconButtonWithOptions.tsx
@@ -1,4 +1,4 @@
-import { forwardRef, useState } from "react";
+import { forwardRef } from "react";
 import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
 import classNames from "classnames";
 import IconButton, { IconButtonProps } from "./shared/IconButton";


filip131311:
@kmagiera I applied your suggestions: 
Screenshot 2025-07-08 at 18 17 51
Screenshot 2025-07-08 at 18 17 33
Screenshot 2025-07-08 at 18 17 45
 
to make the highlights more natural  I switched all of the buttons styles to more of a square-ish shape

@kacperkapusciak what do you think?

kacperkapusciak (commented):
Remove this separator pls.

From the design point of view it is used incorrectly. Separator separates different parts of the UI where here it is used inside a button which should be whole.

image

Use spacing instead. Margin between the down arrow and the refresh icon should be smaller than the spacings between different buttons.

jwajgelt:
I'm with @kacperkapusciak on this one, the separator does look out of place if you don't already have the context that reload and the drop-down are a part of a single button.
Maybe we could have it only appear on hover?

PR #1308, opened 2025-07-08
Allow selecting from multiple launch configurations

jwajgelt:
- allows configuring multiple Radon launch configurations at the same time
- adds the custom launch configurations to the "app root select" drop-down
- allows configuring each LaunchConfig separately
- fixes eas build profile / xcode scheme suggestions in Launch Configuration modal to always suggest values for selected app root

### How Has This Been Tested: 
- open a monorepo in Radon
- if you had previously configured a launch config, Radon should use it, and you should see it in the approot drop-down
- removing the config from `launch.json` should remove it from the drop-down
- close the `launch.json` file
- press "Add custom launch config" in the approot menu
- (BONUS POINTS) selecting an AppRoot should add appropriate options to "IOS Scheme" and "EAS Build" profiles selects
- after selecting the options, press "Save"
- the config should be added to the `launch.json` file and should be visible in the approot select
- add multiple configs
- verify selecting different configs restarts the selected device, stops background devices, and correctly applies the configured options

#1312

### DIFF
--- a/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
+++ b/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
@@ -63,6 +63,7 @@ function LaunchConfigurationView({ launchConfigToUpdate }: LaunchConfigurationVi
         </h2>
         <div className="launch-configuration-button-group">
           <Button
+            className="button-secondary"
             onClick={() =>
               openModal(
                 "Launch Configuration",
@@ -72,7 +73,6 @@ function LaunchConfigurationView({ launchConfigToUpdate }: LaunchConfigurationVi
             Cancel
           </Button>
           <Button
-            className="launch-configuration-delete"
             onClick={() => {
               project.createOrUpdateLaunchConfiguration(undefined, launchConfigToUpdate);
               closeModal();
@@ -148,7 +148,7 @@ function LaunchConfigurationView({ launchConfigToUpdate }: LaunchConfigurationVi
           <Button
             className="button-secondary"
             onClick={() => {
-              openModal("Delete Launch Configuration", <DeleteConfirmationModal />);
+              openModal("", <DeleteConfirmationModal />);
             }}>
             Delete
           </Button>


packages/vscode-extension/src/panels/AppRootConfigController.ts
```diff
@@ -140,11 +77,18 @@ export class LaunchConfigController implements Disposable, LaunchConfig {
     return applicationRoots.map(readApplicationRoot);
   }
 
-  async getAvailableXcodeSchemes() {
-    const appRoot = this.appRootFolder;
-    const sourceDir = getIosSourceDir(appRoot);
+  async getAvailableXcodeSchemes(appRoot?: string) {
+    if (!appRoot) {
```
filip131311: this does not look correct to me, 
we are inside `AppRootConfigController` which seems to be a stateful component I think it should be a source of truth about the current appRoot to avoid situations when methods return misaligned results. 

If it is a utility function that is to be used for multiple appRoots regardless of the current one I still think we should force  `appRoot` and only the caller could perform `findAppRootCandidates` if necessary

packages/vscode-extension/src/panels/AppRootConfigController.ts
```diff
@@ -140,11 +77,18 @@ export class LaunchConfigController implements Disposable, LaunchConfig {
...
+      const appRootCandidates = findAppRootCandidates();
+      if (appRootCandidates.length === 0) {
+        return [];
+      }
+      appRoot = appRootCandidates[0];
+    }
+    const absoluteAppRoot = path.resolve(workspace.workspaceFolders![0].uri.fsPath, appRoot);
```
filip131311: this seems like a utility  that is needed in multiple places could you move it to a separate abstraction for easier decoupling from vscode in the future?

packages/vscode-extension/src/panels/AppRootConfigController.ts
```diff
@@ -159,28 +103,17 @@ export class LaunchConfigController implements Disposable, LaunchConfig {
     return await findXcodeScheme(xcodeProject);
   }
 
-  async getAvailableEasProfiles(): Promise<EasBuildConfig> {
-    const appRoot = this.appRootFolder;
-    const easConfig = await readEasConfig(appRoot);
+  async getAvailableEasProfiles(appRoot?: string): Promise<EasBuildConfig> {
+    if (!appRoot) {
+      const appRootCandidates = findAppRootCandidates();
```
filip131311: same comments as before

filip131311 (commented):
Left some initial coments inline

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -257,50 +252,57 @@ interface appRootConfigurationProps {
...
-  applicationRoots,
-  addCustomApplicationRoot,
-}: appRootConfigurationProps) {
+function AppRootConfiguration({ appRoot, update, applicationRoots }: appRootConfigurationProps) {
   const customAppRootInputRef = useRef<HTMLInputElement>(null);
 
-  const [customAppRootButtonDisabled, setCustomAppRootButtonDisabled] = useState(true);
+  function getInitialSelectedValue(): string {
+    if (appRoot === undefined) {
+      return "Auto";
```
filip131311, line 307: Note: while we are working on this part o=f radon we should change Auto to `DefaultAppRootName (auto)` and same for all configurations 

(not in this PR)

filip131311 (commented):
While testing I noticed there is no UI way of removing custom configurations, I think It might be useful to add them.

packages/vscode-extension/src/panels/AppRootConfigController.ts
```diff
@@ -0,0 +1,108 @@
...
+      displayName: appRootConfig.displayName,
+    };
+  }
+  throw new Error("Could not read config file.");
+}
+
+function readApplicationRoot(appRootPath: string): ApplicationRoot {
+  const appRootAbsolutePath = path.resolve(workspace.workspaceFolders![0].uri.path, appRootPath);
+  try {
+    return readApplicationRootFromStaticConfig(appRootPath, appRootAbsolutePath + "/app.json");
```
kmagiera, line 33: would be better to use path.join here (not to break windows support 😉 )

packages/vscode-extension/src/panels/AppRootConfigController.ts
jwajgelt: You don't want to get the schemes for the "current" appRoot, you want the ones for given launch config options.
If the launch config options don't specify one, we need to somehow figure out the "default" one. The frontend cannot call `findAppRootCandidates`, and I don't want to depend on the `ApplicationRoot`s returned from `getAvailableApplicationRoots` having the correct order. 
We could probably pass the "default" appRoot to the frontend (somewhere), but that seems overly complicated. I like the current behaviour better.

packages/vscode-extension/src/panels/AppRootConfigController.ts
```diff
@@ -0,0 +1,108 @@
...
+    path: appRootPath,
+    name: path.basename(appRootPath),
+  };
+}
+
+export class AppRootConfigController implements AppRootConfig {
+  async getAvailableApplicationRoots() {
+    const workspacePath = workspace.workspaceFolders![0].uri.fsPath;
+    const applicationRootsCandidates = findAppRootCandidates().map((candidate) => {
+      return "./" + path.relative(workspacePath, candidate);
```
kmagiera, line 58: do we need to prefix relative paths with `"./"` – typically no prefix means relative and `'/'` or `C:\\` etc is absolute

packages/vscode-extension/src/panels/AppRootConfigController.ts
```diff
@@ -0,0 +1,108 @@
...
+import { ApplicationRoot, AppRootConfig } from "../common/AppRootConfig";
+import { extensionContext, findAppRootCandidates } from "../utilities/extensionContext";
+import { findXcodeProject, findXcodeScheme } from "../utilities/xcode";
+import { Logger } from "../Logger";
+import { getIosSourceDir } from "../builders/buildIOS";
+import { readEasConfig } from "../utilities/eas";
+import { EasBuildConfig } from "../common/EasConfig";
+import { requireNoCache } from "../utilities/requireNoCache";
+
+const CUSTOM_APPLICATION_ROOTS_KEY = "custom_application_roots_key";
```
kmagiera, line 12: what is this for? I can only see that we read from it but never write?

packages/vscode-extension/src/panels/WebviewController.ts
```diff
@@ -43,7 +43,7 @@ export class WebviewController implements Disposable {
       ["Project", () => this.ide.project as object],
       ["DeviceSessionsManager", () => this.ide.project.deviceSessionsManager as object],
       ["WorkspaceConfig", () => this.ide.workspaceConfigController as object],
-      ["LaunchConfig", () => this.ide.project.launchConfigurationController as object],
+      ["LaunchConfig", () => this.ide.project.appRootConfigController as object],
```
kmagiera: do we want to rename `appRootConfigController` to `LaunchConfigController`?

packages/vscode-extension/src/panels/AppRootConfigController.ts
jwajgelt: We now set the value for "Auto" on the frontend, instead of leaving the value empty value in the launch config.
This means we can just pass the value we want to set here.

filip131311 (approved):
Looks much better now, but pleas update and link documentation before merging

packages/vscode-extension/src/webview/providers/ApplicationRootsProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+  );
+}
+
+const ApplicationRootsContext = createContext<ApplicationRoot[]>([]);
+
+export default function ApplicationRootsProvider({ children }: PropsWithChildren) {
+  const [applicationRoots, setApplicationRoots] = useState<ApplicationRoot[]>([]);
+
+  useEffect(() => {
+    appRootConfigProxy.getAvailableApplicationRoots().then(setApplicationRoots);
```
kmagiera, line 45: this won't allow us to get the updated list of app roots ever. Are you sure you want this to work in that way?

kmagiera (commented):
See the PR is already accepted but added a couple of inline comments from myself

packages/vscode-extension/src/panels/AppRootConfigController.ts
jwajgelt: This code is copied over from the old `LaunchConfigController`, I didn't change it now.

packages/vscode-extension/src/panels/AppRootConfigController.ts
jwajgelt: 1. This code is copied over from the old `LaunchConfigController`, I didn't change it now.
2. this path is only used for display purposes, and the leading `./` helps readability (IMO)

packages/vscode-extension/src/panels/AppRootConfigController.ts
jwajgelt: 1. This code is copied over from the old LaunchConfigController
2. previously, we used it to allow users to add and use custom appRoots without needing to configure them in `launch.json`. I removed that functionality (IMO it didn't improve the user experience over just having a "Custom" option + text field in the Launch Configurations modal), but we should still read it to avoid breaking existing setups

packages/vscode-extension/src/panels/WebviewController.ts
jwajgelt: No, but we probably want to change the binding name from `"LaunchConfig"` to `"AppRootConfig"`.
The new `AppRootConfig` is the parts of the old `LaunchConfig`/`LaunchConfigController` that were not `LaunchConfig`-related (since those were moved to the new `LaunchConfigManager` and the are interacted with through `Project`/`ProjectState` now instead)

packages/vscode-extension/src/webview/providers/ApplicationRootsProvider.tsx
jwajgelt: At this time, it's never updated anyway.
If in the future we decide to update it (by watching the workspace files and searching for the new approots whenever something changes), we will have to change it here as well.

PR #1309, opened 2025-07-08
Add a fallback for simulator server image when no is provided

filip131311:
This PR adds a fallback for when simulator server is killed but the webview still tries to read the stream. 

To do this in a way that will not disrupt our  ability to diagnose problems within the IDE we introduce new `IS_DEV` value passed down to the webview and conditionally render the visible (and annoying)  "NO IMAGE" warning if we are in the development mode. 

In production we replace the stream with a black screen to mask any imperfections in our state management and provide better user experience.

Dev: 

Screenshot 2025-07-08 at 12 22 42

Production: 

https://github.com/user-attachments/assets/e81cf7bd-a84e-4fc8-8820-7cf70e6da321

### How Has This Been Tested: 

- Run any test application and wait for the stream to be visible 
- find `simulator-server-macos` process in activity monitor and kill it 
- do it both for prod and dev environment.

Not applicable

### DIFF
--- a/packages/vscode-extension/src/webview/Preview/MjpegImg.tsx
+++ b/packages/vscode-extension/src/webview/Preview/MjpegImg.tsx
@@ -1,6 +1,13 @@
 import { useEffect, forwardRef, RefObject } from "react";
+import { IS_DEV } from "../providers/UtilsProvider";
 
-const NO_IMAGE_DATA = "data:,";
+const NO_IMAGE_DATA_PRODUCTION =
+  "data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' width='200' height='200'><rect width='100%' height='100%' fill='black'/></svg>";
+
+const NO_IMAGE_DATA_DEV =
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/webview/Preview/MjpegImg.tsx
```diff
@@ -61,7 +68,14 @@ const MjpegImg = forwardRef<HTMLImageElement, React.ImgHTMLAttributes<HTMLImageE
       };
     }, [ref]);
 
-    return <img ref={ref} {...rest} />;
+    function handleError(e: React.SyntheticEvent<HTMLImageElement, Event>) {
+      const img = e.currentTarget;
+      if (img.src !== NO_IMAGE_DATA) {
+        img.src = NO_IMAGE_DATA;
+      }
```
jwajgelt, line 75: NIT: You could just 
```suggestion
        img.src = NO_IMAGE_DATA;
```
it should be noop if the condition isn't met.

jwajgelt (approved):

PR #1310, opened 2025-07-08
Add device rotation in the IDE

pFornagiel:
## Descriptions

This PR introduces device rotation option during development, visual rotation preview, vscode rotate commands with keyboard shortcuts.

Introduced features:
- triggering device rotation in the simulation server from extension level
- proper display of rotated device in preview
- changes to preview Device layout
- scrolling support, when device is set to fixed size in the preview
- "Rotate device" dropdown option in device settings
- "Rotate device clockwise" and "Rotate device anticlockwise" RadonIDE commands
- shortcuts for above commands
- saving the rotation configuration in vscode workspace options

image

image

https://github.com/user-attachments/assets/53b031e2-e913-4cac-aada-8f091e93f024

### How Has This Been Tested: 

- run Radon IDE extension panel, and launch app in selected device. Rotation should be possible via device settings and commands.

Updated docs for Device Settings nad Keyboard Shortcuts in #@pFornagiel/docs-rotation


kmagiera (commented):
Haven't reviewed the code just yet, but we need to add the following to the test plan:
1) interacting with the app via touch (multi touch in particular) after rotation) – this can work differently on Android and iOS and not sure if the touch point positions needs to be rotated along with the devide or we should keep sending coordinates relative to top/left
2)  element inspector after rotation (element inspector both uses the touch coordinates and also overlays some content of the device screen which may behave differently after this change)
3) react-scan after rotation (react scan also overlays things over the screen when rotated)
4) test debugger overlay when paused on breakpoints (again, this is an UI that gets overlayed)
4) test reloads / app restarts to verify that the content and progress bars look ok on the rotated device.

packages/vscode-extension/src/common/Project.ts
```diff
@@ -128,6 +129,8 @@ export type AppPermissionType = "all" | "location" | "photos" | "contacts" | "ca
 
 export type DeviceButtonType = "home" | "back" | "appSwitch" | "volumeUp" | "volumeDown";
 
+export type DeviceRotationType = "Portrait" |  "PortraitUpsideDown" | "LandscapeLeft"  | "LandscapeRight";
```
jwajgelt: NIT: I'd use an enum for this instead
```suggestion
export enum DeviceRotation {
  Portrait = "Portrait",
  PortraitUpsideDown = "PortraitUpsideDown",
  LandscapeLeft = "LandscapeLeft",
  LandscapeRight = "LandscapeRight",           
}
```
this way you get IDE suggestions when typing the values and it's harder to make a typo (probably intellisense and/or copilot can handle it just fine with the union type though, and types make sure that typos are rejected, so it's not that important)
If we ever changed the values for whatever reason, it would also mean we don't have to hunt down every single "Portrait" string in the codebase, you just change the enum value. This is probably a stronger reason for enum vs union of string literal types

packages/vscode-extension/src/devices/DeviceBase.ts
```diff
@@ -169,7 +175,7 @@ export abstract class DeviceBase implements Disposable {
 
   public sendKey(keyCode: number, direction: "Up" | "Down") {
     // iOS simulator has a buggy behavior when sending cmd+V key combination.
-    // It sometimes triggers paste action but with a very low success rate.
+    // It sometimes triggers paste action but with a very §w success rate.
```
jwajgelt: I think this was committed by mistake

packages/vscode-extension/src/devices/DeviceBase.ts
```diff
@@ -209,6 +215,17 @@ export abstract class DeviceBase implements Disposable {
     this.preview?.sendWheel(point, deltaX, deltaY);
   }
 
+  public sendRotate(rotation: DeviceRotationType) {
+    // Preview may not be started yet, but we still wish to be able to rotate during init of device.
+    // The rotation is set in constructor of device and is gotten from the workspace configuration anyways.
+    // if(!this.preview) {
+    //   throw new Error("Preview not started");
+    // }
```
jwajgelt:
```suggestion
```

packages/vscode-extension/src/extension.ts
```diff
@@ -374,6 +378,33 @@ async function captureScreenshot() {
...
+async function rotateDeviceAnticlockwise(){
+  const configuration = workspace.getConfiguration("RadonIDE");
+  const rotation = configuration.inspect<DeviceRotationType>("deviceRotation")?.workspaceValue;
+  if(!rotation){
+    await configuration.update("deviceRotation", "Portrait", false);
+    return;
+  }
+  const currentIndex = ROTATIONS.indexOf(rotation);
+  const newIndex = (currentIndex - 1 + ROTATIONS.length) % ROTATIONS.length;
+  await configuration.update("deviceRotation", ROTATIONS[newIndex], false);
```
jwajgelt: Personally, I don't like the idea of controlling the rotation by writing to the config, I'd leave that to the implementation.
Instead, you can access `Project` here by using `IDE.getInstanceIfExists()?.project` and calling `dispatchRotate` on it directly.
To get the current rotation value, you should be able to get the `projectState` from it as well.

packages/vscode-extension/src/extension.ts
```diff
@@ -374,6 +378,33 @@ async function captureScreenshot() {
...
+const configuration = workspace.getConfiguration("RadonIDE");
+  const rotation = configuration.inspect<DeviceRotationType>("deviceRotation")?.workspaceValue;
+  if(!rotation){
+    await configuration.update("deviceRotation", "Portrait", false);
+    return;
+  }
+  const currentIndex = ROTATIONS.indexOf(rotation);
+  const newIndex = (currentIndex + 1) % ROTATIONS.length;
+  await configuration.update("deviceRotation", ROTATIONS[newIndex], false);
+}
```
jwajgelt: This is a nitpick:
You could extract the common code between these two functions to some `async function rotateDevice(rotation)` and then have `rotateDevice(Anti)Clockwise` call
```
async function rotateDeviceClockwise() {
  const currentIndex = ...;
  const newIndex = (currentIndex + 1) % ROTATIONS.length;
  await rotateDevice(ROTATIONS[newIndex]);
}
```
or even have `rotateDevice(direction: -1 | 1)` and handle selecting the rotation inside it:
```
async function rotateDeviceClockwise() {
  await rotateDevice(1);
}

async function rotateDeviceAntiClockwise() {
  await rotateDevice(-1);
}
```

it's not important though, feel free to leave this as is.

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -59,6 +64,22 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
...
+  }> = [
+    { label: "Portrait", value: "Portrait", icon: "device-mobile", rotation: "0deg" },
+    { label: "Landscape Left", value: "LandscapeLeft", icon: "device-mobile", rotation: "-90deg" },
+    {
+      label: "Landscape Right",
+      value: "LandscapeRight",
+      icon: "device-mobile",
+      rotation: "90deg",
+    },
+  ];
```
jwajgelt: NIT: I'd move this outside of the component:
1. so that you don't create new objects each time the component renders
2. that's how the other options here are handled
3. I personally think it's just nicer that way

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -105,11 +107,33 @@ function Preview({
...
+        newX = y;
+        newY = 1 - x;
+        break;
+      default:
+        break;
+    }
+
+    let clampedX = clamp(newX, 0, 1);
+    let clampedY = clamp(newY, 0, 1);
+
```
jwajgelt: I think this function deserves either a different name (along the lines of `getNormalizedDeviceCoordinates`), or a comment explaining what it does -- or even both.

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -445,6 +469,25 @@ function Preview({
...
 
+  useEffect(() => {
+    const loadingBackgroundElement = loadingBackgroundRef.current;
+    if (!loadingBackgroundElement) {
+      return;
+    }
+
+    const resizeObserver = new ResizeObserver((entries) => {
+      for (const entry of entries) {
+        setLoadingBackgroundHeight(entry.contentRect.height);
```
jwajgelt: I wonder, should we throttle this? I'm not sure how often the `ResizeObserver` fires, but the `Preview` component is pretty heavy at the moment, I'm not sure we want to render it each time the observer fires.
It's possible it's fine as is, just wanted to make sure we're considering this, and not just overlooking a potential issue.

packages/vscode-extension/src/webview/components/PreviewLoader.css
```diff
@@ -1,5 +1,121 @@
...
+    &.small {
+      .preview-loader-load-info {
+        font-size: 0.6em;
+      }
+
+      .preview-loader-waiting-actions {
+        font-size: 0.4em;
+        flex-direction: row !important;
+        /* gap: 0.5em; */
+        /* text-align: left; */
```
jwajgelt:
```suggestion
```
This should probably be removed before merging.

packages/vscode-extension/src/webview/components/PreviewLoader.css
```diff
@@ -1,5 +1,121 @@
...
+.preview-loader-wrapper.rotated.small,
+.preview-loader-wrapper.rotated.medium {
+  width: 100%;
+}
+
+/* .preview-loader-wrapper.small.rotated {
+}
+
+.preview-loader-wrapper.medium.rotated {
+} */
```
jwajgelt:
```suggestion
```
This should probably be removed before merging.

packages/vscode-extension/src/webview/components/PreviewLoader.tsx
```diff
@@ -20,20 +20,52 @@ const startupStageWeightSum = StartupStageWeight.map((item) => item.weight).redu
   0
 );
 
+const BREAKPOINT_SMALL_ROTATED = 450;
+const BREAKPOINT_MEDIUM_ROTATED = 700;
```
jwajgelt: NIT:
```suggestion
const BREAKPOINT_SMALL_LANDSCAPE = 450;
const BREAKPOINT_MEDIUM_LANDSCAPE = 700;
```
makes more sense to me to have PORTRAIT-LANDSCAPE dichotomy, instead of `PORTRAIT`-`ROTATED`.

packages/vscode-extension/src/webview/providers/WorkspaceConfigProvider.tsx
```diff
@@ -30,10 +32,12 @@ const WorkspaceConfigContext = createContext<WorkspaceConfigContextType>({
 
 export default function WorkspaceConfigProvider({ children }: PropsWithChildren) {
   const [config, setConfig] = useState<WorkspaceConfigProps>(INITIAL_WORKSPACE_CONFIG);
+  const { project } = useProject();
 
   useEffect(() => {
     function watchConfigChange(e: WorkspaceConfigProps) {
       document.body.setAttribute("data-use-code-theme", `${e.themeType === "vscode"}`);
+      project.dispatchRotate(e.deviceRotation);
```
jwajgelt: I don't like that we apply the initial config in `DeviceBase`, but watch for updates here. I'd prefer if both happened in the same place.
Personally, I feel the best place would be in `DeviceSession`, so that `DeviceBase` doesn't have to care about `vscode`-specific APIs.

packages/vscode-extension/src/common/Project.ts
pFornagiel: Refactored in 0c10c7b.

packages/vscode-extension/src/devices/DeviceBase.ts
pFornagiel: Fixed in 47baec2c9aa71728ff1a2d92d76396052a2dae55

packages/vscode-extension/src/devices/DeviceBase.ts
pFornagiel: Comment removed  in 47baec2c9aa71728ff1a2d92d76396052a2dae55

packages/vscode-extension/src/webview/components/PreviewLoader.css
pFornagiel: Comment removed in 47baec2c9aa71728ff1a2d92d76396052a2dae55

packages/vscode-extension/src/webview/components/PreviewLoader.tsx
pFornagiel: Renamed in 47baec2c9aa71728ff1a2d92d76396052a2dae55

packages/vscode-extension/src/webview/components/PreviewLoader.css
pFornagiel: Comment removed in 47baec2c9aa71728ff1a2d92d76396052a2dae55

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
pFornagiel: Refactored in 6d7090e3e2fd267f944e0a1c469de82f7299b828

packages/vscode-extension/src/webview/components/Preview.tsx
pFornagiel: Given that the `loadingBackgroundHeight` state is updated whenever the height changes, this could be a performance issue, that's an oversight from my side. 

To avoid the unnecessary serenaders in Preview, I've moved both the state and the observer to the PreviewLoader component (in order to re-render only it's subtree, not the entire Preview component) and made it so the state changes happen only whenever given threshold is passed. The observer should be performant in this case, without triggering too much re-renders.

Implemented in 46beb74a

packages/vscode-extension/src/webview/providers/WorkspaceConfigProvider.tsx
pFornagiel: Rotation configuration reading and updates have been disconnected from DeviceBase and WorkspaceConfigProvider and moved to `project.ts` file. The configuration now updates upon invoking project.dispatchRotate, making Devices independent from said vscode API. Changes have been made to adjust files using the rotation functionality.

Implemented in 07fdfbc5.

packages/vscode-extension/src/extension.ts
pFornagiel: Rotation configuration reading and updates have been moved to `project.ts` file. The updates to rotation are now performed by invoking dispatchRotate as requested.

Implemented in 07fdfbc5.

packages/vscode-extension/src/extension.ts
pFornagiel: Code has been refactored to include enum RotationDirection and rotateDevice function:
```ts
enum RotationDirection {
  Clockwise = -1,
  Anticlockwise = 1
}

async function rotateDevice(direction: RotationDirection) {
  const project = IDE.getInstanceIfExists()?.project;
  if (!project) {
    throw new Error("Radon IDE is not initialized yet.");
  }
  const projectState = await project.getProjectState();
  const rotation = projectState.rotation;
  const currentIndex = ROTATIONS.indexOf(rotation);
  const newIndex = (currentIndex - direction + ROTATIONS.length) % ROTATIONS.length;
  project.dispatchRotate(ROTATIONS[newIndex]);
}

async function rotateDeviceAnticlockwise() {
  await rotateDevice(RotationDirection.Anticlockwise)
}

async function rotateDeviceClockwise() {
  await rotateDevice(RotationDirection.Clockwise)
}
```

Implemented in 07fdfbc5

packages/vscode-extension/lib/instrumentation.js
```diff
@@ -88,7 +88,7 @@ function getWindowRect() {
       height: screenHeight,
     };
   } else {
-    const { width: windowWidth, height: windowHeight } = Dimensions.get("window");
+    const { width: windowWidth, height: windowHeight } = Dimensions.get("screen");
```
jwajgelt:
```suggestion
    const { width: screenWidth, height: screenHeight } = Dimensions.get("screen");
```
Then, since this is the same call in both branches of the `if`/`else`, you can lift it up.

packages/vscode-extension/lib/orientation.js
```diff
@@ -0,0 +1,178 @@
...
+// the orientation change is not reported.
+
+// On iOS, the orientation events are fired when the DEVICE ROTATES,
+// because the API is based on device sensors.
+// Hence, orientation change event is not fired when the app starts,
+// and the app's initial orientation is not known until the first rotation occurs - this
+// is why we need to use the Dimensions API and appOrientationInit to properly handle it on the frontend.
+// Additionally, when app does not support given orientation, the event still fires (new orientation is reported
+// even though the app remains in previous mode, only because the device was rotated). This is why, we need to
+// infer some information about the app's orientation based on the device orientation and DO NOT SUPPORT PORTRAIT-SECONDAY
```
jwajgelt:
```suggestion
// infer some information about the app's orientation based on the device orientation and ASSUME THAT PORTRAIT-SECONDARY IS NOT SET
```
1. typo
2. I would underline that this is an assumption that we make about the application's behavior.

packages/vscode-extension/lib/orientation.js
```diff
@@ -0,0 +1,178 @@
...
+// the other one is supported as well.
+
+// For IOS the orientation event is fired after the DEVICE rotates, but before the APP actually rotates.
+// This means, we are experiencing a lag between current orientation and information from Dimensions API,
+// which is why additional listener is added to the Dimensions API to amend the effect.
+
+// Mapping from namedOrientationDidChangeEvent names to DeviceRotationType names or "Landscape" specific case
+// sent to the extension.
+// "Landscape" case happens when we cannot infer the detailed landscape orientation of the app and
+// leave the interpretation to the frontend of the extension.
```
jwajgelt:
```suggestion
/**
 * Mapping from namedOrientationDidChangeEvent names to DeviceRotationType names or "Landscape" specific case
 * sent to the extension.
 * "Landscape" case happens when we cannot infer the detailed landscape orientation of the app and
 * leave the interpretation to the frontend of the extension.
 **/
```
You can turn these kinds of comments into doc-comments, then you can see them when you hover over the commented item (`orientationMapping`) in VSCode (any many other IDEs).

packages/vscode-extension/lib/orientation.js
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    // for ios landscape-primary -> LandscapeRight
+    if (orientation === "landscape-primary") {
+      return "LandscapeRight";
+    }
+  }
+
+  // for android simply return the mapping
+  return orientationMapping[orientation];
```
jwajgelt: Maybe you could rename `orientationMapping` -> `androidOrientationMapping` since it only seems to be used for that.

packages/vscode-extension/lib/orientation.js
```diff
@@ -0,0 +1,178 @@
...
+// "Landscape" case happens when we cannot infer the detailed landscape orientation of the app and
+// leave the interpretation to the frontend of the extension.
+const orientationMapping = {
+  "portrait-primary": "Portrait",
+  "portrait-secondary": "PortraitUpsideDown",
+  "landscape-primary": "LandscapeLeft",
+  "landscape-secondary": "LandscapeRight",
+};
+
+let currentMappedOrientation = null;
```
jwajgelt: NIT: I'd rename this to 
```suggestion
let currentAppOrientation = null;
```
to highlight that this is meant to store the way that the app itself is rotated (rather than the device).
Similarly, elsewhere where you refer to the "mapped orientation", I'd rename it to "app orientation".

packages/vscode-extension/lib/orientation.js
```diff
@@ -0,0 +1,178 @@
...
+  if (setupCompleted) {
+    return cleanup; // Return cleanup function if already set up
+  }
+  setupCompleted = true;
+
+  initializeOrientationAndSendInitMessage();
+
+  
+  // The below line may throw a warning on iOS, because RCTUIManager does not implement the addListener method
+  orientationEventEmitter = new NativeEventEmitter(RCTUIManager);
```
jwajgelt: I don't know if you've already done that, but we should check on an application running the old RN architecture if the `RCTUIManager` behaves the same there.
At least our `expo-51` app uses it, you should be able to test with that.

packages/vscode-extension/lib/orientation.js
```diff
@@ -0,0 +1,178 @@
...
+  orientationEventEmitter.addListener("namedOrientationDidChange", handleOrientationChange);
+
+  if(Platform.OS === "ios") {
+    console.log("Above warnings may be ignored - RCTUIManager does not implement the addListener method, but is required to succesfully listen to namedOrientationDidChange events on iOS.");
+  }
```
jwajgelt: Does this mean the app logs warnings at this step?
If so, we should find a way to suppress them.

packages/vscode-extension/lib/orientation.js
```diff
@@ -0,0 +1,178 @@
...
+
+  // Dimension change lags behind Orientation change in the IOS case, so we add a second listener to amend the effect
+  // Explained in the context above.
+  dimensionEventSubscription = Dimensions.addEventListener("change", handleDimensionsChange);
+
+  return cleanup;
+}
+
+// Cleanup function to remove listeners and reset state
+export function cleanup() {
```
jwajgelt:
```suggestion 
function cleanup() {
```
This doesn't need to be exported, right?

jwajgelt:
```suggestion
  return function cleanup() {
```
You can move the `cleanup()` function into the body of `setup()`. Then you can declare `orientationEventEmitter` and `dimensionEventSubscription` inside `setup` as well, instead of having them at the module scope.

packages/vscode-extension/lib/orientation.js
```diff
@@ -0,0 +1,178 @@
...
+};
+
+let setupCompleted = false;
+let orientationEventEmitter;
+let dimensionEventSubscription;
+
+export function setup() {
+  if (setupCompleted) {
+    return cleanup; // Return cleanup function if already set up
+  }
```
jwajgelt: I'm not sure this is correct.
`useEffect` in the `wrapper` can be called multiple times, in which case the first one will cleanup the listeners and break all the other applications that expect it to still work.
You could instead keep a counter that is incremented each time `setup` is called and decremented inside `cleanup`, and only clear the event emitters when the counter is zeroed.

packages/vscode-extension/lib/wrapper.js
```diff
@@ -45,6 +44,9 @@ const InternalImports = {
   get setupRenderOutlinesPlugin() {
     return require("./render_outlines").setup;
   },
+  get setupOrientationPlugin() {
+    return require("./orientation").setup;
+  },
```
jwajgelt:
```suggestion
  get setupOrientationListeners() {
    return require("./orientation").setup;
  },
```
the "plugins" are specifically code related to our "tools", while the orientation is a core Radon thing -- I'd call it something more specific.

packages/vscode-extension/lib/wrapper.js
```diff
@@ -341,6 +349,8 @@ export function AppWrapper({ children, initialProps, fabric }) {
 
     InternalImports.setupRenderOutlinesPlugin();
     InternalImports.setupNetworkPlugin();
+    // Initialised eventListeners, cleanup needed
```
jwajgelt: VERY NIT:
I think this comment is excessive, the line below explains itself well enough in my opinion.

packages/vscode-extension/src/common/Project.ts
```diff
@@ -138,6 +140,24 @@ export type AppPermissionType = "all" | "location" | "photos" | "contacts" | "ca
 
 export type DeviceButtonType = "home" | "back" | "appSwitch" | "volumeUp" | "volumeDown" | "power";
 
+export enum DeviceRotationType {
```
jwajgelt: NIT:
```suggestion
export enum DeviceRotation {
```
I'd avoid adding the "Type" suffix to type names.
The type (of a variable) describes _what_ the variable holds. Thus, when you write
```
const rotation = XXX.Portrait;
```
I'd say the `rotation` variable holds a value describing a "rotation", not a "type of rotation" -- hence to me, it reads more naturally to have it be
```
const rotation = DeviceRotation.Portrait;
```
instead of
```
const rotation = DeviceRotationType.Portrait;
```

Same goes for `AppOrientationType` below.

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -355,6 +361,29 @@ export class DeviceSession
         this.emitStateChange();
       }
     });
+    devtools.onEvent("appOrientationChanged", (orientation: AppOrientationType) => {
+      const isLandscape =
+        this.rotation === DeviceRotationType.LandscapeLeft ||
+        this.rotation === DeviceRotationType.LandscapeRight;
+
+      // if the app orientation is equal to "Landscape", it means we do not have enuogh
+      // information on the application site to infer the detailed orientation.
```
jwajgelt:
```suggestion
      // information on the application side to infer the detailed orientation. 
```
typo

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -355,6 +361,29 @@ export class DeviceSession
...
     });
+    devtools.onEvent("appOrientationChanged", (orientation: AppOrientationType) => {
+      const isLandscape =
+        this.rotation === DeviceRotationType.LandscapeLeft ||
+        this.rotation === DeviceRotationType.LandscapeRight;
+
+      // if the app orientation is equal to "Landscape", it means we do not have enuogh
+      // information on the application site to infer the detailed orientation. 
+      // "Landscape" will only be messaged upon the initialisation - after that the app can infer the orientation
+      // based on our assumptions.
```
jwajgelt: I feel that this isn't really relevant here.
Here, we don't really care that the "Landscape" orientation will only be passed during initialization, that's a detail of how the code on the application side works -- all we need is the information that "Landscape" is a partial information about app orientation and we need to heuristically pick something ourselves.

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -355,6 +361,29 @@ export class DeviceSession
...
+      // If the device is in landscape mode, we assume that the app orientation is the same as the device rotation.
+      // Otherwise, it means the app is initialised and does not have previous orientation -
```
jwajgelt:
```suggestion
      // If the device is in landscape mode, we assume that the app orientation is the same as the device rotation.
      // Otherwise, we pick `LandscapeLeft`
```
I think this comment should say _why_ we pick `LandscapeLeft`, not detail the reasons why `"Landscape"` orientation was sent.
The reason, as I understand it, is that the application is forced into starting in landscape mode while the device itself was in portrait, and in that case `LandscapeLeft` seems to be the default, so I'd write something along the lines of

```suggestion
      // If the device is in landscape mode, we assume that the app orientation is the same as the device rotation.
      // Otherwise, we assume the application was started in Landscape mode while the device was oriented in Portrait, and we pick `LandscapeLeft` as the default orientation in that case.
```

jwajgelt: Also, I think it might be nicer to move these comments inside the `if` cases below, so they're directly above the `this.appOrientation = ` assignments they relate to.

packages/vscode-extension/src/project/project.ts
```diff
@@ -415,6 +457,32 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
+        case DeviceRotationType.PortraitUpsideDown:
+          // 180° map (x,y) to (1-x, 1-y)
+          array[i] = { xRatio: 1 - x, yRatio: 1 - y };
+          break;
+        default:
+          // Portrait mode: no transformation needed
+          break;
+      }
+    });
+
```
jwajgelt: I don't think this logic belongs here.
Project is mostly a facade that the frontend can call, and (currently) handles some workspace-specific stuff (like watching the workspace configuration).
This is more about translating between the representation used by the JS part of Radon and the representation that the sim-server requires, which I think _should_ happen in `DeviceBase` (since that's the module that directly talks to it).
I'd move it down to `DeviceBase`.

jwajgelt (changes requested):
Left a first batch of comments, will continue reviewing later.

packages/vscode-extension/src/webview/components/Preview.css
```diff
@@ -82,6 +132,10 @@
 
 .phone-frame {
   pointer-events: none;
+  /* Uncomment if something breaks */
+  /* position: absolute; */
```
jwajgelt:
```suggestion
```
remove before merging

packages/vscode-extension/src/webview/components/Preview.css
```diff
@@ -1,36 +1,85 @@
...
   outline: none;
 }
 
+.phone-wrapper {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  min-width: var(--phone-wrapper-min-width); 
+  min-height: var(--phone-wrapper-min-height); 
+  width: var(--phone-wrapper-width);
```
jwajgelt: It seems like this is never set.

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -105,19 +106,29 @@ function Preview({
...
   const openRebuildAlert = useNativeRebuildAlert();
 
-  function getTouchPosition(event: MouseEvent<HTMLDivElement>) {
+  function getNormalizedTouchCoordinates(event: MouseEvent<HTMLDivElement>) {
+    /*
+    Converts mouse event coordinates to normalized touch coordinates ([0-1] range)
+    relative to the device preview image.
+    Coordinate transformation handling when the device when the device is rotated,
+    done in the dispatchTouches method of project.
+    */
```
jwajgelt: Put this outside the body of the function, as a doc comment
```
/**
 * blah blah blah
 **/
```

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -105,19 +106,29 @@ function Preview({
...
     const imgRect = previewRef.current!.getBoundingClientRect();
+
+    // Normalize coordinates to [0-1] range
     const x = (event.clientX - imgRect.left) / imgRect.width;
     const y = (event.clientY - imgRect.top) / imgRect.height;
-    const clampedX = clamp(x, 0, 1);
-    const clampedY = clamp(y, 0, 1);
+
+    // Ensure coordinates are clamped to [0, 1] range
```
jwajgelt: This comment is unnecessary, if you can't tell what `clamp(x, 0, 1)` does, this comment won't help you.

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -575,6 +582,16 @@ function Preview({
           />
         </div>
       </div>
+
+      {/* Hack needed for css to cache those images */}
```
jwajgelt: You could add to this comment what issues _not_ having this causes.

```suggestion
      {/* Hack needed for css to cache those images. Without this, when the device rotates ... */}
```

packages/vscode-extension/src/webview/hooks/useCanvasRenderer.tsx
```diff
@@ -0,0 +1,107 @@
...
+interface TransformationConfig {
+  angle: number;
+  isPortrait: boolean;
+  rotation: DeviceRotationType;
+}
+
+function useCanvasRenderer(
+  rotation: DeviceRotationType,
+  canvasRef: RefObject<HTMLCanvasElement | null>
+) {
```
jwajgelt: NIT: you can probably move this to `MediaCanvas.tsx`, since that seems to be the only place that should use it.

packages/vscode-extension/src/webview/hooks/useCanvasRenderer.tsx
```diff
@@ -0,0 +1,107 @@
...
+      }
+      const canvas = canvasRef.current;
+      const ctx = canvas.getContext("2d");
+      if (!ctx) {
+        return;
+      }
+
+      const imgWidth =
+        sourceImg instanceof HTMLVideoElement ? sourceImg.videoWidth : sourceImg.width;
+      const imgHeight =
```
jwajgelt: NIT: I'd rename these to `sourceWidth`, `sourceHeight`

packages/vscode-extension/src/webview/hooks/useCanvasRenderer.tsx
```diff
@@ -0,0 +1,107 @@
...
+      }
+
+      // Clear canvas
+      ctx.clearRect(0, 0, newWidth, newHeight);
+
+      if (transformConfig.rotation === DeviceRotationType.Portrait) {
+        // Direct draw for portrait mode
+        ctx.drawImage(sourceImg, 0, 0);
+      } else {
+        // Apply cached transformation
```
jwajgelt: Cached?

packages/vscode-extension/src/webview/Preview/Device/Device.tsx
```diff
@@ -1,43 +1,318 @@
...
+    : `${zoomHeight}px`;
+  const phoneContentMinWidth = `${phoneContentMinWidthValue}px`;
+  const phoneContentMinHeight = `${phoneContentMinHeightValue}px`;
+  const phoneScreenHeight = `${(device.screenWidth / frame.width) * 100}%`; // Swapped for landscape
+  const phoneScreenWidth = `${(device.screenHeight / frame.height) * 100}%`; // Swapped for landscape
+  const phoneTop = `${(frame.offsetX / frame.width) * 100}%`; // Swapped for landscape
+  const phoneLeft = `${(frame.offsetY / frame.height) * 100}%`; // Swapped for landscape
+  const phoneMaskImage = `url(${device.landscapeScreenImage})`;
+  const phoneFrameImage = `url(${frame.imageLandscape})`;
+  const phoneAspectRatio = `${1 / aspectRatio}`;
```
jwajgelt: NIT: I'd probably inline these into the returned object, I don't think it's more readable this way.
Same goes for the other `get...()` functions in this file.

packages/vscode-extension/src/webview/Preview/MediaCanvas.tsx
```diff
@@ -0,0 +1,102 @@
...
+  src?: string;
+}
+
+const MediaCanvas = forwardRef<HTMLCanvasElement, MediaCanvasProps>(
+  ({ mediaRef, src, ...rest }, ref) => {
+    const { projectState } = useProject();
+    const rotation = projectState.rotation;
+
+    const isRunningRef = useRef(false);
+    const canvasRef = ref as RefObject<HTMLCanvasElement>;
```
jwajgelt, line 24: Is this cast necessary?

packages/vscode-extension/src/webview/Preview/MediaCanvas.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const drawToCanvas = useCanvasRenderer(rotation, canvasRef);
+    const drawToCanvasRef = useRef(drawToCanvas);
+
+    useEffect(() => {
+      drawToCanvasRef.current = drawToCanvas;
+    }, [drawToCanvas]);
```
jwajgelt: If you moved the code from `useCanvasRenderer` into this component, you could probably get rid of some extra `useRef`/`useEffect`s like this one.

packages/vscode-extension/src/webview/utilities/deviceContants.ts
```diff
@@ -27,11 +27,44 @@ import iphone16pro from "../../assets/iphone_16_pro/skin.png";
...
+
+import pixel7Landscape from "../../assets/pixel_7/landscape/skin.webp";
+import pixel7LandscapeMask from "../../assets/pixel_7/landscape/mask.png";
+import pixel7LandscapeBezel from "../../assets/pixel_7/landscape/bezel.png";
+import pixel7LandscapeScreen from "../../assets/pixel_7/landscape/screen.png";
+
+import pixel6aLandscape from "../../assets/pixel_6a/landscape/skin.webp";
+import pixel6aLandscapeMask from "../../assets/pixel_6a/landscape/mask.png";
+import pixel6aLandscapeBezel from "../../assets/pixel_6a/landscape/bezel.png";
+import pixel6aLandscapeScreen from "../../assets/pixel_6a/landscape/screen.png";
```
jwajgelt: I'd move these next to the portrait images imports, like you have for the iphone16 and iphoneSE

packages/vscode-extension/src/webview/utilities/deviceContants.ts
```diff
@@ -27,11 +27,44 @@ import iphone16pro from "../../assets/iphone_16_pro/skin.png";
...
 import iphone16proscreen from "../../assets/iphone_16_pro/screen.png";
+import iphone16proLandscape from "../../assets/iphone_16_pro/landscape/skin.png";
+import iphone16proLandscapeMask from "../../assets/iphone_16_pro/landscape/mask.png";
+import iphone16proLandscapeBezel from "../../assets/iphone_16_pro/landscape/bezel.png";
+import iphone16proLandscapeScreen from "../../assets/iphone_16_pro/landscape/screen.png";
+
+import iphone15proLandscape from "../../assets/iphone_15_pro/landscape/skin.png";
+import iphone15proLandscapeMask from "../../assets/iphone_15_pro/landscape/mask.png";
+import iphone15proLandscapeBezel from "../../assets/iphone_15_pro/landscape/bezel.png";
+import iphone15proLandscapeScreen from "../../assets/iphone_15_pro/landscape/screen.png";
```
jwajgelt, line 53: I'd move these next to the portrait images imports, like you have for the iphone16 and iphoneSE

packages/vscode-extension/src/webview/utilities/translateAppCoordinates.ts
```diff
@@ -0,0 +1,176 @@
...
+    actualPortaitAppRight,
+    actualUpsideDownAppPortrait,
+    actualUpsideDownAppLeft,
+    actualUpsideDownAppRight,
+    actualLeftAppPortrait,
+    actualRightAppPortrait,
+  };
+}
+
+export function translateAppToPreviewCoordinates(
```
jwajgelt: NIT:
I'd rename this to 
```suggestion
export function transformAppToPreviewCoordinates(
```
since translation is the operation of moving in some direction by a set distance (that is, you add a vector to each point), while what is happening here is actually rotating everything around a point.
I would also be fine with just
```suggestion
export function appToPreviewCoordinates(
```
Same applies to the other function.

You may also want to add a short doc comment on what this function does and why it's needed when rotation is applied.

jwajgelt (changes requested):

packages/vscode-extension/lib/orientation.js
pFornagiel: Addressed in d92222f0.

```
// infer information about the app's orientation based on the device orientation and ASSUME THAT PORTRAIT-SECONDARY IS NOT SET.

// We also assume that, if one of the landscape orientations is supported, the other one is supported as well.
```

packages/vscode-extension/lib/orientation.js
pFornagiel: Addressed in d92222f0.
```ts
/**
 * Mapping from namedOrientationDidChangeEvent names to DeviceRotation names or "Landscape" specific case
 * sent to the extension.
 * "Landscape" case happens when we cannot infer the detailed landscape orientation of the app and
 * leave the interpretation to the frontend of the extension.
 */
const androidOrientationMapping = {
  "portrait-primary": "Portrait",
  "portrait-secondary": "PortraitUpsideDown",
  "landscape-primary": "LandscapeLeft",
  "landscape-secondary": "LandscapeRight",
};

/**
 * The function is used to map the orientation to a common format
 * used by the AppOrientation in the extension, as the landscape orientation
 * from the eventListener object is different on iOS and Android.
 * iOs -> "landscape-secondary" === LandscapeLeft
 * Android -> "landscape-primary" === LandscapeRight
 */
const getMappedOrientation = (orientation, isLandscape) => {
  // ...
}
```

packages/vscode-extension/lib/orientation.js
pFornagiel: Changed in d92222f0
```ts
const androidOrientationMapping = {
  "portrait-primary": "Portrait",
  "portrait-secondary": "PortraitUpsideDown",
  "landscape-primary": "LandscapeLeft",
  "landscape-secondary": "LandscapeRight",
};
```

packages/vscode-extension/lib/orientation.js
pFornagiel: Changed in d92222f0
```ts
let currentAppOrientation = null;
```

packages/vscode-extension/lib/orientation.js
pFornagiel: Addressed in d92222f0

The Native Module was not supported in the older expo versions. To amend this, we now use `UIManager` built-in module from `react-native` import.

```ts
const { Dimensions, Platform, UIManager, NativeEventEmitter } = require("react-native");

//...

const orientationEventSubscription = new NativeEventEmitter(UIManager).addListener(
    "namedOrientationDidChange",
    handleOrientationChange
  );
```

packages/vscode-extension/lib/orientation.js
pFornagiel: Fixed in d92222f0

```ts
const { Dimensions, Platform, UIManager, NativeEventEmitter } = require("react-native");

// ...

// Used to supress warnings about UIManager not implementing below methods
// when passed to NativeEventEmitter.
UIManager.addListener = () => {};
UIManager.removeListeners = () => {};
```

packages/vscode-extension/lib/orientation.js
pFornagiel: Addressed in d92222f0

```ts
export function setup() {
  initializeOrientationAndSendInitMessage();

  const orientationEventSubscription = new NativeEventEmitter(UIManager).addListener(
    "namedOrientationDidChange",
    handleOrientationChange
  );

  // Dimension change lags behind Orientation change in the IOS case, so we add a second listener to amend the effect
  // Explained in the context above.
  const dimensionEventSubscription = Dimensions.addEventListener("change", handleDimensionsChange);

  return function cleanup() {
    if (orientationEventSubscription) {
      orientationEventSubscription.remove();
    }
    if (dimensionEventSubscription) {
      dimensionEventSubscription.remove();
    }
  };
}
```

pFornagiel: Addressed in d92222f0

```ts
export function setup() {
  initializeOrientationAndSendInitMessage();

  const orientationEventSubscription = new NativeEventEmitter(UIManager).addListener(
    "namedOrientationDidChange",
    handleOrientationChange
  );

  // Dimension change lags behind Orientation change in the IOS case, so we add a second listener to amend the effect
  // Explained in the context above.
  const dimensionEventSubscription = Dimensions.addEventListener("change", handleDimensionsChange);

  return function cleanup() {
    if (orientationEventSubscription) {
      orientationEventSubscription.remove();
    }
    if (dimensionEventSubscription) {
      dimensionEventSubscription.remove();
    }
  };
}
```

packages/vscode-extension/lib/orientation.js
pFornagiel: Fixed in d92222f0.

We now only get rid of subscriptions returned by `addListener` and `addEventListener` methods, so only event-listeners registered in the corresponding re-render are affected.

```ts
export function setup() {
  initializeOrientationAndSendInitMessage();

  const orientationEventSubscription = new NativeEventEmitter(UIManager).addListener(
    "namedOrientationDidChange",
    handleOrientationChange
  );

  // Dimension change lags behind Orientation change in the IOS case, so we add a second listener to amend the effect
  // Explained in the context above.
  const dimensionEventSubscription = Dimensions.addEventListener("change", handleDimensionsChange);

  return function cleanup() {
    if (orientationEventSubscription) {
      orientationEventSubscription.remove();
    }
    if (dimensionEventSubscription) {
      dimensionEventSubscription.remove();
    }
  };
}
```

packages/vscode-extension/lib/wrapper.js
pFornagiel: Renamed in d92222f0

```ts
get setupOrientationListeners() {
    return require("./orientation").setup;
  },
```

packages/vscode-extension/lib/wrapper.js
pFornagiel: Removed in d92222f0

```ts
InternalImports.setupRenderOutlinesPlugin();
InternalImports.setupNetworkPlugin();
const orientationListenersCleanup = InternalImports.setupOrientationListeners();
```

packages/vscode-extension/src/project/deviceSession.ts
pFornagiel: Fixed in d92222f0.

```ts
devtools.onEvent("appOrientationChanged", (orientation: AppOrientationType) => {
  const isLandscape =
    this.rotation === DeviceRotationType.LandscapeLeft ||
    this.rotation === DeviceRotationType.LandscapeRight;

  // if the app orientation is equal to "Landscape", it means we do not have enough
  // information on the application side to infer the detailed orientation.
  // In this case:
  // - if the device is in landscape mode, we assume that the app orientation is correct with device rotation
  // - if the device is not in landscape mode we set app orientation to the last known orientation. 
  //   If the last orientation is not known, we assume the application was started in Landscape mode 
  //   while the device was oriented in Portrait, and we pick `LandscapeLeft` as the default orientation in that case.
  if (orientation === "Landscape") {
    if (isLandscape) {
      this.appOrientation = this.rotation;
    } else {
      this.appOrientation = this.appOrientation ?? DeviceRotationType.LandscapeLeft;
    }
  } else {
    this.appOrientation = orientation;
  }

  this.emitStateChange();
});
return devtools;
}
```

packages/vscode-extension/src/project/deviceSession.ts
pFornagiel: Addressed in d92222f0.

```ts
devtools.onEvent("appOrientationChanged", (orientation: AppOrientationType) => {
  const isLandscape =
    this.rotation === DeviceRotationType.LandscapeLeft ||
    this.rotation === DeviceRotationType.LandscapeRight;

  // if the app orientation is equal to "Landscape", it means we do not have enough
  // information on the application side to infer the detailed orientation.
  // In this case:
  // - if the device is in landscape mode, we assume that the app orientation is correct with device rotation
  // - if the device is not in landscape mode we set app orientation to the last known orientation. 
  //   If the last orientation is not known, we assume the application was started in Landscape mode 
  //   while the device was oriented in Portrait, and we pick `LandscapeLeft` as the default orientation in that case.
  if (orientation === "Landscape") {
    if (isLandscape) {
      this.appOrientation = this.rotation;
    } else {
      this.appOrientation = this.appOrientation ?? DeviceRotationType.LandscapeLeft;
    }
  } else {
    this.appOrientation = orientation;
  }

  this.emitStateChange();
});
return devtools;
}
```

packages/vscode-extension/src/project/deviceSession.ts
pFornagiel: Moved to `if` statements in 78820032

```ts
devtools.onEvent("appOrientationChanged", (orientation: AppOrientationType) => {
  const isLandscape =
    this.rotation === DeviceRotationType.LandscapeLeft ||
    this.rotation === DeviceRotationType.LandscapeRight;

  if (orientation === "Landscape") {
    // if the app orientation is equal to "Landscape", it means we do not have enough
    // information on the application side to infer the detailed orientation.
    if (isLandscape) {
      // if the device is in landscape mode, we assume that the app orientation is correct with device rotation
      this.appOrientation = this.rotation;
    } else {
      // if the device is not in landscape mode we set app orientation to the last known orientation.
      // if the last orientation is not known, we assume the application was started in Landscape mode
      // while the device was oriented in Portrait, and we pick `LandscapeLeft` as the default orientation in that case.
      this.appOrientation = this.appOrientation ?? DeviceRotationType.LandscapeLeft;
    }
  } else {
    this.appOrientation = orientation;
  }

  this.emitStateChange();
});
return devtools;
```

packages/vscode-extension/lib/instrumentation.js
pFornagiel: Moved in b162c6c6

```ts
function getWindowRect() {
  const { width: screenWidth, height: screenHeight } = Dimensions.get("screen");
  if (Platform.OS === "android") {
    const statusBarHeight = StatusBar.currentHeight || 0;
    return {
      x: 0,
      y: statusBarHeight,
      width: screenWidth,
      height: screenHeight,
    };
  } else {
    return {
      x: 0,
      y: 0,
      width: screenWidth,
      height: screenHeight,
    };
  }
}
```

packages/vscode-extension/src/common/Project.ts
pFornagiel: Refactored in 043b7436

```ts
export enum DeviceRotation {
  Portrait = "Portrait",
  PortraitUpsideDown = "PortraitUpsideDown",
  LandscapeLeft = "LandscapeLeft",
  LandscapeRight = "LandscapeRight"
}

export type AppOrientation = DeviceRotation | "Landscape";

export function isOfEnumDeviceRotation(value: any): value is DeviceRotation {
  return Object.values(DeviceRotation).includes(value);
}

```

packages/vscode-extension/src/project/project.ts
pFornagiel: Refactored in f0636d61

The logic has been moved down to `preview.ts` instead of `deviceBase`, for the transform to happen as close to sim-server communication as possible. Additionally, some of the logic dealing with translation to of coordinates to sim-server was already in happening in `preview.ts`.

`preview.ts`:
```ts
public sendTouches(
    touches: Array<TouchPoint>,
    type: "Up" | "Move" | "Down",
    rotation: DeviceRotation
  ) {
    // transform touch coordinates to account for different device orientations before
    // translation and sending them to the simulator-server.
    touches.forEach((touch, i, array) => {
      const { xRatio: x, yRatio: y } = touch;
      switch (rotation) {
        // 90° anticlockwise map (x,y) to (1-y, x)
        case DeviceRotation.LandscapeLeft:
          array[i] = { xRatio: 1 - y, yRatio: x };
          break;
        case DeviceRotation.LandscapeRight:
          // 90° clockwise map (x,y) to (y, 1-x)
          array[i] = { xRatio: y, yRatio: 1 - x };
          break;
        case DeviceRotation.PortraitUpsideDown:
          // 180° map (x,y) to (1-x, 1-y)
          array[i] = { xRatio: 1 - x, yRatio: 1 - y };
          break;
        default:
          // Portrait mode: no transformation needed
          break;
      }
    });

    const touchesCoords = touches.map((pt) => `${pt.xRatio},${pt.yRatio}`).join(" ");
    this.subprocess?.stdin?.write(`touch ${type} ${touchesCoords}\n`);
  }
```

packages/vscode-extension/src/webview/components/Preview.css
pFornagiel: Removed in f0636d61.

```css
.phone-frame {
  pointer-events: none;
  rotate: var(--frame-rotation);
  width: 100%;
}
```

packages/vscode-extension/src/webview/components/Preview.css
pFornagiel: Addressed in f0636d61

The variable is now used to fix problem with device preview in Landscape mode not properly resizing to "Fit", when "Show Device Frame" option was turned off.

```ts
function getLandscapeDimensions(
  config: DeviceLayoutConfig,
  device: DeviceProperties,
  frame: DevicePropertiesFrame,
  zoomLevel: ZoomLevelType
) {
  // ...

  return {
    // ...
    phoneWrapperHeight: "fit-content",
    phoneWrapperWidth: isFitSet ? "100%" : "fit-content",
    // ...
  };
}
```

```ts
export default function Device({
  device,
  zoomLevel,
  children,
  wrapperDivRef,
}: DeviceProps) {
  // ...

  const phoneDimensions = layoutConfig.isLandscape
    ? getLandscapeDimensions(layoutConfig, device, frame, zoomLevel)
    : getPortraitDimensions(layoutConfig, device, frame, zoomLevel);

  // ...

  return {
    // ...
    "--phone-wrapper-height": phoneDimensions.phoneWrapperHeight,
    "--phone-wrapper-width": phoneDimensions.phoneWrapperWidth,
    // ...
  };
  // ...
}
```

```css
.phone-wrapper {
  display: flex;
  align-items: center;
  justify-content: space-around;
  min-width: var(--phone-wrapper-min-width); 
  min-height: var(--phone-wrapper-min-height); 
  width: var(--phone-wrapper-width); 
  height: var(--phone-wrapper-height);
}
```

packages/vscode-extension/src/webview/components/Preview.tsx
pFornagiel: Added in f0636d61

```ts
  /**
   * Converts mouse event coordinates to normalized touch coordinates ([0-1] range)
   * relative to the device preview image.
   * Coordinate transformation handling when the device when the device is rotated,
   * done in the sendTouches method in preview.ts.
   */
  function getNormalizedTouchCoordinates(event: MouseEvent<HTMLDivElement>) {
  // ...
  }
```

packages/vscode-extension/src/webview/components/Preview.tsx
pFornagiel: Removed in f0636d61

packages/vscode-extension/src/webview/components/Preview.tsx
pFornagiel: Added in f0636d61

```ts
    {/* Hack needed for css to cache those images. By default, all the images are not cached on the frontend,
    so without this in place, when the device rotates, the images are re-fetched from the file system
    which causes the device preview to flicker. */}
```

packages/vscode-extension/src/webview/hooks/useCanvasRenderer.tsx
pFornagiel: Addressed in 1bacd459

The logic of `useCanvasRenderer` hook has been integrated into `MediaCanvas` component and the hook has been removed.

packages/vscode-extension/src/webview/hooks/useCanvasRenderer.tsx
pFornagiel: Changed in 1bacd459

```ts
const sourceWidth =  sourceImg instanceof HTMLVideoElement ? sourceImg.videoWidth : sourceImg.width;
const sourceHeight =  sourceImg instanceof HTMLVideoElement ? sourceImg.videoHeight : sourceImg.height;
```

packages/vscode-extension/src/webview/hooks/useCanvasRenderer.tsx
pFornagiel: Addressed in 1bacd459

The comment was meant to highlight the usage of useMemo to memoize transformConfig. The comment has been changed to:

```ts
if (transformConfig.rotation === DeviceRotation.Portrait) {
  // Direct draw for portrait mode
  ctx.drawImage(sourceImg, 0, 0);
} else {
  // Apply transformation from transformConfig
  ctx.save();
  ctx.translate(newWidth / 2, newHeight / 2);
  ctx.rotate(transformConfig.angle);
  ctx.translate(-sourceWidth / 2, -sourceHeight / 2);
  ctx.drawImage(sourceImg, 0, 0);
  ctx.restore();
}
```

packages/vscode-extension/src/webview/Preview/Device/Device.tsx
pFornagiel: Addressed in f0636d61

All the values have been declared inline in the return objects of getX functions. 

```ts
function getLandscapeDimensions(
  config: DeviceLayoutConfig,
  device: DeviceProperties,
  frame: DevicePropertiesFrame,
  zoomLevel: ZoomLevelType
) {
  const { isFitSet, parentHeight, aspectRatio } = config;

  const zoomHeight = frame.height * (zoomLevel as number) * DEVICE_DEFAULT_SCALE;

  return {
    phoneWrapperMinWidth: "auto",
    phoneWrapperMinHeight: "auto",
    phoneWrapperHeight: "fit-content",
    phoneWrapperWidth: isFitSet ? "100%" : "fit-content",
    phoneContentHeight: "auto",
    phoneContentWidth: isFitSet
      ? `calc(min(100%,  ${parentHeight / aspectRatio}px) * ${CSS_MARGIN_FACTOR})`
      : `${zoomHeight}px`,
    phoneContentMinWidth: `${MIN_HEIGHT}px`,
    phoneContentMinHeight: `${MIN_HEIGHT * aspectRatio}px`,
    phoneScreenHeight: `${(device.screenWidth / frame.width) * 100}%`, // Swapped for landscape
    phoneScreenWidth: `${(device.screenHeight / frame.height) * 100}%`, // Swapped for landscape
    phoneTop: `${(frame.offsetX / frame.width) * 100}%`, // Swapped for landscape
    phoneLeft: `${(frame.offsetY / frame.height) * 100}%`, // Swapped for landscape
    phoneMaskImage: `url(${device.landscapeScreenImage})`,
    phoneFrameImage: `url(${frame.imageLandscape})`,
    phoneAspectRatio: `${1 / aspectRatio}`,
  };
}
```

packages/vscode-extension/src/webview/Preview/MediaCanvas.tsx
pFornagiel: Left unchanged.

The cast is necessary for Typescript to properly interpret Ref passed to the component as RefObject - otherwise it treats it as `((instance: HTMLCanvasElement | null) => void) | RefObject<HTMLCanvasElement | null>`, which causes type errors.

packages/vscode-extension/src/webview/Preview/MediaCanvas.tsx
pFornagiel: Left unchanged.

The Ref helps to prevent unnecessary cleanups. If the ref was removed, the `useEffect` registering listeners would have to be dependent on `rotation` because of closure, which would cause additional registration / cleanup whenever it changed.

packages/vscode-extension/src/webview/utilities/deviceContants.ts
pFornagiel: Changed in aa6c4456

packages/vscode-extension/src/webview/utilities/deviceContants.ts
pFornagiel: Changed in aa6c4456

packages/vscode-extension/src/webview/utilities/translateAppCoordinates.ts
pFornagiel: Addressed in cb8b8150

Renamed the functions, renamed the file, added comment-docs.

```ts
/**
 * Transform coordinates and rects from app's coordinate system to preview's coordinate system.
 * @param appOrientation - Current orientation of the app.
 * @param deviceOrientation - Current orientation of the device.
 * @param frameRect - coordinates and frame rects in app coordinate system.
 * @returns coordinates and frame rects in preview coordinate system.
 *  The transform is needed to account of change of origin point after both - the device preview rotation
 *  and app orientation - change, synchronizing the app's coordinate system with the preview's coordinate system.
 * */
export function appToPreviewCoordinates(
  appOrientation: DeviceRotation | undefined,
  deviceOrientation: DeviceRotation,
  frameRect: NormalizedFrameRect
): NormalizedFrameRect {
  //...
}
```

packages/vscode-extension/src/webview/Preview/MediaCanvas.tsx
jwajgelt: I think this component doesn't need the `forwardRef` anymore, the components using this one don't seem to use the `ref` for anything.
Then, if you changed it so the ref is created inside this component, you'll have it typed correctly as `RefObject`.

packages/vscode-extension/vite.config.js
jwajgelt, line 1: Please revert this to avoid breaking the IDE preview for everyone.

packages/vscode-extension/lib/orientation.js
```diff
@@ -0,0 +1,159 @@
+const { Dimensions, Platform, UIManager, NativeEventEmitter } = require("react-native");
+const inspectorBridge = require("./inspector_bridge");
+
+// Used to supress warnings about UIManager not implementing below methods
+// when passed to NativeEventEmitter.
+UIManager.addListener = () => {};
+UIManager.removeListeners = () => {};
```
jwajgelt: Don't spend too much time on this one, it should be fine as is, but:

This is a bit risky, since we're overwriting properties of RN's exports.
I think I'd prefer we did this directly before the call that causes the warnings to be logged, and restore the original values directly after, like
```tsx
const originalAdd = UIManager.addListener;
const originalRemove = UIManager.removeListener;
UIManager.addListener = () => {};
UIManager.removeListeners = () => {};
/* ... do the stuff ... */
UIManager.addListener = originalAdd;
UIManager.removeListeners = originalRemove;
```
Or maybe even replace the `console.warn` method temporarily instead.

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -194,7 +204,34 @@ export class Preview implements Disposable {
...
+          break;
+        case DeviceRotation.PortraitUpsideDown:
+          // 180° map (x,y) to (1-x, 1-y)
+          array[i] = { xRatio: 1 - x, yRatio: 1 - y };
+          break;
+        default:
+          // Portrait mode: no transformation needed
+          break;
+      }
+    });
```
jwajgelt:
```suggestion
    touches = touches.map((touch) => {
      const { xRatio: x, yRatio: y } = touch;
      switch (rotation) {
        // 90° anticlockwise map (x,y) to (1-y, x)
        case DeviceRotation.LandscapeLeft:
          return { xRatio: 1 - y, yRatio: x };
        case DeviceRotation.LandscapeRight:
          // 90° clockwise map (x,y) to (y, 1-x)
          return { xRatio: y, yRatio: 1 - x };
        case DeviceRotation.PortraitUpsideDown:
          // 180° map (x,y) to (1-x, 1-y)
          return { xRatio: 1 - x, yRatio: 1 - y };
        case DeviceRotation.Portrait:
          // Portrait mode: no transformation needed
          return touch;
      }
    });
```
Use `Array.map` in cases where you want to calculate some function for each element of the array and collect the results.
Use `Array.forEach` when you want to call some function where you only care about some side effect of the function, rather than performing some pure calculation.

jwajgelt: If you go this way, it might be better creating a new variable `const transformedTouches` for the results of the map instead of reassigning to `touches`.

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -140,6 +144,7 @@ export class DeviceSession
       useParentDebugSession: true,
     });
     this.watchProjectSubscription = watchProjectFiles(this.onProjectFilesChanged);
+    // this.appOrientation = this.rotation;
```
jwajgelt:
```suggestion
```
remove the commented-out code before merging

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -178,6 +282,9 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
               </DropdownMenu.SubContent>
             </DropdownMenu.Portal>
           </DropdownMenu.Sub>
+
+         
+
```
jwajgelt:
```suggestion
```

jwajgelt (changes requested):
The last batch of change requests, overall looks very good 👍

packages/vscode-extension/vite.config.js
pFornagiel: Reverted vite.config host change in 9b248f6e

The MediaCanvas needs forwardRef in case of MjpegImg and Preview components: the preview uses ref of canvas to properly handle user touches.

The cast is not correct in general-case usage, but in our use cases it will always be of type `RefObject<HTMLCanvasElement>`, so it seems sensible. Leaving unchanged.

```ts
function Preview({
  //...
}: Props) {
    const previewRef = useRef<HTMLCanvasElement>(null);
   
    // ...

    function getNormalizedTouchCoordinates(event: MouseEvent<HTMLDivElement>) {
        const imgRect = previewRef.current!.getBoundingClientRect();
    
        // Normalize coordinates to [0-1] range
        const x = (event.clientX - imgRect.left) / imgRect.width;
        const y = (event.clientY - imgRect.top) / imgRect.height;
    
        let clampedX = clamp(x, 0, 1);
        let clampedY = clamp(y, 0, 1);
        return { x: clampedX, y: clampedY };
  }
  // ...
```

```tsx
 <MjpegImg
    src={previewURL}
    ref={previewRef}
    style={{
      cursor: isInspecting ? "crosshair" : "default",
    }}
    className="phone-screen"
  />
```

packages/vscode-extension/lib/orientation.js
pFornagiel: Changed to overwriting console.warn in f0ae7c87, because of the issue explained in the code comment:

```ts
export function setup() {
  initializeOrientationAndSendInitMessage();

  // Suppress warnings about UIManager not implementing proper methods by overwriting console.warn temporarily.
  // This is needed, because UIManager.addListener and UIManager.removeListeners methods
  // are undefined. When the same trick is applied to those methods (overwriting and assigning the original 
  // methods later), we begin to get Render and Console errors in LogBox after the assignment:
  // "_this$_nativeModule2.removeListeners is not a function (it is undefined)"
  const originalConsoleWarn = console.warn;
  console.warn = () => {};

  const orientationEventSubscription = new NativeEventEmitter(UIManager).addListener(
    "namedOrientationDidChange",
    handleOrientationChange
  );

  // Dimension change lags behind Orientation change in the IOS case, so we add a second listener to amend the effect.
  // Explained in the context above.
  const dimensionEventSubscription = Dimensions.addEventListener("change", handleDimensionsChange);

  // Restore the original console.warn function
  console.warn = originalConsoleWarn; 

  // ...
}
```

packages/vscode-extension/src/devices/preview.ts
pFornagiel: Changed in 5db3f3ac, 5c54d7e5

```ts
public sendTouches(
    touches: Array<TouchPoint>,
    type: "Up" | "Move" | "Down",
    rotation: DeviceRotation
  ) {
    // transform touch coordinates to account for different device orientations before
    // translation and sending them to the simulator-server.
    const transformedTouches = touches.map((touch, i) => {
      const { xRatio: x, yRatio: y } = touch;
      switch (rotation) {
        // 90° anticlockwise map (x,y) to (1-y, x)
        case DeviceRotation.LandscapeLeft:
          return { xRatio: 1 - y, yRatio: x };
        case DeviceRotation.LandscapeRight:
          // 90° clockwise map (x,y) to (y, 1-x)
          return { xRatio: y, yRatio: 1 - x };
        case DeviceRotation.PortraitUpsideDown:
          // 180° map (x,y) to (1-x, 1-y)
          return { xRatio: 1 - x, yRatio: 1 - y };
        default:
          // Portrait mode: no transformation needed
          return touch;
      }
    });

    const touchesCoords = transformedTouches.map((pt) => `${pt.xRatio},${pt.yRatio}`).join(" ");
    this.subprocess?.stdin?.write(`touch ${type} ${touchesCoords}\n`);
  }
  ```

packages/vscode-extension/src/project/deviceSession.ts
pFornagiel: Removed alongside other formatting changes in 07d83c18

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
pFornagiel: Removed alongside other formatting changes in 07d83c18

jwajgelt (approved):
Looks good 👍 
Great work pushing this over the finish line.

PR #1313, opened 2025-07-08
Reload MCP tool schema upon restoring backend connection

latekvo:
This PR adds the `ConnectionListener`, which monitors connection to the backend.
When connection to the backend is lost, the `ConnectionListener` begins pinging the server, until one of the pings is acknowledged, and the connection is restored. 
When connection is restored, the MCP server entry along with the tool schema are reloaded.

#### Scenario 1 - Editor started with no internet:

1. Disable internet on your computer
2. Start VSCode or Cursor
3. Notice how every 5 seconds, a ping to the backend is being sent. This can be seen in the logs
4. Enable internet
5. Within 5 seconds:
    - The tool schema is fetched
    - On `Cursor`, `mcp.json` is updated
    - On `VSCode`, the `lm.registerMcpServerDefinitionProvider` update event is fired

#### Scenario 2 - Connection lost during editor usage:

1. Start backend server locally - adjust `BACKEND_URL` appropriately
2. Start VSCode or Cursor
3. Kill backend server after editor initialization
4. Send a request to the server by prompting the AI to use one of the tools
5. Notice how every 5 seconds, a ping to the backend is being sent. This can be seen in the logs
6. Start backend server again
7. One ping should is received, the pinging stops immediately after that
8. Within 5 seconds:
    - The tool schema is fetched
    - On `Cursor`, `mcp.json` is updated
    - On `VSCode`, the `lm.registerMcpServerDefinitionProvider` update event is fired

### DIFF
--- a/packages/vscode-extension/src/ai/mcp/index.ts
+++ b/packages/vscode-extension/src/ai/mcp/index.ts
@@ -9,6 +9,7 @@ import { MCP_LOG } from "./utils";
 import "../../../vscode.mcpConfigurationProvider.d.ts";
 import { extensionContext } from "../../utilities/extensionContext";
 import { ConnectionListener } from "../shared/ConnectionListener";
+import { disposeAll } from "../../utilities/disposables";
 
 async function updateMcpConfig(port: number) {
   const mcpConfig = (await readMcpConfig()) || newMcpConfig();
@@ -16,7 +17,7 @@ async function updateMcpConfig(port: number) {
   await writeMcpConfig(updatedConfig);
 }
 
-function directLoadRadonAI(connectionListener: ConnectionListener): Disposable[] {
+function directLoadRadonAI(connectionListener: ConnectionListener): Disposable {
   // Version suffix has to be incremented on every MCP server reload.
   let versionSuffix = 0;
 
@@ -45,7 +46,9 @@ function directLoadRadonAI(connectionListener: ConnectionListener): Disposable[]
     },
   });
 
-  return [connectionChangeListener, licenseChangeListener, mcpServerEntry];
+  return new Disposable(() =>
+    disposeAll([connectionChangeListener, licenseChangeListener, mcpServerEntry])
+  );
 }
 
 async function fsLoadRadonAI(connectionListener: ConnectionListener) {
@@ -72,13 +75,13 @@ function isDirectLoadingAvailable() {
   );
 }
 
-export default function registerRadonAi(): Disposable[] {
+export default function registerRadonAi(): Disposable {
   const connectionListener = new ConnectionListener();
 
   if (isDirectLoadingAvailable()) {
     const disposables = directLoadRadonAI(connectionListener);
 
-    return [...disposables, connectionListener];
+    return new Disposable(() => disposeAll([disposables, connectionListener]));
   } else {
     const connectionChangeListener = connectionListener.onConnectionRestored(() => {
       fsLoadRadonAI(connectionListener);
@@ -88,6 +91,8 @@ export default function registerRadonAi(): Disposable[] {
       fsLoadRadonAI(connectionListener);
     });
 
-    return [connectionListener, connectionChangeListener, licenseObserver];
+    return new Disposable(() =>
+      disposeAll([connectionListener, connectionChangeListener, licenseObserver])
+    );
   }
 }
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -312,7 +312,7 @@ export async function activate(context: ExtensionContext) {
   );
 
   // Initializes MCP part of Radon AI
-  context.subscriptions.push(...registerRadonAi());
+  context.subscriptions.push(registerRadonAi());
 
   // You can configure the chat in package.json under the `chatParticipants` key
   registerRadonChat(context);


packages/vscode-extension/src/ai/shared/api.ts
```diff
@@ -119,3 +137,9 @@ export async function getSystemPrompt(userPrompt: string): Promise<SystemRespons
 
   return await response.json();
 }
+
+export async function isServerOnline(): Promise<boolean> {
+  const url = new URL("/api/ping/", BACKEND_URL);
+  const response = await fetch(url);
+  return response.status === 200;
```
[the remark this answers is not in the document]
latekvo: Fixed in [98bfd18](https://github.com/software-mansion/radon-ide/pull/1313/commits/98bfd1865129eed0b8ad79673ff449a549d19735)

packages/vscode-extension/src/ai/shared/ConnectionListener.ts
```diff
@@ -0,0 +1,67 @@
...
+    this.connectionListeningInterval = setInterval(async () => {
+      const isOnline = await isServerOnline();
+
+      if (isOnline && this.connectionListeningInterval) {
+        this.connectionSuccessEmitter.fire(isOnline);
+        clearInterval(this.connectionListeningInterval);
+      }
+    }, PING_INTERVAL);
+
+    extensionContext.subscriptions.push(new Disposable(this.tryClearListeningInterval));
```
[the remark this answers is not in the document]
latekvo: Fixed in [064a9fa](https://github.com/software-mansion/radon-ide/pull/1313/commits/064a9faa3eedbf4a465b37d105fede7c2dcc9e90)

packages/vscode-extension/src/ai/shared/ConnectionListener.ts
```diff
@@ -0,0 +1,49 @@
...
+        this.connectionRestoredEmitter.fire();
+      }
+    }, PING_INTERVAL);
+
+    extensionContext.subscriptions.push(new Disposable(() => this.tryClearListeningInterval()));
+  }
+
+  public onConnectionRestored(callback: () => unknown) {
+    extensionContext.subscriptions.push(this.connectionRestoredEmitter.event(callback));
+  }
```
jwajgelt: I would prefer it if the event subscription was returned to the caller, and the caller was responsible for calling `dispose()` on it when it's no longer needed.

packages/vscode-extension/src/ai/shared/ConnectionListener.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.connectionListeningInterval = setInterval(async () => {
+      const isOnline = await isServerOnline();
+
+      if (isOnline && this.connectionListeningInterval) {
+        this.tryClearListeningInterval();
+        this.connectionRestoredEmitter.fire();
+      }
+    }, PING_INTERVAL);
+
+    extensionContext.subscriptions.push(new Disposable(() => this.tryClearListeningInterval()));
```
jwajgelt: This will push a new `Disposable` each time this is called.
Instead, you register a call to `tryClearListeningInterval` once, in the constructor.
Even better, you could make `ConnectionListener implement Disposable`, call `clearInterval` in the `dispose()` method, and have whatever creates the `ConnectionListener` instance be responsible for clearing it up.

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -72,12 +75,18 @@ function isDirectLoadingAvailable() {
...
+
   if (isDirectLoadingAvailable()) {
-    directLoadRadonAI();
+    directLoadRadonAI(connectionListener);
   } else {
+    connectionListener.onConnectionRestored(() => {
+      fsLoadRadonAI(connectionListener);
+    });
+
     extensionContext.subscriptions.push(
```
jwajgelt: Instead of registering the `extensionContext.subscriptions` here, I would prefer if this function returned a `Disposable` that `extension.ts` (or whatever calls it) can then disposed.

jwajgelt (changes requested):
Left some comments re cleanup.
Generally, I'd prefer if we implemented/returned `Disposable`s wherever possible and handle cleanup in the caller, instead of registering with the extension wherever we have something we may want to dispose of.
This way it's harder to register something unnecessarily (like with `tryClearListeningInterval` in this PR) or clean up too late.
It also means that the "disposable" components are not effectively singletons, like they are now, where they register themselves with a _global_ extension context (and sure, in this case they likely will always be created once and held for the lifetime of the extension, but you should have to think about it while writing or calling this function).

packages/vscode-extension/src/ai/shared/ConnectionListener.ts
latekvo: I changed how it works in [10c989d](https://github.com/software-mansion/radon-ide/pull/1313/commits/10c989d40ca516cb8faf8a0a6c5352be5dfae1da), let me know if this is what you had in mind.

packages/vscode-extension/src/ai/shared/ConnectionListener.ts
latekvo: Done in [10c989d](https://github.com/software-mansion/radon-ide/pull/1313/commits/10c989d40ca516cb8faf8a0a6c5352be5dfae1da)

packages/vscode-extension/src/ai/mcp/index.ts
latekvo: Done in [10c989d](https://github.com/software-mansion/radon-ide/pull/1313/commits/10c989d40ca516cb8faf8a0a6c5352be5dfae1da)

latekvo: The same issue also occurs with `RadonChat` (`registerRadonChat` in `extension.ts`).
I'll open a separate PR with that fixed, since there are more issues i found with `RadonChat` implementation, which are out of the scope of this PR.

packages/vscode-extension/src/ai/shared/ConnectionListener.ts
```diff
@@ -0,0 +1,56 @@
...
+      if (isOnline && this.connectionListeningInterval) {
+        this.tryClearListeningInterval();
+        this.connectionRestoredEmitter.fire();
+      }
+    }, PING_INTERVAL);
+  }
+
+  onConnectionRestored(callback: () => unknown) {
+    this.connectionRestoredEmitter.event(callback, undefined, this.listenerDisposables);
+  }
```
jwajgelt:
```suggestion
  onConnectionRestored(callback: () => unknown): Disposable {
    return this.connectionRestoredEmitter.event(callback);
  }
```
so that you can unregister it in the same place where you register for the callback.

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -8,49 +8,52 @@ import { startLocalMcpServer } from "./server";
...
-      versionSuffix += 1;
       didChangeEmitter.fire();
     })
   );
 
+  connectionListener.onConnectionRestored(() => {
+    didChangeEmitter.fire();
+  });
+
   extensionContext.subscriptions.push(
```
jwajgelt: I'd rather you returned this to `extension` and register there

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -8,49 +8,52 @@ import { startLocalMcpServer } from "./server";
...
+  // Version suffix has to be incremented on every MCP server reload.
+  let versionSuffix = 0;
+
   const didChangeEmitter = new EventEmitter<void>();
 
-  // version suffix is incremented whenever we get auth token update notification
-  // this way we request vscode to reload the tool on regular basis but also immediately
-  // after the user inputs the license token
-  let versionSuffix = 0;
   extensionContext.subscriptions.push(
```
jwajgelt: I'd rather you returned this to `extension` and register in the `subscriptions` there

packages/vscode-extension/src/ai/shared/ConnectionListener.ts
latekvo: Fixed this in [03f3aa0](https://github.com/software-mansion/radon-ide/pull/1313/commits/03f3aa0752a4075286026a6b84c431a5acaba8ac), although I'm not sure if i see the point of carrying this list of disposables all the way to the `extension.ts`. I think there isn't any functional difference between `ConnectionListener` handling it, or returning it, except that the latter option requires more code.

Note that I'm not suggesting usage of `context.subscriptions.push` within `ConnectionListener`, I acknowledge that it's cleaner to have all usages of `context.subscriptions.push` be organised within `extension.ts`.
Instead, since `ConnectionListener` is already a `Disposable`, and is already registered within `context.subscriptions`, it's possible for `ConnectionListener` to store references to all the `Disposable` it creates. Wouldn't it be cleaner for `ConnectionListener` to later dispose all of it's own artifacts instead of passing them through all it's method callers up to the `extension.ts`?

By returning the `Disposable` we're also making it quite difficult for any function which in future might be using `onConnectionRestored` to return anything other than `Disposables`.

packages/vscode-extension/src/ai/mcp/index.ts
latekvo: Fixed in [4f9e1c4](https://github.com/software-mansion/radon-ide/pull/1313/commits/4f9e1c4b429a6955cc5c32ffa7519797a39e90a8).

packages/vscode-extension/src/ai/mcp/index.ts
latekvo: Fixed in [4f9e1c4](https://github.com/software-mansion/radon-ide/pull/1313/commits/4f9e1c4b429a6955cc5c32ffa7519797a39e90a8).

packages/vscode-extension/src/ai/shared/ConnectionListener.ts
latekvo: Also, not sure if this is relevant, but I can see the exact same pattern of `private disposables: Disposable[]` being used in [`Tabpanel.ts`](https://github.com/software-mansion/radon-ide/blob/45bb5f8997fecef38b1eb0e95da6dd35b7f3b900/packages/vscode-extension/src/panels/Tabpanel.ts#L24C1-L25C1), [`Connector.ts`](https://github.com/software-mansion/radon-ide/blob/45bb5f8997fecef38b1eb0e95da6dd35b7f3b900/packages/vscode-extension/src/connect/Connector.ts#L28), and [`+7`](https://github.com/search?q=repo%3Asoftware-mansion%2Fradon-ide%20%3A%20Disposable%5B%5D&type=code) other classes within `Radon IDE`.

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -8,49 +8,50 @@ import { startLocalMcpServer } from "./server";
...
+          "RadonAI",
+          Uri.parse(`http://127.0.0.1:${port}/mcp`),
+          {},
+          extensionContext.extension.packageJSON.version + `.${versionSuffix++}`
+        ),
+      ];
+    },
+  });
+
+  return [connectionChangeListener, licenseChangeListener, mcpServerEntry];
```
jwajgelt: You can wrap this in a single `Disposable` for easier consumption


```suggestion
  return new Disposable(() => [connectionChangeListener, licenseChangeListener, mcpServerEntry].forEach(d => d.dispose()));
```

similarly elsewhere where you return `Disposable[]`

jwajgelt (approved):
The caller of the function shouldn't really have to deal with `Disposable[]`. It's better to return a single `Disposable` that wraps all of the resources that need cleaning up, and return that instead.

Other than that, I don't have any further comments. Feel free to merge.

packages/vscode-extension/src/ai/shared/ConnectionListener.ts
jwajgelt: > Instead, since ConnectionListener is already a Disposable, and is already registered within context.subscriptions, it's possible for ConnectionListener to store references to all the Disposable it creates. Wouldn't it be cleaner for ConnectionListener to later dispose all of it's own artifacts instead of passing them through all it's method callers up to the extension.ts?

The issue with that is, it's not really `ConnectionListener` that "creates" the `Disposable`, it's the caller of the method -- the caller subscribes for the event, and should be responsible for unsubscribing -- otherwise, the caller has no way of telling the `ConnectionListener` that they no longer care about the event. It also means that, once the caller registers for the event, the `ConnectionListener` will retain a reference to it (in order to call the callback), creating potential memory leaks.

latekvo: Thanks, that makes sense

packages/vscode-extension/src/ai/mcp/index.ts
latekvo: Fixed in [0c8559d](https://github.com/software-mansion/radon-ide/pull/1313/commits/0c8559d3b4b92dbb0b52e8aad526bfa3b4998441), will be merging after i finish testing it one last time.

PR #1314, opened 2025-07-09
Add UI for Radon Connect and enable in public builds

kmagiera:
This PR enables Radon Connect in public builds (previously was only available in dev mode) and adds necessary UI elements for Radon Connect to be controlled from the panel.

This PR introduces several changes across the codebase:
1) we expose radon connect to the webview via a state variable, this way the webview is able to update according to that state and also control radon connect (enable / disable)
2) we introduce additional logic that controls device sessions when radon connect is enabled / disabled: When user enables radon connect we terminate all sessions, and when they disable radon connect we force the initial session to boot up.
3) we add new pieces of the UI that are used when Radon Connect is enabled (see below).
4) we change the device selector such that it displays radon connect as option. When radon connect is selected, we disable some buttons in the panel UI that aren't used in radon connect mode + we also hide the project select menu which isn't used.
5) we update the debugger logic such that for Radon Connect it doesn't hide the debugger controls that can be needed because the on-device debugger only has "resume" button but no step-over
6) we update the activity item text to use consistent wording
7) we disable code that was checking for dev mode when enabling radon connect

### New UI elements:

**1) Device selection dropdown now has _Radon Connect_ option**
image

**2) Waiting for connection panel view**
image

**3) Connected panel view***
image

**4) Activity item when waiting for connection**
image

**5) Activity item when connected**
image

**6) We now use activity item instead of the "Open IDE Panel" text in the editor context action menu**
image

### How Has This Been Tested: 
1) Run example app, open IDE panel
2) Test enabling radon connect from Activity Item menu (should result in both the activity item and the panel updating)
3) Disable radon connect from the panel by selecting device – should disconnect the debug session and update activity item
4) Test disabling radon connect from activity item menu
5) When in Radon connect test turning off device and starting it again to verify that it re-connects properly.

### DIFF
--- a/packages/vscode-extension/src/webview/views/PreviewView.css
+++ b/packages/vscode-extension/src/webview/views/PreviewView.css
@@ -66,8 +66,6 @@
   background-color: var(--swm-separator);
 }
 
-
-
 .device-select {
   min-width: 200px;
 }


packages/vscode-extension/src/common/Project.ts
```diff
@@ -117,12 +117,18 @@ export interface DeviceSessionsManagerState {
   deviceSessions: Record<DeviceId, DeviceSessionState>;
 }
 
+export type ConnectState = {
+  enabled: boolean;
+  connected: boolean;
+};
```
jwajgelt, line 123: Maybe this could be a tri-state enum?
```suggestion
export enum ConnectState {
Disabled,
Disconnected,
Connected
};
```
Since I assume we never want to be `!enabled && connected`

packages/vscode-extension/src/project/project.ts
```diff
@@ -92,6 +95,21 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
+
+    connector.delegate = {
+      onConnectStateChanged: (connectState) => {
+        this.updateProjectState({ connectState });
+        if (connectState.enabled) {
+          this.deviceSessionsManager.terminateAllSessions();
+        } else {
+          this.deviceSessionsManager.findInitialDeviceAndStartSession();
+        }
+      },
```
jwajgelt: This approach feels awkward to me.
Usually, we have the `delegate` passed in the ctor, instead of having whatever wants to listen to the "delegating" object register itself on it like we do here.
I think in those cases it's more natural to have an `EventEmitter` + `onConnectStateChanged(cb)` listener registration. 
When reading `Connector` before, I was a bit confused when I didn't see any place assigning the `delegate` and had to hunt it down, while if we had the regular event setup it would be more obvious that registering the listener happens elsewhere.

packages/vscode-extension/src/connect/Connector.ts
```diff
@@ -169,7 +179,14 @@ export class Connector implements Disposable {
...
 
-  private updateStatusBarItem() {
+  private handleStateChange() {
+    // trigger delegate
+    this.delegate?.onConnectStateChanged({
+      enabled: this.isEnabled,
+      connected: this.isConnected,
+    });
+
+    // udpate status bar item
```
jwajgelt:
```suggestion
    // update status bar item
```

packages/vscode-extension/src/project/project.ts
```diff
@@ -84,6 +85,8 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
       this
     );
 
+    const connector = Connector.getInstance();
```
jwajgelt, line 88: Can we inject this in the constructor, instead of making it a global/Singleton?

packages/vscode-extension/src/common/Project.ts
kmagiera: that's right, however, separating these feels more logical as the enabled state is controlled by the user whereas connected state is set automatically when Radon finds available metro instance. We currently disconnect connected metro when user disables connect, so it isn't possible for it to be in connected and disabled state, but still feels more natural to have distinct variables

packages/vscode-extension/src/webview/components/DeviceSelect.tsx
```diff
@@ -119,21 +125,21 @@ function DeviceSelect() {
...
+
   return (
-    <Select.Root
-      onValueChange={handleDeviceDropdownChange}
-      // NOTE: we pass the placeholder text as the value because passing `undefined` causes
-      // issues with the "manage devices" option.
-      // See https://github.com/software-mansion/radon-ide/pull/1231#pullrequestreview-2910304970
-      value={selectedDevice?.id ?? placeholderText}>
+    <Select.Root onValueChange={handleDeviceDropdownChange} value={value}>
       <Select.Trigger className="device-select-trigger" disabled={hasNoDevices}>
```
```suggestion
      <Select.Trigger className="device-select-trigger">
```
since you can always select Connect, right?

packages/vscode-extension/src/connect/Connector.ts
```diff
@@ -72,32 +77,37 @@ export class Connector implements Disposable {
...
+  public disable() {
     extensionContext.workspaceState.update(RADON_CONNECT_ENABLED_KEY, false);
     this.disconnect();
     this.stopScanner();
-    this.updateStatusBarItem();
+    this.handleStateChange();
   }
 
   public get isEnabled() {
     return extensionContext.workspaceState.get(RADON_CONNECT_ENABLED_KEY, true);
```
jwajgelt:
```suggestion
    return extensionContext.workspaceState.get(RADON_CONNECT_ENABLED_KEY, false);
```
It should probably be disabled by default, for new users?

PR #1315, opened 2025-07-09
Add warning when the currently selected custom config is removed

jwajgelt:
Displays an alert when the currently selected custom config is removed from the `launch.json` file, either because the file is edited by hand, or through the "Delete" option in the UI.

image

### How Has This Been Tested: 
- add a custom launch configuration and select it
- verify that removing it in various ways makes the alert appear:
  - remove it via UI (launch config select -> gear icon -> "Delete" button in modal)
  - remove it from the `launch.json` file manually
  - change it in the `launch.json` file manually
- verify the alert can be dismissed with the "close" button

### DIFF
--- a/packages/vscode-extension/src/project/launchConfigurationsManager.ts
+++ b/packages/vscode-extension/src/project/launchConfigurationsManager.ts
@@ -41,7 +41,8 @@ export function launchConfigurationFromOptions(
 
 function launchConfigFromOptionsWithDefaultAppRoot(
   options: LaunchConfigurationOptions,
-  defaultAppRoot: string | undefined
+  defaultAppRoot: string | undefined,
+  launchConfigurationKind: LaunchConfigurationKind = LaunchConfigurationKind.Custom
 ): LaunchConfiguration {
   if ((options.appRoot ?? defaultAppRoot) === undefined) {
     const maybeName =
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/project/launchConfigurationsManager.ts
```diff
@@ -104,7 +109,10 @@ export class LaunchConfigurationsManager implements Disposable {
     if (this._launchConfigurations.length > 0) {
       return this._launchConfigurations[0];
     }
-    return launchConfigFromOptionsWithDefaultAppRoot({}, findDefaultAppRoot(true));
+    return launchConfigFromOptionsWithDefaultAppRoot(
+      { kind: LaunchConfigurationKind.Detected } as LaunchConfigurationOptions,
```
filip131311: why want you just change the type of "options argument or add it as a sperate one instead?

packages/vscode-extension/src/webview/components/AppRootSelect.tsx
```diff
@@ -103,6 +109,34 @@ function renderCustomLaunchConfigurations(
...
+
+  const alertId = "unknown-launch-configuration-alert";
+
+  useEffect(() => {
+    if (shouldOpen && !isOpen(alertId)) {
+      openAlert({
+        id: alertId,
+        title: "Unknown launch configuration",
+        description:
+          "The selected launch configration is no longer specified in the workspace's launch.json file. " +
```
filip131311: I think sice it can be triggered by just a modification and not only by deletion of the config this message should say more something along the lines of 

"The selected launch configration was modified or deleted..."

filip131311 (approved):
The description of the PR has outdated screenshot now.

PR #1320, opened 2025-07-10
Restore last used launch config on IDE start

jwajgelt:
Restores the last used launch configuration on start.
The implementation checks if the last used config is still configured, and if not, falls back to the first config on the list.

### How Has This Been Tested: 
- open a workspace
- add two (or more) launch configs
- select any custom config
- close and reopen the IDE
- the same config should be selected

### DIFF
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -48,32 +48,10 @@ import { LaunchConfiguration, LaunchConfigurationOptions } from "../common/Launc
 
 const PREVIEW_ZOOM_KEY = "preview_zoom";
 const DEEP_LINKS_HISTORY_KEY = "deep_links_history";
-const INITIAL_LAUNCH_CONFIGURATION_KEY = "initialLaunchConfiguration";
 
 const DEEP_LINKS_HISTORY_LIMIT = 50;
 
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/project/project.ts
```diff
@@ -48,11 +48,33 @@ import { LaunchConfiguration, LaunchConfigurationOptions } from "../common/Launc
 
 const PREVIEW_ZOOM_KEY = "preview_zoom";
 const DEEP_LINKS_HISTORY_KEY = "deep_links_history";
+const INITIAL_LAUNCH_CONFIGURATION_KEY = "initialLaunchConfiguration";
 
 const DEEP_LINKS_HISTORY_LIMIT = 50;
 
 const MAX_RECORDING_TIME_SEC = 10 * 60; // 10 minutes
 
+function getInitialLaunchConfig(launchConfigsManager: LaunchConfigurationsManager) {
```
filip131311: I really don't like that this is happening in project, we have multiple classes for managing launch cofiguratuon and this seems to be in their domain

jwajgelt: `project` saves the launch configuration to `workspaceState`, and I don't want to have multiple places in the code where we write and read this specific `workspaceState` entry.
That being said, we can probably move this to launch configs manager together with the saving code.

PR #1322, opened 2025-07-11
select launch config when started from vscode UI

jwajgelt:
When opening Radon via the launch action in the "Run and Debug" menu, selects the run launch configuration in Radon automatically.

### How Has This Been Tested: 
- open vscode window (with Radon closed)
- configure multiple launch configurations
- "Run and Debug" menu
- select a config from the dropdown and press "Start debugging"
- Radon should open with the chosen configuration selected
- select a different config and press "Start debugging"
- Radon should switch to the chosen configuration

### DIFF
--- a/packages/vscode-extension/src/extension.ts
+++ b/packages/vscode-extension/src/extension.ts
@@ -325,10 +325,16 @@ export async function activate(context: ExtensionContext) {
 }
 
 class LaunchConfigDebugAdapterDescriptorFactory implements vscode.DebugAdapterDescriptorFactory {
-  createDebugAdapterDescriptor(
+  async createDebugAdapterDescriptor(
     session: vscode.DebugSession
-  ): vscode.ProviderResult<vscode.DebugAdapterDescriptor> {
-    commands.executeCommand("RNIDE.openPanel");
+  ): Promise<vscode.DebugAdapterDescriptor> {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/extension.ts
```diff
@@ -325,10 +325,16 @@ export async function activate(context: ExtensionContext) {
...
-  createDebugAdapterDescriptor(
+  async createDebugAdapterDescriptor(
     session: vscode.DebugSession
-  ): vscode.ProviderResult<vscode.DebugAdapterDescriptor> {
-    commands.executeCommand("RNIDE.openPanel");
+  ): Promise<vscode.DebugAdapterDescriptor> {
+    await commands.executeCommand("RNIDE.openPanel");
+    const ide = IDE.attach();
+    try {
+      ide.project.selectLaunchConfiguration(session.configuration);
```
filip131311: I don't like this implementation as it exposes us to the risk of overidding the configuration already selected by the "openPanel". 

It would be much better solution if we wnet with passing arguments to the "OpenPanel" command so there is not risk of "flickering"

filip131311 (changes requested):
inline comments

filip131311 (approved):
👍

PR #1323, opened 2025-07-11
Refactor launch configuration modal

kmagiera:
This PR updates the launch configuration modal to use vscode-elements and provide better and more intuitive guidance when configuring the launch settings.

One of the main logical change is that we show/hide fields depending on the selection:
1) we group Android and iOS build settings under separate tabs
2) for each platform, we first ask for the build type to be selected and allow the following types: Standard, EAS (cloud), EAS Loccal, Custom
3) Depending on the build type we ask for different configuration to be provided, as it only makes sense for certain fields to be present when a given configuration is open:
 - For Standard configuration we ask for build scheme, configuration name (on iOS), and for flavor on Android
 - For EAS cloud, we ask for profile name and UUID matching scheme
 - For EAS Local we only ask for profile name
 - For Custom build we ask for build command and fingerprint command

On top of that:
1) we change the form to use uncontrolled components and use form data to extract the settings details prior to saving
2) we use vscode-elements and use single-select with "createable" flag for fields like appRoot where we give initial selection but allow it to be a free-form field eventually
3) we add "Edit in launch.json" button that redirects to launch configuration
4) we use text description from the launch config scheme in package.json to provide description for the fields.
5) we edit some of those fields to provide a more condensed description.
6) since the settings use components from vscode-elements, we rely on --vscode color variables in the styles. This is because for example tab panel from vscode-elements has a background color set using the vscode color variables

**Before**
image

**After**
image

**After light theme**
image

### How Has This Been Tested: 
1) Open random project
2) Add configuration using the new form, and save
3) See the config being created (open launch.json)
4) Edit the config and change single field, see it being reflected onto the UI
5) Add another config with some different settings to see them being added to the file as well
6) Test under different themes including the radon themes

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -230,7 +230,7 @@ export interface ProjectInterface {
    */
   createOrUpdateLaunchConfiguration(
     newLaunchConfiguration: LaunchConfigurationOptions | undefined,
-    oldLaunchConfiguration?: LaunchConfiguration
+    oldLaunchConfiguration?: LaunchConfigurationOptions
   ): Promise<void>;
   selectLaunchConfiguration(launchConfig: LaunchConfigurationOptions): Promise<void>;
 
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -230,7 +230,7 @@ export interface ProjectInterface {
    */
   createOrUpdateLaunchConfiguration(
     newLaunchConfiguration: LaunchConfigurationOptions | undefined,
-    oldLaunchConfiguration?: LaunchConfigurationOptions
+    oldLaunchConfiguration?: LaunchConfiguration
   ): Promise<void>;
   selectLaunchConfiguration(launchConfig: LaunchConfigurationOptions): Promise<void>;
 
--- a/packages/vscode-extension/src/webview/hooks/useFormValidityTrigger.tsx
+++ b/packages/vscode-extension/src/webview/hooks/useFormValidityTrigger.tsx
@@ -1,0 +1,25 @@
+import { useEffect, RefObject } from "react";
+
+function useFormValidityTrigger(formRef: RefObject<HTMLFormElement | null>, callback: () => void) {
+  useEffect(() => {
+    const form = formRef.current;
+    if (!form) return;
+
+    callback(); // Initial check
+
+    // Watch for changes to the form structure (i.e. new required fields added
+    // or values changed)
+    const observer = new MutationObserver(callback);
+    observer.observe(form, { childList: true, subtree: true, attributes: true });
+
+    // Watch for input changes
+    form.addEventListener("input", callback);
+
+    return () => {
+      observer.disconnect();
+      form.removeEventListener("input", callback);
+    };
+  }, [callback]);
+}
+
+export default useFormValidityTrigger;
--- a/packages/vscode-extension/src/webview/hooks/useFormValidityTrigger.tsx
+++ b/packages/vscode-extension/src/webview/hooks/useFormValidityTrigger.tsx
@@ -8,8 +8,10 @@ function useFormValidityTrigger(formRef: RefObject<HTMLFormElement | null>, call
     callback(); // Initial check
 
     // Watch for changes to the form structure (i.e. new required fields added
-    // or values changed)
-    const observer = new MutationObserver(callback);
+    // or values changed). We use Mutation Observer which blocks layout and
+    // hence we delay the actual callback to run after the next tick.
+    const delayedCallback = () => setTimeout(callback, 0);
+    const observer = new MutationObserver(delayedCallback);
     observer.observe(form, { childList: true, subtree: true, attributes: true });
 
     // Watch for input changes
--- a/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
+++ b/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
@@ -24,7 +24,7 @@ import {
 import extensionPackageJSON from "../../../package.json";
 
 /**
- * Vscode element components are controller, this is a simple wrapper allowing
+ * Vscode element components are controlled, this is a simple wrapper allowing
  * it to be used as not-controlled component with an initial value.
  */
 function TextField({
--- a/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
+++ b/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
@@ -324,7 +324,7 @@ function BuildConfiguration({
         <SingleSelect
           value={buildType}
           name={`buildType.${platform}`}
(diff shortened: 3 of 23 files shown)


packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -1,57 +1,159 @@
...
+  VscodeFormHelper as FormHelper,
+  VscodeButton as Button,
+  VscodeTabPanel as TabPanel,
+  VscodeTabs as Tabs,
+  VscodeTabHeader as TabHeader,
+} from "@vscode-elements/react-elements";
+import extensionPackageJSON from "../../../package.json";
+
+/**
+ * Vscode element components are controller, this is a simple wrapper allowing
```
jwajgelt:
```suggestion
 * Vscode element components are controlled, this is a simple wrapper allowing
```

packages/vscode-extension/src/common/Project.ts
```diff
@@ -230,7 +230,7 @@ export interface ProjectInterface {
    */
   createOrUpdateLaunchConfiguration(
     newLaunchConfiguration: LaunchConfigurationOptions | undefined,
-    oldLaunchConfiguration?: LaunchConfiguration
+    oldLaunchConfiguration?: LaunchConfigurationOptions
```
jwajgelt: What's the reason for this change?
You cannot actually pass the `Options` here, it will mess up the equality check in the implementation. Typing it like that opens us up to making this mistake and introducing bugs.

jwajgelt (changes requested):
I have only one real issue, other than that it looks ok.

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -84,351 +186,322 @@ function LaunchConfigurationView({ launchConfigToUpdate }: LaunchConfigurationVi
...
+            secondary
             onClick={() => {
               openModal("", <DeleteConfirmationModal />);
             }}>
+            <span className="codicon codicon-trash" />
             Delete
           </Button>
         )}
-        <Button onClick={save}>Save{isEditingSelectedConfig ? " and restart device" : ""}</Button>
+        <Button onClick={save}>Save{isEditingSelectedConfig ? " and restart" : ""}</Button>
```
jwajgelt: Previously we checked if the edited launch config was already running.
Now we display "Save and restart" if any existing config is being edited, even though we don't _actually_ restart any devices in that case.
Why did you make that change?

kmagiera: that's an error on my side. I misunderstood the condition that was used before. will fix this

packages/vscode-extension/src/common/Project.ts
kmagiera: fixed

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -84,351 +182,322 @@ function LaunchConfigurationView({ launchConfigToUpdate }: LaunchConfigurationVi
...
+          <Label>
+            <span className="setting-item-category">iOS:</span> Scheme name
+          </Label>
+          <FormHelper>
+            {launchConfigAttrs?.properties?.ios?.properties?.scheme?.description}
+          </FormHelper>
+          <SingleSelect initialValue={config?.ios?.scheme ?? "auto"} name="ios.scheme">
+            <Option disabled value="auto">
+              Detect automatically
+            </Option>
```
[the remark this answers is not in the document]
kmagiera: there's no placeholder prop on SingleSelect 🙄

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -1,57 +1,152 @@
...
+          buildCommand: data[`customBuild.${platform}.buildCommand`],
+          fingerprintCommand: data[`customBuild.${platform}.fingerprintCommand`],
+        },
+      };
+    } else if (buildType === "eas") {
+      newConfig.eas = {
+        ...newConfig.eas,
+        [platform]: {
+          profile: data[`eas.${platform}.profile`],
+          buildUUID: data[`eas.${platform}.buildUUID`],
```
jwajgelt: Shouldn't this be `undefinedIfEmpty`, since this is optional?

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -1,57 +1,152 @@
...
+          buildType: undefinedIfEmpty(data["android.buildType"]),
+          productFlavor: undefinedIfEmpty(data["android.productFlavor"]),
+        };
+      }
+    } else if (buildType === "custom") {
+      newConfig.customBuild = {
+        ...newConfig.customBuild,
+        [platform]: {
+          buildCommand: data[`customBuild.${platform}.buildCommand`],
+          fingerprintCommand: data[`customBuild.${platform}.fingerprintCommand`],
```
jwajgelt: Shouldn't this be `undefinedIfEmpty`, since this is optional?

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -1,57 +1,152 @@
...
+        newConfig.android = {
+          buildType: undefinedIfEmpty(data["android.buildType"]),
+          productFlavor: undefinedIfEmpty(data["android.productFlavor"]),
+        };
+      }
+    } else if (buildType === "custom") {
+      newConfig.customBuild = {
+        ...newConfig.customBuild,
+        [platform]: {
+          buildCommand: data[`customBuild.${platform}.buildCommand`],
```
jwajgelt, line 109: We may want to check if this (and other required values) are empty, and disable the `save` button in that case.

jwajgelt (commented):
Left some comments re `serializeLaunchConfig`

packages/vscode-extension/src/webview/hooks/useFormValidityTrigger.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+    // Watch for input changes
+    form.addEventListener("input", callback);
+
+    return () => {
+      observer.disconnect();
+      form.removeEventListener("input", callback);
+    };
+  }, [callback]);
+}
```
jwajgelt: NIT:
`useFormValidityTrigger` doesn't seem to have anything to do with the form's validity, it just watches the form for any changes and calls an arbitrary callback.
I feel like something like
```ts
// returns true if form is valid, false otherwise
function useFormValidity(formRef): boolean {/* ... */}
/* 
  ... 
*/
// somewhere in the component
const isValid = useFormValidity(formRef);
```
would be better in the sense that it encapsulates the extra state variable and is easier to read in the component code.

jwajgelt (approved):
Looking good 👍

PR #1327, opened 2025-07-14
hotifx: don't serialize launch config `kind` field

jwajgelt:
Fixes the `kind` launch configuration field being serialized despite being meant for internal use

### How Has This Been Tested: 
- save a launch config through the UI
- verify the entry in `launch.json` does not have a `kind` field


filip131311 (approved):
👍

PR #1330, opened 2025-07-15
Add env variable editor to launch config modal

kmagiera:
This PR adds an env variable editor to launch configuration modal. Previously, the ability to edit it was missing from the dialog.

The new editor resembles similar record-based input elements from VScode settings providing a familiar interface.

Eventually, it is backed by a hidden form field allowing the form serialization logic to access its value and to provide initial state.

**Env editor side-by-side with json content**
image

**Adding new entry**
image

**Editing row**
image

### How Has This Been Tested: 
1. Open launch config editor
2. Add new entries to env -> see them show up in the launch.json
3. Try editing rows and deleting rows -> again, see changes reflected in json file.


filip131311 (changes requested):
I tried editing the envs but the ide does not save them 

https://github.com/user-attachments/assets/37f5c7dc-fc73-4347-93e9-6f2e3bb5b821

kmagiera:
@filip131311 this was due to a bad merge, I think I fixed all the merge related issues and also improved the generated form code

packages/vscode-extension/src/webview/views/EnvEditor.css
```diff
@@ -0,0 +1,18 @@
...
+  justify-content: flex-end;
+  gap: 4px;
+	margin: 4px 0;
+}
+
+.env-editor-add-form {
+  display: flex;
+  gap: 4px;
+  justify-content: flex-end;
+}
```
filip131311: nit: You missed the end-line character here

filip131311 (approved):
the modal looks very good now

packages/vscode-extension/src/webview/views/EnvEditor.tsx
```diff
@@ -0,0 +1,163 @@
...
+                    </>
+                  )}
+                </div>
+              </VscodeTableCell>
+            </VscodeTableRow>
+          ))}
+        </VscodeTableBody>
+      </VscodeTable>
+      {!showAddForm && <Button onClick={() => startEditing(null)}>Add Variable</Button>}
+      <input type="hidden" name="env" value={JSON.stringify(env)} />
```
jwajgelt, line 159: I don't know how I feel about this vs having a controlled prop / returning the value via callback.

packages/vscode-extension/src/webview/views/EnvEditor.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const [editingKeyValue, setEditingKeyValue] = useState<string>("");
+  const [editingValue, setEditingValue] = useState<string>("");
+  const [showAddForm, setShowAddForm] = useState<boolean>(false);
+
+  const updateEnv = (newEnv: Record<string, string>) => {
+    setEnv(newEnv);
+    onChange?.(newEnv);
+  };
+
+  const startEditing = (key: string | null) => {
```
jwajgelt, line 32: Could we have this focus the `key` text field for the new/edited entry as well?

jwajgelt (approved):
It looks nice and seems to work well. Good work 👍

packages/vscode-extension/src/webview/views/EnvEditor.tsx
kmagiera: I wanted to do that but for some reason autofocus prop doesn't work in vscode-elements. Would need to add a bunch of more workarounds for this to work apparently.

packages/vscode-extension/src/webview/views/EnvEditor.tsx
kmagiera: this is mainly for consistency reasons with all the other fields that we now collect via the FormData. Otherwise this would be the only one controlled component which would feel off.

PR #1338, opened 2025-07-16
Add additional telemetry

filip131311:
This PR adds additional telemetry to help us understand new functionality introduced in version 1.9.0 namely "radon connect".

Additionally we add events for inspector functionality which were missed from previous versions.

### DIFF
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -18,7 +18,7 @@ import {
 import { useResizableProps } from "../hooks/useResizableProps";
 import ZoomControls from "./ZoomControls";
 import { throttle } from "../../utilities/throttle";
-import { Platform } from "../providers/UtilsProvider";
+import { Platform, useUtils } from "../providers/UtilsProvider";
 import DimensionsBox from "./DimensionsBox";
 import ReplayUI from "./ReplayUI";
 import MjpegImg from "../Preview/MjpegImg";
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -166,6 +168,9 @@ function Preview({
     if (type === "Leave") {
       return;
     }
+    if (type === "RightButtonDown") {
+      sendTelemetry("inspector:right-button-down", {});
```
kacperkapusciak:
```suggestion
      sendTelemetry("inspector:show-hierarchy", {});
```
maybe its better to describe what is does and not what is it

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -341,7 +341,12 @@ function PreviewView() {
           tooltip={{
             label: "Select an element to inspect it",
           }}
-          onClick={() => setIsInspecting(!isInspecting)}
+          onClick={() => {
+            sendTelemetry("inspector:inspector-button-clicked", {
```
kacperkapusciak: i'm thinking whether `inspector:enabled` would be better but there are different ways to enable inspector so idk 🤔 

just 

`inspector:button-clicked` would be enough tho

filip131311: it is not the same as "enabled" because using this button can also disable the inspector

PR #1343, opened 2025-07-17
refactor: store and cleanup log outputs in IDE, uniform API for focusing output in project

jwajgelt:
- introduces `OutputChannelRegistry` which stores all the created `vscode` `LogOutputChannel` instances
- changes `Project` to expose a single function `focusOutput(name)` instead of separate `focusBuildOutput`, `focusExtensionLogsOutput`
- removes the build output channel from `BuildManager` (instead it gets its channel from `Registry`), reducing the amount of state it holds.
- fixes a bug where `Project.focusBuildOutput` would always focus the last started build, instead of the build for the currently selected device

### How Has This Been Tested: 
- open Radon IDE
- verify build logs go to the correct output tab
- verify pressing on "Booting" while the device starts opens the IDE logs
- verify that if the project build fails, the button in the error alert leads to the correct logs output

### DIFF
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -75,7 +75,7 @@ export class IosSimulatorDevice extends DeviceBase {
   }
 
   private get nativeLogsOutputChannel() {
-    return this.outputChannelRegistry.getChannel(Output.IosDevice);
+    return this.outputChannelRegistry.getOrCreateOutputChannel(Output.IosDevice);
   }
 
   public dispose() {
--- a/packages/vscode-extension/src/project/OutputChannelRegistry.ts
+++ b/packages/vscode-extension/src/project/OutputChannelRegistry.ts
@@ -7,7 +7,7 @@ export class OutputChannelRegistry implements Disposable {
     [Output.Ide, Logger.rawOutputChannel],
   ]);
 
-  getChannel(channel: Output) {
+  getOrCreateOutputChannel(channel: Output) {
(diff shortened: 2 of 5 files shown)


packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -67,17 +68,18 @@ export class IosSimulatorDevice extends DeviceBase {
     return this._deviceInfo;
   }
 
-  get lockFilePath(): string {
+  public get lockFilePath(): string {
```
filip131311, line 71: seems unrelated to that PR

packages/vscode-extension/src/project/OutputChannelRegistry.ts
```diff
@@ -0,0 +1,24 @@
+import { Disposable, LogOutputChannel, window } from "vscode";
+import { Output } from "../common/OutputChannel";
+import { Logger } from "../Logger";
+
+export class OutputChannelRegistry implements Disposable {
+  private channelByName = new Map<Output, LogOutputChannel>([
+    [Output.Ide, Logger.rawOutputChannel],
+  ]);
+
+  getChannel(channel: Output) {
```
filip131311: this name could indcate that the new chanel can be created

filip131311 (approved):
very small nits it the comments

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
jwajgelt: `public` is the default visibility in TS, so this doesn't actually change anything. It just aligns this class closer with others (where we use it explicitely)

PR #1344, opened 2025-07-17
fix: unknown launch config always displays for detected configs

jwajgelt:
After #1339, selecting a "detected" AppRoot always displays the "unknown launch config" alert. This is due to the `kind` field sent by the frontend being now stripped from options before creating the full launch config. This PR fixes that by explicitly asking for the kind when selecting the config.

### How Has This Been Tested: 
- open Radon IDE
- pick a "Detected Application Root"
- verify the "Unknown Launch Config" alert doesn't appear
- verify the approot is displayed as "selected" in the launch config selector


filip131311 (commented):
why can you just go like this: 

```js
const launchConfigurationKind = options.kind ?? LaunchConfigurationKind.Custom
// only pick the keys that we care about from `options`
  options = _.pick(options, LAUNCH_CONFIG_OPTIONS_KEYS);
```

?

jwajgelt:
> why can you just go like this:
> 
> ```js
> const launchConfigurationKind = options.kind ?? LaunchConfigurationKind.Custom
> // only pick the keys that we care about from `options`
>   options = _.pick(options, LAUNCH_CONFIG_OPTIONS_KEYS);
> ```
> 
> ?

`kind` is not a part of the `LaunchConfigOptions` interface, so this is a bit of an abuse of the method's interface.
Instead, we could introduce an extra `LaunchConfigurationOptionsWithMaybeKind` and have the functions accept that, but I refuse.

filip131311 (approved):
:+1:

PR #1346, opened 2025-07-18
refactor: remove BuildManager's dependency on DependencyManager

jwajgelt:
Moves installing / checking of dependencies out of `BuildManager`, removing the dependency on `DependencyManager`
### How Has This Been Tested: 
- open app in Radon
- verify it still builds
- check that removing `ios` or `android` folder in a standard build causes errors
- check that pods are installed if not present

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -4,11 +4,11 @@ import assert from "assert";
 import _ from "lodash";
 import {
   commands,
+  window,
   DebugSessionCustomEvent,
   Disposable,
   extensions,
   Uri,
-  window,
   workspace,
 } from "vscode";
 import { MetroLauncher, MetroDelegate } from "./metro";
@@ -49,6 +49,9 @@ import { focusSource } from "../utilities/focusSource";
 import { ApplicationContext } from "./ApplicationContext";
 import { BuildCache } from "../builders/BuildCache";
 import { watchProjectFiles } from "../utilities/watchProjectFiles";
+import { BuildConfig, BuildType } from "../common/BuildConfig";
+import { OutputChannelRegistry } from "./OutputChannelRegistry";
+import { Output } from "../common/OutputChannel";
 
 const MAX_URL_HISTORY_SIZE = 20;
 const CACHE_STALE_THROTTLE_MS = 10 * 1000; // 10 seconds
@@ -124,7 +127,8 @@ export class DeviceSession
   constructor(
     private readonly applicationContext: ApplicationContext,
     private readonly device: DeviceBase,
-    private readonly deviceSessionDelegate: DeviceSessionDelegate
+    private readonly deviceSessionDelegate: DeviceSessionDelegate,
+    private readonly outputChannelRegistry: OutputChannelRegistry
   ) {
     this.devtools = this.makeDevtools();
     this.metro = new MetroLauncher(this.devtools, this);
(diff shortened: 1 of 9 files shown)


packages/vscode-extension/src/dependency/DependencyManager.ts
```diff
@@ -94,6 +102,30 @@ export class DependencyManager implements Disposable, DependencyManagerInterface
     this.checkEasCliInstallationStatus();
   }
 
+  public async ensureDependenciesForBuild(
+    buildConfig: BuildConfig,
+    outputChannel: OutputChannel,
```
filip131311, line 107: This is a bit confessing to me as this can only ever be an ios output channel. Could you pas the full registry here or type it in such a way that it is obvious that only ios output chanel may be relavan here?

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -722,17 +724,46 @@ export class DeviceSession
...
+      clean || buildDependenciesChanged,
       launchConfiguration,
       buildType
     );
+    const buildOutputChannel = this.outputChannelRegistry.getOrCreateOutputChannel(
+      this.platform === DevicePlatform.IOS ? Output.BuildIos : Output.BuildAndroid
+    );
+
+    const dependencyManager = this.applicationContext.dependencyManager;
+    await dependencyManager.ensureDependenciesForBuild(
```
filip131311, line 761: if this throws an error It should be wrapped as `BuildError` is this behavior kept after this changes?

packages/vscode-extension/src/dependency/DependencyManager.ts
jwajgelt: It can be any channel, the function shouldn't care.

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: Yes, the errors are caught outside of this function.

packages/vscode-extension/src/dependency/DependencyManager.ts
filip131311: Maybe I reframe my problem, as it is typed right now types allow for situation in which `outputChannel` is mismatched with the type of the build that is requested. It's not a big issue but it is annoying  as only ever one platform might use `outputChannel` in this function

jwajgelt: We may wish to handle different kinds of builds in this function in the future (for example, we may want to run prebuild here for expo projects).
We could create the channels inside this function, but then you duplicate the logic of picking the right channel for the build type / build config.
We _could_ have a function mapping a `BuildConfig` to the correct output channel, but I don't really like the idea -- we may want to split different build steps into different outputs, which then requires us to change all of this again, while with the current approach you have a single place (`DeviceSession`) which is responsible for setting "build-related" stuff up, and you can just create the channel there.
It would also mean that everywhere we may want to call that function, we need to access the `OutputChannelRegistry`, increasing coupling between components.

I don't understand why you're set on this being an issue. Sure, you _could_ pass the wrong output channel here, but you also _could_ have a bug in this function and create the wrong channel internally, and I feel like it's even more likely, since it's _harder_ for the caller of this function to accidentally pass a different channel to it than to the rest of the build process (as it would need to create two channels, where it should probably just use one), than it is for us to mismatch the channels we use outside of this function and inside it when we make changes here in the future.

PR #1347, opened 2025-07-18
Support device audio

balins:
This PR updates the Simulator Server to handle volume up / volume down buttons on iOS Simulator and provides UI controls to control the volume.

https://github.com/user-attachments/assets/34f6da2d-1a59-4ae0-b212-0e3104689a7f

Notice that the audio is outputted by the underlying iOS Simulator/Android Emulator, so it won't be present on videos recorded by the Simulator Server.

### How Has This Been Tested: 

Simply run iOS Simulator/Android Emulator, ensure volume is non-zero and the sound will be outputted by your default speakers. You can hold mouse down to change the volume faster (similarly to holding a button on a real device).

TBD

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -100,6 +100,30 @@
         "enablement": "RNIDE.extensionIsActive"
       },
       {
+        "command": "RNIDE.deviceVolumeIncreaseButtonDown",
+        "title": "Trigger volume increase button down on the current device",
+        "category": "Radon IDE",
+        "enablement": "RNIDE.extensionIsActive"
+      },
+      {
+        "command": "RNIDE.deviceVolumeIncreaseButtonUp",
+        "title": "Trigger volume increase button up on the current device",
+        "category": "Radon IDE",
+        "enablement": "RNIDE.extensionIsActive"
+      },
+      {
+        "command": "RNIDE.deviceVolumeDecreaseButtonDown",
+        "title": "Trigger volume decrease button down on the current device",
+        "category": "Radon IDE",
+        "enablement": "RNIDE.extensionIsActive"
+      },
+      {
+        "command": "RNIDE.deviceVolumeDecreaseButtonUp",
+        "title": "Trigger volume decrease button up on the current device",
+        "category": "Radon IDE",
+        "enablement": "RNIDE.extensionIsActive"
+      },
+      {
         "command": "RNIDE.captureReplay",
         "title": "Capture Replay",
         "category": "Radon IDE",
--- a/packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
+++ b/packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
@@ -149,6 +149,7 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
             Location
           </DropdownMenu.Item>
           <LocalizationItem />
+          <VolumeItem />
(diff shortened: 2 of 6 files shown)


packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -330,4 +331,57 @@ const CameraItem = () => {
...
+    project.runCommand("RNIDE.deviceVolumeIncreaseButtonUp");
+  };
+
+  const handleVolumeDecreaseDown = () => {
+    project.runCommand("RNIDE.deviceVolumeDecreaseButtonDown");
+  };
+
+  const handleVolumeDecreaseUp = () => {
+    project.runCommand("RNIDE.deviceVolumeDecreaseButtonUp");
+  };
```
jwajgelt: We shouldn't run commands directly in the webview.
SInce "commands" are a vscode-specific feature, it couples the frontend code more with vscode, which we're trying to move away from.
It also makes the API between the "extension" and the "webview" less defined.
Just call `project.dispatchButton` here.

packages/vscode-extension/package.json
```diff
@@ -99,6 +99,30 @@
...
+        "title": "Trigger volume decrease button down on the current device",
+        "category": "Radon IDE",
+        "enablement": "RNIDE.extensionIsActive"
+      },
+      {
+        "command": "RNIDE.deviceVolumeDecreaseButtonUp",
+        "title": "Trigger volume decrease button up on the current device",
+        "category": "Radon IDE",
+        "enablement": "RNIDE.extensionIsActive"
+      },
```
jwajgelt: Commands are exposed to the user via the command palette, and having separate "Button down" and "Button up" commands is not very useful there.
If we want to keep commands for volume control (and I think we do, it can be useful to bind them to keyboard shortcuts), we should have "Increase volume" and "Decrease volume" commands instead (which internally trigger "button down", "button up" actions -- that API is fine for the `Project` class)

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -330,4 +331,57 @@ const CameraItem = () => {
...
+          onMouseLeave={handleVolumeDecreaseUp}>
+          <span className="codicon codicon-chevron-down" />
+        </button>
+        <button
+          title="Volume Up"
+          className="volume-button"
+          onMouseDown={handleVolumeIncreaseDown}
+          onMouseUp={handleVolumeIncreaseUp}
+          onMouseLeave={handleVolumeIncreaseUp}>
+          <span className="codicon codicon-chevron-up" />
```
jwajgelt:
```suggestion
          <span className="codicon codicon-dash" />
        </button>
        <button
          title="Volume Up"
          className="volume-button"
          onMouseDown={handleVolumeIncreaseDown}
          onMouseUp={handleVolumeIncreaseUp}
          onMouseLeave={handleVolumeIncreaseUp}>
          <span className="codicon codicon-add" />
```
I think I'd prefer "+" and "-" here over the chevrons.

balins: Not sure if it's better but this is how it looks after the update (I've used `codicon-remove` instead of `codicon-dash` since `codicon-dash` is too short and imo doesn't look good next to with `codicon-add`).

Screenshot 2025-07-18 at 14 07 32

jwajgelt: I think it's better this way.
So far, I think we've used chevron icons exclusively for scrolling and drop-downs, so they looked out of place here for me.

PR #1359, opened 2025-07-21
fix: Logger output channel is disposed when panel closes, DeviceSessionsManager is not disposed correctly

jwajgelt:
- fixes a bug where the output channel for Logger would be disposed with IDE, causing issues when the IDE panel is closed and reopened
- fixes a bug where the DeviceSessionsManager would not be cleaned up when the IDE is disposed

### How Has This Been Tested: 
- open the IDE panel
- close it
- reopen it
- the IDE should work


filip131311 (approved):
I think we should move output chanel registry higher in the future but it okay for now

PR #1361, opened 2025-07-22
feat: load environment variables from project`s dotenv files

### LINKED ISSUE #1350: Read the dot env files

jwajgelt:
Loads the environment variables for the application root using `@expo/env` and appends it to the system and configured environment that is passed to the other parts of the extension.
Also changes the build cache key to use a hash of the environment, instead of the full stringified env.

Closes #1350 

### How Has This Been Tested: 
- check that adding a `.env` file in the app root folder results in the variables being added to the `ApplicationContext.launchConfig` object
- check that `.env` contents don't override values set in the system or custom launch config
- check that switching app roots doesn't carry over the variables loaded from the previously chosen approot

### DIFF
--- a/packages/vscode-extension/src/project/ApplicationContext.ts
+++ b/packages/vscode-extension/src/project/ApplicationContext.ts
@@ -21,15 +21,14 @@ export type ResolvedLaunchConfig = LaunchOptions & {
   env: Record<string, string>;
 };
 
-function resolveLaunchConfig(configuration: LaunchConfiguration): ResolvedLaunchConfig {
-  const appRoot = configuration.appRoot;
-  const absoluteAppRoot = path.resolve(workspace.workspaceFolders![0].uri.fsPath, appRoot);
-
-  const configuredEnv = configuration.env || {};
+function resolveEnvironment(
+  appRoot: string,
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/project/ApplicationContext.ts
```diff
@@ -18,12 +21,38 @@ export type ResolvedLaunchConfig = LaunchOptions & {
...
   const appRoot = configuration.appRoot;
   const absoluteAppRoot = path.resolve(workspace.workspaceFolders![0].uri.fsPath, appRoot);
+
+  const configuredEnv = configuration.env || {};
+
   return {
-    env: {},
     ...configuration,
+    get env() {
+      return resolveEnvironment(absoluteAppRoot, configuredEnv);
```
filip131311, line 54: why as a getter?

filip131311 (approved):
Looks good

packages/vscode-extension/src/project/ApplicationContext.ts
jwajgelt: In order to call `resolveEnvironment` each time the env is read.

PR #1362, opened 2025-07-22
fix: element inspector not working with Fragments wrapping empty components on Paper

### LINKED ISSUE #980: Code inspection does not always work with Fragment + empty component
**Description**
I've noticed that code inspector doesn't always work for our project. We often use empty components with some sideeffects. I was able to track it down and create a minimal repro example.

Code inspection does not work for the following example:
Right Clicking `View A` does not open the inspector and Left clicking doesn't navigate to code

```

function C() {
	return (
		<React.Fragment>
			<B />
			<A />
		</React.Fragment>
	);
}

function A() {
	return <View style={{ height: 100, backgroundColor: 'red' }} />;
}

function B() {
	return null;
}
```
It works correctly if we switch the order 
```
function C() {
	return (
		<React.Fragment>
			<A />
			<B />
		</React.Fragment>
	);
}
```
or wrap in View
```
function C() {
	return (
		<View>
			<B />
			<A />
		</View>
	);
}
```

**Environment**

Radon IDE : 1.3.1
React-native: 0.74
React: 18.2.0
Expo: 51

jwajgelt:
Fixes #980

### How Has This Been Tested: 
- open a project which uses the Paper (old) renderer in Radon
- render the C component:
```tsx
function C() {
  return (
    <React.Fragment>
      <B />
      <A />
    </React.Fragment>
  );
}

function A() {
  return <View style={{ height: 100, backgroundColor: "red" }} />;
}

function B() {
  return null;
}
```
- right click the red `View`
- element inspector should show the element stack with `<View>,  <A>, <C>` components

image

### DIFF
--- a/packages/vscode-extension/lib/wrapper.js
+++ b/packages/vscode-extension/lib/wrapper.js
@@ -149,16 +149,16 @@ function getInspectorDataForCoordinates(mainContainerRef, x, y, requestStack, ca
         inspectorDataStack.map(
           (inspectorData) =>
             new Promise((res, rej) => {
+              const source = {
+                fileName: inspectorData.source.fileName,
+                line0Based: inspectorData.source.lineNumber - 1,
+                column0Based: inspectorData.source.columnNumber - 1,
+              };
               try {
                 inspectorData.measure((_x, _y, viewWidth, viewHeight, pageX, pageY) => {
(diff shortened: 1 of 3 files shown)


filip131311 (approved):
👍

PR #1364, opened 2025-07-22
refactor: Start moving to legend state

filip131311:
This PR is the first one in an attempt to consolidate state in extension and webview runtimes, to achieve that we do the following:
- we move state management in the webview from multiple contexts to one that is using [legend state](https://legendapp.com/open-source/state/v3/intro/introduction/)
- we crate an universal way of reporting state changes in the extension side (look into `stateManager.ts` for implementation)

additionally this PR moves the workspace configuration management to the new system to show case all the functionality: 

- getting the state from the extension
- setting the state and updating the extension 
- updating state on the extension side and sending the information over 

### How Has This Been Tested: 

- use `radon-ide-test-apps` repository and play with it 
- change the "showDeviceFrame" property and using the debugger make sure that only the `Partial<State>` with the changes is send over not the whole state
- open the settings UI and change the settings see the change propagated to the webview side

This is an internal change


packages/vscode-extension/src/project/StateManager.ts
```diff
@@ -0,0 +1,100 @@
...
+
+type EventMap<T> = {
+  setState: Partial<T>;
+};
+
+export abstract class StateManager<T extends object> implements Disposable {
+  abstract setState(partialState: Partial<T>): void;
+  abstract getState(): T;
+
+  protected eventEmitter: EventEmitter = new EventEmitter();
```
jwajgelt: Use `vscode.EventEmitter` instead. They're nicer to work with (since they already return `Disposable`) and we've been moving to them in code we've written recently (see `Connector` or `RadonCDPProxyDelegate` for examples).

packages/vscode-extension/src/project/StateManager.ts
```diff
@@ -0,0 +1,100 @@
...
+    this.state = _.merge(this.state, partialState);
+    this.eventEmitter.emit("setState", partialState);
+  }
+
+  getState(): T {
+    return this.state;
+  }
+}
+
+export class DerivedStateManager<T extends object, K extends object> extends StateManager<T> {
```
jwajgelt:
```suggestion
class DerivedStateManager<T extends object, K extends object> extends StateManager<T> {
```
Is there a reason why this is exported? I'd see this as an implementation detail.

packages/vscode-extension/src/project/StateManager.ts
```diff
@@ -0,0 +1,100 @@
...
+      this.eventEmitter.off(event, listener);
+    });
+  }
+
+  public dispose() {
+    disposeAll(this.disposables);
+  }
+}
+
+export class RootStateManager<T extends object> extends StateManager<T> {
```
jwajgelt:
```suggestion
class RootStateManager<T extends object> extends StateManager<T> {
```
I wouldn't export this class.
Instead, we can have a `static createState<T>(initial: T): StateManager<T>` method on the abstract `StateManager` class

packages/vscode-extension/src/project/StateManager.ts
```diff
@@ -0,0 +1,100 @@
...
+import { EventEmitter } from "stream";
+import { Disposable } from "vscode";
+import { disposeAll } from "../utilities/disposables";
+
+type EventMap<T> = {
+  setState: Partial<T>;
+};
+
+export abstract class StateManager<T extends object> implements Disposable {
+  abstract setState(partialState: Partial<T>): void;
```
jwajgelt: I'm not sure how we would delete keys (for states with dynamic keys, like `DeviceSession`s or `LaunchConfiguration`s) with this API.
Maybe it's something we can worry about later though.

packages/vscode-extension/src/project/StateManager.ts
```diff
@@ -0,0 +1,100 @@
...
+export class DerivedStateManager<T extends object, K extends object> extends StateManager<T> {
+  constructor(
+    private parent: StateManager<K>,
+    private keyInParent: keyof K
+  ) {
+    super();
+    this.disposables.push(
+      parent.on("setState", (partialParentState: Partial<K>) => {
+        const partialState = partialParentState[this.keyInParent] as T | undefined;
+        if (!partialState) {
```
jwajgelt:
```suggestion
        if (partialState !== undefined) {
```
`T` can be `number`/`string`/`boolean`/whatever, in which case it can be falsy while being set.

jwajgelt:
```suggestion
        if (partialState === this.previousStateThatYouAddHere) {
```
You can set the same thing, in which case we probably don't want to fire the event and continue the walk down the state tree.

packages/vscode-extension/src/panels/WorkspaceConfigController.ts
```diff
@@ -1,86 +1,73 @@
...
-  ) {
-    this.eventEmitter.addListener(eventType, listener);
-  }
+    this.stateManager.on("setState", async (partialState) => {
+      const partialStateEntries = Object.entries(partialState);
+
+      const config = workspace.getConfiguration("RadonIDE");
+
+      for (const partialStateEntry of partialStateEntries) {
+        if (config.inspect(partialStateEntry[0] as string)?.workspaceValue) {
```
jwajgelt, line 28:
```suggestion
        if (config.inspect(partialStateEntry[0] as string)?.workspaceValue === undefined) {
```
Do we want this branch on falsy values as well?

packages/vscode-extension/src/panels/WorkspaceConfigController.ts
```diff
@@ -1,86 +1,73 @@
...
-      const config = workspace.getConfiguration("RadonIDE");
+    this.stateManager.setState(workspaceConfig);
 
-      const newConfig = {
-        panelLocation: config.get<PanelLocation>("panelLocation")!,
-        showDeviceFrame: config.get<boolean>("showDeviceFrame")!,
-        stopPreviousDevices: config.get<boolean>("stopPreviousDevices")!,
-      };
+    this.disposables.push(
+      workspace.onDidChangeConfiguration((event: ConfigurationChangeEvent) => {
```
jwajgelt:
```suggestion
      workspace.onDidChangeConfiguration(this.onConfigurationChange)
```
while we're here, we could clean this up a bit.

packages/vscode-extension/src/panels/WebviewController.ts
```diff
@@ -62,26 +99,37 @@ export class WebviewController implements Disposable {
...
     commands.executeCommand("setContext", "RNIDE.panelIsOpen", false);
     disposeAll(this.disposables);
+    this.ide.off("stateChanged", this.onStateUpdated);
     this.ide.detach();
   }
 
   private setWebviewMessageListener(webview: Webview) {
     webview.onDidReceiveMessage(
-      (message: WebviewEvent) => {
+      async (message: WebviewEvent) => {
```
jwajgelt: Why `async`?

packages/vscode-extension/src/project/ide.ts
```diff
@@ -6,27 +6,48 @@ import { Utils } from "../utilities/utils";
...
+import { EventEmitter } from "stream";
+import { RootStateManager } from "./StateManager";
 import { OutputChannelRegistry } from "./OutputChannelRegistry";
 
 interface InitialOptions {
   initialLaunchConfig?: LaunchConfiguration;
 }
 
-export class IDE implements Disposable {
+export class IDE extends EventEmitter implements Disposable {
```
jwajgelt: I'd strongly prefer if you just had the emitter in a field.
Also, `vscode.EventEmitter` is love.

jwajgelt (changes requested):

packages/vscode-extension/src/project/StateManager.ts
filip131311: yes, we discussed it internally and this is something we can worry about when improving the exposer of "non-selected" states, right now we do not have that problem sice we expose only one state

packages/vscode-extension/src/project/StateManager.ts
filip131311: I will make it explicit but it is not an issue as T extends object

filip131311: I don't quite understand when use set I want the event to be fired regardless if I was the one causing the state to be set (set itself does not fire except in the root component). We could introduce a "quietSet" functionality if its necessery but right now i don't see the need

packages/vscode-extension/src/panels/WorkspaceConfigController.ts
filip131311: yes this is a filed that informs us if this is a workspace or user setting

packages/vscode-extension/src/panels/WebviewController.ts
```diff
@@ -4,18 +4,54 @@ import { getTelemetryReporter } from "../utilities/telemetry";
 import { IDE } from "../project/ide";
 import { disposeAll } from "../utilities/disposables";
 import { RENDER_OUTLINES_PLUGIN_ID } from "../common/RenderOutlines";
-import { PanelLocation } from "../common/WorkspaceConfig";
+import { PanelLocation, RecursivePartial, State } from "../common/State";
 
-type CallArgs = {
+interface EventBase {
+  [key: string]: unknown;
+}
```
jwajgelt: Isn't this just

```suggestion
type EventBase  = Record<string, unknown>
```
?

packages/vscode-extension/src/panels/WebviewController.ts
```diff
@@ -4,18 +4,54 @@ import { getTelemetryReporter } from "../utilities/telemetry";
 import { IDE } from "../project/ide";
 import { disposeAll } from "../utilities/disposables";
 import { RENDER_OUTLINES_PLUGIN_ID } from "../common/RenderOutlines";
-import { PanelLocation } from "../common/WorkspaceConfig";
+import { PanelLocation, RecursivePartial, State } from "../common/State";
 
-type CallArgs = {
+interface EventBase {
```
jwajgelt: What's this interface needed for? Why must the various ...Command interfaces extend it?

packages/vscode-extension/src/project/StateManager.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export abstract class StateManager<T extends object> implements Disposable {
+  static create<T extends object>(initialState: T): StateManager<T> {
+    return new RootStateManager(initialState);
+  }
+
+  abstract setState(partialState: RecursivePartial<T>): void;
+  abstract getState(): T;
+
+  protected onSetStateEmitter: EventEmitter<RecursivePartial<T>> = new EventEmitter<
```
jwajgelt:
```suggestion
  protected onSetStateEmitter = new EventEmitter<
```
you can let typescript infer the type when you initialize it directly.

packages/vscode-extension/src/project/ide.ts
```diff
@@ -35,7 +55,14 @@ export class IDE implements Disposable {
       this.outputChannelRegistry,
       initialLaunchConfig
     );
-    this.workspaceConfigController = new WorkspaceConfigController();
+
+    this.project.appRootConfigController.getAvailableApplicationRoots().then((applicationRoots) => {
+      this.stateManager.setState({ applicationRoots });
+    });
```
jwajgelt, line 61: NIT: `IDE` filling random fields of of the state in its constructor feels awkward to me, but it's fine.
It does feel weird that it calls through `this.project.appRootConfigController` though. Maybe we could lift the `appRootConfigController` up to `IDE` instead?

packages/vscode-extension/src/utilities/mergeAndCalculateChanges.ts
```diff
@@ -0,0 +1,48 @@
...
+      continue;
+    }
+    if (oldNode[key] === newNode[key]) {
+      result[key] = oldNode[key];
+    } else {
+      wasChanged = true;
+      result[key] = newNode[key];
+      changes[key] = newNode[key];
+    }
+  }
```
jwajgelt, line 43: You could probably use `_.map` to make this cleaner.
https://lodash.com/docs/4.17.15#map

jwajgelt (approved):

packages/vscode-extension/src/panels/WebviewController.ts
filip131311: so ts lets me do this        
```
 // ignore dispatchTouches and log calls from being logged as "Message from webview"
        if (!message.method || (message.method !== "dispatchTouches" && message.method !== "log")) {
 ```

packages/vscode-extension/src/project/ide.ts
filip131311: This is an artefact of my previous attempt of unifying launch configuration management for this initial stage of the move to legend state, as workspaceConfig was a better example it was left here only half finished. I will manage it in a nicer way once I will move all of it into the new aproach

packages/vscode-extension/src/utilities/mergeAndCalculateChanges.ts
filip131311: eee it outputs an array would be cleaner if it was able to generate the "result' object for me

PR #1365, opened 2025-07-23
fix: saving initial launch configuration when it's detected

filip131311:
This PR fixes an issue with detected launch configuration never being used as initial launch configuration. 
This was caused by the initial config detector always checking if the saved configuration is still present in the custom configurations list even if the configuration wan never part of that list. 

To fix that we first check if the configurations kind is `Detected` and if it is and the selected appRoot still exists we select it.

### How Has This Been Tested: 

- run `radon-ide-test-apps` directory 
- select one of the detected configs 
- turn of vscode 
- turn it back on and check if the above config is the new initial config

Internal fix

### DIFF
--- a/packages/vscode-extension/src/panels/AppRootConfigController.ts
+++ b/packages/vscode-extension/src/panels/AppRootConfigController.ts
@@ -1,72 +1,15 @@
 import path from "path";
 import { workspace } from "vscode";
-import { ApplicationRoot, AppRootConfig } from "../common/AppRootConfig";
-import { extensionContext, findAppRootCandidates } from "../utilities/extensionContext";
+import { AppRootConfig } from "../common/AppRootConfig";
 import { findXcodeProject, findXcodeScheme } from "../utilities/xcode";
 import { Logger } from "../Logger";
 import { getIosSourceDir } from "../builders/buildIOS";
 import { readEasConfig } from "../utilities/eas";
 import { EasBuildConfig } from "../common/EasConfig";
-import { requireNoCache } from "../utilities/requireNoCache";
-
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/panels/AppRootConfigController.ts
```diff
@@ -51,23 +51,27 @@ function readApplicationRoot(appRootPath: string): ApplicationRoot {
...
-      return [];
-    }
+  if (!applicationRoots) {
+    Logger.debug(`Could not find any application roots.`);
+    return [];
+  }
 
-    return applicationRoots.map(readApplicationRoot);
+  return applicationRoots.map(readApplicationRoot);
+}
```
jwajgelt: This should be moved to a file under `utilities/`

packages/vscode-extension/src/project/launchConfigurationsManager.ts
```diff
@@ -131,6 +132,18 @@ export class LaunchConfigurationsManager implements Disposable {
...
+          (availableAppRoot) => availableAppRoot.path === savedLaunchConfig.appRoot
+        )
+      ) {
+        return savedLaunchConfig;
+      }
+    }
+
     if (
       savedLaunchConfig &&
       this._launchConfigurations.find((config) => _.isEqual(config, savedLaunchConfig))
```
jwajgelt: We could probably check the kind here as well.


```suggestion
if (
savedLaunchConfig?.kind === LaunchConfigurationKind.Custom &&
this._launchConfigurations.find((config) => _.isEqual(config, savedLaunchConfig))
```

jwajgelt (approved):
🆒 🔥

PR #1372, opened 2025-07-25
feat: automatically reconnect debugger when application is closed and reopened

### LINKED ISSUE #1368: Automatically reconnect debugger
Currently, we only connect the debugger when we launch the application or when certain reload actions happen.
This can cause issues when:
- the debugger connection is lost for whatever reason
- the application is stopped on the device and restarted from the device's launcher (not from Radon IDE)

jwajgelt:
Closes #1368

### How Has This Been Tested: 
- open an app in Radon
- close the application through the device's App Switcher
- open the application from the device's system launcher
- check that the debugger connects automatically to the application

#### Check for potential regression -- Expo Go sometimes misbehaves on JS reload #985 
- open an app which uses Expo Go
- reload JS
- verify that the debugger still works correctly (e.g. by triggering a breakpoint)

### DIFF
--- a/packages/vscode-extension/src/connect/ConnectSession.ts
+++ b/packages/vscode-extension/src/connect/ConnectSession.ts
@@ -22,7 +22,9 @@ class DebugSessionInspectorBridge extends BaseInspectorBridge {
   }
 
   protected send(message: unknown) {
-    this.debugSession.dispatchRadonAgentMessage(message);
+    this.debugSession.evaluateExpression({
+      expression: `globalThis.__radon_dispatch(${JSON.stringify(message)});`,
+    });
   }
 }
 
--- a/packages/vscode-extension/src/connect/ConnectSession.ts
+++ b/packages/vscode-extension/src/connect/ConnectSession.ts
@@ -1,7 +1,8 @@
-import { DebugSession } from "../debugging/DebugSession";
-import { Metro } from "../project/metro";
 import { Disposable } from "vscode";
+import { DebugSession, DebugSessionImpl } from "../debugging/DebugSession";
+import { Metro } from "../project/metro";
 import { BaseInspectorBridge } from "../project/bridge";
+import { disposeAll } from "../utilities/disposables";
 
 export interface ConnectSessionDelegate {
   onSessionTerminated: () => void;
@@ -20,13 +21,14 @@ class DebugSessionInspectorBridge extends BaseInspectorBridge {
     }
   }
 
-  protected send(message: any) {
+  protected send(message: unknown) {
     this.debugSession.dispatchRadonAgentMessage(message);
   }
 }
 
 export default class ConnectSession implements Disposable {
-  private debugSession: DebugSession;
+  private debugSession: DebugSession & Disposable;
+  private debugEventsSubscription: Disposable;
   public readonly inspectorBridge: DebugSessionInspectorBridge;
 
   public get port() {
@@ -37,20 +39,11 @@ export default class ConnectSession implements Disposable {
     private readonly metro: Metro,
     private readonly delegate: ConnectSessionDelegate
   ) {
-    this.debugSession = new DebugSession(
-      {
-        onDebugSessionTerminated: () => {
-          this.delegate.onSessionTerminated();
-        },
-        onBindingCalled: (event: any) => {
-          this.inspectorBridge.onBindingCalled(event);
-        },
-      },
-      {
-        suppressDebugToolbar: false,
-        displayName: "Radon Connect Debugger",
-      }
-    );
+    this.debugSession = new DebugSessionImpl({
+      suppressDebugToolbar: false,
+      displayName: "Radon Connect Debugger",
+    });
+    this.debugEventsSubscription = this.registerDebugSessionListeners();
     this.inspectorBridge = new DebugSessionInspectorBridge(this.debugSession);
   }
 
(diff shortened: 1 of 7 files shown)


packages/vscode-extension/src/connect/ConnectSession.ts
```diff
@@ -20,13 +21,16 @@ class DebugSessionInspectorBridge extends BaseInspectorBridge {
     }
   }
 
-  protected send(message: any) {
-    this.debugSession.dispatchRadonAgentMessage(message);
+  protected send(message: unknown) {
```
filip131311, line 24: I am not sure I like this change previously dispatchRadonAgentMessage method allowed for communication with the application runtime, now we need to know the implementation details of the communication mechanisms to do it.

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -299,6 +299,19 @@ export class DeviceSession
     }
   };
 
+  private registerDebugSessionListeners(): Disposable {
```
filip131311, line 302: I am not quite sure what is the advantage of this aproach in this scenario? 
Do you register listeners form more places? 

This is quite a bit more "boilerplaty"  and is a differnt approach then widlly used in the code base where we mostlly just use delegates

packages/vscode-extension/src/debugging/ReconnectingDebugSession.ts
```diff
@@ -0,0 +1,118 @@
...
+      throw new Error("Ping timeout");
+    });
+    return Promise.race([resultPromise, timeout]).catch((_e) => false);
+  }
+
+  private maybeReconnect = async () => {
+    if (!this.isRunning || this.cancelReconnect !== undefined) {
+      return;
+    }
+    this.cancelReconnect = new CancelToken();
```
filip131311: nit: I would prefer reconnectCancelToken ore something like that, cancelReconnect looks like a method that cancels reconnect

packages/vscode-extension/src/debugging/ReconnectingDebugSession.ts
```diff
@@ -0,0 +1,118 @@
...
+  public async startJSDebugSession(configuration: JSDebugConfiguration) {
+    this.isRunning = true;
+    return this.debugSession.startJSDebugSession(configuration);
+  }
+
+  private async pingJsDebugSessionWithTimeout() {
+    const resultPromise = this.debugSession
+      .evaluateExpression({ expression: "('ping')" })
+      .then((response) => {
+        return response.result.value === "ping";
```
filip131311, line 35:
```suggestion
        return true;
```

I think it would be just a bit cleaner (can you ever get a repsonse and not return true?

packages/vscode-extension/src/debugging/ReconnectingDebugSession.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+    this.cancelReconnect = new CancelToken();
+    while (this.isRunning && !this.cancelReconnect.cancelled) {
+      try {
+        const connected = await this.pingJsDebugSessionWithTimeout();
+        if (connected) {
+          // if we're connected to a responsive session, we can break
+          break;
+        }
+        const websocketAddress = await this.metro.getDebuggerURL(undefined, this.cancelReconnect);
```
filip131311: we should look for websockets first as if there is no present there is no point trying to ping anything

filip131311: (we can save on the timeout)

packages/vscode-extension/src/debugging/ReconnectingDebugSession.ts
```diff
@@ -0,0 +1,118 @@
...
+    return Promise.race([resultPromise, timeout]).catch((_e) => false);
+  }
+
+  private maybeReconnect = async () => {
+    if (!this.isRunning || this.cancelReconnect !== undefined) {
+      return;
+    }
+    this.cancelReconnect = new CancelToken();
+    while (this.isRunning && !this.cancelReconnect.cancelled) {
+      try {
```
filip131311, line 49: we should have a separet sleep independent of the ping method if we introduce my sugesstion below: 
"we should look for websockets first as if there is no present there is no point trying to ping anything"

filip131311 (commented):
Left inline qestions

packages/vscode-extension/src/connect/ConnectSession.ts
jwajgelt: That's the point.
If one of `DebugAdapter`, `DebugSession` and `ConnectSession` needs to know about `RadonAgent`, it should be the `ConnectSession` (which implements the `InspectorBridge`)

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: > Do you register listeners form more places?

Yes, the `ReconnectingDebugSession` decorator needs to register to at least the `terminate` event.
We could have the `Reconnecting` accept a `Delegate`, and create its own which wraps the `terminate` method, but I feel like that is more more confusing than this approach, especially when you try to actually figure out what's called on each event.

We're kind of moving away from the `Delegate` as a way of passing messages in other parts as well.
This will also be useful in the future, if we decide to move away from the React DevTools protocol to Debugger-based communication.

packages/vscode-extension/src/debugging/ReconnectingDebugSession.ts
jwajgelt: 1. this is the way it worked before
2. I don't know, but I'd rather not risk it backfiring.

packages/vscode-extension/src/debugging/ReconnectingDebugSession.ts
jwajgelt: nah

PR #1373, opened 2025-07-28
feat: add "Restart Metro" reload action

### LINKED ISSUE #1369: Option to restart Metro
**Description**
Since Radon handles the metro bundler automatically - i need an easy option to restart it.

I use several environmental variables for development, when i make any changes to them - i need to restart metro for them to apply. With the current options in the reload app menu, none of them seem to do this, either just the app/emulator is restarted, or a full rebuild is triggered.

**Environment**

- Radon IDE v1.9.1
- RN 0.76, Expo 52.

jwajgelt:
- Adds a "Restart Metro server" reload action to the restart dropdown
- No longer restarts the whole device for metro restart actions

Closes #1369 

### How Has This Been Tested: 
- open `expo-52-prebuilt` test app
- check that both "Restart Metro" and "Clear Metro cache" options work
- check that Redux / React Query devtools work after restarting metro this way

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -59,7 +59,6 @@ const CACHE_STALE_THROTTLE_MS = 10 * 1000; // 10 seconds
 
 type RestartOptions = {
   forceClean: boolean;
-  cleanCache: boolean;
 };
 
 export type DeviceSessionDelegate = {
@@ -424,7 +423,7 @@ export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
     }
   }
 
-  public async performReloadAction(type: ReloadAction): Promise<boolean> {
+  public async performReloadAction(type: ReloadAction): Promise<void> {
     try {
       this.resetStartingState();
 
@@ -432,18 +431,13 @@ export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
         platform: this.platform,
         method: type,
       });
-      const result = await this.performReloadActionInternal(type);
-      if (result) {
-        this.status = "running";
-        this.emitStateChange();
-      } else {
-        window.showErrorMessage("Failed to reload, you may try another reload option.", "Dismiss");
-      }
-      return result;
+      await this.performReloadActionInternal(type);
+      this.status = "running";
+      this.emitStateChange();
     } catch (e) {
       if (e instanceof CancelError) {
         // reload got cancelled, we don't show any errors
-        return false;
+        return;
       } else if (e instanceof BuildError) {
         this.status = "fatalError";
         this.fatalError = {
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -453,51 +447,99 @@ export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
...
+  private async restartMetro({ resetCache }: { resetCache: boolean }) {
+    if (this.cancelToken) {
+      this.cancelToken.cancel();
+    }
+
+    const cancelToken = new CancelToken();
+    this.cancelToken = cancelToken;
+
+    this.status = "starting";
+    this.fatalError = undefined;
```
filip131311, line 500: this is interesting change, what is `fatalError` is  unrelated to metro and for example is caused by build system, before this refactor it would be picked up agin after restart because we restarted everything, but if we only restart metro won't we loos the issue that is breaking the application?

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -453,51 +447,99 @@ export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
...
+    this.updateStartupMessage(StartupMessage.StartingPackager);
```
filip131311, line 501: if we skip the build and the booting stage shouldn't we go straight to the installing phase?

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -453,51 +447,99 @@ export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
...
+
+    Logger.debug(`Launching metro`);
+    await this.metro.start({
+      resetCache,
+      launchConfiguration: this.applicationContext.launchConfig,
+      dependencies: [],
+    });
+
+    await cancelToken.adapt(this.restartDebugger());
+    if (!this.buildResult) {
```
filip131311: how can this ever happen?

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -687,6 +697,9 @@ export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
       Promise.all([
         this.metro.ready(),
         this.device.startPreview().then((url) => {
+          if (!url) {
```
filip131311, line 700: can it ever resolve empty?

jwajgelt: If it does, we should throw anyway.
We checked against undefined previously, I just changed it to throw in that (and empty) case instead of returning `false` from the function.

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: It can't, missed that we have a getter which throws when the `buildResult` is not available.
As for if it's possible for the build result to not be available at this point, you could trigger this reload action while the app is first building, cancelling the build.

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: We're literally starting the packager at this step, I don't know what message is more appropriate.
After metro starts, you will get the appropriate "building" or "installing" phase set by `this.buildApp` or `this.installApp`

jwajgelt: Wait, "Packager" is meant to be the package manager, not the bundler here?

jwajgelt: No, it's fine, this is the same thing that we set when calling `waitForMetroReady`.

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: It should be fine. Depending on what went wrong previously, I would expect one of the steps here to fail in that case.

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: We should probably clear the build result when starting building, to not use the old build when a restart is cancelled.

packages/vscode-extension/src/project/deviceSession.ts
filip131311: > Wait, "Packager" is meant to be the package manager, not the bundler here?

yes this is the first step of the normal pipeline npm checks

jwajgelt: > yes this is the first step of the normal pipeline npm checks

No, it's not...
> No, it's fine, this is the same thing that we set when calling `waitForMetroReady`.

And it's only set there.
The only state set before checking node modules is `StartupMessage.InitializingDevice`

PR #1375, opened 2025-07-28
refactor: move dependency management to the new state system 

filip131311:
This PR moves the old DependencyManager and DependenciesProvider functionality to the new state solution and introduces the following changes as the result:

1) we devide `DependencyManager` into `EnvironmentDependencyManager` and `ApplicationDependencyManager` to isolate dependencies that require application specific information in the application context
2) we move the functionality of the package manager into a separate callas to further improve the overblown `ApplicationDependencyManager` part of `DependencyManager` while doing so we introduce similar patterns as `pods` class with dedicated cancelToken and output channel.
3) we introduce `useDependencyErrors` to provide a functionality previously defined in DependencyProvider 
4) we move the `runAllChecks` method into project which shapes to be our stable API for the backend part of the extension and remove `DependenciesProvider`

5) I fix my mistake in `mergeAndCalculateChanges` from the previous refactor when I assumed that `oldNode` always contains the necessary key.

### How Has This Been Tested: 

- remove node_modules and see them installed
- brak something like for example remove the android dir and check if it shows up in the doctor 

This is internal change


packages/vscode-extension/src/common/OutputChannel.ts
```diff
@@ -4,4 +4,5 @@ export enum Output {
   BuildAndroid = "Radon IDE (Android build)",
   AndroidDevice = "Radon IDE (Android Emulator Logs)",
   IosDevice = "Radon IDE (iOS Simulator Logs)",
+  PackageManager = "Radon IDE (Package Manager)",
```
jwajgelt:
```suggestion
  PackageManager = "Radon IDE (JS Package Manager)",
```
Maybe? It's fine as is, just not sure if this could cause some confusion.

packages/vscode-extension/src/common/State.ts
```diff
@@ -12,15 +12,58 @@ export type WorkspaceConfiguration = {
...
+  | "ios"
+  | "android"
+  | "pods"
+  | "reactNative"
+  | "expo"
+  | "expoRouter"
+  | "storybook"
+  | "easCli";
+
+export type InstallationStatus = "installed" | "notInstalled" | "installing";
```
jwajgelt, line 33: NIT: string `enum`s are generally preferred over union of string literal types.

packages/vscode-extension/src/common/State.ts
```diff
@@ -12,15 +12,58 @@ export type WorkspaceConfiguration = {
...
+
+export type DependencyStatus = {
+  status: InstallationStatus;
+  isOptional: boolean;
+  details?: string;
+};
+
+export type EnvironmentDependencyRecord = Partial<Record<EnvironmentDependency, DependencyStatus>>;
+export type ApplicationDependencyRecord = Partial<Record<ApplicationDependency, DependencyStatus>>;
```
jwajgelt: NIT: I'd avoid including `Record` in the type name, it doesn't really mean anything.
Something like
```suggestion
export type EnvironmentDependencyStatuses = Partial<Record<EnvironmentDependency, DependencyStatus>>;
export type ApplicationDependencyStatuses = Partial<Record<ApplicationDependency, DependencyStatus>>;
```
would be preferable.

packages/vscode-extension/src/common/State.ts
```diff
@@ -12,15 +12,58 @@ export type WorkspaceConfiguration = {
...
+
+export type ProjectStore = {
+  applicationContext: ApplicationContextState;
+};
+
+export type ApplicationContextState = {
+  applicationDependencies: ApplicationDependencyRecord;
+};
```
jwajgelt:
```suggestion
export type ProjectStore = {
  applicationDependencies: ApplicationDependencyRecord;
};
```
I don't think we need to mirror the way components compose in the extension code in the shape of the shared state.

packages/vscode-extension/src/dependency/packageManager.ts
```diff
@@ -1,16 +1,106 @@
...
+        nodeModulesPromise: Promise<void>;
+        cancelToken: CancelToken;
+      }
+    | undefined;
+  constructor(private readonly launchConfiguration: ResolvedLaunchConfig) {}
+
+  public async isPackageManagerInstalled() {
+    // the resolvePackageManager function in getPackageManager checks
+    // if a package manager is installed and otherwise returns undefined
+    return Boolean(await resolvePackageManager(this.launchConfiguration));
```
jwajgelt:
```suggestion
    return (await resolvePackageManager(this.launchConfiguration)) !== undefined;
```

packages/vscode-extension/src/project/project.ts
```diff
@@ -183,6 +193,11 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
     });
   }
 
+  public async runDependencyChecks(): Promise<void> {
+    await this.applicationContext.applicationDependencyManager.runAllDependencyChecks();
+    await this.environmentDependencyManager.runAllDependencyChecks();
```
jwajgelt: Shouldn't these be run in parallel?

```suggestion
    await Promise.all([
      this.applicationContext.applicationDependencyManager.runAllDependencyChecks(),
      this.environmentDependencyManager.runAllDependencyChecks()
    ])
```

packages/vscode-extension/src/utilities/testCommand.ts
```diff
@@ -0,0 +1,15 @@
+import { command } from "./subprocess";
+
+export async function testCommand(cmd: string) {
```
jwajgelt: A doc comment and an explicit return type would be nice.
Previously this was used internally as part of the implementation, so you were assumed to know what it does while working on the code surrounding it, but it's useful to not have to look up the implementation of a util we use throughout the code base every time you encounter it.

packages/vscode-extension/src/webview/hooks/useDependencyErrors.ts
```diff
@@ -1,102 +1,33 @@
...
-      runDiagnostics,
-      errors: getErrors(depsState),
-    };
-  }, [depsState, runDiagnostics, getErrors]);
-
-  return (
-    <DependenciesContext.Provider value={contextValue}>{children}</DependenciesContext.Provider>
+type ErrorType = "ios" | "simulator" | "emulator" | "android" | "common";
+type Dependency = ApplicationDependency | EnvironmentDependency;
+type DependencyRecord = ApplicationDependencyRecord | EnvironmentDependencyRecord;
```
jwajgelt: Same comment about the type name as for `ApplicationDependencyRecord`, `EnvironmentDependencyRecord`

jwajgelt (approved):

packages/vscode-extension/src/common/State.ts
filip131311: this approach allows me to not have multiple stateManagers in one class, which is very convenient now that I move a lot of functionality around, if it will be for some reason a better approach to flatten the state object at the end of this refactor I will do it then

packages/vscode-extension/src/common/State.ts
filip131311: It was like that before introducing enums requires more sophisticated changes on the frontend error handling side and I'll just leave it for now

PR #1376, opened 2025-07-29
refactor: telemetry state

filip131311:
This PR moves the telemetry to the new state management solution and removes old unnecessary providers  

### How Has This Been Tested: 

- open settings dropdown and test the send feedback button 
- open vscode settings and toggle `Telemetry: Telemetry Level` setting 
- check out the send feedback button again

this is an internal change


packages/vscode-extension/src/project/telemetryManager.ts
```diff
@@ -0,0 +1,19 @@
+import { Disposable, env } from "vscode";
+import { StateManager } from "./StateManager";
+import { TelemetryState } from "../common/State";
+
+export class TelemetryManager implements Disposable {
```
jwajgelt: I'd prefer a less generic name for this, you can't really tell from just the name that `TelemetryManager` only really watches the `vscode.env` and updates the `telemetry.enabled` boolean.

packages/vscode-extension/src/project/telemetryManager.ts
```diff
@@ -0,0 +1,19 @@
...
+  constructor(private readonly stateManager: StateManager<TelemetryState>) {
+    const isTelemetryEnabled = env.isTelemetryEnabled;
+
+    this.stateManager.setState({ enabled: isTelemetryEnabled });
+
+    env.onDidChangeTelemetryEnabled((telemetryEnabled) => {
+      this.stateManager.setState({ enabled: telemetryEnabled });
+    });
```
jwajgelt: Doesn't this listener have to be somehow cleaned up?

packages/vscode-extension/src/utilities/utils.ts
```diff
@@ -29,11 +29,7 @@ type KeybindingType = {
 export class Utils implements UtilsInterface {
   private eventEmitter = new EventEmitter();
 
-  constructor() {
-    vscode.env.onDidChangeTelemetryEnabled((telemetryEnabled) => {
-      this.eventEmitter.emit("telemetryEnabledChanged", telemetryEnabled);
-    });
-  }
+  constructor() {}
```
jwajgelt:
```suggestion
```
You can safely remove an empty constructor.

jwajgelt (changes requested):

packages/vscode-extension/src/project/telemetryManager.ts
filip131311: nice catch

packages/vscode-extension/src/project/telemetryManager.ts
filip131311: I like the name as it matches the rest of the managers in the ide scope, 
more over I beleave that it will perform more relavant role in near future when will try to decouple telemetry behavior from vscode

PR #1378, opened 2025-07-29
fix: attaching debugger on relaunch

filip131311:
after #1373 the application would get stuck in "attaching debugger" state after using "restart app process" or "reinstall" options. It was happening because the new "reconnecting" functionality would cancel the launch proces attempt to connect to the debugger and that would cause the launch proces to never end. 

This fixes it in a crude way by just checking if  the attempt to connect to the debugger was in fact cancelled and not thrown more critical exception. 

The real fix would be to rework the debugger session to only start new session and cancel the previous try if the configuration has changed and even the probably instead of cancelling It should just become the new awaited session for the old start request.  

### How Has This Been Tested: 

use the reinstall functionality 

Internal change


filip131311:
closed in favor of #1380

PR #1382, opened 2025-07-30
remove excessive logging when loading dotenv files

jwajgelt:
Removes the excessive "loaded environment variables from .env files" logs which currently happen each time anything accesses the project's environment.

### How Has This Been Tested: 
- open a project in Radon
- don't get flooded by "loaded environment" logs

### DIFF
--- a/packages/vscode-extension/src/project/ApplicationContext.ts
+++ b/packages/vscode-extension/src/project/ApplicationContext.ts
@@ -10,6 +10,7 @@ import { LaunchConfiguration, LaunchOptions } from "../common/LaunchConfig";
 import { StateManager } from "./StateManager";
 import { ApplicationContextState } from "../common/State";
 import { ApplicationDependencyManager } from "../dependency/ApplicationDependencyManager";
+import { Logger } from "../Logger";
 
 /**
  * Represents a launch configuration that has been resolved with additional properties.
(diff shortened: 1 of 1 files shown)


filip131311 (changes requested):
Can you just move the log to the debug type and leave it there

packages/vscode-extension/src/project/ApplicationContext.ts
```diff
@@ -23,37 +23,37 @@ export type ResolvedLaunchConfig = LaunchOptions & {
...
 
+  let lastLoadedEnvFiles: string[] | undefined;
+
+  function resolveEnvironment(): Record<string, string> {
+    const systemEnv = process.env as NodeJS.ProcessEnv;
+    const mergedEnv = { ...systemEnv, ...configuredEnv };
+    // load the dotenv files for the project into `mergedEnv`
+    const loadEnvResult = loadProjectEnv(absoluteAppRoot, { force: true, systemEnv: mergedEnv });
+
+    if (loadEnvResult.result === "loaded" && !_.isEqual(loadEnvResult.files, lastLoadedEnvFiles)) {
```
filip131311, line 40: shouldn't we log when the env is updeted has changed and not when the file is different? Is it achievable?

jwajgelt: We only log the files, I don't see how logging that we've loaded the same files again is helpful.
The only reason to look at the log is to check if some file we expect to be loaded does not get loaded, we don't care about the specific values.

filip131311: yes and we might want to expect the log after we changed the file

PR #1383, opened 2025-07-30
fix: workspace state fast updates

filip131311:
This PR fixes how we perform the sync with workspace settings.  Before a consecutive fast updates would cause an infinite 
loop of synchronization between `workspaceConfiguration` and our state manager. 

This change checks if the update that comes back is the same as the one just set before updating it. 

Additionally after this change updates that take no effect are not send at all.

### How Has This Been Tested: 

- run radon and use `control` +`option` + 0/9 keybinding (rotate the device) in very short succession 
- before this update it would trigger infinite spiral of updates 

internal change


pFornagiel (approved):
looks good, approved

PR #1384, opened 2025-07-30
Add proper screenshot and video rotation upon saving

pFornagiel:
## Description

This PR introduces rotated screenshots and video saving (media now having proper dimensions), according to the current orientation of device in the IDE.

An example of recording the app and the resulting video after the changes:

https://github.com/user-attachments/assets/e0558842-b6bf-4b98-a517-5a789a876643

https://github.com/user-attachments/assets/6208cec0-b2db-4953-9ec9-f13d99a18ef0

The PR also fixes unrecorded issue - When the device's rotation is set to PortraitUpsideDown, the Replay preview in ReplayUI component is incorrectly rotated (appears upside down). 

### How Has This Been Tested: 

- run Radon IDE extension panel, and launch app in selected device, start recording / take screenshot when screen is in orientation different from portrait.

Expected behaviour - does not need documentation in my opinion.


pFornagiel:
> Something went wrong with the replay option :(
> 
>  Screen.Recording.2025-07-30.at.20.07.52.mov

On my end the replay option works fine, could you provide steps to reproduce? Was the simulator-server updated to the newest version on main branch?

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
```diff
@@ -15,7 +15,7 @@ export async function screenshotToolExec(): Promise<ToolResponse> {
     );
   }
 
-  const screenshot = await project.deviceSession.captureScreenshot();
+  const screenshot = await project.deviceSession.captureScreenshot(project.getDeviceRotation());
```
filip131311, line 18: why not just `project.captureScreenshot`?

packages/vscode-extension/src/devices/preview.ts
```diff
@@ -58,7 +62,8 @@ export class Preview implements Disposable {
 
     const newPromiseHandler = { resolve: resolvePromise!, reject: rejectPromise! };
     this.multimediaPromises.set(id, newPromiseHandler);
-    stdin.write(`${type} ${id} ${type === "video" ? "save" : ""}\n`);
+    stdin.write(`${type} ${id} ${type === "video" ? "save " : ""}-r ${rotation}\n`);
+    // stdin.write(`${type} ${id} ${type === "video" ? "save" : ""}\n`);
```
filip131311: seems like a left over

packages/vscode-extension/src/webview/Preview/MediaCanvas.tsx
```diff
@@ -1,59 +1,163 @@
...
+      dimensionsSwapped: false,
+      shouldDrawWithoutTransform: true, // Native orientation
+    },
+    [DeviceRotation.PortraitUpsideDown]: {
+      angle: -Math.PI / 2, // Rotate 90° counter-clockwise
+      dimensionsSwapped: true,
+      shouldDrawWithoutTransform: false,
+    },
+  },
+} as const;
```
filip131311, line 136: this could probably be replaced with a function that calculates the angle and swap based on the media angle, but as you already wrote all the cases Let's leave it

filip131311 (approved):
My skill issue, I left some inline comments but its nice

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
pFornagiel: `screenshotToolExec` is used to give Copilot/Cursor in agent mode ability to take and view screenshot internally using MCP.

`deviceSession.captureScreenshot(rotation)` returns a promise (created in `saveMultimediaWithID` method of `Preview` class), which allows for saving the photo in temporary location without user interaction.

`project.captureScreenshot()` abstracts this functionality, does not return anything and additionally prompts user with SaveDialog to confirm or change save destination.

We do not wish for `screenshotToolExec` to initialise communication with user, only take screenshot internally for chat to use, which is why we use internal `DeviceSession` method by accessing `project.deviceSession.captureScreenshot` and passing it current orientation from `project`.

packages/vscode-extension/src/devices/preview.ts
pFornagiel: Removed in b7c0369c

PR #1386, opened 2025-07-31
feat: Add build log viewer tool to AI

latekvo:
### Changes:

- add a `view_application_logs` tool, which the AI Agent may use to access the build and bundling logs, as well as screenshot if it's available.
- add `ReadableLogOutputChannel`, which makes the regular `LogOutputChannel` readable
- expose `DeviceSession.previewReady` and `DeviceSession.deviceRotation`
- add dedicated output channel for `Metro`, which is hidden from the user, but available to the AI
- refactor `OutputChannelRegistry` to be a global singleton (fixes an issue already present on `main`)
- add hidden log output channel for the logs generated by the `Metro` bundler

### How Has This Been Tested: 

1. Request the AI agent to use the `view_debugging_data` tool, which displays the build logs, as well as bundling logs and a screenshot of the screen, if the build succeeded.
2. Verify that the tool output contains all recent build, bundler and app logs.

### Example usage in different scenarios:

Build Failing | Cocapods issue |Runtime error |Mostly fine app   
---|---|---|---
image|image|image|image

Updated documentation with: https://github.com/software-mansion/radon-ide/pull/1398


packages/vscode-extension/src/project/ReadableLogOutputChannel.ts
```diff
@@ -0,0 +1,32 @@
...
+    clear: () => {
+      logRegistry.length = 0;
+      outputChannel.clear();
+    },
+    append: (value: string) => {
+      logRegistry.push(value);
+      outputChannel.append(value);
+    },
+    appendLine: (value: string) => {
+      logRegistry.push(value);
```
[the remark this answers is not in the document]
latekvo: Due to how we're serving this data, this newline is already added to the output.

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
```diff
@@ -20,12 +22,65 @@ export async function screenshotToolExec(): Promise<ToolResponse> {
...
+  if (!bundlerLogs.isEmpty()) {
+    combinedLogs.push("\n\n=== BUNDLING PROCESS STARTED ===\n", ...bundlerLogs.readAll());
+  }
+
+  if (!deviceLogs.isEmpty()) {
+    combinedLogs.push("\n\n=== APPLICATION STARTED ===\n", ...deviceLogs.readAll());
+  }
+
+  // TODO: Are `bundlerLogs` and `deviceLogs` cleared before build?
+  //       ^ If not, then store timestamps of the latest build, bundle with logs of bundler if they occured after the build.
```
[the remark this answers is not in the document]
latekvo: still in the process of testing if this is actually needed

packages/vscode-extension/src/ai/mcp/utils.ts
```diff
@@ -12,13 +12,42 @@ export function getEditorType(): EditorType {
   return EditorType.VSCODE;
 }
 
+export function base64ToToolContent(data: string): ImageContent {
+  return {
+    type: "image",
+    data,
+    mimeType: "image/png",
+  };
+}
```
jwajgelt: NIT: isn't the method name too generic? You may want to send non-Image binary content as base64. Why not

```suggestion
export function pngToToolContent(base64Encoded: string): ImageContent {
  return {
    type: "image",
    data: base64Encoded,
    mimeType: "image/png",
  };
}
```

packages/vscode-extension/src/project/CircularBuffer.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  public clear() {
+    this.headIndex = 0;
+    this.buffer.length = 0;
+  }
+
+  public readAll(): T[] {
+    return [
+      ...this.buffer.slice(this.headIndex, this.size),
+      ...this.buffer.slice(0, this.headIndex),
```
jwajgelt: This requires that either `headIndex === 0` or `this.size === this.capacity`. It seems to be true, looking at the implementation, but it might not be obvious to someone making changes in this class. Would be nice to have an `assert` in this function for that.

packages/vscode-extension/src/project/CircularBuffer.ts
```diff
@@ -0,0 +1,31 @@
...
+  public write(value: T) {
+    this.buffer[this.headIndex] = value;
+    this.headIndex = (this.headIndex + 1) % this.capacity;
+    this.size = Math.min(this.size + 1, this.capacity);
+  }
+
+  public clear() {
+    this.headIndex = 0;
+    this.buffer.length = 0;
+  }
```
jwajgelt:
```suggestion
  public clear() {
    this.headIndex = 0;
    this.buffer.length = 0;
    this.size = 0;
  }
```

packages/vscode-extension/src/project/CircularBuffer.ts
```diff
@@ -0,0 +1,31 @@
+export class CircularBuffer<T> {
```
jwajgelt, line 1: This is a piece of pure logic, would be nice to have some unit tests for it...

packages/vscode-extension/src/project/CircularBuffer.ts
latekvo: These assumptions are not correct. This code works well with both `headIndex != 0` and `this.size != this.capacity`. Here's an example:


```ts

const buffer = new CircularBuffer<number>(10);

for (let i = 0; i < 7; i++) {
  buffer.write(i);
}

// at the time of reading, `headIndex = 7` and `this.size < this.capacity`
console.log(buffer.readAll());

// outputs: [0, 1, 2, 3, 4, 5, 6]
```

The part of code you highlighted makes sure every element at position `< this.headIndex` is moved at the front of the array before being returned, since these are the latest additions. 

- With `headIndex = 0`, the `slice(0, this.headIndex)` is simply ignored, and `slice(this.headIndex, this.size)` returns the entire array. 
- With `this.size = this.headIndex`, `slice(this.headIndex, this.size)` is ignored, and `slice(0, this.headIndex)` returns the entire array.
- With `this.size = this.capacity` and `this.headIndex != 0`, both `slice` operations return two smaller portions of the array, with the total amount of elements returned being exactly `this.size`.

I tested this code with input values `5 - 30` at `capacity = 10` to confirm i didn't miss anything.

packages/vscode-extension/src/ai/mcp/utils.ts
latekvo: Good idea, done in [7164380](https://github.com/software-mansion/radon-ide/pull/1386/commits/71643801c1fca49cd99352ee402a26b89a4182ed)

packages/vscode-extension/src/project/CircularBuffer.ts
latekvo: Fixed in [2a4b1e7](https://github.com/software-mansion/radon-ide/pull/1386/commits/2a4b1e7bcbe62fdddddb00a21b9968453047ca95).

packages/vscode-extension/src/project/CircularBuffer.ts
latekvo: added basic unit tests in [dca3fee](https://github.com/software-mansion/radon-ide/pull/1386/commits/dca3fee5b558bab8db41a9f4bc7236db10c43fba)

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
```diff
@@ -20,12 +22,62 @@ export async function screenshotToolExec(): Promise<ToolResponse> {
...
   };
 }
+
+export async function buildLogsToolExec(): Promise<ToolResponse> {
+  const ideInstance = IDE.getInstanceIfExists();
+
+  const errorMessage = "Could not view the build logs! Radon IDE extension has not been opened.";
+
+  if (!ideInstance) {
+    return textToToolResponse(errorMessage);
```
kmagiera: would be better to add more actionable error messages. This state here likely means that the user is not running the IDE panel. We should write: "Couldn't retrieve build logs, Radon IDE is not launched. Open Radon IDE first"

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
```diff
@@ -20,12 +22,62 @@ export async function screenshotToolExec(): Promise<ToolResponse> {
...
+  }
+
+  const registry = ideInstance.outputChannelRegistry;
+  const session = ideInstance.project.deviceSession;
+
+  if (!session) {
+    return textToToolResponse(errorMessage);
```
kmagiera: This error means, that the user runs the IDE but there is no "active session" which means that they should select a project configuration or a device. I'd write "Radon IDE hasn't run any build. You need to select a project and device in Radon IDE panel".

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
```diff
@@ -20,12 +22,62 @@ export async function screenshotToolExec(): Promise<ToolResponse> {
...
+  const packageManagerLogs = registry.getOrCreateOutputChannel(Output.PackageManager);
+
+  const deviceLogs = registry.getOrCreateOutputChannel(
+    isAndroid ? Output.AndroidDevice : Output.IosDevice
+  );
+
+  const combinedLogs = [];
+
+  if (!buildLogs.isEmpty()) {
+    combinedLogs.push("=== BUILD PROCESS STARTED ===\n", ...buildLogs.readAll());
```
kmagiera: it seems like the MCP Api accepts multi part and fragmented responses. Wouldn't it be better if we send the individual logs as separate entries in the response. We can also provide details about the log output like for example: "BUILD LOGS (50 first log entries and 200 latest entries, full logs are in ${filepath} in lines {startLine}-{endLine}:"

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
```diff
@@ -20,12 +22,62 @@ export async function screenshotToolExec(): Promise<ToolResponse> {
...
-        type: "image",
-        data: contents,
-        mimeType: "image/png",
-      },
-    ],
+    content: [pngToToolContent(contents)],
   };
 }
+
+export async function buildLogsToolExec(): Promise<ToolResponse> {
```
kmagiera: this is not just reading build logs and deserves a different name

packages/vscode-extension/src/project/ReadableLogOutputChannel.ts
```diff
@@ -0,0 +1,66 @@
...
+export interface ReadableLogOutputChannel extends LogOutputChannel {
+  readAll: () => string[];
+  isEmpty: () => boolean;
+}
+
+export function createReadableOutputChannel(channel: string): ReadableLogOutputChannel {
+  const outputChannel = window.createOutputChannel(channel, { log: true });
+
+  const logTail: string[] = [];
+  const logHeadBuffer = new CircularBuffer<string>(KEEP_LAST_N);
```
kmagiera: this is reversed. Typically the tail of the logs are the latest ones and head are the oldest logs (this matches the convention that the unix head and tail commands are using)

packages/vscode-extension/src/Logger.ts
```diff
@@ -1,6 +1,10 @@
-import { LogOutputChannel, window } from "vscode";
+import { Output } from "./common/OutputChannel";
+import {
+  createReadableOutputChannel,
+  ReadableLogOutputChannel,
+} from "./project/ReadableLogOutputChannel";
 
-const outputChannel = window.createOutputChannel("Radon IDE", { log: true });
+const outputChannel = createReadableOutputChannel(Output.Ide);
```
kmagiera: shouldn't we always access the channel via OutpuChannelRegistry? It seems like its purpose is to not recreate the channel, but here we create a new one and whoever is accessing it via the registry will get Logger.rawOutputChannel

packages/vscode-extension/src/project/ReadableLogOutputChannel.ts
```diff
@@ -0,0 +1,66 @@
...
+    ...outputChannel,
+    readAll,
+    isEmpty: () => {
+      return !logTail.length;
+    },
+    clear: () => {
+      droppedLogsCounter = 0;
+      logTail.length = 0;
+      logHeadBuffer.clear();
+      outputChannel.clear();
```
kmagiera: I wonder if we could re-create the output channel here and whether that'd help with resetting the logs. Currently the log output channel doesn't support clear method and the logs stay there across re-builds. Maybe if we dismiss the channel here and create a new one this could fix the error?

packages/vscode-extension/src/Logger.ts
latekvo: Fixed this in [2f02829](https://github.com/software-mansion/radon-ide/pull/1386/commits/2f028295da888783b57bb3ba09715d64f8ab1783).

Doing this will prevent `Logger` from logging until the `IDE` instance is created, since `outputChannelRegistry` is constructed only after the `IDE` instance is constructed. But I tested this, and this doesn't seem to cause any issues in practice.

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
latekvo: >Wouldn't it be better if we send the individual logs as separate entries in the response?

Good idea, done in [93a44ad](https://github.com/software-mansion/radon-ide/pull/1386/commits/93a44ad74e552b2e0628ddd6220afa33974a162a).

>(50 first log entries and 200 latest entries)

I think this adds unnecessary context, the LLM doesn't need to know that.
We already attach `...\n\n[SKIPPED ${droppedLogsCounter} LINES OF LOGS]\n\n...` to the message so that the LLM knows that the logs are partial. The logs which we provide also contain nearly 100% of the useful data, so far i haven't found a single case where this `50 + 200` model isn't enough to capture the entire warnings block, and the entire stack trace if one appears. Anything beyond that is usually linking status, and occasional warnings within the libraries used by the user.

> full logs are in ${filepath} 

The parts we supply are already more than the model needs for reading the complete error messages, stack traces and warnings. The LLM won't be able to read these logs in full, they are `>40k` lines long, at `4MB` of size - that's over 1 million tokens - way too much for any model to handle.

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
latekvo: Renamed in [e2f2103](https://github.com/software-mansion/radon-ide/pull/1386/commits/e2f2103e5a8f85cc29f60014da51665f94e3c15f)

packages/vscode-extension/src/project/ReadableLogOutputChannel.ts
latekvo: Fixed in [8371c95](https://github.com/software-mansion/radon-ide/pull/1386/commits/8371c95347d67d0df21f1cea5e000f3d5c09f2b8)

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
latekvo: Done in [bef8e5e](https://github.com/software-mansion/radon-ide/pull/1386/commits/bef8e5e98e2b65d8afd5154ae198e293669731ab)

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
latekvo: Done in [f170a0e](https://github.com/software-mansion/radon-ide/pull/1386/commits/f170a0e2063cbc8d805db87dcaec195c3f4ecadf). 

Here's the full message i used:

`Couldn't retrieve build logs - Radon IDE hasn't run any build. You need to select a project and a device in Radon IDE panel.`

packages/vscode-extension/src/project/CircularBuffer.ts
jwajgelt: You're right, I confused the meaning of `headIndex` (I thought it was the head index, not the tail index...) as well as the arguments to `slice` (while reading the PR I thought the second one was length of the slice, not the element-after-last), which coincidentally cancelled-out to a correct implementation (assuming no `pop` operation, which we don't have).

latekvo: My bad, `headIndex` is a bad name choice. I'll rename it to `tailIndex`. Alternatively, I think `nextIndex` also sounds like a good name, and perhaps it's a bit clearer in it's function, let me know if you prefer it over `tailIndex`.

packages/vscode-extension/src/project/ReadableLogOutputChannel.ts
latekvo: This is an issue with VSCode that's currently being worked on ([link](https://github.com/microsoft/vscode/issues/204946#issuecomment-2705935992)). I don't think duplicating their work is a good idea, especially since we haven't ever received any issues regarding logs not being cleared.

kmagiera: if re-creating the channel fixes the problem we should use it as a workaround

latekvo: Re-creating the channel unfortunately does not work as a workaround, the logs persist through removals. 
Although directly removing the `.log` file does work, so I'll take some time to figure out how (and if it is possible) to access `context` from within `OutputChannelRegistry` (or more broadly, the `IDE` class). The `ExtensionContext` is required for getting `logUri`.

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
```diff
@@ -20,12 +22,68 @@ export async function screenshotToolExec(): Promise<ToolResponse> {
...
+    const screenshot = await session.captureScreenshot(session.deviceRotation);
+    const contents = readFileSync(screenshot.tempFileLocation, { encoding: "base64" });
+
+    return {
+      content: [...combinedLogsContent, pngToToolContent(contents)],
+    };
+  }
+
+  return {
+    content: [...combinedLogsContent],
```
kmagiera: we do we clone the array here again?

packages/vscode-extension/src/project/ReadableLogOutputChannel.ts
kmagiera: ok, if dropping the channel doesn't work lets not change it here.

packages/vscode-extension/src/Logger.ts
```diff
@@ -1,6 +1,9 @@
-import { LogOutputChannel, window } from "vscode";
+import { Output } from "./common/OutputChannel";
+import { IDE } from "./project/ide";
 
-const outputChannel = window.createOutputChannel("Radon IDE", { log: true });
+const getIdeOutputChannel = () => {
+  return IDE.getInstanceIfExists()?.outputChannelRegistry.getOrCreateOutputChannel(Output.Ide);
```
kmagiera: this isn't good as it will implicitly create IDE instance. While logs can be produces before we can instantiate the IDE. I think it'd be better to convert the OutputChannelRegistry to be accessible as singleton. The class is stateless, so it should be fine to make the exception in this case

kmagiera (changes requested):

packages/vscode-extension/src/Logger.ts
latekvo: > it will implicitly create IDE instance

I don't think that's the case. The `getInstanceIfExists()` method does not create an `IDE` instance, it returns `null` when no `IDE` instance exists. When it returns `null`, the `getOrCreateOutputChannel()` method is not called either.

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
latekvo: It's a mistake. Fixed in [aa1caea](https://github.com/software-mansion/radon-ide/pull/1386/commits/aa1caea0c8b94801137a53b2d7366f6966d23eec)

latekvo:
Added metro logging support in [f8f951a](https://github.com/software-mansion/radon-ide/pull/1386/commits/f8f951ad4b812d1fb9ecbdf49b708d7ffad0d64f).

packages/vscode-extension/src/project/CircularBuffer.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  public write(value: T) {
+    this.buffer[this.tailIndex] = value;
+    this.tailIndex = (this.tailIndex + 1) % this.capacity;
+    this.size = Math.min(this.size + 1, this.capacity);
+  }
+
+  public clear() {
+    this.tailIndex = 0;
+    this.buffer.length = 0;
```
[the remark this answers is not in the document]
latekvo: that's not true, setting `buffer.length = 0` will clear the array

packages/vscode-extension/src/project/ReadableLogOutputChannel.ts
```diff
@@ -0,0 +1,75 @@
...
+  let droppedLogsCounter = 0;
+
+  const storeLog = (value: string) => {
+    if (logHead.length < KEEP_FIRST_N) {
+      logHead.push(value);
+      return;
+    }
+
+    logTailBuffer.write(value);
+    droppedLogsCounter++;
```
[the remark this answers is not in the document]
latekvo: this is also completely false, the value is incremented only once the head array is filled

packages/vscode-extension/src/Logger.ts
kmagiera: this makes it even worse. Because we will never log things until IDE instance is created. This may result in some important error logs getting missing in some initialization phase.

latekvo: Fixed in [2902d92](https://github.com/software-mansion/radon-ide/pull/1386/commits/2902d92d77675095ab8645cd6e0f9a3ed1227467). Since `IDE` depended on `Logger` during initialization, and `Logger` depended on `IDE` due to dependency on `OutputChannelRegistry`, I fixed this circular dependency by making `OutputChannelRegistry` a singleton that's initialized on `extension.ts:activate()` and disposed on `extension.ts:deactivate()`.

packages/vscode-extension/src/project/metro.ts
```diff
@@ -515,19 +526,25 @@ export class MetroLauncher extends Metro implements Disposable {
...
           } else if (event.type === "client_log" && event.level === "error") {
-            Logger.error(stripAnsi(event.data[0]));
+            const err = stripAnsi(event.data[0]);
+            Logger.error(err);
+            metroOutputChannel?.appendLine(err);
           } else {
             Logger.debug("Metro", line);
           }
 
+          let log;
```
kmagiera: why is this here?. It should just be const defined in the case block. I think you need to use `{ }` braces around the case in order to be able to define variables there, like so:

```ts
case "initialize_done": {
  this._port = event.port;
  const log = `Metro started on port ${this._port}`;
  metroOutputChannel.appendLine(log):
  Logger.into(log);
  resolve();
}
```

packages/vscode-extension/src/project/metro.ts
```diff
@@ -515,19 +526,25 @@ export class MetroLauncher extends Metro implements Disposable {
...
           switch (event.type) {
             case "RNIDE_expo_env_prelude_lines":
               this._expoPreludeLineCount = event.lineCount;
               Logger.debug("Expo prelude line offset was set to: ", this._expoPreludeLineCount);
               break;
             case "initialize_done":
               this._port = event.port;
-              Logger.info(`Metro started on port ${this._port}`);
+              log = `Metro started on port ${this._port}`;
+              metroOutputChannel?.appendLine(log);
```
kmagiera: wouldn't it be better if we also used log levels here instead of `appendLine` ? Seems like that would be more consistent with other output channel that we use:
```
metroOutputChannel.info(log);
```

packages/vscode-extension/src/project/metro.ts
```diff
@@ -515,19 +526,25 @@ export class MetroLauncher extends Metro implements Disposable {
               this.bundleProgressEventEmitter.fire({ bundleProgress });
             }
           } else if (event.type === "client_log" && event.level === "error") {
-            Logger.error(stripAnsi(event.data[0]));
+            const err = stripAnsi(event.data[0]);
+            Logger.error(err);
+            metroOutputChannel?.appendLine(err);
```
kmagiera: why not use `metroOutputChannel.error()` here?

packages/vscode-extension/src/project/metro.ts
```diff
@@ -477,6 +479,15 @@ export class MetroLauncher extends Metro implements Disposable {
       EXPO_EDITOR: FAKE_EDITOR,
       ...(isExtensionDev ? { RADON_IDE_DEV: "1" } : {}),
     };
+
+    const metroOutputChannel =
```
kmagiera, line 483: I think we can bail here if IDE instance doesn't exists. Then the outputChannel would never be undefined

packages/vscode-extension/src/Logger.ts
```diff
@@ -1,6 +1,10 @@
-import { LogOutputChannel, window } from "vscode";
+import { Output } from "./common/OutputChannel";
+import {
+  createReadableOutputChannel,
+  ReadableLogOutputChannel,
+} from "./project/ReadableLogOutputChannel";
 
-const outputChannel = window.createOutputChannel("Radon IDE", { log: true });
+const outputChannel = createReadableOutputChannel(Output.Ide, true);
```
kmagiera: I don't think we need to change this at all, we never access the readable functions of this channel anyway

kmagiera (approved):
Left a couple of comments inline. Otherwise looks good

packages/vscode-extension/src/project/metro.ts
latekvo: The `.info` is not conventional, we only use it in `Logger.ts`. 
In all other places we're using `.appendLine` (in `buildAndroid.ts`, `packageManager.ts`, `pods.ts`, `AndroidEmulatorDevice.ts`, `easCommand.ts`, `customBuild.ts`, and 3 other files).

So I don't think there's a point in using `info` in this case, as these logs will be available exclusively to the AI, and their contents are already well described, without needing additional `[info]` or `[error]` tags.

Lastly, while this isn't such a big issue, values passed to `trace`, `warn`, `info`, `error` and `debug` aren't being tracked within `ReadableLogOutputChannel`, and would all have to be overwritten within `createReadableOutputChannel` to make this insignificant addition, which would add quite a bit of bearly used code, reducing maintainability in the future.

packages/vscode-extension/src/project/metro.ts
latekvo: Same as the comment above.

packages/vscode-extension/src/project/metro.ts
latekvo: Yes, done in [ba3343c](https://github.com/software-mansion/radon-ide/pull/1386/commits/ba3343ce1cb0de8b4e5b3ffa4d09b20f8276ad76)

packages/vscode-extension/src/project/metro.ts
latekvo: Fixed in [ba3343c](https://github.com/software-mansion/radon-ide/pull/1386/commits/ba3343ce1cb0de8b4e5b3ffa4d09b20f8276ad76). This was left behind from older versions where we had multiple log definitions. Also, no need for a code block, regular `case` syntax works, since there's only a single `log` definition.

packages/vscode-extension/src/Logger.ts
latekvo: Using `createReadableOutputChannel` avoids a lot of typing issues that I wasn't able to easily bypass.

packages/vscode-extension/src/project/metro.ts
kmagiera: All the other places where we append to channel are where we directly want to pipe output of another process. Here, we create log entries on our own, so I wouldn't compare those two cases. This one is more similar to the Logger case, we just need the log to appear in another channel

Either way, I'm ok keeping this as is. Doesn't seem too imporant for the whole PR to be blocked on the log formatting output.

packages/vscode-extension/src/Logger.ts
kmagiera: what issues? You can safely revert all the changes in this file. I suppose there were no typing issues before you made the changes

latekvo: The typing issues occur in `OutputChannelRegistry`, which holds all output channels, including this one, in one `Map`. 

Although now looking at it, I will just remove `[Output.Ide, Logger.rawOutputChannel]` entry from the `channelByName` `Map`, since was never accessed there, and never will be, not sure why it was there in the first place. This resolves the problem.

Done in [02a4a98](https://github.com/software-mansion/radon-ide/pull/1386/commits/02a4a98f2092b4a96c58134a334b86ed9656fe3b)

PR #1387, opened 2025-07-31
refactor: extract application specific code from DeviceSession to ApplicationSession

jwajgelt:
Splits the `DeviceSession` into two parts:
- `DeviceSession` which is meant to model a "running device" and controls both the device and the application building process
- `ApplicationSession` which models a "running application" and provides controls for the application (JS reload, profiling, debugger controls) and handles the resources for a single "run" of the application (DebugSession, ToolsManager)

### How Has This Been Tested: 
- open app in Radon
- check the app launches correctly on both iOS and Android
- check that all the restart options still work
- check that CPU and React profiler still works
- check that the debugger still works and reconnects after the app is restarted from the phone's launcher
- check that switching devices / projects still works
- check launching a device in the background behaves correctly

### DIFF
--- a/packages/vscode-extension/src/project/applicationSession.ts
+++ b/packages/vscode-extension/src/project/applicationSession.ts
@@ -61,14 +61,14 @@ export class ApplicationSession implements ToolsDelegate, Disposable {
 
   public static async launch(
     { applicationContext, device, buildResult, metro, devtools }: LaunchApplicationSessionDeps,
-    initiallyActive: boolean,
+    getIsActive: () => boolean,
     onLaunchStage: (stage: StartupMessage) => void,
     cancelToken: CancelToken
   ): Promise<ApplicationSession> {
     const packageNameOrBundleId =
       buildResult.platform === DevicePlatform.IOS ? buildResult.bundleID : buildResult.packageName;
     const session = new ApplicationSession(device, metro, devtools, packageNameOrBundleId);
-    if (initiallyActive) {
+    if (getIsActive()) {
       // we need to start the parent debug session asap to ensure metro errors are shown in the debug console
       await session.setupDebugSession();
     }
@@ -84,7 +84,7 @@ export class ApplicationSession implements ToolsDelegate, Disposable {
       Promise.withResolvers<void>();
     const bundleErrorSubscription = metro.onBundleError(({ source }) => {
       resolveBundleError();
-      if (initiallyActive) {
+      if (getIsActive()) {
         focusSource(source);
       }
     });
@@ -97,6 +97,17 @@ export class ApplicationSession implements ToolsDelegate, Disposable {
 
       onLaunchStage(StartupMessage.WaitingForAppToLoad);
       await cancelToken.adapt(Promise.race([devtools.appReady(), bundleErrorPromise]));
+
+      if (getIsActive()) {
+        const activatePromise = session.activate();
+        const hasBundleError = session.bundleError !== undefined;
+        // NOTE: if an initial bundle error occurred, the app won't connect to Metro
+        // and we won't be able to attach the debugger anyway, so there's no point in waiting
+        if (!hasBundleError) {
+          onLaunchStage(StartupMessage.AttachingDebugger);
+          await cancelToken.adapt(activatePromise);
+        }
+      }
 
       return session;
     } catch (e) {
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -499,23 +499,13 @@ export class DeviceSession implements Disposable {
         metro: this.metro,
         devtools: this.devtools,
       },
-      this.isActive,
+      () => this.isActive,
       this.updateStartupMessage.bind(this),
       cancelToken
     ).then(async (applicationSession) => {
-      // if an initial bundle error occurred, the app won't connect to Metro anyway,
-      // so there's no point in trying to attach a debugger
-      const hasBundleError = applicationSession.state.bundleError !== undefined;
-      try {
-        if (this.isActive && !hasBundleError) {
-          this.updateStartupMessage(StartupMessage.AttachingDebugger);
-          await cancelToken.adapt(applicationSession.activate());
-        } else if (!this.isActive) {
-          await cancelToken.adapt(applicationSession.deactivate());
-        }
-      } catch (e) {
+      if (cancelToken.cancelled) {
         applicationSession.dispose();
-        throw e;
+        throw new CancelError("Application launch was cancelled");
       }
 
       this.applicationSession = applicationSession;


packages/vscode-extension/src/devices/DeviceBase.ts
```diff
@@ -225,17 +230,17 @@ export abstract class DeviceBase implements Disposable {
...
   }
 
-  async startPreview() {
+  public async startPreview() {
     if (!this.previewStartPromise) {
       this.preview = this.makePreview();
       this.previewStartPromise = this.preview.start();
       this.previewStartPromise.then(() => {
         this.applyPreviewSettings();
+        this.preview?.rotateDevice(this._rotation);
```
filip131311: why int it one of the preview settings?

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -0,0 +1,459 @@
...
+
+interface LaunchApplicationSessionDeps {
+  applicationContext: ApplicationContext;
+  device: DeviceBase;
+  buildResult: BuildResult;
+  metro: MetroLauncher;
+  devtools: Devtools;
+}
+
+export class ApplicationSession implements ToolsDelegate, Disposable {
```
filip131311, line 44: could you abstract an interface that ApplicationSession is implementing? It is difficult for me to grasp what it is in such a big class

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -76,36 +59,27 @@ export class DeviceBootError extends Error {
   }
 }
 
-export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
+export class DeviceSession implements Disposable {
```
filip131311, line 62: Same comment as in application session Could we have an interface that those sessions are implementing, I know we didn't before but now that I need to understand what exactly is where I see it would be useful

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -167,12 +135,11 @@ export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
         stageProgress: this.stageProgress,
       };
     } else if (this.status === "running") {
```
filip131311, line 137: shouldn't "running" status be infered from the application state (or it's existence) after this change feels wrong to keep this as deviceSession property

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -552,64 +376,47 @@ export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
   }
 
   private async reloadJS() {
-    if (!this.devtools.hasConnectedClient) {
+    if (this.applicationSession === undefined) {
```
filip131311, line 377: is it imposible for applicationSession to exist and `!this.devtools.hasConnectedClient === true` ?

filip131311: If it's not we should probably perform an extra check before running `this.updateStartupMessage(StartupMessage.WaitingForAppToLoad);`

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -428,27 +265,22 @@ export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
...
-      this.emitStateChange();
-      this.debugSessionEventSubscription.dispose();
-      await this.debugSession.dispose();
+      await this.applicationSession?.deactivate();
+    }
+  }
+
+  private cancelOngoingOperations() {
+    if (this.cancelToken) {
+      this.cancelToken.cancel();
```
filip131311: this never happens as you changed the type of cancelToken:

```ts
 private cancelToken: CancelToken = new CancelToken();
 ```

File: packages/vscode-extension/src/project/deviceSession.ts [hunk 7]
```diff
@@ -680,75 +487,74 @@ export class DeviceSession implements Disposable, MetroDelegate, ToolsDelegate {
...
+      getTelemetryReporter().sendTelemetryEvent(
+        "app:launch:completed",
+        { platform: this.platform },
+        { durationSec: launchDurationSec }
+      );
+    });
+
+    const launchConfig = this.applicationContext.launchConfig;
+    const shouldWaitForAppLaunch = launchConfig.preview.waitForAppLaunch;
+    const waitForAppReady = shouldWaitForAppLaunch ? applicationSessionPromise : Promise.resolve();
```
filip131311, line 527: we should remove `shouldWaitForAppLaunch `completely` as it does not work by accident and nobody noticed that would allow as to move most of the logic here to ApplicationSession where its supposed to be

filip131311 (commented):
I left some inline comments

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: I thought about it. Inferring it makes it harder to understand the transitions while reading the code, since you can't simply look at where `this.status` is assigned. That doesn't mean it's a bad idea, it's just not obviously better.

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: this always happens as I changed the type of cancelToken 🙂

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: I moved the check for that inside `ApplicationSession`

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -0,0 +1,459 @@
...
+    try {
+      onLaunchStage(StartupMessage.Launching);
+      await cancelToken.adapt(
+        device.launchApp(buildResult, metro.port, devtools.port, launchArguments)
+      );
+
+      onLaunchStage(StartupMessage.WaitingForAppToLoad);
+      await cancelToken.adapt(Promise.all([devtools.appReady()]));
+
+      if (getIsActive()) {
```
filip131311, line 101: this should be an independent call

packages/vscode-extension/src/project/deviceSession.ts
filip131311: we can leave it like that untill i get there

filip131311 (approved):
I left one more comment we discusseed but the rest looks nice

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -499,15 +499,18 @@ export class DeviceSession implements Disposable {
...
-    ).then((applicationSession) => {
-      // NOTE: this can happen if the launch finished successfully,
-      // but then the start operation was cancelled before this callback started executing
-      if (cancelToken.cancelled) {
-        applicationSession.dispose();
-        throw new CancelError("Launch cancelled");
+    ).then(async (applicationSession) => {
+      if (this.isActive) {
+        try {
+          this.updateStartupMessage(StartupMessage.AttachingDebugger);
```
filip131311: I don't like it but its okay

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -82,21 +80,30 @@ export class ApplicationSession implements ToolsDelegate, Disposable {
...
     try {
       onLaunchStage(StartupMessage.Launching);
       await cancelToken.adapt(
         device.launchApp(buildResult, metro.port, devtools.port, launchArguments)
       );
 
       onLaunchStage(StartupMessage.WaitingForAppToLoad);
-      await cancelToken.adapt(Promise.all([devtools.appReady()]));
-
-      session.status = "running";
```
filip131311: removing this from here is warring what does set this status?

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -82,21 +80,30 @@ export class ApplicationSession implements ToolsDelegate, Disposable {
     onLaunchStage(StartupMessage.StartingPackager);
     await cancelToken.adapt(metro.ready());
 
+    const { promise: bundleErrorPromise, resolve: resolveBundleError } =
+      Promise.withResolvers<void>();
+    const bundleErrorSubscription = metro.onBundleError(({ source }) => {
+      resolveBundleError();
+      if (initiallyActive) {
```
filip131311: sorry for going back and forth but maybe a getter here would be better as this could happen after switching the device

filip131311 (changes requested):
left some more inline comments

PR #1388, opened 2025-07-31
feat: implement paywall view

kacperkapusciak:
This PR implements the styles of a paywall view. The paywall can be opened using 
```js
const { openPaywall } = usePaywall();
```
hook. Prices on paywall are taken from our payment processor Paddle and can be remotely adjusted. When there's no connection to the internet a generic paywall without prices are displayed.

Also, modals in Radon IDE now support full screen mode. 

https://github.com/user-attachments/assets/9cd05a6b-09d5-4193-81b4-f8b62368da25


packages/vscode-extension/src/webview/components/shared/Modal.tsx
```diff
@@ -2,28 +2,48 @@ import React from "react";
...
 interface ModalProps {
-  title: string;
+  title?: string;
   component: React.ReactNode;
-  open: boolean;
-  setOpen: React.Dispatch<React.SetStateAction<boolean>>;
+  isOpen: boolean;
+  onClose: () => void;
   headerShown?: boolean;
+  isFullScreen?: boolean;
```
jwajgelt: VERY NIT:
```suggestion
  open: boolean;
  onClose: () => void;
headerShown?: boolean;
  fullscreen?: boolean;
```

For me, `<Modal fullscreen />` reads better than `<Modal isFullScreen />`

Then again, the opinions are mixed:
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/boolean-prop-naming.md

So I might be in the wrong here. 🤷

packages/vscode-extension/src/webview/providers/ModalProvider.tsx
```diff
@@ -14,31 +27,51 @@ const ModalContext = createContext<ModalContextProps>({
...
-    // @ts-ignore TODO see this further but i think it's fine
-    setComponent(modalComponent);
-    setOpen(true);
+  const [state, setState] = useState<ModalState>({
+    title: "",
+    component: <></>,
+    open: false,
+    headerShown: true,
+    isFullScreen: false,
+  });
```
jwajgelt: I'd prefer if we used a `null` value instead of a "special" state

```suggestion
  const [state, setState] = useState<ModalState>(null);
```

jwajgelt: Then, you can use this to conditionally render the `<Modal />` component in the `return` statement of this component (you already do that in `Modal` itself -- https://github.com/software-mansion/radon-ide/pull/1388/files#diff-6f597facf5ac34efece731f6f83743d516ac8c4ed237a5be584708bf832abc30R25-R27).

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -308,7 +308,7 @@ function LaunchConfigurationView({
           <Button
             secondary
             onClick={() => {
-              openModal("", <DeleteConfirmationModal />);
+              openModal(<DeleteConfirmationModal />);
```
jwajgelt, line 327: Did you check that (non-fullscreen) modals with no title still look ok?

packages/vscode-extension/src/webview/views/PaywallView.tsx
```diff
@@ -0,0 +1,166 @@
...
+                  )}
+                  onClick={() => setSelectedPlan("monthly")}>
+                  <div className="plan-header">
+                    <span className="plan-name">Monthly Plan</span>
+                    <span className="plan-price">
+                      <span>{radonProMonthlyPrice?.formattedTotals.total}</span> / month
+                    </span>
+                  </div>
+                  <div className="plan-description">Billed monthly</div>
+                </div>
```
jwajgelt: This begs for a `SubscriptionOption` component.

packages/vscode-extension/src/webview/views/PaywallView.tsx
```diff
@@ -0,0 +1,166 @@
...
+  "All the Free features",
+  "Redux UI plugin",
+  "React Query plugin",
+  "CPU Profiler integration",
+  "React Profiler integration",
+  "React Scan integration",
+  "Radon AI assistant",
+  "Early access to new features",
+  "Priority support",
+];
```
jwajgelt, line 26: Do we want to leave this as is?

packages/vscode-extension/src/webview/views/PaywallView.tsx
```diff
@@ -0,0 +1,166 @@
...
+          countryCode: "US",
+        },
+      });
+
+      if (!response) {
+        throw new Error("Failed to fetch prices from Paddle - no response received");
+      }
+
+      setPrices(response);
+      setIsLoadingPrices(false);
```
kacperzolkiewski: I think that the line `setIsLoadingPrices(false)` in the `try` block might be redundant, as we already have this line in the finally block which will execute regardless of whether an error occurs

kacperkapusciak: you are right, fixed

packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
kacperkapusciak: i did but i missed one critical bug still, so thanks to this comment it's now fixed

packages/vscode-extension/src/webview/views/PaywallView.tsx
kacperkapusciak: let's hold the PR in this state for now

packages/vscode-extension/src/webview/providers/ModalProvider.tsx
kacperkapusciak: fixed

packages/vscode-extension/src/webview/components/shared/Modal.tsx
kacperkapusciak: okay 🤷

packages/vscode-extension/src/webview/views/PaywallView.tsx
kacperkapusciak: i _beg_ to differ as it made it more complex

but okay I've extracted it to separate component

PR #1396, opened 2025-08-04
fix: fix touch-area overflow issues

pFornagiel:
### Description

This PR fixes issue with persistent overflow when device is in portrait mode due to problems with touch-area dimensions. Touch area dimensions have been changed, to not include 7px additional margin on both sides, which has been deemed unnecessary because of current implementation of touch-handling.

### How Has This Been Tested: 

- **test whether during window-resizing the overflow persist, check if the device properly handles touches which start beyond the phone screen**

Not applicable.


jwajgelt (approved):
Changes look fine, and the behavior when testing seems to be correct as well. Good job.

PR #1397, opened 2025-08-04
refactor: cleanup `inspectElementAt` extension-side implementation

### LINKED ISSUE #1244: Reload the app with enabled inspector - emulator freeze (Too many callbacks in memory! Something is wrong!)
**Description**
When I tested the Inspector button and switched between devices, after building the application on the newly selected device I had a black screen for a long time (mainly on android devices). Also after a few minutes (when the app emulator loaded) - and inspecting the elements on the screen errors was thrown in output console ("Too many callbacks in memory! Something is wrong!").

**Preconditions**

- Radon IDE extension is available and launched for the react native application project
- iOS and android devices are available in the devices list
- App is launched on some iOS simulator

**Steps to Reproduce**

1. I run the tested application on iOS symulator
2. I clicked on "Select an element to inspect it" and hovered on one items on the device screen
3. Without turning off this option - opened the devices list and selected some android devices
4. Observed that the last log inside the Output console is an information from Metro type: bundle_bould_done
5. But the screen of device is whole black - also after reload the app, wait a longer time or change device settings (like font size)
6. I have also tried to Clear the Metro cache - but also with the same result, a black screen
7. Also the same situation when I try t switch to some other android device and turned off the inspector - a black screen for longer while
8. I rebooted the Radon IDE but also the black screen was displayed
9. Then after a longer while I have clicked on device screen and android logo app was displayed

https://github.com/user-attachments/assets/8c2547bd-023f-442d-b525-e69c7407f033

Then after a longer while I reloaded the application and then error was displayed.

Image

I had to clear metro cache to finally launch this project on android device:

Image

Also when I had active inspector and launched the iOS device - a strange white screen was displayed instead of loaded app (some error from simulator_server was displayed then: Unrecognized argument: setUpKeyboard). After app reload nothing changed. I had to reboot the Radon IDE to load successfully the app on this device.

https://github.com/user-attachments/assets/8641190b-bc25-49d1-902e-fe7dd35f6760

**Actual Behavior**
There is a lot of freezes, hangings when I try to load project on some device with enabled inspecting.

**Expected Behavior**
Switching between the devices should works smoothly (even inspecting elements is opened).

**Environment**
Version of used Radon IDE: 1.7.1
Version of React Native: 0.76

jwajgelt:
- moves the implementation of `inspectElementAt` fully to application session (from `Project` and `DeviceSession`)
- changes the API to return a Promise with InspectData instead of accepting a callback

Fixes #1244

### How Has This Been Tested: 
- open app in Radon
- verify that the element inspector still works
- verify that after hovering a lot of elements in a short time with the element inspector enabled, you don't see "Too many callbacks in memory! Something is wrong!" warnings in the console logs

### DIFF
--- a/packages/vscode-extension/src/project/ApplicationContext.ts
+++ b/packages/vscode-extension/src/project/ApplicationContext.ts
@@ -8,7 +8,7 @@ import { BuildManagerImpl, BuildManager } from "../builders/BuildManager";
 import { BatchingBuildManager } from "../builders/BatchingBuildManager";
 import { LaunchConfiguration, LaunchOptions } from "../common/LaunchConfig";
 import { StateManager } from "./StateManager";
-import { ApplicationContextState } from "../common/State";
+import { ApplicationContextState, WorkspaceConfiguration } from "../common/State";
 import { ApplicationDependencyManager } from "../dependency/ApplicationDependencyManager";
 import { Logger } from "../Logger";
 
@@ -70,6 +70,7 @@ export class ApplicationContext implements Disposable {
 
   constructor(
     private readonly stateManager: StateManager<ApplicationContextState>,
+    private readonly workspaceConfigState: StateManager<WorkspaceConfiguration>, // owned by `Project`, do not dispose
     launchConfig: LaunchConfiguration,
     public readonly buildCache: BuildCache
   ) {
@@ -82,6 +83,10 @@ export class ApplicationContext implements Disposable {
     this.buildManager = buildManager;
 
     this.disposables.push(this.applicationDependencyManager, buildManager);
+  }
+
+  public get workspaceConfiguration(): WorkspaceConfiguration {
+    return this.workspaceConfigState.getState();
   }
 
   public get appRootFolder(): string {
--- a/packages/vscode-extension/src/project/applicationSession.ts
+++ b/packages/vscode-extension/src/project/applicationSession.ts
@@ -54,6 +54,7 @@ export class ApplicationSession implements ToolsDelegate, Disposable {
   private isRefreshing: boolean = false;
   private appOrientation: DeviceRotation | undefined;
   private isActive = false;
+  private inspectCallID = 7621;
 
   private stateChangedEventEmitter = new EventEmitter<void>();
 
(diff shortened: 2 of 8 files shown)


packages/vscode-extension/src/project/ApplicationContext.ts
```diff
@@ -70,6 +70,7 @@ export class ApplicationContext implements Disposable {
 
   constructor(
     private readonly stateManager: StateManager<ApplicationContextState>,
+    private readonly workspaceConfigState: StateManager<WorkspaceConfiguration>, // owned by `Project`, do not dispose
```
filip131311, line 73: I know that you don't like the name manager. But pleas add it to this name as I would like to reserve pure `state` naming for states that are handled directly inside classes

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -448,6 +458,48 @@ export class ApplicationSession implements ToolsDelegate, Disposable {
...
+  ): Promise<InspectData> {
+    const id = this.inspectCallID++;
+    const { promise, resolve, reject } = Promise.withResolvers<InspectData>();
+    const listener = this.devtools.onEvent("inspectData", (payload) => {
+      if (payload.id === id) {
+        listener.dispose();
+        resolve(payload as unknown as InspectData);
+      } else if (payload.id >= id) {
+        listener.dispose();
+        reject("Inspect request was invalidated by a later request");
```
filip131311, line 476: where do we handle this reject?

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -461,3 +513,15 @@ export class ApplicationSession implements ToolsDelegate, Disposable {
     this.device.terminateApp(this.packageNameOrBundleId);
   }
 }
+
+export function isAppSourceFile(filePath: string) {
```
filip131311: I exported it as a util in #1404

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: It will be thrown on the webview, as an unhandled promise rejection.
It should be fine to catch it there and do nothing though.

packages/vscode-extension/src/project/ApplicationContext.ts
jwajgelt: nah

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: 🆒

PR #1399, opened 2025-08-05
Add Disabling Inspector When Unsupported

pFornagiel:
### Description

This PR introduces functionality, which disables inspector-like-overlays (Element Inspector and Render Overlays) if:
- the application is not opened or is in state other than "active"
- the application does not support Edge-to-Edge mode, which causes problems with the extensions

If any of above cases occur, the options become greyed-out with appropriate tooltip on hover and inspection on right click is replaced with information-box information that inspector is not available.

Right click when inspector is unavailable:
Screenshot 2025-08-05 at 11 16 16

Disabled Inspector button:
Screenshot 2025-08-05 at 11 16 49

Disabled Render Outlines button:
Screenshot 2025-08-05 at 11 17 02

Known Issues - Problem with Devtools: On android, when app is closed for longer time, devtools become disconnected and the button becomes grayed-out for good. Not solved here - matter for different PR.

### How Has This Been Tested: 

- **test Inspector and Render Outlines overlays in apps that do not support Edge-to-Edge, outside of developed applications, in properly working apps, etc.**

Added limitation section to Features -> Element inspector docs section:
[@pFornagiel/docs-inspector-availability](https://github.com/software-mansion/radon-ide/tree/%40pFornagiel/docs-inspector-availability)

### DIFF
--- a/packages/vscode-extension/src/webview/components/RenderOutlinesOverlay.tsx
+++ b/packages/vscode-extension/src/webview/components/RenderOutlinesOverlay.tsx
@@ -35,7 +35,8 @@ function useIsEnabled() {
     return false;
   }
   const isToolEnabled = selectedDeviceSession?.toolsState[RENDER_OUTLINES_PLUGIN_ID]?.enabled;
-  const isInspectorAvailable = selectedDeviceSession?.toolsState[RENDER_OUTLINES_PLUGIN_ID]?.pluginAvailable;
+  const isInspectorAvailable =
+    selectedDeviceSession?.toolsState[RENDER_OUTLINES_PLUGIN_ID]?.pluginAvailable;
   return isToolEnabled && isInspectorAvailable;
 }
 


packages/vscode-extension/lib/inspector_availability.js
```diff
@@ -0,0 +1,87 @@
...
+  let appFocusSubscription = null;
+
+  const dimensionEventSubscription = Dimensions.addEventListener("change", handleDimensionsChange);
+  const appStateSubscription = AppState.addEventListener("change", handleAppStateChange);
+
+  // iOS does not support "blur" and "focus" events on AppState
+  if (Platform.OS === "android") {
+    appBlurSubscription = AppState.addEventListener("blur", handleBlurChange);
+    appFocusSubscription = AppState.addEventListener("focus", handleFocusChange);
+  }
```
jwajgelt: Are these triggered differently from the `"change"` event on Android?
Can you be "blurred" but "active" (or "focused" but not "active")?

packages/vscode-extension/src/common/Project.ts
```diff
@@ -33,6 +33,8 @@ export type ToolState = {
   enabled: boolean;
   panelAvailable: boolean;
   label: string;
+  pluginButtonDisabled: boolean;
+  disabledPluginTooltipLabel: string;
```
jwajgelt: I'd make this optional.

packages/vscode-extension/src/common/Project.ts
```diff
@@ -33,6 +33,8 @@ export type ToolState = {
   enabled: boolean;
   panelAvailable: boolean;
   label: string;
+  pluginButtonDisabled: boolean;
```
jwajgelt: I don't like the name.
I'd call this simply "available", but that is already taken for fully hiding stuff...
Having both `disabled` and `enabled` would also be psychotic.
Maybe we could rename the "available" field on `ToolPlugin` to something else ("detected" maybe?) and reuse the name here?

Or perhaps, we could have a `status` field instead of `enabled`, taking three values:
```
enum ToolStatus {
  Enabled,
  Disabled,
  ForceDisabled
}
```

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -355,6 +358,11 @@ export class ApplicationSession implements ToolsDelegate, Disposable {
           this.appOrientation = orientation;
         }
 
+        this.emitStateChange();
+      }),
+      this.devtools.onEvent("inspectorAvailabilityChanged", (isAvailable: boolean) => {
+        this.inspectorAvailability = isAvailable;
+        this.toolsManager.setToolButtonDisabled(RENDER_OUTLINES_PLUGIN_ID, !isAvailable);
```
jwajgelt: Disabling the toggle is a part of the plugin's logic -- only the plugin should know what it needs to work correctly, and it should be responsible for changing its own state.

You can move the listener inside the `RenderOutlinesPlugin` instead.
Then, you also don't need to have the `ToolsManager.setToolButtonDisabled` method either, so that other components which can access `ToolsManager` can't just set the state for the tools.

packages/vscode-extension/src/webview/components/shared/IconButton.tsx
```diff
@@ -77,7 +79,7 @@ const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>((props,
 
   return (
     <Tooltip label={label} side={tooltipSide} type={tooltipType ?? type}>
-      {button}
+      {shouldDisplayLabelWhileDisabled ? <span>{button}</span> : button}
```
jwajgelt: Why is this necessary?

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -52,6 +53,11 @@ interface Point {
   y: number;
 }
 
+interface InspectorUnavailableBoxData {
+  visible: boolean;
```
jwajgelt: `visible` is only ever set to `true`. I'd change it to get rid of this interface and instead just have 
```tsx
  const [inspectorUnavailablePosition, setInspectorUnavailablePosition] = useState<Point | null>(null);
```

packages/vscode-extension/src/plugins/render-outlines/render-outlines-plugin.ts
```diff
@@ -34,6 +36,11 @@ export class RenderOutlinesPlugin implements ToolPlugin, RenderOutlinesInterface
   public readonly label = "Outline Renders";
   public readonly available = true;
   public readonly persist = false;
+  public readonly disabledTooltipLabel = INSPECT_UNAVAILABLE_TOOLTIP_LABEL;
```
jwajgelt: I'd prefer if the label here said "Render Outlines is disabled in ..." rather than "Inspector"

packages/vscode-extension/src/common/Constants.ts
```diff
@@ -3,3 +3,6 @@ export const MinSupportedVersion = {
   expo: "49.0.0",
   storybook: "5.2.0",
 } as const;
+
+export const INSPECT_UNAVAILABLE_TOOLTIP_LABEL =
+  "Inspector is disabled in apps that don\'t support Edge-to-Edge.";
```
jwajgelt: I don't know if this is the right place for this.
Maybe we don't want to have this string be shared anyway, I think it should only be used in one place (the element inspector button label)

jwajgelt (changes requested):

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: Also, `RenderOutlinesPlugin` should probably disable itself when `inspectorAvailabilityChanged(false)` is sent, to avoid getting stuck in a `enabled: true` state without being able to toggle the switch.

packages/vscode-extension/lib/inspector_availability.js
pFornagiel: Added comments in d79d5c09

As per documentation:
image

The events are separate from "change" event and from the app state. On android the app state remains active in situations where on iPad it would become inactive - that is when the app switcher is used, notification drawer is pulled, etc. Upon these actions on android `blur` and `focus` events are fired, which do not change the app state but allow for detecting such situations. 

In code, thanks to the predicate:

```js
const availability = isAppStateActive && isEdgeToEdge && isFocused;
```

We are able to properly handle these situations, as we are also always checking if app remains active.

The `if` is needed, because trying to subscribe to those events on iOS results in an error.

packages/vscode-extension/src/common/Project.ts
pFornagiel: Changed in b398d80d

`panelAvailable` has been changed to `isPanelTool`, as it describes the meaning of the predicate on the frontend nicely.

`pluginButtonDisabled` has been changed to `pluginAvailable` and `disabledPluginTooltipLabel` to `pluginUnavailableTooltip`.

```ts
export type ToolState = {
  enabled: boolean;
  isPanelTool: boolean;
  label: string;
  pluginAvailable?: boolean;
  pluginUnavailableTooltip?: string;
};
```

The enum approach was not chosen, because such three-state logic would need a greater amount of code logic refactoring, which is not necessity as of now. Can be made into different PR later.

packages/vscode-extension/src/project/applicationSession.ts
pFornagiel: > Disabling the toggle is a part of the plugin's logic -- only the plugin should know what it needs to work correctly, and it should be responsible for changing its own state.
> 
> You can move the listener inside the `RenderOutlinesPlugin` instead. Then, you also don't need to have the `ToolsManager.setToolButtonDisabled` method either, so that other components which can access `ToolsManager` can't just set the state for the tools.

Changed in b398d80d

The listener has been moved to `RenderOutlinesPlugin`. Thanks for outlining this possibility.

```ts
export class RenderOutlinesPlugin implements ToolPlugin, RenderOutlinesInterface, Disposable {
  // ...
  constructor(private inspectorBridge: RadonInspectorBridge) {
        // ...
        this.devtoolsListeners.push( this.inspectorBridge.onEvent("inspectorAvailabilityChanged", (isAvailable: boolean) => {
                this._pluginAvailable = isAvailable;
              })
            );
          }
```

pFornagiel: > Also, `RenderOutlinesPlugin` should probably disable itself when `inspectorAvailabilityChanged(false)` is sent, to avoid getting stuck in a `enabled: true` state without being able to toggle the switch.

Changed in 46e69381

The switch now appears disabled, when RenderOutlines plugin is not available, but still remembers the previous state, so that when user has switched it on and then, for example, closed and reopened app (which disables and enables Render Outlines now), it remains to be turned on afterwards.

packages/vscode-extension/src/plugins/render-outlines/render-outlines-plugin.ts
pFornagiel: Changed in b398d80

```ts
public readonly pluginUnavailableTooltip = "Render Outlines is disabled in apps that don\'t support Edge-to-Edge.";
```

packages/vscode-extension/src/common/Constants.ts
pFornagiel: Deleted in b398d80

The element inspector and Outlines Overlay now use separate strings defined locally.

```ts
public readonly pluginUnavailableTooltip = "Render Outlines is disabled in apps that don\'t support Edge-to-Edge.";
```

```ts
tooltip={{
            label: inspectorAvailable
              ? "Select an element to inspect it"
              : "Render Outlines is disabled in apps that don't support Edge-to-Edge.",
          }}
```

packages/vscode-extension/src/common/Project.ts
pFornagiel: Made optional in b398d80

```ts
export type ToolState = {
  enabled: boolean;
  isPanelTool: boolean;
  label: string;
  pluginAvailable?: boolean;
  pluginUnavailableTooltip?: string;
};
```

packages/vscode-extension/src/webview/components/shared/IconButton.tsx
pFornagiel: Added comment in bfc066ef

The button has `pointer-events: none` set in css when disabled,  so if we wish to show label when disabled, we've got to have an element (in this case span), which can react to pointer events, `onHover` included.

packages/vscode-extension/src/webview/components/Preview.tsx
pFornagiel: Refactored and simplified the logic in 8496f3a5

Thanks for pointing this one out.

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: Won't this cause issues where when the backgrounded application renders something while this option was already enabled, you'll get the render outlines even though in the UI the option is disabled, and there's no option to change that?

packages/vscode-extension/src/webview/components/shared/IconButton.tsx
jwajgelt: Then, I think, the right place to add that `<span>` is in the `Tooltip` component, to ensure the actual tooltip is shown even if the wrapped component "misbehaves".

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: https://github.com/software-mansion/radon-ide/pull/1399/files#diff-c355da324ac0bddfb3413da5f867756f61000bfa2d110fec335e11709437aeffR38-R39

Ok, I see you guard against it on the frontend.
_I think_ it would be better if the `ToolPlugin` handled that, so you don't need to repeat the check in two places (with risk of bugs when changes are made), but it's fine to leave it as is.

pFornagiel: This does not cause issues, because the `tools.pluginAvailable` predicate used in `DevToolCheckbox`:
```tsx
    <DevToolCheckbox
      // ...
      checked={tool.enabled && (tool.pluginAvailable ?? true)}
      // ...
    />
```
is synchronised with `selectedDeviceSession?.inspectorAvailability` used in `RenderOutlinesOverlay` component inside `useIsEnabled`, which is used to decide whether to render outlines or the screen or not.

```ts
function useIsEnabled() {
  const { selectedDeviceSession } = useProject();
  if (selectedDeviceSession?.status !== "running") {
    return false;
  }
  const isToolEnabled = selectedDeviceSession?.toolsState[RENDER_OUTLINES_PLUGIN_ID]?.enabled;
  const isInspectorAvailable = selectedDeviceSession?.inspectorAvailability;
  return isToolEnabled && isInspectorAvailable;
}
```

The `useIsEnabled` could possibly be renamed to something like `useShouldRenderOutlines` to avoid confusion or the plugin could be switched off when availability changes to false, but that would require more logic refactoring.

jwajgelt:
image

We may want to change the message while the app is inactive.

packages/vscode-extension/src/webview/components/shared/IconButton.tsx
pFornagiel: Right, refactored in ff83eb93

`Tooltip.tsx`:
```tsx
<RadixTooltip.Trigger asChild>
  {/* Some elements may have `pointer-events: none` set in css when disabled,
  so if we wish to show label, we've got to have an element 
  (in this case <span>), which can react to pointer events. If
  one wishes to disable the Tooltip when `pointer-events: none` is set
  they should declare this explicitly setting the `disabled` prop on the
  Tooltip component. */}
  <span>{children}</span>
</RadixTooltip.Trigger>
```

pFornagiel:
> image
>
> We may want to change the message while the app is inactive.

Introduced in 0588c67f with some notes:

The message handling required redesigning the communication protocol between `extension` and `lib` - instead of boolean,  a string message (interpreted as Enum in the extension) is sent to additionally allow for differentiating between the causes of unavailability.

```ts
export enum InspectorAvailabilityStatus {
  Available = "available",
  UnavailableEdgeToEdge = "unavailableEdgeToEdge",
  UnavailableInactive = "unavailableInactive",
}
```

In order to make this possible, information flow between `tools.ts` and `RenderOutlinesPlugin` class had to be revised, because `RenderOutlinesPlugin` has to notify the `tools.ts` that availability changed and that the `pluginUnavailableTooltip` sent to the frontend should be updated. Current solution for this is adding an optional callback `onAvailabilityChange` to the class constructor, which notifies the tools whenever there is a change. 

`tools.ts`:
```ts
const handleRenderOutlinesAvailabilityChange = () => {
      this.handleStateChange();
    };

this.plugins.set(REDUX_PLUGIN_ID, new ReduxDevtoolsPlugin(inspectorBridge));
this.plugins.set(NETWORK_PLUGIN_ID, new NetworkPlugin(inspectorBridge));
this.plugins.set(
  RENDER_OUTLINES_PLUGIN_ID,
  new RenderOutlinesPlugin(inspectorBridge, handleRenderOutlinesAvailabilityChange)
);
```

`render-outlines-plugin.ts`:
```ts
constructor(
    private inspectorBridge: RadonInspectorBridge,
    private onAvailabilityChange?: () => void
  ) {
    // ...
    this.devtoolsListeners.push(
      this.inspectorBridge.onEvent(
        "inspectorAvailabilityChanged",
        (inspectorAvailability: InspectorAvailabilityStatus) => {
          this.inspectorAvailability = inspectorAvailability;

          if (inspectorAvailability === InspectorAvailabilityStatus.Available) {
            this.activateDueToAvailabilityChange();
          } else {
            this.deactivateDueToAvailabilityChange();
          }

          this.onAvailabilityChange?.();
        }
      )
    );
  }
```


Additionally, thanks to this remaking of flow, the plugin has been made to actually disable, when inspector is not available, instead of just making the option grayed-out on the frontend.

If this commit is considered to be too much logic and information-flow changing, we may revert this change and leave the tooltips as they are.

packages/vscode-extension/src/plugins/render-outlines/render-outlines-plugin.ts
```diff
@@ -9,15 +9,28 @@ import {
...
   private devtoolsListeners: Disposable[] = [];
-  private buttonDisabled = false;
+  private inspectorAvailability: InspectorAvailabilityStatus =
+    InspectorAvailabilityStatus.Available;
+  private wasPreviouslyEnabled = false;
 
-  constructor(private inspectorBridge: RadonInspectorBridge) {
+  constructor(
+    private inspectorBridge: RadonInspectorBridge,
+    private onAvailabilityChange?: () => void
```
jwajgelt:
```suggestion
    private onAvailabilityChange: () => void
```
I don't think it needs to be optional.

packages/vscode-extension/src/webview/components/ToolsDropdown.tsx
```diff
@@ -67,10 +67,10 @@ function ToolsList({
...
       key={key}
       label={tool.label}
-      checked={tool.enabled}
-      panelAvailable={tool.panelAvailable}
-      disabled={tool.pluginButtonDisabled}
-      disabledTooltipLabel={tool.disabledPluginTooltipLabel}
+      checked={tool.enabled && (tool.pluginAvailable ?? true)}
+      isPanelTool={tool.isPanelTool}
+      enabled={tool.pluginAvailable ?? true}
+      pluginUnavailableTooltip={tool.pluginUnavailableTooltip ?? ""}
```
jwajgelt: Why not 
```suggestion
      pluginUnavailableTooltip={tool.pluginUnavailableTooltip}
```
The `pluginUnavailableTooltip` prop is optional, this should work, right?

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -365,7 +378,7 @@ function PreviewView() {
           tooltip={{
             label: inspectorAvailable
               ? "Select an element to inspect it"
-              : INSPECT_UNAVAILABLE_TOOLTIP_LABEL,
+              : INSPECTOR_AVAILABILITY_MESSAGES[inspectorAvailabilityStatus],
```
jwajgelt: NIT:
```suggestion
label: INSPECTOR_AVAILABILITY_MESSAGES[inspectorAvailabilityStatus],
```
And set `INSPECTOR_AVAILABILITY_MESSAGES[Available]` to `"Select an element to inspect it"`

packages/vscode-extension/src/common/Project.ts
jwajgelt: Well, `pluginAvailable` should probably be required, the comment was just for the label. Feel free to keep it as is though.

pFornagiel: Made required in 97179c88

packages/vscode-extension/src/webview/components/ToolsDropdown.tsx
pFornagiel: Changed in 2ee26b9e, passing `undefined` to tooltip results in the same behaviour anyways

packages/vscode-extension/src/webview/views/PreviewView.tsx
pFornagiel: Changed in 1c1f5700

packages/vscode-extension/src/plugins/render-outlines/render-outlines-plugin.ts
pFornagiel: Changed in 0666d257, as well as added a comment explaining one of caveats, which are not immediately obvious:

```ts
const handleRenderOutlinesAvailabilityChange = () => {
      // Note that this enables and disables the tool internally,
      // without updating the tool enabled state (calling updateToolEnabledState)
      // in order to avoid telemetry events spam and allow for storing the previous state
      // without refactoring the entire ToolsManager logic.

      // On the frontend, to determine whether tool is actually enabled, we have
      // to check the inspector availability (this.pluginAvailable) in cases where
      // it is unavailable, because getToolState() will return the active state from before
      // availability change.
      this.handleStateChange();
    };
```

Additionally, introduce additional guard in `RenderOutlinesOverlay.tsx`, to make sure everything works fine:

```ts
function useIsEnabled() {
  const { selectedDeviceSession } = useProject();
  if (selectedDeviceSession?.status !== "running") {
    return false;
  }
  const isToolEnabled = selectedDeviceSession?.toolsState[RENDER_OUTLINES_PLUGIN_ID]?.enabled;
  const isInspectorAvailable =
    selectedDeviceSession?.toolsState[RENDER_OUTLINES_PLUGIN_ID]?.pluginAvailable;
  return isToolEnabled && isInspectorAvailable;
}
```

pFornagiel:
Merged main, as well as added a comment explaining one of caveats, which are not immediately obvious:

```ts
const handleRenderOutlinesAvailabilityChange = () => {
      // Note that this enables and disables the tool internally,
      // without updating the tool enabled state (calling updateToolEnabledState)
      // in order to avoid telemetry events spam and allow for storing the previous state
      // without refactoring the entire ToolsManager logic.

      // On the frontend, to determine whether tool is actually enabled, we have
      // to check the inspector availability (this.pluginAvailable) in cases where
      // it is unavailable, because getToolState() will return the active state from before
      // availability change.
      this.handleStateChange();
    };
```

Additionally, introduce additional guard in `RenderOutlinesOverlay.tsx`, to make sure everything works fine:

```ts
function useIsEnabled() {
  const { selectedDeviceSession } = useProject();
  if (selectedDeviceSession?.status !== "running") {
    return false;
  }
  const isToolEnabled = selectedDeviceSession?.toolsState[RENDER_OUTLINES_PLUGIN_ID]?.enabled;
  const isInspectorAvailable =
    selectedDeviceSession?.toolsState[RENDER_OUTLINES_PLUGIN_ID]?.pluginAvailable;
  return isToolEnabled && isInspectorAvailable;
}
```

jwajgelt (approved):

PR #1404, opened 2025-08-05
refactor: devices state managment

filip131311:
This PR is another one in the series that move us to the new state management system and single unified backend API. 

In this PR we: 
- remove `DevicesProvider` 
  -  move all the state it was responsible for to the new `store$` 
  -  remove `DeviceSessionsManagerInterface` and move its methods to the `ProjectInterface`
  -  remove `DevicesManagerInterface` and move its methods to the `ProjectInterface`
- Adjust the way updates are merged to the `store$` to allow for removing elements from stored arrays 
- Generally clean up the `project` class to make its growth sustainable we: 
  - devide it in clear sections and mark them using `#region` comments 
  - remove helper functions into utilis 
- clean up the `state.ts` file to make its growth sustainable
- remove `DeviceManage`, `DeviceSessionsManager` and `WorkspaceConfig` callable objects from the webview controller callable object list 

### How Has This Been Tested: 

- add a device 
- rename a device 
- remove a device
- change devices theme (to test something still in old state system)
- change rotation 

This is internal change


packages/vscode-extension/src/project/project.ts
```diff
@@ -213,130 +316,213 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
-    getTelemetryReporter().sendTelemetryEvent("recording:start-recording", {
-      platform: this.selectedDeviceSessionState?.deviceInfo.platform,
-    });
-    if (!this.deviceSession) {
-      throw new Error("No device session available");
-    }
-    this.deviceSession.startRecording();
+  async focusOutput(channel: Output): Promise<void> {
+    this.outputChannelRegistry.getOrCreateOutputChannel(channel).show();
+  }
```
jwajgelt: This shouldn't be in the "Debugger" region.

packages/vscode-extension/src/utilities/merge.ts
```diff
@@ -0,0 +1,46 @@
...
+  oldNode: T,
+  newNode: RecursivePartial<T>
+): T {
+  const result: RecursivePartial<T> = {};
+  let wasChanged = false;
+
+  const allKeys = [...Object.keys(oldNode), ...Object.keys(newNode)] as (keyof T)[];
+
+  for (const key of allKeys) {
+    if (newNode[key] === undefined || newNode[key] === null) {
```
jwajgelt: I think we shouldn't special-case `null`, it can be useful to represent "empty" states.

packages/vscode-extension/src/utilities/mergeAndCalculateChanges.ts
```diff
@@ -45,7 +44,7 @@ export function mergeAndCalculateChanges<T extends object>(
   }
 
   if (wasChanged) {
-    return [result, changes] as [T, Partial<T>];
+    return [result, changes] as [T, RecursivePartial<T>];
```
jwajgelt, line 47: Why is this cast needed?

packages/vscode-extension/src/webview/components/DeviceSelect.tsx
```diff
@@ -90,20 +89,21 @@ function DeviceSelect() {
   const stopPreviousDevices = use$(store$.workspaceConfiguration.stopPreviousDevices);
 
   const { selectedDeviceSession, projectState, project } = useProject();
-  const { devices, deviceSessionsManager } = useDevices();
+
+  const devices = use$(store$.devicesState.devices);
```
```suggestion
  const devices = use$(store$.devicesState.devices) ?? [];
```
saves you on null checks elsewhere.

packages/vscode-extension/src/webview/hooks/useFatalErrorAlert.tsx
```diff
@@ -90,12 +88,13 @@ const noErrorAlert = {
 
 function createBuildErrorAlert(
   buildErrorDescriptor: BuildErrorDescriptor,
-  deviceSessionsManager: DeviceSessionsManagerInterface,
   hasSelectedScheme: boolean,
   xcodeSchemes: string[]
 ) {
+  const { project } = useProject();
```
jwajgelt: You cannot use hooks on non-hooks, pass either `project` or `onReload` to this function.

packages/vscode-extension/src/webview/views/CreateDeviceView.tsx
```diff
@@ -63,31 +65,36 @@ function CreateDeviceView({ onCreate, onCancel }: CreateDeviceViewProps) {
...
-  const { iOSRuntimes, androidImages, deviceManager, reload } = useDevices();
 
-  useEffect(() => {
-    reload();
-  }, []);
+  const { project } = useProject();
+
+  const store$ = useStore();
+  const iOSRuntimes = use$(store$.devicesState.iOSRuntimes);
+  const androidImages = use$(store$.devicesState.androidImages);
```
jwajgelt:
```suggestion
  const iOSRuntimes = use$(store$.devicesState.iOSRuntimes) ?? [];
  const androidImages = use$(store$.devicesState.androidImages) ?? [];
```
Why does it return `null` anyway? Does that happen during loading? Maybe we should use `[]` as the initial value?

jwajgelt (changes requested):
Overall looks very good, left a couple smaller comments need to be addressed.

packages/vscode-extension/src/utilities/mergeAndCalculateChanges.ts
filip131311: because, `result` in not a full `T` util the end of the algorithm when it is

packages/vscode-extension/src/utilities/merge.ts
filip131311: I thought we decided for null to strictly mean "not initialized"

packages/vscode-extension/src/utilities/merge.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @param {T} oldNode - The original state object.
+ * @param {RecursivePartial<T>} newNode - A partial state object containing updates.
+ *
+ * @returns {T} The new state object post-merge.
+ */
+export function merge<T extends { [P in keyof T]: T[P] }>(
+  oldNode: T,
+  newNode: RecursivePartial<T>
+): T {
```
jwajgelt: This should just call `mergeAnCalculateChanges` and return the `result` part.

Also, you can move both to the same module, we don't need to have 1 function per file.

PR #1406, opened 2025-08-05
feat: make log indicator more compact

filip131311:
This is a proposal of an alternative approach to indicating that the application generated new logs: 

this will solve two issues: 
- if the number gets to large it hides under the next button
- if the user opens the debug console directly the indicator would not reset and its number would diverge from the actual number of logs generated since last viewed 

| before | after |
|--------|-------|
| Before Image | After Image |

### DIFF
--- a/packages/vscode-extension/src/webview/components/shared/IconButton.css
+++ b/packages/vscode-extension/src/webview/components/shared/IconButton.css
@@ -96,11 +96,12 @@
 }
 
 .icon-button-indicator {
+  box-sizing: border-box;
   position: absolute;
-  top: 2px;
-  right: 2px;
-  width: 8px;
-  height: 8px;
+  top: 1px;
+  right: 1px;
+  width: 10px;
+  height: 10px;
   background-color: var(--swm-button-counter-background, #ff5555);
   border-radius: 50%;
   border: 1px solid var(--swm-button-counter-border, transparent);


packages/vscode-extension/src/webview/components/shared/IconButton.css
```diff
@@ -94,3 +94,14 @@
...
+.icon-button-indicator {
+  position: absolute;
+  top: 2px;
+  right: 2px;
+  width: 8px;
+  height: 8px;
+  background-color: var(--swm-button-counter-background, #ff5555);
+  border-radius: 50%;
+  border: 1px solid var(--swm-button-counter-border, transparent);
+}
```
kacperkapusciak, line 114:
```suggestion
  transform: scale(0);
  transition: transform 175ms ease-in-out;
}

.icon-button-indicator.visible {
  transform: scale(1);
}
```
i love microinteractions


https://github.com/user-attachments/assets/06becc13-a26e-413f-a130-d44fd2a7cfaf

packages/vscode-extension/src/webview/components/shared/IconButton.tsx
```diff
@@ -61,11 +63,14 @@ const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>((props,
...
       {children}
-      {counter !== null && (
+      {counterMode === "full" && counter !== null && (
         <span className={classnames("icon-button-counter", showCounter && "visible")}>
           {displayCounter}
         </span>
       )}
+      {counterMode === "compact" && counter !== null && showCounter && (
+        <span className="icon-button-indicator" />
+      )}
```
kacperkapusciak:
```suggestion
      {counterMode === "compact" && counter !== null && (
        <span className={classnames("icon-button-indicator", showCounter && "visible")} />
      )}
```
i love microinteractions part 2

kmagiera:
add test plan and include testing this on high contrast themes. On high-contrast themes the current indicator is "wider" than on normal themes due to border that is otherwise transparent. In my PR #1408 I use `box-sizing: border-box;` style.

PR #1408, opened 2025-08-06
Use new badge indicator instead of counter for logs button

kmagiera:
This PR replaces the log counter taht we show over the logs button with a small round "new content badge"

There are two main reasons we want to get rid of the counter:
1) it is never accurate as people may also read the logs by just opening the console and we never know about that
2) when the numbers are big, they obstruct other buttons on the tool bar. We could just use a "9+" text but then it still may not be accurate as people may have viewed some of the logs already.

Light theme:
image

Dark theme:
image

Dark high contrast:
image

Light high contrast:
image

### How Has This Been Tested: 
1. Run the extension, generate some logs and test on different themes
2. Click the button to see the indicator gone

### DIFF
--- a/packages/vscode-extension/src/webview/components/shared/IconButton.css
+++ b/packages/vscode-extension/src/webview/components/shared/IconButton.css
@@ -72,25 +72,14 @@
   background-color: var(--swm-button-secondary-hover);
 }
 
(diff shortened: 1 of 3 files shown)


kmagiera:
closing in favor of #1406

Issue #1280, opened 2025-06-27
Simulator stuck on "Waiting for app to load"


danielecalzone:
**Description**
The simulator fails to load on both iOS and Android. It gets stuck at “Waiting for app to load”, and does not proceed further. The console shows the error log included below.

This issue occurs on the latest version 1.8.0, but everything works correctly in version 1.7.1.

**Environment**

- Version of currently used Radon IDE (Extensions tab -> click on Radon IDE extension -> version is located next to the extension name): 1.8.0
- Version of React Native/Expo used in the project: React Native 0.76.9 - Expo 52.0.47
- Any other versions of tools related to the problem: /

`2025-06-27 14:35:06.702 [info] Metro {"type":"client_log","level":"log","data":["__RNIDE_INTERNAL","radon-ide runtime loaded"],"mode":"BRIDGE"}
2025-06-27 14:35:06.703 [error] Error: Radon inspector bridge agent is not installed, js engine: hermes
2025-06-27 14:35:06.703 [info] Metro {"type":"client_log","level":"log","data":["Running \"main\" with {\"rootTag\":1,\"initialProps\":{\"concurrentRoot\":false}}","http://-index.bundle//&platform=ios&dev=true&hot=false&transform.engine=hermes&transform.bytecode=1&transform.routerRoot=app&unstable_transformProfile=hermes-stable","44435","22"],"mode":"BRIDGE"}
2025-06-27 14:35:06.703 [error] Error: Radon inspector bridge agent is not installed, js engine: hermes
2025-06-27 14:35:06.703 [error] TypeError: Cannot read property 'AppWrapper' of undefined, js engine: hermes
2025-06-27 14:35:06.704 [info] Metro {"type":"client_log","level":"info","data":["\n\u001b[7m \u001b[1m💡 JavaScript logs will be removed from Metro in React Native 0.77!\u001b[22m Please use React Native DevTools as your default tool. Tip: Type \u001b[1mj\u001b[22m in the terminal to open (requires Google Chrome or Microsoft Edge).\u001b[27m\n"]}
2025-06-27 14:35:06.704 [error] Error: Radon inspector bridge agent is not installed, js engine: hermes
2025-06-27 14:35:06.704 [error] TypeError: Cannot read property 'AppWrapper' of undefined, js engine: hermes`

jwajgelt:
Hi @danielecalzone @jackmew 
Could you check if clearing the metro cache helps with the problem?
You can do it by selecting the

[...]

PR #1410, opened 2025-08-07
fix: bundle load failing due to agent initalising late

jwajgelt:
Fixes runtime errors during JS bundle loading when the `inspectorBridge` is accessed before the DevTools Radon Agent is set up.

Currently, we load the Devtools Radon Agent in `wrapper.js`, which is only loaded when the application first starts and React Native loads the component registered through `AppRegistry.setWrapperComponentProvider`.
However, some of our plugins import `inspectorBridge` directly, but can be loaded before the application runs, during the initial bundle evaluation:
- `react-query-devtools.js` is loaded whenever the `react-query` module is evaluated.
- `expo_dev_plugins.js` is loaded whenever the `redux-devtools-expo-dev-plugin` module is evaluated.
Which causes the `inspectorBridge` to throw an error when checking if the `agent` is already set up, causing the whole bundle execution to fail.

This PR fixes that by assuming that, when the `inspectorBridge` loads and no `agent` is already registered, we're not in Connect mode and we can simply import the Devtools implementation (this should not break anything since, if we hit that branch in connect mode before connect sets up the CDP agent implementation, we'd throw an error and break something anyway).

Related #1280

### How Has This Been Tested: 
- setup an app which loads `react-query` before `react-native`
- check if the app gets stuck on "Waiting for app to load..."

### DIFF
--- a/packages/vscode-extension/lib/inspector_bridge.js
+++ b/packages/vscode-extension/lib/inspector_bridge.js
@@ -1,7 +1,8 @@
-const agent = globalThis.__radon_agent;
+let agent = globalThis.__radon_agent;
 
 if (!agent) {
-  throw new Error("Radon inspector bridge agent is not installed");
+  // if the agent was not loaded yet, we do it here -- hopefully connect had the time to register its bridge already
+  agent = require("./react_devtools_agent");
 }
 
 const messageListeners = [];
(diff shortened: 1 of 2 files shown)


filip131311 (approved):
Could we encapsulate `inspector_bridge.js` in some kind of module that is able to get an argument informing it if its required in connect mode or regular one? This way we wouldn't have to assume as much. 

THe fix looks perfectly correct tho

PR #1411, opened 2025-08-07
feat: Improve test coverage. GUI interactions and device adding/removing

KeyJayY:
### Description

This PR adds GUI interaction tests for the main application window.  
Specifically, it includes:

- Tests for clicking buttons in the main GUI
- Tests for adding and removing devices

The motivation behind this change is to improve test coverage and ensure that key user actions work as expected.

---

### How Has This Been Tested

- Verified correct behavior of:
  - Main window button clicks
  - Device addition
  - Device removal

To reproduce:

1. Make sure the virtual machine environment is properly set up
2. Run the tests using:
   `npm run run-test-on-VM`
   
Test environment:
  Virtual Machine available via `npm run set-up-VM-environment`

---
  
### How has this change been documented
Not documented yet.


packages/vscode-extension-tester/scripts/runTests.sh
```diff
@@ -53,19 +53,38 @@ done
 
 echo "VM is ready at $VM_IP"
 
-echo "Copying '$LOCAL_PROJECT_PATH' to VM..."
-sshpass -p "$VM_PASSWORD" scp -r "$LOCAL_PROJECT_PATH" "$VM_USER@$VM_IP:/Users/$VM_USER/" || {
-    echo "Failed to copy project files."
-    exit 1
-}
+echo "Running test commands on VM..."
```
filip131311: this is not very informative, it would be better to write something like "cleaning up the vm environment" or "preparing VM environment"

packages/vscode-extension-tester/scripts/runTests.sh
```diff
@@ -53,19 +53,38 @@ done
...
+for item in * .*; do
+    [[ "$item" == "." || "$item" == ".." ]] && continue
+    [[ "$item" == "node_modules" || "$item" == "scripts" || "$item" == ".gitignore" ]] && continue
 
+    echo "Copying: $item"
+    scp -i ./scripts/id_vm_mac -r "$item" "$VM_USER@$VM_IP:/Users/$VM_USER/$REMOTE_PATH/"
+done
+
+    
 echo "Running test commands on VM..."
```
filip131311: this would be better named " installing test dependencies on VM"

filip131311: *** now I see that setup-run-tests also runs the tests so it might be useful to include it in the description

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -324,7 +325,9 @@ function PreviewView() {
             <span slot="start" className="codicon codicon-debug-console" />
           </IconButton>
           <SettingsDropdown project={project} isDeviceRunning={isRunning || radonConnectConnected}>
-            <IconButton tooltip={{ label: "Settings", type: "primary" }}>
+            <IconButton
+              tooltip={{ label: "Settings", type: "primary" }}
+              dataTest="radon-settings-button">
```
filip131311: this seems not specific enough as a name. Future maintainer of this code would not know what it means. I would name it more something like `radon-uotp-bar-settings-droppdown-trigger` or maybe something you feel is more suitable but in a way that allows for easy identification.

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -276,7 +277,7 @@ function PreviewView() {
             onClick={stopProfilingReact}
           />
           <ToolsDropdown disabled={!debuggerToolsButtonsActive}>
-            <IconButton tooltip={{ label: "Tools", type: "primary" }}>
+            <IconButton tooltip={{ label: "Tools", type: "primary" }} dataTest="radon-tools-button">
```
filip131311: same as `radon-settings-button`

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -40,6 +40,7 @@ function ActivateLicenseButton() {
   return (
     <Button
       className="activate-license-button"
+      dataTest="activate-license-button"
```
filip131311: same as `radon-settings-button`

packages/vscode-extension/src/webview/views/ManageDevicesView.tsx
```diff
@@ -220,7 +229,11 @@ function ManageDevicesView() {
           {androidDevices.map(renderRow)}
         </>
       )}
-      <Button autoFocus className="create-button" onClick={() => setCreateDeviceViewOpen(true)}>
+      <Button
+        autoFocus
+        className="create-button"
+        dataTest="create-new-device-button"
```
filip131311: smae as `radon-settings-button`

packages/vscode-extension/src/webview/views/ManageDevicesView.tsx
```diff
@@ -207,7 +216,7 @@ function ManageDevicesView() {
   }
 
   return (
-    <div className="manage-devices-container">
+    <div className="manage-devices-container" data-test="manage-devices-view">
```
filip131311, line 214: okay it seems that there is quite a lot of those tags I would go through each of them and check if it can be ambiguous.

filip131311: sorry for going from last to first

packages/vscode-extension/src/webview/views/ManageDevicesView.tsx
```diff
@@ -24,6 +24,7 @@ interface DeviceRowProps {
   onDeviceDelete: (device: DeviceInfo) => void;
   isSelected: boolean;
   isRunning: boolean;
+  dataTest?: string;
```
filip131311: this will be used multiple times so it might be convenient to create a generic type `propsWithDataTest`

packages/vscode-extension-tester/ui-tests/01-smoke.test.js
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  it("should open Radon IDE webview for a specific project", async function () {
+    const { driver } = get();
+    await openRadonIDEPanel(driver);
+
+    const title = await driver.getTitle();
+    assert.equal(
+      title,
+      "Radon IDE — " + texts.pageTitle,
```
filip131311, line 47: this seems not extendable, as it hard codes the expected title without any relation to the project that is being opened, fixing it might be out of scope, but we should keep it in mind if you were to add more tests like that

packages/vscode-extension-tester/ui-tests/01-smoke.test.js
```diff
@@ -0,0 +1,63 @@
...
+      `Page title should be: ${texts.pageTitle}`
+    );
+
+    const approot = await driver.findElement(
+      By.css('[data-test="approot-select-value"]')
+    );
+    await waitForElement(driver, approot);
+
+    const text = await approot.getText();
```
filip131311, line 56: again, either this test should set up the whole environment in such a way that we know what application is being tested here, or take is as an argument with the expected project name (we need a relation between  `expectedProjectName` and "project we run" (possibly in a follow up )

packages/vscode-extension-tester/ui-tests/02-GUIButtons.test.js
```diff
@@ -0,0 +1,73 @@
...
+  const get = sharedTestLifecycle();
+  let driver;
+
+  beforeEach(async function () {
+    ({ driver } = get());
+    await openRadonIDEPanel(driver);
+  });
+
+  it("Should open device settings window", async function () {
+    await findAndClickElementByTag(driver, "device-settings-button");
```
filip131311: I believe those names are not informative enough I wrote more about it on the extension side.

packages/vscode-extension-tester/ui-tests/interactions.js
```diff
@@ -0,0 +1,131 @@
+import {
+  findAndWaitForElement,
+  findAndClickElementByTag,
+  waitUntilElementGone,
+} from "../utils/helpers.js";
+import { By } from "vscode-extension-tester";
+
+export async function openDeviceCreationModal(driver) {
```
filip131311: There is a function in vs code that allows you to declare regions in your file by placing 
// #region <REGIONS_NAME>
... code in the region
// #endregion <REGIONS_NAME>
(see more in project.ts in the extension) 

I think it might improve this file if we added regions for each test suite

packages/vscode-extension-tester/ui-tests/01-smoke.test.js
KeyJayY: I partly solved it in this PR https://github.com/software-mansion/radon-ide/pull/1470, it's not hard coded when running on GitHub Actions runner. I'll solve this problem for running locally and on local VM in follow up PR.

packages/vscode-extension-tester/.gitignore
```diff
@@ -3,3 +3,4 @@ data/radon-ide-test-apps
 data/*.vsix
 .vscode
 config.json
+data/vscode-extensions
```
filip131311: please add end of line char to this file

packages/vscode-extension/src/common/types.ts
```diff
@@ -0,0 +1 @@
+export type PropsWithDataTest<P = {}> = P & { dataTest?: string };
```
filip131311: lint is shouting here to use `unknown` instead of `{}` 

you can read more about why [here](https://typescript-eslint.io/rules/no-empty-object-type/)

filip131311 (approved):
You have a few conflicts to resolve and I left to more small comments but is looks good and tests are passing correctly good work

PR #1412, opened 2025-08-07
feat: add ping animation to a log indicator

kacperkapusciak:
This PR makes the log indicator indicate whenever there's a new log with a delicate ping animation. Related to https://github.com/software-mansion/radon-ide/pull/1406 

light theme

https://github.com/user-attachments/assets/1085ed95-21bb-43d7-a657-482dd601f835

dark theme

https://github.com/user-attachments/assets/8ae618ab-0233-4282-9f6d-ae1d01e50fdf

high-contrast

https://github.com/user-attachments/assets/49c0eac6-43f7-4abf-8b65-a33e5ab6827c

### DIFF
--- a/packages/vscode-extension/src/webview/components/shared/IconButton.tsx
+++ b/packages/vscode-extension/src/webview/components/shared/IconButton.tsx
@@ -3,6 +3,8 @@ import classnames from "classnames";
 import "./IconButton.css";
 import Tooltip from "./Tooltip";
 
+const PING_DURATION = 300; // ms
+
 export interface IconButtonProps {
   onClick?: React.MouseEventHandler<HTMLButtonElement>;
   children: React.ReactNode;
@@ -38,12 +40,20 @@ const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>((props, 
     ...rest
   } = props;
   const [displayCounter, setDisplayCounter] = useState(counter);
+  const [shouldPing, setShouldPing] = useState(false);
 
   useEffect(() => {
     if (counter !== 0) {
       setDisplayCounter(counter);
     }
-  }, [counter]);
+    if (counterMode === "compact" && counter !== 0) {
+      setShouldPing(true);
+      const timer = setTimeout(() => {
+        setShouldPing(false);
+      }, PING_DURATION);
+      return () => clearTimeout(timer);
+    }
+  }, [counter, counterMode]);
 
   const showCounter = Boolean(counter);
   const button = (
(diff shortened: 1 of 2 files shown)


jwajgelt:
What happens if you spam the logs every, say, 10ms?

packages/vscode-extension/src/webview/components/shared/IconButton.tsx
```diff
@@ -38,12 +40,20 @@ const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>((props,
...
       setDisplayCounter(counter);
     }
-  }, [counter]);
+    if (counterMode === "compact" && counter !== 0) {
+      setShouldPing(true);
+      const timer = setTimeout(() => {
+        setShouldPing(false);
+      }, PING_DURATION);
+      return () => clearTimeout(timer);
+    }
```
jwajgelt: Doesn't that meant that, if we set the `counter` prop to `0`, we'll cancel the timeout and get stuck in the `shouldPing` state?

packages/vscode-extension/src/webview/components/shared/IconButton.tsx
```diff
@@ -38,12 +40,20 @@ const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>((props,
     ...rest
   } = props;
   const [displayCounter, setDisplayCounter] = useState(counter);
```
kmagiera: why is this a state? `const displayCounter = counter !== 0`

packages/vscode-extension/src/webview/components/shared/IconButton.tsx
kmagiera: not stuck, when count goes to 1 it will go back to normal. Coincidentally this is actually fine, because when count goes from 0 to 1 the dot will show again, but it just wouldn't bounce.

I agree it would be cleaner to handle this properly, so we should be setting ping=true when counter != 0 and set it to `false` otherwise.

kacperkapusciak: fixed

packages/vscode-extension/src/webview/components/shared/IconButton.tsx
kacperkapusciak: seems to be useless

kacperkapusciak:
@jwajgelt when one spams with console logs every 20ms the indicator doesn't freak out, so that's a good thing 

https://github.com/user-attachments/assets/7836a4ae-c89f-4392-aa38-bb8252ed03d9

kmagiera:
it wouldn've been nicer if we get some more bouncing in that case though. The current code would just cancel the timer and start new one indefinitely extending the time where the bouncing is stopped. Instead, we could skip cancelling the timeout when getting new counter while bouncing is going on and just let it reset in the originally scheduled time

Issue #1400, opened 2025-08-05
Android disconnects after backgrounded


jwajgelt:
**Description**
Android application disconnects from HMR and DevTools after being backgrounded for a couple seconds and doesn't reconnect after moving to the foreground.
Same thing happens with RN running on an emulator, connected to a standalone Metro server and React Native Devtools

**Environment**

- Version of currently used Radon IDE: 1.9.1
- Version of React Native/Expo used in the project: React Native 0.81

jwajgelt:
This should now work after #1550 when using the CDP devtools implementation.

PR #1413, opened 2025-08-07
feat: show alert when DevTools disconnect with information that tools may not work

jwajgelt:
When the Android application is moved to the background, the Websockets it opened all close. This causes the React DevTools connection to close, breaking Radon's Inspector Bridge which is needed for many tools to work.
This PR shows an alert when that happens to inform the user that features may not work correctly and prompt them to restart the app.
Related #1400

image

### How Has This Been Tested: 
- open an app in Android and go to the launcher / switch application -- the alert should appear after a few seconds
- close the app fully on Android/iOS -- the alert should appear after the app is killed
- open the app again, through the phone's launcher -- the alert should disappear and tools should work again
- trigger any restart action while the alert is shown -- the alert should disappear

### DIFF
--- a/packages/vscode-extension/src/webview/hooks/useApplicationDisconnectedAlert.tsx
+++ b/packages/vscode-extension/src/webview/hooks/useApplicationDisconnectedAlert.tsx
@@ -7,43 +7,39 @@ const alertId = "devtools-disconnected-alert";
 
 export function useApplicationDisconnectedAlert(shouldShow: boolean) {
   const { project } = useProject();
-  const { openAlert, closeAlert, isOpen } = useAlert();
+  const { openAlert, closeAlert } = useAlert();
 
   const close = useCallback(() => {
-    if (isOpen(alertId)) {
-      closeAlert(alertId);
-    }
-  }, [closeAlert, isOpen]);
+    closeAlert(alertId);
+  }, [closeAlert]);
 
   const open = useCallback(() => {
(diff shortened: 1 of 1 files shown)


filip131311 (changes requested):
https://github.com/user-attachments/assets/d8a2357b-a339-447a-9074-64370af9591c

It's undimisable

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -104,6 +106,10 @@ function Preview({
   const showDevicePreview =
     selectedDeviceSession?.previewURL && (showPreviewRequested || isRunning);
 
+  const isAppDisconnected =
+    isRunning && selectedDeviceSession.inspectorBridgeStatus === InspectorBridgeStatus.Disconnected;
+  useApplicationDisconnectedAlert(isAppDisconnected);
+
   useFatalErrorAlert(fatalErrorDescriptor);
```
filip131311, line 113: right now we don't have priority defined for fatalErrors or at least i could not find them is it possble for you new alert to hide the fatal error one?

jwajgelt: I think 0 is the default, but also no, since this will only appear when "running"

filip131311: 👍

PR #1414, opened 2025-08-07
fix: node version check

filip131311:
This PR changes `getMinimumSupportedNodeVersion` logic to return the most restrictive version instead of the first found one. This Solves an issue with projects that have updated rn version but not the engine requirements in their package.json. 

### How Has This Been Tested: 

add a brak point to the `getMinimumSupportedNodeVersion` method run in tin the `react-native-81` test app and check if the return value is 22.14 instead of 18

internal change

### DIFF
--- a/packages/vscode-extension/src/utilities/getMinimumSupportedNodeVersion.ts
+++ b/packages/vscode-extension/src/utilities/getMinimumSupportedNodeVersion.ts
@@ -1,5 +1,8 @@
 import fs from "fs";
 import { Logger } from "../Logger";
+import semver from "semver";
+
+const DEFAULT_MINIMUM_NODE_VERSION = ">=18";
 
 /* 
 This function return a minimum supported node version for given appRoot assuming node_modules are installed,
@@ -8,8 +11,10 @@ of react native, so it should be always used after node_modules installation.
 */
 export function getMinimumSupportedNodeVersion(appRoot: string): string {
   const appPackageJson = require(`${appRoot}/package.json`);
+  let applicationMinimumNodeVersion;
+  let reactNativeMinimumNodeVersion;
   if (appPackageJson.engines && appPackageJson.engines.node) {
-    return appPackageJson.engines.node;
+    applicationMinimumNodeVersion = appPackageJson.engines.node;
   }
 
   if (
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/utilities/getMinimumSupportedNodeVersion.ts
```diff
@@ -8,8 +11,10 @@ of react native, so it should be always used after node_modules installation.
 */
 export function getMinimumSupportedNodeVersion(appRoot: string): string {
   const appPackageJson = require(`${appRoot}/package.json`);
```
jwajgelt: Can we change it to use `requireNoCache` while we're fixing it? (and elsewhere in this file as well)

packages/vscode-extension/src/utilities/getMinimumSupportedNodeVersion.ts
```diff
@@ -8,8 +11,10 @@ of react native, so it should be always used after node_modules installation.
...
+  let applicationMinimumNodeVersion;
```
jwajgelt: Isn't this only assigned once? Also, you should be able to use the nullable member access operator.

```suggestion
  const applicationMinimumNodeVersion = appPackageJson?.engines?.node;
```

packages/vscode-extension/src/utilities/getMinimumSupportedNodeVersion.ts
```diff
@@ -18,15 +23,33 @@ export function getMinimumSupportedNodeVersion(appRoot: string): string {
   ) {
     const reactNativePackageJson = require(`${appRoot}/node_modules/react-native/package.json`);
     if (reactNativePackageJson.engines && reactNativePackageJson.engines.node) {
-      return reactNativePackageJson.engines.node;
+      reactNativeMinimumNodeVersion = reactNativePackageJson.engines.node;
     }
```
jwajgelt: This feels like it could be a function.

packages/vscode-extension/src/utilities/getMinimumSupportedNodeVersion.ts
```diff
@@ -18,15 +23,33 @@ export function getMinimumSupportedNodeVersion(appRoot: string): string {
...
-    "[Minimum Supported node version] Utility function called before node modules were installed, the returned value might, be deprecated."
-  );
-  return ">=18";
+  const ranges = [
+    applicationMinimumNodeVersion,
+    reactNativeMinimumNodeVersion,
+    DEFAULT_MINIMUM_NODE_VERSION,
+  ].filter(Boolean);
+
+  if (ranges.length === 1) {
```
jwajgelt: This took me second to parse.
I think even something like
```suggestion
  if (!applicationMinimumNodeVersion && !reactNativeMinimumNodeVersion) {
```
communicates the intent better.

packages/vscode-extension/src/utilities/getMinimumSupportedNodeVersion.ts
```diff
@@ -18,15 +23,33 @@ export function getMinimumSupportedNodeVersion(appRoot: string): string {
...
+
+  return ranges.reduce((mostRestrictive, current) => {
+    const currentMinVersion = semver.minVersion(current);
+    const restrictiveMinVersion = semver.minVersion(mostRestrictive);
+
+    if (currentMinVersion && restrictiveMinVersion) {
+      return semver.gt(currentMinVersion, restrictiveMinVersion) ? current : mostRestrictive;
+    }
+    return current || mostRestrictive;
+  });
```
jwajgelt: Don't we want to return `semver.minVersion(/* whatever this evals to */)`?

jwajgelt: Eh, let's keep it as it was.

packages/vscode-extension/src/utilities/getMinimumSupportedNodeVersion.ts
jwajgelt: Although, since this is a check for a log that explicitely mentions the RN's minimal supported version, it might be better to just check `reactNativeMinimumNodeVersion` anyway, since the warning still applies in that case.

PR #1418, opened 2025-08-08
fix: state updates lost when frequently updated

jwajgelt:
- Fixes a bug introduced in #1404 which caused some updates to be lost when the extension sent frequent updates to the webview[^1].
- Skips calling `onSetState` when nothing changed in `StateManager`.
- Adds a random `useMemo` in `AppRootSelect` because why not.

[^1]: The bug was caused by changing the legend-state update mode to `"set"` and setting a merge of the current state in the store with the updates received in the listener which synchronizes state changes in the Extension with the Webview's state.
Because the `update` function is called asynchronously, if the listener is called multiple times before the update can be applied, we merge the changes with the stale state stored in the Legend State observable and set that, losing the intermediate updates.
The fix passes a callback to the `update` function which receives the value set at the time the assignment happens, not at the time the changes were received.
The `update` function is typed to expect the actual values, but the implementation _does_ accept the callback and handles it correctly.

### How Has This Been Tested: 
- set a breakpoint in `ide.ts` on `this.stateManager.setState({ applicationRoots });`
- open a workspace in Radon
- after stopping at the breakpoint and waiting a while, resume the debugger
- the application roots should set up correctly

### DIFF
--- a/packages/vscode-extension/src/project/StateManager.ts
+++ b/packages/vscode-extension/src/project/StateManager.ts
@@ -62,8 +62,10 @@ class RootStateManager<T extends object> extends StateManager<T> {
 
   setState(partialState: RecursivePartial<T>): void {
     const [newState, changes] = mergeAndCalculateChanges(this.state, partialState);
-    this.state = newState;
-    this.onSetStateEmitter.fire(changes);
+    if (this.state !== newState) {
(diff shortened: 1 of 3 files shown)


filip131311 (approved):
I trust you lets make it into a package 🚢

PR #1436, opened 2025-08-12
feat: run Expo Prebuild when configured

jwajgelt:
Allows configuring Expo Prebuild to run before building the application.

The prebuild step is skipped if the native directories already exist and the fingerprint didn't change after the last prebuild operation completed successfully.

image

### How Has This Been Tested: 
- open radon test app `expo-52-prebuild-with-plugins`
- delete the `ios` and `android` directories
- open the Launch Configuration editor and set the "Use Expo Prebuild" option to true
- restart the app
- check that prebuild run in the build logs
- check that prebuild doesn't run when rebuilding the app if the fingerprint didn't change

#1439

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -605,7 +605,7 @@
               },
               "usePrebuild": {
                 "type": "boolean",
-                "description": "When set, Radon will run Expo Prebuild before building the application."
+                "description": "When set, Radon will run Expo Prebuild before building the application. Only applies to applications using Expo CLI."
               }
             }
           }
@@ -827,7 +827,7 @@
               },
               "usePrebuild": {
                 "type": "boolean",
-                "description": "When set, Radon will run Expo Prebuild before building the application."
+                "description": "When set, Radon will run Expo Prebuild before building the application. Only applies to applications using Expo CLI."
               }
             }
           }
--- a/packages/vscode-extension/src/dependency/prebuild.ts
+++ b/packages/vscode-extension/src/dependency/prebuild.ts
@@ -1,5 +1,3 @@
-import path from "path";
-import fs from "fs";
 import assert from "assert";
 import { Disposable, OutputChannel } from "vscode";
 import { exec } from "../utilities/subprocess";
@@ -10,6 +8,9 @@ import { FingerprintProvider } from "../project/FingerprintProvider";
 import { lineReader } from "../utilities/subprocess";
 import { CancelError, CancelToken } from "../utilities/cancelToken";
 import { extensionContext } from "../utilities/extensionContext";
+import _ from "lodash";
+import crypto from "crypto";
+import { checkNativeDirectoryExists } from "../utilities/checkNativeDirectoryExists";
 
 class PrebuildProcess {
   private attachedCounter = 0;
@@ -37,8 +38,17 @@ class PrebuildProcess {
   }
 }
 
-function persistKey(platform: DevicePlatform, appRoot: string) {
-  const key = `prebuild:${platform}:${appRoot}`;
+function getEnvHash(env: Record<string, string>) {
+  const sortedEnv = _.sortBy(Object.entries(env), ([key, _value]) => key);
+  const envHash = crypto.createHash("md5");
+  sortedEnv.forEach(([k, v]) => {
+    envHash.update(`${k}=${v}`);
+  });
+  return envHash.digest("hex");
+}
+
+function getPersistKey(platform: DevicePlatform, appRoot: string, env?: Record<string, string>) {
+  const key = `prebuild:${platform}:${appRoot}:${getEnvHash(env ?? {})}`;
   return key;
 }
 
@@ -53,17 +63,6 @@ export class Prebuild implements Disposable {
     });
   }
 
-  private async nativeDirectoryExists(buildConfig: IOSLocalBuildConfig | AndroidLocalBuildConfig) {
-    const directoryName = buildConfig.platform === DevicePlatform.Android ? "android" : "ios";
-    const nativeDirectoryPath = path.join(buildConfig.appRoot, directoryName);
-    try {
-      const stat = await fs.promises.stat(nativeDirectoryPath);
-      return stat.isDirectory();
-    } catch {
-      return false;
-    }
-  }
-
   public async runPrebuildIfNeeded(
     buildConfig: IOSLocalBuildConfig | AndroidLocalBuildConfig,
     outputChannel: OutputChannel,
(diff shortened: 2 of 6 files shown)


packages/vscode-extension/package.json
```diff
@@ -602,6 +602,10 @@
                     "description": "Defaults to `true`, this should only be set to `false` for brownfield setups when React Native components aren't rendered immediately after the app launches."
                   }
                 }
+              },
+              "usePrebuild": {
+                "type": "boolean",
+                "description": "When set, Radon will run Expo Prebuild before building the application."
```
filip131311: This description does not provide a full context for this setting. It will run prebuild before build but only if it is necessery, and it will also affect if the build will run at all (instead of expo go) It will also affect only expo projects.

packages/vscode-extension/package.json
```diff
@@ -820,6 +824,10 @@
                     "description": "Defaults to `true`, this should only be set to `false` for brownfield setups when React Native components aren't rendered immediately after the app launches."
                   }
                 }
+              },
+              "usePrebuild": {
```
filip131311, line 828: same

packages/vscode-extension/src/dependency/prebuild.ts
```diff
@@ -0,0 +1,170 @@
...
+      }
+    });
+  }
+
+  public cancel() {
+    this.cancelToken.cancel();
+  }
+}
+
+function persistKey(platform: DevicePlatform, appRoot: string) {
```
filip131311: the name of this function seems to be a verb and I would expect it to persist Key 

can you make it into `getPersistKey` or `generatePersistKey`?

packages/vscode-extension/src/dependency/prebuild.ts
```diff
@@ -0,0 +1,170 @@
...
+  const key = `prebuild:${platform}:${appRoot}`;
```
filip131311: does `expo prebuild` command take env arguments? if so maybe we want to invalidate on env change as well?

packages/vscode-extension/src/dependency/prebuild.ts
```diff
@@ -0,0 +1,170 @@
...
+  return key;
+}
+
+export class Prebuild implements Disposable {
+  private prebuildProcessByPlatform = new Map<DevicePlatform, PrebuildProcess>();
+
+  constructor(private fingerprintProvider: FingerprintProvider) {}
+
+  dispose() {
+    this.prebuildProcessByPlatform.forEach((prebuildProcess) => {
```
filip131311, line 61: I lately started to do: 
```
this.disposables.push(new Disposable(()=>{
logic that does something 
})
```

in order to unify our approach to disposable classes. 

I am not attached to the Idea but would love to know your opinion?

packages/vscode-extension/src/dependency/prebuild.ts
```diff
@@ -0,0 +1,170 @@
...
+      prebuildProcess.cancel();
+    });
+  }
+
+  private async nativeDirectoryExists(buildConfig: IOSLocalBuildConfig | AndroidLocalBuildConfig) {
```
filip131311: don't we do that logic in other places when we detect if project is expo go for example?
It does not seem to depend on any class fields. Could it be a shared util?

packages/vscode-extension/src/dependency/prebuild.ts
```diff
@@ -0,0 +1,170 @@
...
+      ongoingPrebuild === undefined && lastSuccessfulFingerprint === currentFingerprint;
+    const isCurrentlyPrebuilding = ongoingPrebuild !== undefined;
+
+    const canSkipPrebuild =
+      !buildConfig.forceCleanBuild &&
+      nativeDirectoryExists &&
+      (hasAlreadyPrebuild || isCurrentlyPrebuilding);
+    if (canSkipPrebuild) {
+      if (isCurrentlyPrebuilding) {
+        ongoingPrebuild.attach(cancelToken);
```
filip131311, line 90: scenario to consider: 

- ongoing prebuild exists 
- then another proces wants to run prebuild and attaches here so attach count is at 2 
- then another proces wants a prebuild but it requires a fresh prebuild for some reason 
- this causes `ongoingPrebuild?.cancel();` to be run and it does not stop the prebuild process because attach count was at 2. 
- is it intended? Why?

packages/vscode-extension/src/dependency/prebuild.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  private async runPrebuild(
+    buildConfig: IOSLocalBuildConfig | AndroidLocalBuildConfig,
+    outputChannel: OutputChannel,
+    cancelToken: CancelToken
+  ) {
+    const appRoot = buildConfig.appRoot;
+    const cliPath = getExpoCliPath(appRoot);
+    if (!cliPath) {
+      throw new BuildError(
```
filip131311, line 130: should we make usePrebuild option disabled in the launch configuration ui when the XLI does not exist?

packages/vscode-extension/src/dependency/prebuild.ts
```diff
@@ -0,0 +1,170 @@
...
+    args.push("--no-install");
+    if (buildConfig.forceCleanBuild) {
+      args.push("--clean");
+    }
+
+    if (cancelToken.cancelled) {
+      throw new CancelError("Prebuild was cancelled");
+    }
+
+    const process = exec("node", args, { cwd: appRoot });
```
filip131311, line 147: somewhere in code we have two checks 

`ensureDependenciesForBuild` and `ensureDependenciesForStart` the second one checks node version compatibility with the project. we might make sure it is run before this moment (or we should make a separate check for just expo cli compatibility)

packages/vscode-extension/package.json
jwajgelt: > It will run prebuild before build but only if it is necessery

That's an implementation detail.

>  it will also affect if the build will run at all (instead of expo go)

This is implied, that's the same if the user run `prebuild` before `run`.

> It will also affect only expo projects

This could probably be mentioned.

packages/vscode-extension/src/dependency/prebuild.ts
jwajgelt: I feel it's good when you have `Disposable` components, but in this case (or whenever you have actual logic you want to run, and that's it) I feel it just creates noise.
It's a preference thing though.

packages/vscode-extension/src/dependency/prebuild.ts
jwajgelt: Maybe? 🤷

packages/vscode-extension/src/dependency/prebuild.ts
jwajgelt: `cancel()` will always fully cancel the prebuild process, it calls the `CancelToken` internal to the `PrebuildProcess` which wraps the actual process.

packages/vscode-extension/src/dependency/prebuild.ts
jwajgelt: No.
I don't like the idea of the Config menu reading the user's environment, it should relate to the application -- especially since we intend the configs to be shared between people working on a project.

packages/vscode-extension/src/dependency/prebuild.ts
jwajgelt: `ForStart` runs before `ForBuild`. 
If your node doesn't run `@expo/cli` correctly this will just error, it's fine.

packages/vscode-extension/package.json
filip131311: > That's an implementation detail.

I am not sure as we luck "forcePrebuild" option it might be interpreted as such 

> This is implied, that's the same if the user run prebuild before run.

our poor communication in the past is not a good argument. It affects the whole experience and could just mean "would you like us to do it for you?" type of thing I would like it mentioned

packages/vscode-extension/src/dependency/prebuild.ts
filip131311: I can live with that

packages/vscode-extension/src/dependency/prebuild.ts
jwajgelt: I guess.

PR #1438, opened 2025-08-12
feat: improve inspector and replay overlay rotation handling

pFornagiel:
### Description

This PR introduces improved app orientation detection by utilising popular react-native libraries (if they are currently available) and using orientation information from the Info.plist file in case of iOS applications. This is meant to battle edge-cases not handled by previous approach.

This PR also brings back `PortraitUpsideDown` device orientation, as most issues connected with its handling have been fixed.

### How Has This Been Tested: 

- **test whether element inspector works correctly on different device rotations**
- **test whether the behaviour of element inspector is correct on android and iOS using [react-native-orientation-locker](https://www.npmjs.com/package/react-native-orientation-locker) and [expo-screen-orientation](https://www.npmjs.com/package/expo-screen-orientation) on different devices**

Not applicable.


packages/vscode-extension/lib/orientation/DefaultOrientationStrategy.js
```diff
@@ -1,36 +1,29 @@
...
+ * Android: Orientation events fire after app rotation and are distinct from device orientation. In case of
+ * android, the event fires on app start and can be relied on during the app lifecycle.
+ *
+ * iOS: Events fire when DEVICE rotates (based on sensors), not when APP rotates, creating not only
+ * timing issues, but also firing events event when the application's UI is not rotated. This means
+ * we cannot rely on the events entirely and the frontend must properly interpret the messages sent.
+ *
+ * To handle iOS limitations, we use DimensionsObserver for additional tracking and make assumptions
+ * about supported orientations (e.g., portrait-secondary typically not supported).
+ */
```
jwajgelt: Since Android "just works" -- could split this file into two:
- `AndroidOrientationStrategy.js` which uses just the `UIManager` event listener and sends the event
- `DefaultIosOrientationStrategy.js` which is what we currently do here
and when selecting the strategy, on Android always use the first one, and only try to use the package-based ones or the default one on iOS?

packages/vscode-extension/src/builders/BuildCache.ts
```diff
@@ -124,8 +124,11 @@ function getAppPath(build: BuildResult) {
   return build.platform === DevicePlatform.Android ? build.apkPath : build.appPath;
 }
 
-async function getAppHash(appPath: string) {
-  return (await calculateMD5(appPath)).digest("hex");
+async function getAppHash(appBuild: AndroidBuildResult | IOSBuildResult) {
+  const fileHash = (await calculateMD5(getAppPath(appBuild))).digest("hex");
+  const supportedOrientationsHash =
+    appBuild.platform === DevicePlatform.IOS ? appBuild.supportedInterfaceOrientations.join() : "";
+  return `${fileHash}${supportedOrientationsHash}`;
```
jwajgelt: Is this necessary? I would expect that different settings produce different `fileHash`.

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -252,6 +256,51 @@ export class ApplicationSession implements ToolsDelegate, Disposable {
...
+        return DeviceRotation.LandscapeLeft;
+      } else {
+        return DeviceRotation.LandscapeRight;
+      }
+    }
+
+    if (orientation === "Portrait") {
+      // iPhone case - expo always reports portraitUpsideDown as portrait on iPads
+      if (
+        this.device.deviceInfo.deviceType === "Tablet" &&
```
jwajgelt: Why not
```suggestion
        this.device.deviceInfo.deviceType === DeviceType.Tablet &&
```
?

jwajgelt (approved):
Overall looks very good, left a couple comments.

packages/vscode-extension/src/builders/BuildCache.ts
pFornagiel: This handles the case, where user updates to the new version of the extension whilst still having the same options. We wish to invalidate the old cache, so that the app will not use an old cached build without defined supportedInterfaceOrientations field.

If it's fitting, we could also add some constant string of characters to the hash for iOS builds instead, this would achieve the same effect.

packages/vscode-extension/src/project/applicationSession.ts
pFornagiel: Changed in fb354e74

packages/vscode-extension/lib/orientation/DefaultOrientationStrategy.js
pFornagiel: Introduced `IosDefaultOrientationStrategy` and `AndroidDefaultOrientationStrategy` in cdb5e15.

```ts
const getOrientationStrategy = () => {
  if(Platform.OS === "android") {
    return AndroidDefaultOrientationStrategy.getStrategy();
  }

  if (ExpoOrientationStrategy) {
    return ExpoOrientationStrategy.getStrategy();
  }

  if (OrientationLockerStrategy) {
    return OrientationLockerStrategy.getStrategy();
  }

  return IosDefaultOrientationStrategy.getStrategy();
};

const orientationStrategy = getOrientationStrategy();
```

packages/vscode-extension/src/builders/BuildCache.ts
jwajgelt: I'd prefer we changed the cache key rather than the hash we store in the result.

pFornagiel: Changed in 96814d47

jwajgelt (approved):

PR #1451, opened 2025-08-13
feat: GitHub Actions 

KeyJayY:
This PR introduces working tests in GitHub Actions.  
These are simple smoke tests that validate basic functionality.  

Screenshot 2025-08-13 at 15 57 51

### How Has This Been Tested:
- The workflow was executed in GitHub Actions on branch `@KeyJayY/GithubActions`.
- Verified that the smoke tests ran successfully and completed without errors.
- Also confirmed that tests designed to fail indeed fail

- No documentation yet

### DIFF
--- a/.github/workflows/ui-tests.yaml
+++ b/.github/workflows/ui-tests.yaml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build-and-test:
-    runs-on: macos-latest
+    runs-on: macos-latest-xlarge
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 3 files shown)


.github/workflows/ui-tests.yaml
```diff
@@ -0,0 +1,55 @@
+name: Build VSIX and Run Tests (macOS)
+
+on:
+  workflow_dispatch:
+    inputs:
+      node:
+        description: Node.js version to use
+        required: false
+        default: "20"
```
filip131311: I don't know it the 20 is correct default as it might clash with the new react native 81,  I just run this on the node 22 and it passed so it should not be a problem to bump this

PR #1453, opened 2025-08-14
feat: allow sending files to device

jwajgelt:
Allows sending files to the selected device via Radon.
- adds a "Send file" option to the device settings dropdown which opens the system file picker to select the file(s) to send
- allows drag-n-dropping files over the preview in order to send them over to the device

UI to be done in a separate PR.

https://github.com/user-attachments/assets/afd426d1-53a0-471d-8dd6-4cb6ab65ddf2

Limitations: 
- currently, only media files are supported on iOS due to issues with the "Files" app not working correctly in Radon Simulators. The default behaviour of the Apple Simulator app when dropping a (non-media) file over the simulator seems to be the same as `simctl openurl file://${path-to-file}` which opens the "Files" app on the device and allows selecting where the file should be saved. However, on the Radon Simulator, the "Files" app is unable to save the file (the button is disabled), and other functions (e.g. creating a new folder) also don't work -- I don't know why.

### How Has This Been Tested: 
- open the `react-native-image-picker` example app
- use "send files" option in the Device Settings dropdown and select a photo
- drag another photo to the preview
- verify both photos are selectable in the image picker

WIP

### DIFF
--- a/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
@@ -725,6 +725,10 @@ export class AndroidEmulatorDevice extends DeviceBase {
   async getClipboard() {
     // No need to copy clipboard, Android Emulator syncs it for us whenever a user clicks on 'Copy'
   }
+
+  public async sendFile(filePath: string): Promise<void> {
+    throw new Error("Method not implemented.");
+  }
 }
 
 export async function createEmulator(
--- a/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
@@ -727,7 +727,21 @@ export class AndroidEmulatorDevice extends DeviceBase {
   }
 
   public async sendFile(filePath: string): Promise<void> {
-    throw new Error("Method not implemented.");
+    const args = ["push", "-q", filePath, `/sdcard/Download/${path.basename(filePath)}`];
+    await exec(ADB_PATH, ["-s", this.serial!, ...args]);
+    // Notify the media scanner about the new file
+    await exec(ADB_PATH, [
+      "-s",
+      this.serial!,
+      "shell",
+      "am",
+      "broadcast",
+      "-a",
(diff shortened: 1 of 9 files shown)


packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -725,6 +725,24 @@ export class AndroidEmulatorDevice extends DeviceBase {
   async getClipboard() {
     // No need to copy clipboard, Android Emulator syncs it for us whenever a user clicks on 'Copy'
   }
+
+  public async sendFile(filePath: string): Promise<void> {
+    const args = ["push", "-q", filePath, `/sdcard/Download/${path.basename(filePath)}`];
```
filip131311, line 730: is the `/sdcard/Download/` only place files can be saved?  can you leave the explainer comment here?

filip131311: can the usere ever want the file to be saved somehere else?

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -519,6 +520,26 @@ export class IosSimulatorDevice extends DeviceBase {
       getOrCreateDeviceSet(this.deviceUDID),
     ]);
   }
+
+  public async sendFile(filePath: string): Promise<void> {
+    if (!isMediaFile(filePath)) {
+      throw new Error("Only media file transfer is supported on iOS.");
```
filip131311, line 526: this is not handled anywhere will it as part of the UI design?

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
jwajgelt: If anyone asks, we can iterate.
Downloads just seems like a reasonable place to save files to, since its usually indexed by the media pickers by default (at least on every phone I remember using, it did).

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
jwajgelt: It will throw from the promise in the webview and we should handle it there (like any other failure in this process).

PR #1454, opened 2025-08-14
add stepInto/stepOut controls to debugger overlay

### LINKED ISSUE #283: Step In and Step Out Debugging options
**Description**
I would like to see step-in and step-out functions in the debugging panel options . Ideally, I would love to see these options to enter into functions and exit from them. The step option there works perfectly well, but that also means that I would have to set up so many break points so that the debugger just does not run through my code.

**Environment**
- react-native-ide 0.0.14
- React Native 0.73

image

jwajgelt:
Adds "Step Into" / "Step Out" controls to the debugger overlay.
Closes #283 

### How Has This Been Tested: 
- open (ReactNative74, ReactNative81) test apps in Radon (to test both old and new debugger implementations)
- modify `printLogs` function to call something else:
```
function getText() {
  return 'console.log';
}

function printLogs() {
  // put breakpoint on the next line
  const text = getText();
  console.log(text);
}
```
- place a breakpoint inside `printLogs`
- verify "Step Into" and "Step Out" buttons correctly step through the code.

nah

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -303,6 +303,8 @@ export interface ProjectInterface {
 
   resumeDebugger(): Promise<void>;
   stepOverDebugger(): Promise<void>;
+  stepIntoDebugger(): Promise<void>;
+  stepOutDebugger(): Promise<void>;
   focusDebugConsole(): Promise<void>;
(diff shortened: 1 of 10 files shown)


filip131311 (approved):
tested on expo 53 👍

PR #1459, opened 2025-08-18
feat: add fps recording feature 

filip131311:
NOTE: 
After internal changes we decided not to go with this approach but we still wan to merge it as it contains useful info-structure that we can use in the future. This PR does not expose any UI option to use the frame rate repoerter. 

This PR adds a new fps reporting functionality enabled by https://github.com/software-mansion-labs/simulator-server/commit/f22a46abeafee0055687f39ea0fc2f16399c3824. To do this we:

- integrate with the new sim server function 
- update our new state management system to allow for having states of multiple device sessions at once (note: the `use$` function form legend state seems to only work correctly when observing primitives if the object in question is not part of the initial observable for example then we use `Record` type. This is not a big limitation but will require the structure of the frontend to look close to `useFrameReporting` hook)
- update the UI to allow for frame rate inspecting. 

https://github.com/user-attachments/assets/b69e2c06-303f-4ef1-b775-31f6e96dbdc6

### How Has This Been Tested: 

exactly as on video above

-- this is backend only for now


packages/vscode-extension/src/devices/preview.ts
```diff
@@ -106,6 +111,29 @@ export class Preview implements Disposable {
...
+          let frameRateData: FramerateReport;
+          try {
+            frameRateData = JSON.parse(jsonString) as FramerateReport;
+          } catch (error) {
+            Logger.error("[Preview] Error parsing frame rate JSON:", error);
+            return;
+          }
+          if (!this.fpsReportListener) {
+            Logger.warn(
+              "[Preview] No FPS report listener registered but received frame rate data, this should not be reachable."
```
jwajgelt: Isn't this _kind of_ expected when you disable reporting?
Before sim-server can read our `fps false` command it can write one or more records that we will read here.

packages/vscode-extension/src/webview/hooks/useFrameReporting.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+  const enabled = use$(
+    store$.projectState.deviceSessions[selectedSessionId!]?.frameReporting?.enabled ?? false
+  );
+  const fps = use$(
+    store$.projectState.deviceSessions[selectedSessionId!]?.frameReporting?.frameReport?.fps ??
+      undefined
+  );
+
+  return { enabled, fps };
```
jwajgelt: The returned object should probably be `useMemo`ed

packages/vscode-extension/src/devices/DeviceBase.ts
```diff
@@ -144,6 +144,14 @@ export abstract class DeviceBase implements Disposable {
     this.preview?.dispose();
   }
 
+  public startFrameRateReporting(onFpsReport: (report: FramerateReport) => void) {
```
jwajgelt:
```suggestion
  public startReportingFrameRate(onFpsReport: (report: FramerateReport) => void) {
```
start-sth-doing sounds off to me.

jwajgelt: Similarly elsewhere and `stopFrameRateReporting`

packages/vscode-extension/src/common/Project.ts
```diff
@@ -326,6 +326,9 @@ export interface ProjectInterface {
   captureScreenshot(): void;
   saveMultimedia(multimediaData: MultimediaData): Promise<boolean>;
 
+  startFrameRateReporting(): void;
```
jwajgelt: It's `FrameRateReporting` but `FramerateReport` ("Rate" vs "rate"). Stick to one.

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -283,16 +266,23 @@ function PreviewView() {
...
             title="Stop profiling React"
             onClick={stopProfilingReact}
           />
+          {/* Note: This is now hidden from the UI as it points to something that might be misleading to the user, we might use it 
+          in the future when we have a way of counting dropped frames on ios tho. */}
+          {/* <ActiveToolButton
+            toolState={frameReportingEnabled ? "profiling" : "stopped"}
+            title={"FPS: " + (fps ?? 0)}
+            onClick={stopFrameRateReporting}
+          /> */}
```
jwajgelt: Maybe let's just remove it instead of leaving it in the code.

packages/vscode-extension/src/devices/preview.ts
filip131311: yes you are right

PR #1461, opened 2025-08-19
Reactor build to make builds depend on device and also fix cache to rely on device config

kmagiera:
This PR refactors the build logic to let the build configuration access the device instance. In particular, `xcodebuild` may need to know the device it builds for. But also to address two main flows of the build strategy:
1) The current implementations leaves orphaned entries in the global state. The reason is that the cache key depends on some parameters that can be adjusted by the user and in particular we also changed the key prefix recently when we added support for iPads.
2) We currently use too few parameters for fingerprinting and hence it may happen that we will attempt to reuse a build that should not be reused. In particular, we only use a subset of launch config options while all of them may impact the output of the build process.

To address the above, we are introducing a new concept of build fingerprint. Build fingerprint is an extension of the fingerprinting mechanism we used before, but also include the additional parameters that can impact the build process and are specific to Radon. The idea is that build fingerpring should represent all input that took part in producing the build artifact. This includes the project fingerprint (representing all the native code and dependencies), but also the build settings and environment constants that are set in the launch configuration.

To calculate the build fingerprint we therefore take the project fingerprint and add a deterministic hash of the build configuration. We also place a platform prefix in front for the fingerprint to be better readable.

In order to manage the build cache, we are introducing a new GC mechanism. GC scans the entries in global store and tries to delete ones that no longer has corresponding backing build artifacts. We also consider the case where many keys point to a single build artifact, in which case we cleanup the ones that expect a different artifact hash.

Finally, In order for build config to remain stable, we had to remove `forceCleanBuild` from it and move it to `BuildOption`. This is an option that can be set for some builds and not for others, and does not impact the build artifact but only forces executing the build even if the fingerprint matches.

### Test plan
1) run iOS and Android builds
2) try force clean build option
3) make sure cached builds are used after restarting the IDE
4) run unit tests

### DIFF
--- a/packages/vscode-extension/src/builders/BuildCache.ts
+++ b/packages/vscode-extension/src/builders/BuildCache.ts
@@ -42,6 +42,9 @@ export class BuildCache {
 
   private async collectGarbageCacheEntriesNow() {
     const keys = await extensionContext.globalState.keys();
+    Logger.info(
+      `Collecting garbage cache entries. ${keys.length} keys present in the global state.`
+    );
     const keysToRemove = keys.filter((key) => key.startsWith(GLOBAL_STATE_BUILD_CACHE_KEY_PREFIX));
     const buildResultsForAppPaths = new Map<string, [BuildResult]>();
     const promises = [];
--- a/packages/vscode-extension/src/builders/BuildCache.ts
+++ b/packages/vscode-extension/src/builders/BuildCache.ts
@@ -6,6 +6,7 @@ import { BuildFingerprint, BuildResult } from "./BuildManager";
 import { DevicePlatform } from "../common/State";
 
 const GLOBAL_STATE_BUILD_CACHE_KEY_PREFIX = "build-cache";
+const GLOBAL_STATE_LGACY_BUILD_CACHE_PREFIXES = ["android_build_cache", "ios_build_cache"];
 
 function globalStateKey(fingerprint: BuildFingerprint) {
   return `${GLOBAL_STATE_BUILD_CACHE_KEY_PREFIX}:${fingerprint}`;
@@ -24,6 +25,16 @@ export class BuildCache {
 
   public async clearCache(fingerprint: BuildFingerprint) {
     await extensionContext.globalState.update(globalStateKey(fingerprint), undefined);
+  }
+
+  public static async clearLegacyCacheEntries() {
+    const keys = await extensionContext.globalState.keys();
+    const keysToRemove = keys.filter((key) =>
+      GLOBAL_STATE_LGACY_BUILD_CACHE_PREFIXES.some((prefix) => key.startsWith(prefix))
+    );
+    await Promise.all(
+      keysToRemove.map((key) => extensionContext.globalState.update(key, undefined))
+    );
   }
 
   private async collectGarbageCacheEntriesNow() {
--- a/packages/vscode-extension/src/builders/BuildCache.ts
+++ b/packages/vscode-extension/src/builders/BuildCache.ts
@@ -13,6 +13,9 @@ function globalStateKey(fingerprint: BuildFingerprint) {
 }
 
 export class BuildCache {
+  constructor() {
+    BuildCache.clearLegacyCacheEntries();
+  }
   /**
    * Passed fingerprint should be calculated at the time build is started.
    */
--- a/packages/vscode-extension/src/builders/BuildCache.ts
+++ b/packages/vscode-extension/src/builders/BuildCache.ts
@@ -1,14 +1,15 @@
 import fs from "fs";
-import assert from "assert";
 import crypto from "crypto";
+import stableStringify from "fast-json-stable-stringify";
 import { Logger } from "../Logger";
 import { extensionContext } from "../utilities/extensionContext";
 import { IOSBuildResult } from "./buildIOS";
 import { AndroidBuildResult } from "./buildAndroid";
 import { calculateMD5 } from "../utilities/common";
 import { BuildResult } from "./BuildManager";
-import { FingerprintOptions, FingerprintProvider } from "../project/FingerprintProvider";
+import { FingerprintProvider } from "../project/FingerprintProvider";
 import { DevicePlatform } from "../common/State";
+import { BuildConfig } from "../common/BuildConfig";
 
 const ANDROID_BUILD_CACHE_KEY = "android_build_cache";
 const BASE_IOS_BUILD_CACHE_KEY = "ios_build_cache";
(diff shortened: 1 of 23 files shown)


packages/vscode-extension/src/builders/BuildCache.ts
```diff
@@ -30,25 +30,15 @@ export type BuildCacheInfo = {
...
-
-function stringifyCacheKey({ platform, appRoot, env }: CacheKey) {
+function buildCacheKey(buildConfig: BuildConfig) {
   const keyPrefix =
-    platform === DevicePlatform.Android ? ANDROID_BUILD_CACHE_KEY : IOS_BUILD_CACHE_KEY;
-  const envHash = hashEnvironment(env);
+    buildConfig.platform === DevicePlatform.Android ? ANDROID_BUILD_CACHE_KEY : IOS_BUILD_CACHE_KEY;
+
+  const buildConfigHash = crypto.createHash("md5");
+  buildConfigHash.update(JSON.stringify(buildConfig));
```
jwajgelt: This doesn't account for the fact that "equal" objects can have their keys in a different order.
This produces different hashes for `{a:1, b:2}` and `{b:2, a:1}`.

Instead, you could take the approach we had here previously for `hashEnvironment` where we serialize objects into lists of sorted key-value pairs and run that recursively on the `buildConfigHash` object.

Or, you could `update` the hash here manually
```
buildConfigHash.update(buildConfig.appRoot)
buildConfigHash.update(buildConfig.platform)
buildConfigHash.update(hashEnvironment(buildConfig.env))
buildConfigHash.update(buildConfig.runtimeId)
buildConfigHash.update(buildConfig.deviceTypeId)
// etc...
```

Or even better, have a list of keys we want to take into account for the cache key and use that:
```
const RELEVANT_KEYS = [
"appRoot", "platform", "runtimeId", // ect...
]
RELEVANT_KEYS.forEach(k => {
  buildConfigHash.update(buildConfig[k])
});
```

packages/vscode-extension/src/builders/buildIOS.ts
```diff
@@ -228,20 +261,30 @@ async function buildLocal(
...
   Logger.debug(`Xcode build will use "${scheme}" scheme`);
 
+  const { simulatorUdid, cleanupSimulator } = await getOrMakeSimulator(buildConfig);
+  if (cleanupSimulator) {
+    if (cancelToken.cancelled) {
+      cleanupSimulator();
+    } else {
+      cancelToken.onCancel(cleanupSimulator);
+    }
+  }
```
jwajgelt: I think It'd be best to just wrap the whole of this function's body after the temp simulator is created in a `try`/`finally { cleanup(); }`, that way there shouldn't be any risk of it failing us.

Right now, if `await buildProcess` [at line 292](https://github.com/software-mansion/radon-ide/pull/1461/files#diff-c1465ea393e4601b8aa787aa23cbffe33457e2fe8e62608a248b33ba0a39eaafL249) fails, it looks to me that the simulator is never cleaned up.

packages/vscode-extension/src/dependency/prebuild.ts
```diff
@@ -100,7 +98,7 @@ export class Prebuild implements Disposable {
 
     const prebuildCancelToken = new CancelToken();
     const prebuildProcess = new PrebuildProcess(
-      this.runPrebuild(buildConfig, outputChannel, prebuildCancelToken),
+      this.runPrebuild(buildConfig, buildOptions),
```
jwajgelt: You need to pass `prebuildCancelToken` here, not the one from `buildOptions`


```suggestion
      this.runPrebuild(buildConfig, {...buildOptions, cancelToken: prebuildCancelToken}),
```

packages/vscode-extension/src/builders/BuildCache.ts
kmagiera: this works ok now, because the configs are created as objects and never modified. In JS this guarantees you the order of items. If we were to use `obj.key = 1` or `obj['key'] = 1` later on on the configs which would result in a new key being added, then we cannot rely on the content being ordered consistently.

I think I'll just restore sorting mechanism back as it is too fragile to rely on the ordering. I don't like whitelisting here as we have a dozen of different config types for Eas, Android Local, Prebuild etc and it'd just generate an implicit coupling between these types and the code here.

packages/vscode-extension/src/dependency/prebuild.ts
kmagiera: will just restore the previous code here

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -660,6 +624,10 @@ export class DeviceSession implements Disposable {
 
   public async start() {
     try {
+      const launchConfiguration = this.applicationContext.launchConfig;
+      const buildType = await inferBuildType(this.platform, launchConfiguration);
+      this.buildConfig = createBuildConfig(this.device, launchConfiguration, buildType);
```
jwajgelt: Since `buildConfig` stores the `env` read from `launchConfiguration`, doing it this way will prevent us from reading a new `env` when restarting / reloading the app.
We should create a new `BuildConfig` each time we may want to rebuild the app.
I'd suggest:
- having a field `lastBuildConfig` which is used for the `onProjectFilesChanged` cache stale check _only_
- add a `BuildConfig` parameter to `DeviceSession.buildApp` and set `this.lastBuildConfig` whenever the build succeeds
- have places which call `buildApp` create a new `BuildConfig` where appropriate (e.g. `autoReload` should create a new one before checking cache staleness) and pass it down to `buildApp`

packages/vscode-extension/src/builders/BuildManager.ts
```diff
@@ -182,28 +193,33 @@ export async function inferBuildType(
 }
 
 export class BuildManagerImpl implements BuildManager {
-  constructor(private readonly buildCache: BuildCache) {}
+  constructor(
+    private readonly buildCache: BuildCache,
+    private readonly fingerprintProvider: FingerprintProvider
+  ) {}
+
+  public async calculateBuildFingerprint(buildConfig: BuildConfig): Promise<BuildFingerprint> {
```
jwajgelt, line 201: Is there a reason this is a part of `BuildManager` rather than `BuildCache`? It made more sense to me to have this logic in the `Cache`

kmagiera: It feels more natural to me to be in build manager. BuildCache does not use this method directly and it was only being called from BuildManager via BuildCache. I included fingerprint in BuildResult which IMO is handy because by owning the BuildResult instance in device session we are able to tell whether this result is stale. As a consequence it became more natural to have the BuildFingerprint type defined in BuildManager along BuildResult type. Finally, we'd need to have cyclic dependencies (which are ok, but I tend to avoid them as they typically confuse me when I look at the code after some time) because we'd need to import BuildFingerprint from BuildManager in BuildCache, but also import BuildCache in order to be able to call the fingerprint method.

That being said, I don't feel strongly about this particular part of the change. Will merge this as is and can revert this move if you prefer so in a follow up.

PR #1464, opened 2025-08-19
Skip installing iOS app when it is already there

kmagiera:
This PR implements a fast-path for the case when an app is already installed on iOS simulator.

App installation is one of the necessary step when starting Radon IDE. After looking into it, the test-apps typically install within 5-8 seconds on my Mac M2. This can be improved by skipping installation when the app is already installed.

This PR introduces a mechanism that allows us to determine if app in the exact installed form already exists:
1) we calculated md5 hash for the build directory that gives us the hash of the output
2) we use app container location (place in local filesystem where app is "installed")
3) we place a .dotfile with the hash in that directory
4) when testing whether app is installed, we take the hash of the new app, and compare it with what's in the .dotfile
5) this works, because when over-installing, simctl wipes out the whole app container directory and the newly installed app is located in a new directory (this holds true even if we install exactly the same app binary)

As a result, we replace 5-8sec installation step (on Mac M2) with the check which takes 0.3sec (on the same machine)

### How Has This Been Tested: 
1) install test app on iOS sim
2) restart IDE, see the installation takes less time (we print "skipping installation" in the logs)
3) test force re-installation which should print "App installed" message in logs
4) change something in app.json which result in the app rebuilding and you should see "app installed" message again.

### DIFF
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -452,10 +452,44 @@ export class IosSimulatorDevice extends DeviceBase {
     }
   }
 
+  async locateInstalledAppBuildHashFile(build: IOSBuildResult) {
+    const deviceSetLocation = getOrCreateDeviceSet(this.deviceUDID);
+    const { stdout: appContainerLocation } = await exec("xcrun", [
+      "simctl",
+      "--set",
+      deviceSetLocation,
+      "get_app_container",
+      this.deviceUDID,
+      build.bundleID,
+      "app",
+    ]);
+    return path.join(appContainerLocation, ".radonide.buildhash");
+  }
+
+  async checkInstalledAppBuildHashFile(build: IOSBuildResult) {
+    const buildHashFileLocation = await this.locateInstalledAppBuildHashFile(build);
+    try {
+      const buildHash = await fs.promises.readFile(buildHashFileLocation, "utf8");
+      return buildHash === build.buildHash;
+    } catch (error) {
+      if ((error as NodeJS.ErrnoException).code === "ENOENT") {
+        return null;
+      } else {
+        throw error;
+      }
+    }
+  }
+
+  async updateInstalledAppBuildHash(build: IOSBuildResult) {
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -452,10 +452,44 @@ export class IosSimulatorDevice extends DeviceBase {
...
+  async checkInstalledAppBuildHashFile(build: IOSBuildResult) {
+    const buildHashFileLocation = await this.locateInstalledAppBuildHashFile(build);
+    try {
+      const buildHash = await fs.promises.readFile(buildHashFileLocation, "utf8");
+      return buildHash === build.buildHash;
+    } catch (error) {
+      if ((error as NodeJS.ErrnoException).code === "ENOENT") {
+        return null;
+      } else {
+        throw error;
```
jwajgelt: Do we want to throw here? Maybe we can just log and ignore errors, since this is not critical for the app to run.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -467,7 +501,19 @@ export class IosSimulatorDevice extends DeviceBase {
...
       }
+    } else {
+      const isInstalled = await this.checkInstalledAppBuildHashFile(build);
+      if (isInstalled) {
+        const skipInstallDurationSec = (performance.now() - startTime) / 1000;
+        Logger.info(
+          `App is already installed, skipping installation. Took ${skipInstallDurationSec.toFixed(
+            2
+          )} sec.`
+        );
```
jwajgelt, line 508: You can remove this before merging.

kmagiera: I wanted to have it here. We have similar logs for build and some other phases but not for installation step

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
kmagiera: makes sense, will update

PR #1456, opened 2025-08-14
draft ui

kacperkapusciak:
Start by providing description for the PR: please include a summary of the changes, related issue, motivation, and context behind the change. Provide screenshots or recordings when appropriate. 

Fixes # (issue)

### How Has This Been Tested: 

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

### How Has This Change Been Documented:

If applicable (for UI changes or new features) provide a link to a PR that updates the relevant documentation pages.

PR #1465, opened 2025-08-20
feat: allow sending certificate files to the iOS simulator

### LINKED ISSUE #486: Ability to install Root Certificate in simulator
**Description**
As a user of Proxyman, I need the ability to install a Root certificate.

I understand that there are other ways to listen for traffic, but we have very complex GraphQL queries which we need to rewrite and Proxyman does this with the most grace.

This should be quite easy if we can just navigate to the home screen.

**Environment**
- Version of currently used React Native IDE (Extensions tab -> click on React Native IDE extension -> version is located next to the extension name)
- Version of React Native/Expo used in the project.

jwajgelt:
Adds support for sending `.cer` and `.pem` files to iOS Simulator devices.
Closes #486

### How Has This Been Tested: 
- open an iOS device in Radon
- try sending a `.cer` or `.pem` file with a CA Certificate to the device (either by drag-n-dropping while holding Shift or using the "Send File" option in the device settings dropdown)
- Safari should open and ask you to install a "security profile"

WIP, waiting on #1456

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -952,10 +952,21 @@ export class DeviceSession implements Disposable {
   }
 
   public async sendFileToDevice(fileName: string, data: ArrayBuffer): Promise<void> {
+    let canSafelyRemove = true;
     const tempDir = await this.getTemporaryFilesDirectory();
     const tempFileLocation = path.join(tempDir, fileName);
-    await fs.promises.writeFile(tempFileLocation, new Uint8Array(data));
-    await this.sendFile(tempFileLocation);
+    try {
+      await fs.promises.writeFile(tempFileLocation, new Uint8Array(data));
+      const result = await this.sendFile(tempFileLocation);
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -952,16 +952,32 @@ export class DeviceSession implements Disposable {
...
+
+  private tempDir: string | undefined;
+  /**
+   * Returns the path to a temporary directory, creating it if it does not already exist.
+   * The directory is created using the system's temporary directory and is cleaned up
+   * automatically when the device session is disposed. Subsequent calls return the same directory path.
+   *
+   * @returns {Promise<string>} The path to the temporary directory.
+   */
+  private async getTemporaryFilesDirectory(): Promise<string> {
```
filip131311, line 980: I don't like that in the new aproach the files will only be removed after the device session will dispose. If someone sends pictures to the device they might do that in large batches and this will leak taken storage space.

filip131311 (approved):
👍

PR #1469, opened 2025-08-20
feat: expo54 support - part one 

filip131311:
This PR adds suport for expo 54 by doing the following things:
- clean up react versions in the renderers for RN 81 and RN 80 
- add expo router 6 support 
- override the new terminal reporter location 

This PR does not solve all the issues with the expo 54 as there are still some problems with profilinng we'll work on in a follow up.

### How Has This Been Tested: 

- run the new `expo-54` test app and run it

after the follow up


packages/vscode-extension/lib/expo_router/expo_router_v6_plugin.js
jwajgelt, line 1: Can we have this be a symlink to v5 instead of copying the content?

jwajgelt (approved):
🆒 🆒 🆒

PR #1471, opened 2025-08-20
Use destination specifier when building ios

kmagiera:
This PR changes the set of specifiers we use when building xcode. Apparently, some of our users reported that under specific conditions the builds are failing. The root cause turned out to be the fact that xcodebuild doesn't allow for `ONLY_ACTIVE_ARCH` to ever be overwritten manually. This flag is important for certain xcode projects when building for simulator in case the projects contain targets that cannot build for simulator (i.e. apple watch targets). When specifying simulator as destination, this is detected automatically by xcode build and the `ONLY_ACTIVE_ARCH` is set. Apparently this flag is AFAIK the only one that cannot be overwritten by the options passed to xcodebuild (https://developer.apple.com/documentation/xcode/build-settings-reference#Build-Active-Architecture-Only) – it can only be switched off with destination specifier.

As a consequence, we need to use `destination` for xcode build, but this wasn't as easy. In Radon we use a custom device set. While simctl supports this via `--set` flag, `xcodebuild` always uses the default device set. To workaround that, we need to use a device that exists in the default set. Before starting the build we proceed as follows:
1) we take the runtime id as argument for the build and look for existing device with this runtime id – in case of iOS, we can pick any device, as we override the target family later on and thus we don't need to handle iPhones and iPads separately
2) in no such device exists (which is unlikely), we create a temporary device that we delete after the build process is over.

### How Has This Been Tested: 
1) run iOS build, see in the logs that it uses the destination specifier
2) this has been verified on the users' project where the problem with ONLY_ACTIVE_ARCH surfaced


packages/vscode-extension/src/builders/buildIOS.ts
```diff
@@ -262,6 +316,10 @@ async function buildLocal(
     throw new Error(
       "The iOS app build was successful, but the app file could not be accessed. See the build logs for details."
     );
+  } finally {
+    if (cleanupSimulator) {
+      cleanupSimulator();
+    }
```
jwajgelt: This `try` block should also wrap the `buildProcess`, right now if the build fails (without being cancelled), we won't clean up.

jwajgelt: And perhaps we should even only wrap the `buildProcess`, I don't think we need the simulator to exist here.

packages/vscode-extension/src/builders/buildIOS.ts
```diff
@@ -199,6 +203,39 @@ export async function buildIos(
...
 
+async function getOrMakeSimulator(buildConfig: IOSLocalBuildConfig) {
+  const { runtimeId } = buildConfig;
+  const devices = await listSimulators(SimulatorDeviceSet.Default); // use default location becuase xcodebuild can't use custom device sets
+  const matchingDevice = devices.find(
+    (deviceInfo) => deviceInfo.runtimeInfo?.identifier === runtimeId && deviceInfo.available
+  );
+  if (matchingDevice) {
+    Logger.info("Using existing simulator for iOS build:", matchingDevice);
+    return { simulatorUdid: matchingDevice.UDID };
```
jwajgelt: NIT:
```suggestion
    return { simulatorUdid: matchingDevice.UDID, cleanupSimulator: () => {} };
```
making the `cleanupSimulator` required (but noop in this case) would make calling it more readable.
You could rename it to `maybeCleanupSimulator` if you want to highlight it sometimes does nothing, but I don't think it's necessary.

packages/vscode-extension/src/builders/buildIOS.ts
```diff
@@ -199,6 +203,39 @@ export async function buildIos(
...
+    "Radon IDE Temp Simulator",
+    runtimeId,
+    SimulatorDeviceSet.Default
+  );
+  if (simulatorUdid) {
+    return {
+      simulatorUdid,
+      cleanupSimulator: () => {
+        removeIosSimulator(simulatorUdid, SimulatorDeviceSet.Default);
+      },
```
jwajgelt: I think we should `await` this (here and inside `buildLocal`), to avoid a situation where a new build starts before the temp simulator is fully removed, and `xcodebuild` is passed a simulator which no longer exists.

packages/vscode-extension/src/builders/buildIOS.ts
```diff
@@ -199,6 +203,39 @@ export async function buildIos(
...
+  // It is unlikely we will reach this point as it may only happen when user doesn't have simulator for the selected runtime available
+  // The above is possible but when Xcode installs new SDK, it also creates a set of simulators for it
+  // Thereofre, this can only happen when the user has deleted all of the simulator or if they deleted the runtime
+  // in which case we won't be able to create a new simulator for it anyways and the build should fail.
+  const simulatorUdid = await createSimulatorWithRuntimeId(
+    "iPhone 16 Pro", // device type doesn't matter as long as it supports the runtime, we choose iPhone 16 as it is recent and will be available for a while
+    "Radon IDE Temp Simulator",
+    runtimeId,
+    SimulatorDeviceSet.Default
+  );
```
jwajgelt, line 237: How long does this take? I wouldn't expect this to impact build times much, but maybe it'd be better to create this temp device once (when it's first needed), and then remove it once the whole extension is shutting down, to avoid a situation where the device is created and deleted over and over when a user rebuilds the app?

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -658,13 +656,13 @@ export enum SimulatorDeviceSet {
   RN_IDE,
 }
 
-export async function createSimulator(
-  modelId: string,
+export async function createSimulatorWithRuntimeId(
```
jwajgelt, line 735: VERY NIT:
I don't see the point of making this a separate exported function. Can't the caller just call `createSimulator(...).UDID` if they want just the `UDID`?
Right now you have two similarly called functions where it's not clear what's the difference between them without reading the implementation.

jwajgelt (changes requested):
Overall looks very good.
Left a few suggestions, and one potential issue (temp simulator cleanup on build errors) which should be addressed.

packages/vscode-extension/src/builders/buildIOS.ts
kmagiera: on my mac M2 it takes consistently below 100ms

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
kmagiera: one method takes `runtimeId: string` as argument whereas the other takes an object. The problem here is that when we create temp simulator, we only care about the UDID and so even if we wanted to only extract that we'd still need to have an instance of runtime. In order to get instance of runtime we'd have to pass it down to the build likely via build config and this would impact caching etc. We could also create runtime in place but that'd require an extra simctl call to retrieve the runtime metadata

packages/vscode-extension/src/builders/buildIOS.ts
kmagiera: no, apparently we need that at the later step where we read the build settings. This way we learn the location of the build

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
jwajgelt: I see, I missed this. I thought the parameters for both were the same.
It's fine then.

PR #1472, opened 2025-08-21
fix: cacheing ios installed applications

filip131311:
This PR solves a problem with a new installed application cacheing system, which would throw an error if the application was never installed on the device before. 

### How Has This Been Tested: 

- create a new ios device 
- open any non expo go application 

internal change

### DIFF
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -466,7 +466,6 @@ export class IosSimulatorDevice extends DeviceBase {
       ]);
       return path.join(appContainerLocation, ".radonide.buildhash");
     } catch (error) {
-      Logger.debug("[iosSimulatorDevice] Error locating app build hash file location:", error);
       return undefined;
     }
   }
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -478,6 +486,9 @@ export class IosSimulatorDevice extends DeviceBase {
 
   async updateInstalledAppBuildHash(build: IOSBuildResult) {
     const buildHashFileLocation = await this.locateInstalledAppBuildHashFile(build);
+    if (buildHashFileLocation === undefined) {
+      return;
+    }
```
kmagiera: I'd reverse the condition here. Typically prefer to avoid early returns. In this case we should just not write if the buildHashFileLocation isn't set

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -454,20 +454,28 @@ export class IosSimulatorDevice extends DeviceBase {
...
+        "--set",
+        deviceSetLocation,
+        "get_app_container",
+        this.deviceUDID,
+        build.bundleID,
+        "app",
+      ]);
+      return path.join(appContainerLocation, ".radonide.buildhash");
+    } catch (error) {
+      Logger.debug("[iosSimulatorDevice] Error locating app build hash file location:", error);
```
kmagiera: we either should not log here at all (since this is expected) or don't say "Error" in the message as it indicates something went wrong while it is ok for us to reach this point in code in particular when the app isn't installed

kmagiera (approved):

PR #1473, opened 2025-08-21
feat: network inspector - introduce sorting and filters

pFornagiel:
### Description

This PR introduces **sorting** and **filtering** functionality into the Network Inspector.

#### Sorting
Sorting can be toggled by clicking on the table header. Succeeding clicks on the same header toggle filtering in the pattern: Ascending -> Descending -> None.

https://github.com/user-attachments/assets/0a161751-083f-440a-a660-82b6cc7a94ec

#### Filtering

Filtering has been enhanced, now allowing for searching for patterns global-wise or target specific column by using the following conventions:
- column:value
- column:"value with spaces"

Typing hints, upon starting to write column names, have been added for UX purposes, as well as hint accepting and filter defining using keyboard navigation.

Every applied column filter is turned into a badge, as this approach is more readable and manageable with greater amount of filters applied. The flow of badges and input is as expected from the input-field-experience from user perspective. Keyboard navigation between badges and filters has also been introduced.

Filters may be *inverted* using checkbox next to input field.

Additionally, if table header is wide enough, upon hovering a filtering icon appears, which will paste the empty column filter value into input, set focus and appropriate cursor position. Clicking on the button does not trigger sorting and the button is made to not obstruct column labels.

https://github.com/user-attachments/assets/0a55322c-9f14-4dbd-9663-1bd1d790ec00

### How Has This Been Tested: 

- run Radon IDE extension panel, open `Network Inspector`
- **test applying various filters and sorting the table**

WIP


packages/vscode-extension/src/network/components/FilterInput.tsx
```diff
@@ -0,0 +1,594 @@
...
+const INPUT_LEFT_PADDING = 10;
+const INPUT_UPDATE_TIMEOUT = 10;
+const BADGE_HIGHLIGHT_TIMEOUT = 600;
+
+/**
+ * Get autocomplete suggestion for partial filter text
+ */
+function getFilterAutocompleteSuggestion(filterText: string): string {
+  // No suggestion if empty or ends with whitespace
+  if (!filterText || filterText !== filterText.trimEnd()) {
```
jwajgelt: Would prefer if you checked the last character explicitly rather than calling `trim` and comparing.
For example:

```suggestion
  if (!filterText || filterText.at(-1).trim() === "") {
```
Or even better, having a utility function for readability
```suggestion
function isWhitespace(str: string) {
  return str.trim() === "";
}

  if (!filterText || isWhitespace(filterText.at(-1))) {
```

packages/vscode-extension/src/network/components/FilterInput.tsx
```diff
@@ -0,0 +1,594 @@
...
+    setFilterText,
+  } = useNetworkFilter();
+  const containerRef = useRef<HTMLDivElement>(null);
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  const inputWithSuggestionRef = useRef<HTMLDivElement>(null);
+  const inputOccupiedSpaceRef = useRef<number>(0);
+  const [isFocused, setIsFocused] = useState(false);
+  const [suggestion, setSuggestion] = useState("");
+  const [inputWidth, setInputWidth] = useState<number>(20);
+  const [focusedBadgeIndex, setFocusedBadgeIndex] = useState<number>(-1); // -1 means no badge is focused
```
jwajgelt:
```suggestion
  const [focusedBadgeIndex, setFocusedBadgeIndex] = useState<number|null>(null);
```
Let's avoid using magic values like that -- I know that's what `findIndex` returns, but I think it's still better to use `null` in these cases.

packages/vscode-extension/src/network/components/FilterInput.tsx
```diff
@@ -0,0 +1,594 @@
...
+      scrollInputIntoView();
+    }, INPUT_UPDATE_TIMEOUT);
+
+    return true;
+  };
+
+  const removeBadge = (index: number) => {
+    const badgeId = filterBadges[index].id;
+    const newBadges = filterBadges.filter((badge) => badge.id !== badgeId);
+    setFilterBadges(newBadges);
```
jwajgelt: Why not something like
```suggestion
    const newBadges = filterBadges.toSpliced(index, 1);
    setFilterBadges(newBadges);
```
?

packages/vscode-extension/src/network/components/FilterInput.tsx
```diff
@@ -0,0 +1,594 @@
...
+      focusFilterInput();
+    } else if (index === focusedBadgeIndex) {
+      const newFocusIndex = index === filterBadges.length - 1 ? index - 1 : index;
+      setFocusedBadgeIndex(newFocusIndex);
+    } else {
+      const newBadgeIndex = newBadges.findIndex(
+        (badge) => badge.id === filterBadges[focusedBadgeIndex].id
+      );
+      setFocusedBadgeIndex(newBadgeIndex);
+    }
```
jwajgelt: I'm pretty sure this is equivalent to
```suggestion
    } else {
      let newFocusIndex = focusedBadgeIndex > index ? index - 1 : index;
      newFocusIndex = Math.max(newFocusIndex, newBadges.length - 1);
      setFocusedBadgeIndex(newFocusIndex);
    }
```
You can decide if that's clearer though.

packages/vscode-extension/src/network/components/NetworkBar.tsx
```diff
@@ -59,13 +59,12 @@ function NetworkBar() {
...
-          <VscodeTextfield
-            value={filters.url ?? ""}
-            onInput={(e) => {
-              // @ts-ignore it works, types seem to be incorrect here
-              setFilters({ ...filters, url: e.target.value });
-            }}
-            placeholder="Filter domain"
+          <FilterInput placeholder="Filter: search all columns or <column>:<value>" />
+          <VscodeCheckbox
+            onChange={handleInvertChange}
```
jwajgelt: Why the extra arrow function?
```suggestion
            onChange={toggleInvert}
```

packages/vscode-extension/src/network/utils/networkLogFormatters.ts
jwajgelt, line 1: Should probably remove this.

packages/vscode-extension/src/network/utils/networkLogUtils.ts
```diff
@@ -0,0 +1,211 @@
...
+            KB: 1024,
+            MB: Math.pow(1024, 2),
+            GB: Math.pow(1024, 3),
+            TB: Math.pow(1024, 4),
+          };
+          return value * (multipliers[unit] || 0);
+        };
+
+        comparison = getSizeInBytes(aValue) - getSizeInBytes(bValue);
+        break;
```
jwajgelt: You've got the original `NetworkLog` here, why do you operate on the formatted value? Can't you just compare `log.encodedDataLength`?

packages/vscode-extension/src/network/utils/networkLogUtils.ts
```diff
@@ -0,0 +1,211 @@
...
+      }
+      case NetworkLogColumn.Time: {
+        // Extract numeric value from time string ("150 ms" -> 150)
+        const aNum = parseInt(aValue) || 0;
+        const bNum = parseInt(bValue) || 0;
+        comparison = aNum - bNum;
+        break;
```
jwajgelt: Again, can't you just use `log.timeline?.durationMs`?

packages/vscode-extension/src/network/components/FilterInput.tsx
pFornagiel: Added isWhitespace in bee13f99

```ts
function isWhitespace(str: string) {
  return str.trim() === "";
}

/**
 * Get autocomplete suggestion for partial filter text
 */
function getFilterAutocompleteSuggestion(filterText: string): string {
  // No suggestion if empty or ends with whitespace
  if (!filterText || isWhitespace(filterText.at(-1) ?? "")) {
    return "";
  }
  // ...
}
```

packages/vscode-extension/src/network/components/FilterInput.tsx
pFornagiel: Changed in 6361cab7

packages/vscode-extension/src/network/components/FilterInput.tsx
pFornagiel: Changed in fbbe3280

```ts
const removeBadge = (index: number) => {
  const newBadges = filterBadges.toSpliced(index, 1);
  setFilterBadges(newBadges);

  // Set proper focus
  if (focusedBadgeIndex === null) {
    focusFilterInput();
    scrollInputIntoView();
  } else if (newBadges.length === 0) {
    setFocusedBadgeIndex(null);
    focusFilterInput();
  } else {
    let newFocusIndex = focusedBadgeIndex > index ? index - 1 : index;
    newFocusIndex = Math.min(newFocusIndex, newBadges.length - 1);
    setFocusedBadgeIndex(newFocusIndex);
  }
};
```

packages/vscode-extension/src/network/components/FilterInput.tsx
pFornagiel: Changed in fbbe3280. Pretty sure you meant `Math.min`, instead of `Math.max`, but it still sure is cleaner, thanks.

```ts
const removeBadge = (index: number) => {
  const newBadges = filterBadges.toSpliced(index, 1);
  setFilterBadges(newBadges);

  // Set proper focus
  if (focusedBadgeIndex === null) {
    focusFilterInput();
    scrollInputIntoView();
  } else if (newBadges.length === 0) {
    setFocusedBadgeIndex(null);
    focusFilterInput();
  } else {
    let newFocusIndex = focusedBadgeIndex > index ? index - 1 : index;
    newFocusIndex = Math.min(newFocusIndex, newBadges.length - 1);
    setFocusedBadgeIndex(newFocusIndex);
  }
};
```

packages/vscode-extension/src/network/utils/networkLogFormatters.ts
pFornagiel: An oversight on my site, removed in afd574f2839ed7071f6c9527b55fd57f9b199dc1

packages/vscode-extension/src/network/components/NetworkBar.tsx
pFornagiel: Changed in faf47118

```tsx
<VscodeCheckbox
            onChange={toggleInvert}
            className="invert-checkbox"
            label="Invert"
            checked={filterInvert}
          />
```

packages/vscode-extension/src/network/utils/networkLogUtils.ts
pFornagiel: Changed in b9cd4173, thanks for headsup.

```ts
const compare = (a: NetworkLog, b: NetworkLog) => {
  const aValue = getNetworkLogValue(a, column);
  const bValue = getNetworkLogValue(b, column);
  let comparison = 0;

  // Handle numeric sorting for Status, Size, and Time columns
  switch (column) {
    case NetworkLogColumn.Status: {
      const aNum = parseInt(aValue) || 0;
      const bNum = parseInt(bValue) || 0;
      comparison = aNum - bNum;
      break;
    }
    case NetworkLogColumn.Size: {
      const aSize = a.encodedDataLength ?? 0;
      const bSize = b.encodedDataLength ?? 0;
      comparison = aSize - bSize;
      break;
    }
    case NetworkLogColumn.Time: {
      const aTime = a.timeline?.durationMs ?? 0;
      const bTime = b.timeline?.durationMs ?? 0;
      comparison = aTime - bTime;
      break;
    }
    default: {
      // String comparison for other columns
      comparison = aValue.localeCompare(bValue);
      break;
    }
  }

  return direction === SortDirection.Asc ? comparison : -comparison;
};
```

packages/vscode-extension/src/network/utils/networkLogUtils.ts
pFornagiel: Changed in b9cd4173, thanks for headsup.

```ts
const compare = (a: NetworkLog, b: NetworkLog) => {
  const aValue = getNetworkLogValue(a, column);
  const bValue = getNetworkLogValue(b, column);
  let comparison = 0;

  // Handle numeric sorting for Status, Size, and Time columns
  switch (column) {
    case NetworkLogColumn.Status: {
      const aNum = parseInt(aValue) || 0;
      const bNum = parseInt(bValue) || 0;
      comparison = aNum - bNum;
      break;
    }
    case NetworkLogColumn.Size: {
      const aSize = a.encodedDataLength ?? 0;
      const bSize = b.encodedDataLength ?? 0;
      comparison = aSize - bSize;
      break;
    }
    case NetworkLogColumn.Time: {
      const aTime = a.timeline?.durationMs ?? 0;
      const bTime = b.timeline?.durationMs ?? 0;
      comparison = aTime - bTime;
      break;
    }
    default: {
      // String comparison for other columns
      comparison = aValue.localeCompare(bValue);
      break;
    }
  }

  return direction === SortDirection.Asc ? comparison : -comparison;
};
```

jwajgelt (approved):

PR #1480, opened 2025-08-25
feat: send files UI

jwajgelt:
This PR is a continuation of #1453 adding a UI whenever a user dragges files over the simulator preview.
It adapts the UI changes made in #1456 but moves file transfer state management to legend state.

The UI mimics in style the debug overlay interface.

🚨 Important! You need to hold shift to drag files onto the panel
VSCode displays a "Hold shift to drop into editor" message when Preview is in the Editor Tab
but it doesn't show this when Preview is in the Side Panel

https://github.com/user-attachments/assets/7963a836-72de-449c-ab6e-c43cd7d58c32

WIP

### DIFF
--- a/packages/vscode-extension/src/webview/components/SendFilesOverlay.tsx
+++ b/packages/vscode-extension/src/webview/components/SendFilesOverlay.tsx
@@ -22,7 +22,7 @@ export function SendFilesOverlay() {
   const sentFiles = use$(store$.fileTransfer.sentFiles);
   const isLoading = sendingFiles.length > 0;
   const isError = erroredFiles.length > 0;
-  const isSuccess = sentFiles.length > 0;
+  const isSuccess = !isError && sentFiles.length > 0;
   const isVisible = isDragging || isLoading || isError || isSuccess;
 
   const resetOverlayState = useCallback(() => {
@@ -113,7 +113,8 @@ export function SendFilesOverlay() {
             <span className="codicon codicon-error error-icon"></span>
           </div>
         ),
-        message: erroredFiles[0].errorMessage || "Failed to send files. Check logs for details.",
+        message:
+          erroredFiles[0].errorMessage || "Failed to send some files. Check logs for details.",
       };
     }
 


filip131311 (changes requested):
this PR seems to break the debugger overlay: 

Uploading Screen Recording 2025-08-25 at 15.25.20.mov…

packages/vscode-extension/src/webview/components/SendFilesOverlay.tsx
```diff
@@ -0,0 +1,157 @@
+import { useState, useMemo, useEffect, useCallback } from "react";
+import { VscodeProgressRing } from "@vscode-elements/react-elements";
+import "./Preview.css";
+import "./SendFilesOverlay.css";
+import { use$ } from "@legendapp/state/react";
+import classNames from "classnames";
+import { useProject } from "../providers/ProjectProvider";
+import { useSelectedDeviceSessionState } from "../hooks/selectedSession";
+
+const RETAIN_SUCCESS_SCREEN = 1000; // ms
```
filip131311: nit: I would prefer 1 * 1000 and 5 * 1000

filip131311: If i am nitting already I would also prefer RETAIN_ERROR_SCREEN_TIMEOUT

packages/vscode-extension/src/webview/components/SendFilesOverlay.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const resetOverlayState = useCallback(() => {
+    store$.fileTransfer.erroredFiles.set([]);
+    store$.fileTransfer.sentFiles.set([]);
+  }, []);
+
+  // Hide overlay after success and error animations
+  useEffect(() => {
+    if (!isLoading && (erroredFiles.length > 0 || sentFiles.length > 0)) {
+      const timer = setTimeout(
+        resetOverlayState,
```
filip131311, line 37: is there a way of calling it as a user? In fackt is there a way of not waiting for the load at all?

jwajgelt: No.
We assume the overlay will stay there short enough to not be too annoying.
At the worst, if the user feels stuck, they can restart the device.

packages/vscode-extension/src/project/FileTransfer.ts
```diff
@@ -0,0 +1,116 @@
...
+    this.stateManager.dispose();
+    if (this.tempDir) {
+      fs.promises.rm(this.tempDir, { recursive: true }).catch((_e) => {
+        /* silence the errors, it's fine */
+      });
+    }
+  }
+
+  public async sendFile(filePath: string) {
+    getTelemetryReporter().sendTelemetryEvent("device:send-file", {
```
filip131311, line 27: I was trying to move all telemetry that is "user request related" to the project class so we have a clear idea of the covrage. I know that we can not do that for how long builds last and stuff like that but if it is part of the API it should be reported in the project

filip131311 (approved):
I left one more comment, but except the flash at the end It works now

filip131311 (changes requested):
If you try to send more then one file at once error message stays up for shorter amount of time

packages/vscode-extension/src/webview/components/SendFilesOverlay.tsx
```diff
@@ -0,0 +1,157 @@
...
+export function SendFilesOverlay() {
+  const { project } = useProject();
+  const [isDragging, setIsDragging] = useState(false);
+  const store$ = useSelectedDeviceSessionState();
+  const sendingFiles = use$(store$.fileTransfer.sendingFiles);
+  const erroredFiles = use$(store$.fileTransfer.erroredFiles);
+  const sentFiles = use$(store$.fileTransfer.sentFiles);
+  const isLoading = sendingFiles.length > 0;
+  const isError = erroredFiles.length > 0;
+  const isSuccess = sentFiles.length > 0;
```
filip131311: this is incorrect way of checking for success as only part of the payload might fail

packages/vscode-extension/src/webview/components/SendFilesOverlay.tsx
```diff
@@ -0,0 +1,157 @@
...
+// VSCode displays a "Hold shift to drop into editor" message when Preview is in the Editor Tab
+// but it doesn't show this when Preview is in the Side Panel
+export function SendFilesOverlay() {
+  const { project } = useProject();
+  const [isDragging, setIsDragging] = useState(false);
+  const store$ = useSelectedDeviceSessionState();
+  const sendingFiles = use$(store$.fileTransfer.sendingFiles);
+  const erroredFiles = use$(store$.fileTransfer.erroredFiles);
+  const sentFiles = use$(store$.fileTransfer.sentFiles);
+  const isLoading = sendingFiles.length > 0;
```
filip131311, line 23: the problem with white overlay flashing at the end seem sto be this prop still being true, why can't we just reset it with the rest?

filip131311 (changes requested):
one more qestion

filip131311 (approved):

PR #1481, opened 2025-08-25
refactor: application state

filip131311:
I this PR we continue to move the project state to the new stateManager, this time by moving application session! 

With that change we do not loose the state of the application when switching devices. 

### How Has This Been Tested: 

Test: 
- logCounter
- react profiling 
- cpu profiling 
- outline renders
- network inspector 
- bundle errors 
- disabling inspector when application has non standard size 
- device rotation 
- devtools disconnected alert 

internal change

### DIFF
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -110,11 +110,11 @@ function Preview({
 
   const isRunning = selectedDeviceSession?.status === "running";
 
-  const isRefreshing = use$(
-    isRunning ? selectedDeviceSessionState.applicationSession.isRefreshing : false
+  const isRefreshing = use$(() =>
+    isRunning ? selectedDeviceSessionState.applicationSession.isRefreshing.get() : false
   );
-  const debugPaused = use$(
-    isRunning ? selectedDeviceSessionState.applicationSession.isDebuggerPaused : false
+  const debugPaused = use$(() =>
+    isRunning ? selectedDeviceSessionState.applicationSession.isDebuggerPaused.get() : false
   );
 
   const previewURL = selectedDeviceSession?.previewURL;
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/common/State.ts
```diff
@@ -93,6 +152,7 @@ export type ScreenCaptureState = {
 // #region Device Session
 
 export type DeviceSessionStore = {
+  applicationSession: ApplicationSessionState;
```
jwajgelt, line 155: This should be optional.
Right now we only have the application session state when the `DeviceSession.status` is `running`

packages/vscode-extension/src/webview/components/Preview.tsx
```diff
@@ -96,21 +109,26 @@ function Preview({
...
 
   const isRunning = selectedDeviceSession?.status === "running";
-  const isRefreshing = isRunning && selectedDeviceSession.isRefreshing;
-  const debugPaused = isRunning && selectedDeviceSession.isDebuggerPaused;
+
+  const isRefreshing = use$(
+    isRunning ? selectedDeviceSessionState.applicationSession.isRefreshing : false
+  );
+  const debugPaused = use$(
+    isRunning ? selectedDeviceSessionState.applicationSession.isDebuggerPaused : false
```
jwajgelt: Shouldn't these use the function form of `use$`?

```suggestion
    () => isRunning ? selectedDeviceSessionState.applicationSession.isRefreshing : false
  );
  const debugPaused = use$(
    () => isRunning ? selectedDeviceSessionState.applicationSession.isDebuggerPaused : false
```

filip131311: I don't know if there is a diffrence as long as isRunning is causing rerenders but I'll change it for the future

packages/vscode-extension/src/common/State.ts
filip131311: I don't think it has to be optional we just ignor this state when it is not needed (doesn't make sense)

PR #1482, opened 2025-08-25
fix: network inspector - payload and response tabs

pFornagiel:
### Description

This PR fixes issues with Payload Tab in network inspector, which would cause error throws if the payload was not a proper JSON.

It also introduces:
- Refactors made to Payload Tab to be more in style of Response Tab, included *Copy to Clickboard* button as well
- Deserialising data in `Network plugin` for proper showcase of data sent as UInt8Array

Screenshot 2025-08-25 at 15 50 19

### How Has This Been Tested: 

- **check if Payload and Response tabs in Network Inspector work correctly and do not cause crashes**

Not applicable.

### DIFF
--- a/packages/vscode-extension/lib/network.js
+++ b/packages/vscode-extension/lib/network.js
@@ -103,7 +103,6 @@ function deserializeDataContent(data, contentType) {
     return data;
   }
 
-  
   // Handle native typed Uint8Arrays
   if (data instanceof Uint8Array) {
     return shouldDecodeAsText(contentType) ? decode(data) : dataToBase64(data);
@@ -155,13 +154,15 @@ export function setup() {
 
   let enabled = false;
   messageBridge.addMessageListener("cdp-message", (message) => {
-    if (message.method === "Network.enable" && !enabled) {
-      enabled = true;
-      enableNetworkInspect(messageBridge);
-    } else if (message.method === "Network.disable" && enabled) {
-      enabled = false;
-      disableNetworkInspect();
-    }
+    try {
+      if (message.method === "Network.enable" && !enabled) {
+        enabled = true;
+        enableNetworkInspect(messageBridge);
+      } else if (message.method === "Network.disable" && enabled) {
+        enabled = false;
+        disableNetworkInspect();
+      }
+    } catch (error) {}
   });
 }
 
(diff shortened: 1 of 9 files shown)


packages/vscode-extension/lib/network.js
```diff
@@ -44,6 +45,78 @@ function readResponseBodyContent(xhr) {
...
+      typeof obj.length === "number" &&
+      Object.keys(obj).every((key) => !isNaN(parseInt(key)))
+    );
+  };
+
+  const dataToBase64 = (array) => {
+    return btoa(String.fromCharCode.apply(null, Array.from(array)));
+  };
+  const decode = (array) => {
+    return new TextDecoder().decode(array);
```
jwajgelt, line 90: This throws an error
```
ReferenceError: Property 'TextDecoder' doesn't exist
```

Sadly, Hermes (the RN javascript engine) only provides an implementation for `TextEncoder`, not `TextDecoder`. We'll have to use some polyfill for this if we want to accept utf-8 encoded strings.

jwajgelt (changes requested):

packages/vscode-extension/lib/network.js
jwajgelt, line 1: We should make sure no errors are thrown (and not caught) here, to avoid throwing errors in the users' applications.

packages/vscode-extension/lib/network.js
pFornagiel: Added a local polyfill in dfab8cc4

`lib/polyfills/index.js`
```ts
module.exports = {
  get TextDecoder() {
    if (typeof global.TextDecoder !== "undefined") {
      return global.TextDecoder;
    }
    return require("./third-party/text_decoder").TextDecoder;
  },
};
```

`network.js`
```ts
const TextDecoder = require("./polyfills").TextDecoder;
```

packages/vscode-extension/lib/network.js
pFornagiel: Added try-catch blocks in 554804b3

packages/vscode-extension/lib/polyfills/index.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  get TextDecoder() {
+    if (typeof global.TextDecoder !== "undefined") {
```
jwajgelt: Why this over 

```suggestion
    if (global.TextDecoder !== undefined) {
```
?

packages/vscode-extension/lib/network.js
```diff
@@ -179,115 +180,133 @@ function enableNetworkInspect(networkProxy) {
...
+          .then((body) => {
+            try {
+              networkProxy.sendMessage(
+                "cdp-message",
+                JSON.stringify({
+                  id: message.id,
+                  result: { body },
+                })
+              );
+            } catch (error) {}
```
jwajgelt: Any error here should get caught by the outer `.catch(error => {})`

jwajgelt (approved):

packages/vscode-extension/lib/polyfills/index.js
pFornagiel: Changed in 70c67c35 to suggested

packages/vscode-extension/lib/network.js
pFornagiel: Removed unneeded try-catch in c45ae78f

PR #1491, opened 2025-08-27
fix: no feedback while reading file to be sent

jwajgelt:
Adds feedback while the webview is reading the file to be sent to the device.

### DIFF
--- a/packages/vscode-extension/src/webview/components/SendFilesOverlay.tsx
+++ b/packages/vscode-extension/src/webview/components/SendFilesOverlay.tsx
@@ -7,6 +7,7 @@ import classNames from "classnames";
 import _ from "lodash";
 import { useProject } from "../providers/ProjectProvider";
 import { useSelectedDeviceSessionState } from "../hooks/selectedSession";
+import { merge } from "../../common/Merge";
 
 const RETAIN_SUCCESS_SCREEN_TIMEOUT = 1000; // ms
 const RETAIN_ERROR_SCREEN_TIMEOUT = 5000; // ms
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/webview/components/SendFilesOverlay.tsx
```diff
@@ -17,22 +18,36 @@ export function SendFilesOverlay() {
...
+  const isLoading = inProgressCount > 0;
+  const isError = erroredFiles.length > 0;
+  const isSuccess = !isLoading && !isError && sentFiles.length > 0;
+
   const isVisible = isDragging || isLoading || isError || isSuccess;
 
   const resetOverlayState = useCallback(() => {
-    store$?.fileTransfer.erroredFiles.set([]);
-    store$?.fileTransfer.sentFiles.set([]);
+    store$.fileTransfer.erroredFiles.set([]);
```
filip131311: Is it always called with `store$` defined, because if its not it will recreate the whole tree to set the property 
in this case if `store$ === undefined ` if would send a set with changes to the path 
```
[(...), deviceSessions, null, fileTransfer, erroredFiles]
```

filip131311: https://legendapp.com/open-source/state/v3/usage/observable/#undefined

jwajgelt: [eb84312](https://github.com/software-mansion/radon-ide/pull/1491/commits/eb84312a69674f3c4077294a755ee8973ff11a94) checks if the state is `undefined` in the `.set()` callback.

PR #1497, opened 2025-08-28
refactor: split Devtools into Server and Connection

jwajgelt:
Splits the current `Devtools` class into `DevtoolsServer` (a server the RN application connects to) and `DevtoolsConnection` (an active connection between a running app and Radon, created when the app connects to the `DevtoolsServer`).

This is motivated by two things:
- we may wish to reuse the same `DevtoolsServer` between different `Application`/`DeviceSession`s. We're planning on reusing the same Metro server for all currently running applications, but because we use Metro to provide the port the application should connect to React DevTools with, we need to run a single Server for all apps using that Metro instance.
- we may want to move the React DevTools communication from the WebSocket implementation we currently use to CDP, which is the default for the new RN DevTools (RN >0.75). This abstracts the communication channel away from how Radon uses the RDT to talk with the app.

### How Has This Been Tested: 
- open expo-52-prebuild-with-plugins test app in Radon
- check that the devtools (Network, Redux, React Query) still work
- check that element inspection still works
- open an app running Expo Go on Android
- check that debugger works correctly after JS Reloads

Internal change

### DIFF
--- a/packages/vscode-extension/src/project/applicationSession.ts
+++ b/packages/vscode-extension/src/project/applicationSession.ts
@@ -50,9 +50,7 @@ function waitForAppReady(inspectorBridge: RadonInspectorBridge, cancelToken?: Ca
   cancelToken?.onCancel(() => {
     reject(new CancelError("Cancelled while waiting for the app to be ready."));
   });
-  const appReadyListener = inspectorBridge.onEvent("appReady", () => {
-    resolve();
-  });
+  const appReadyListener = inspectorBridge.onEvent("appReady", resolve);
   promise.finally(() => {
     appReadyListener.dispose();
   });
--- a/packages/vscode-extension/src/project/devtools.ts
+++ b/packages/vscode-extension/src/project/devtools.ts
@@ -39,7 +39,6 @@ export class DevtoolsInspectorBridge extends BaseInspectorBridge implements Disp
 
   constructor(devtoolsServer: DevtoolsServer) {
     super();
-    this.devtoolsConnection = devtoolsServer.connection;
     if (devtoolsServer.connection) {
       this.setupBridge(devtoolsServer.connection);
     }
@@ -73,9 +72,10 @@ export class DevtoolsInspectorBridge extends BaseInspectorBridge implements Disp
 }
 
 export class DevtoolsConnection implements Disposable {
-  bridge: FrontendBridge;
-  store: Store;
-  connected: boolean = true;
+  private bridge: FrontendBridge;
+  private store: Store;
+
+  public connected: boolean = true;
 
   private readonly ideMessageEventEmitter: EventEmitter<IdeMessage> = new EventEmitter();
   private readonly disconnectedEventEmitter: EventEmitter<void> = new EventEmitter();
@@ -153,7 +153,7 @@ export abstract class DevtoolsServer implements Disposable {
   private _connection: DevtoolsConnection | undefined;
 
   protected setConnection(connection: DevtoolsConnection | undefined) {
-    this._connection?.dispose();
+    this._connection?.disconnect();
     this._connection = connection;
     if (connection) {
       this.connectionEventEmitter.fire(connection);
@@ -204,15 +204,15 @@ class WebSocketDevtoolsServer extends DevtoolsServer implements Disposable {
         },
       };
 
-      const session = new DevtoolsConnection(wall);
+      const connection = new DevtoolsConnection(wall);
       ws.on("close", () => {
-        session.disconnect();
+        connection.disconnect();
       });
       ws.on("error", () => {
-        session.disconnect();
+        connection.disconnect();
       });
 
-      super.setConnection(session);
+      super.setConnection(connection);
     });
   }
 


packages/vscode-extension/src/project/devtools.ts
```diff
@@ -152,11 +82,90 @@ export class Devtools extends BaseInspectorBridge implements Disposable {
...
+      address !== null && typeof address !== "string",
+      "The address is an instance of `AddressInfo`"
+    );
+    const serverPort = address.port;
+    Logger.info(`Devtools started on port ${serverPort}`);
+    resolve(serverPort);
+  });
+
+  const port = await promise;
+  return { port, devtoolsServer };
```
kmagiera: wouldn't it be cleaner for the port to be accessible via the devtoolsServer instance. It is being passed around along with the server in multiple places. In some places we only pass the port number, which is ok, but in most of the other places it looks like it complicates things (we need to fields, await two times to get server and then port etc)

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -94,11 +103,16 @@ export class ApplicationSession implements Disposable {
...
+        device.launchApp(buildResult, metro.port, devtoolsPort, launchArguments)
       );
 
       onLaunchStage(StartupMessage.WaitingForAppToLoad);
-      await cancelToken.adapt(Promise.race([devtools.appReady(), bundleErrorPromise]));
+      const appReadyPromise = new Promise<void>((resolve, reject) => {
+        devtoolsServer.onConnection((devtools) => {
+          devtools.appReady.then(resolve, reject);
+        });
+      });
```
kmagiera: I think this actually changes the behavior and can introduce some issues during concurrent startup.

Before we'd always get appReady promise resolved as it was captured by the server and the first incoming connection that happened there. 

After this change, we wait for a new connection to be established before awaiting app ready event.

The difference is that now, if the device is already connected before we reach this place, we will just wait indefinitely (until reload), while before, the appReady promise would resolve immediately letting the code pass.

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -125,13 +139,25 @@ export class ApplicationSession implements Disposable {
...
     private readonly device: DeviceBase,
     private readonly metro: MetroLauncher,
-    private readonly devtools: Devtools,
+    private readonly devtoolsServer: DevtoolsServer,
     private readonly packageNameOrBundleId: string,
     private readonly supportedOrientations: DeviceRotation[]
   ) {
-    this.registerDevtoolsListeners();
+    this.disposables.push(
+      this.devtoolsServer.onConnection((devtools) => {
```
kmagiera, line 160: we again here rely on the connection being established after the instance of this class is made. This could be an important change comparing to the previous version of the code.
I wonder whether the application session shouldn't be associated with a single devtools connection?

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -159,15 +185,15 @@ export class ApplicationSession implements Disposable {
   // #region Tools
 
   public async updateToolEnabledState(toolName: ToolKey, enabled: boolean) {
-    this.toolsManager.updateToolEnabledState(toolName, enabled);
+    this.toolsManager?.updateToolEnabledState(toolName, enabled);
```
kmagiera: it may be worth printing some message here instead of silently swallowing the error when toolsManager isn't set (same for the below methods)

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -344,70 +370,73 @@ export class ApplicationSession implements Disposable {
...
         }
       ),
-      this.devtools.onEvent("disconnected", () => {
+      devtools.onEvent("disconnected", () => {
         if (
           this.stateManager.getState().inspectorBridgeStatus === InspectorBridgeStatus.Connected
         ) {
           this.stateManager.setState({ inspectorBridgeStatus: InspectorBridgeStatus.Disconnected });
         }
+        this.devtools = undefined;
```
kmagiera: shouldn't we check here whether `this.devtools == devtools` – seems like this could happen as a result of a new devtools instance connecting in which case we will reset the newly assigned devtools instead of the old ones.

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: > application session shouldn't be associated with a single devtools connection

It's not, and it shouldn't. When a new devtools connection is made, the application session should switch to using that connection, which is what this listener does.
With the Websocket devtools connection, the app reconnects on JS reloads and whenever the in-app dev-menu (the one you open by pressing "D" in the regular emulator/simulator) is opened.
With the Fusebox/CDP connection, it can also happen when the debugger connection breaks and the debugger reconnects.

The listener in `ApplicationSession` `constructor` is fine, since it's set up before the application is launched on the device. The one in `ApplicationSession.launch` should be moved before the `device.launchApp` call.

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: Like I mentioned in another comment, the idea is you shouldn't reach this place "if the device is already connected", since this is the function that actually launches the app. To me, this also means that, if there already was a connection made, we wouldn't want to `await` its `appReady` promise anyway, since it would be related to an "old" Application Session.
That said, since this listener is created _after_ `device.launchApp` is `await`ed here, there is a risk of the connection being made before this listener is set up. Moving it before the call should be a sufficient fix.

packages/vscode-extension/src/project/devtools.ts
jwajgelt: `port` is specific to the `WebsocketDevtoolsServer`, so adding it to the `DevtoolsServer` interface doesn't make sense.
That said, we can probably just return the `WebsocketDevtoolsServer` from here and have the `port` property on that one.
Then, `DeviceSession` can use the concrete `Websocket` implementation instead of the `DevtoolsServer` abstraction, and access the port that way.

packages/vscode-extension/src/project/applicationSession.ts
kmagiera: what I'm concerned the most with this change is that we now have onconnection listeners in a couple of different places. I'd feel safer if we only have one place where we wait for new connections. This way we can be sure that the other places are not fragile to timing issues like the one I described above

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -152,11 +82,93 @@ export class Devtools extends BaseInspectorBridge implements Disposable {
...
+      address !== null && typeof address !== "string",
+      "The address is an instance of `AddressInfo`"
+    );
+    const serverPort = address.port;
+    Logger.info(`Devtools started on port ${serverPort}`);
+    resolve(serverPort);
+  });
+
+  const port = await promise;
+  const devtoolsServer = new WebSocketDevtoolsServer(server, port);
```
kmagiera: nit: but we don't actually need to pass the port separately as it is accessible via server instance given we already await for the listen event to happen

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -69,7 +69,8 @@ export class DeviceSession implements Disposable {
   private isActive = false;
   private metro: MetroLauncher;
   private maybeBuildResult: BuildResult | undefined;
-  private devtools: Devtools;
+  private devtoolsServer: Promise<DevtoolsServer & { port: number }>;
```
kmagiera: wonder if it wouldn't be easier to just await DevtoolsServer creation instead of using promise here. The server startup should be relatively quick as the only part we wait for is binding in order to get the port number. It should make it easier to follow.

I see that we now call it in constructor which cannot be async. So the option is to either make it sync, or inject ready instance into deviceSession because `startOrActivateSessionForDevice` where we create it is already async and awaits a few things before creating an instance of session.

I'd slightly lean toward the second option as I understand we'd want to have a shared instance of the server in the future anyway?

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -125,14 +142,44 @@ export class ApplicationSession implements Disposable {
     private readonly applicationContext: ApplicationContext,
     private readonly device: DeviceBase,
     private readonly metro: MetroLauncher,
-    private readonly devtools: Devtools,
+    private readonly devtoolsServer: DevtoolsServer,
     private readonly packageNameOrBundleId: string,
     private readonly supportedOrientations: DeviceRotation[]
```
kmagiera, line 157: don't like the fact that both device and application sessions own and maintain its own devtools instance. Especially that device session own application session. What is the purpose of that? Would it make sense for these to "share" that instance (meaning they won't both listed to onconnection calls). Or perhaps we need some abstraction for "devtools session" that abstracts away individual connections but can support re-connecting upol reloads?

packages/vscode-extension/src/project/deviceSession.ts
jwajgelt: I wanted to change it in a later PR, but yes, the goal would be to inject the `DevtoolsServer` and the `MetroLauncher` instances to the `deviceSession` in the future.

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: Currently, `DeviceSession` doesn't use the `DevtoolsConnection` directly, including listening to `onconnection`.
I removed any uses of devtools from `DeviceSession` in https://github.com/software-mansion/radon-ide/pull/1497/commits/eed26fad347e759897b2f32a6d66ca5ce43fb02b
Now it uses the `InspectorBridge` it gets from the `ApplicationSession` instead, which kind of works like the "`DevtoolsSession`" you suggest.

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -33,26 +32,55 @@ import {
...
+
+function waitForAppReady(inspectorBridge: RadonInspectorBridge, cancelToken?: CancelToken) {
+  // set up `appReady` promise
+  const { promise, resolve, reject } = Promise.withResolvers<void>();
+  cancelToken?.onCancel(() => {
+    reject(new CancelError("Cancelled while waiting for the app to be ready."));
+  });
+  const appReadyListener = inspectorBridge.onEvent("appReady", () => {
+    resolve();
+    appReadyListener.dispose();
```
kmagiera: this is already called in `finally` callback below

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -152,11 +122,109 @@ export class Devtools extends BaseInspectorBridge implements Disposable {
...
+
+  function onErrorCallback(error: Error) {
+    Logger.error("Devtools server error:", error);
+    reject(new Error(`Could not start the React Devtools server`));
+  }
+  server.on("error", onErrorCallback);
+
+  await listenPromise;
+
+  server.off("error", onErrorCallback);
```
kmagiera: we could keep the error callback on in order to get the errors logged at some later time (past initialization)

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -152,11 +122,109 @@ export class Devtools extends BaseInspectorBridge implements Disposable {
     return promise;
   }
 
+  public send(event: string, payload: unknown) {
+    this.wall.send(event, payload);
+  }
+
+  public close() {
+    this.disconnectedEventEmitter.fire();
```
kmagiera: this isn't a disconnection event but "close" event. When the connection is disconnected as a result of error of the other end dropping the connection, this listener won't get fired.

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -2,126 +2,96 @@ import http from "http";
...
-      listener.dispose();
+type IdeMessageListener = <K extends keyof RadonInspectorBridgeEvents>(event: {
+  type: K;
+  data: RadonInspectorBridgeEvents[K];
+}) => void;
+type IdeMessage = Parameters<IdeMessageListener>[0];
+
+/**
+ * InspectorBridge implementation that uses the React DevTools frontend to receive messages from the application.
+ * It connects to the application via the last DevtoolsConnection provided by a DevtoolsServer.
```
kmagiera: this isn't precise. Similarly to my other comments we rely on this code being run before any devtools server connection is established. The comment says that it uses "last", which suggest that the one that happened before the instance was created. It also doesn't explain that when new connection arrives it switches to that one

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -822,17 +823,17 @@ export class DeviceSession implements Disposable {
   }
 
   public openNavigation(id: string) {
-    this.inspectorBridge.sendOpenNavigationRequest(id);
+    this.inspectorBridge?.sendOpenNavigationRequest(id);
```
kmagiera, line 826: looks like we don't actually use `inspectorBridge` in this class apart from accessing it for these action handlers. Given we only access `inspectorBridge` now via `applicationSession`, maybe we can just delegate this remaining few calls to `applicationSession` as we already do with the other handlers like `resumeDebugger`, `openTool` etc?

jwajgelt: We also use it for `navigationChanged` and `navigationRouteListUpdated` events, since we want to keep the navigation history between application restarts.
I don't want to move the rest in this PR, since it's already a large refactor.

packages/vscode-extension/src/project/devtools.ts
jwajgelt: > The comment says that it uses "last", which suggest that the one that happened before the instance was created. It also doesn't explain that when new connection arrives it switches to that one

By "the last DevtoolsConnection" I meant "the latest" or "the newest" connection returned by the Server _after_ the `InspectorBridge` was created. It was clear to me when I was writing this comment. I'll try to rephrase it.

jwajgelt: > Similarly to my other comments we rely on this code being run before any devtools server connection is established.

Yes, but I don't think that's an issue when we always create it before launching the application.

jwajgelt: > Similarly to my other comments we rely on this code being run before any devtools server connection is established.

I caved. I added a `connection` property to the `Server` that `InspectorBridge` uses on construction.

packages/vscode-extension/src/project/devtools.ts
jwajgelt: Pretty sure the "close" event should fire even when the connection closes due to an error (at least the spec suggests so)
https://websockets.spec.whatwg.org/#closeWebSocket
but I added an "error" handler anyway.

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -33,26 +32,54 @@ import {
...
+
+function waitForAppReady(inspectorBridge: RadonInspectorBridge, cancelToken?: CancelToken) {
+  // set up `appReady` promise
+  const { promise, resolve, reject } = Promise.withResolvers<void>();
+  cancelToken?.onCancel(() => {
+    reject(new CancelError("Cancelled while waiting for the app to be ready."));
+  });
+  const appReadyListener = inspectorBridge.onEvent("appReady", () => {
+    resolve();
+  });
```
kmagiera:
```suggestion
  const appReadyListener = inspectorBridge.onEvent("appReady", resolve);
```

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -152,11 +126,126 @@ export class Devtools extends BaseInspectorBridge implements Disposable {
...
+          return () => {
+            ws.off("message", listener);
+          };
+        },
+        send(event, payload) {
+          ws.send(JSON.stringify({ event, payload }));
+        },
+      };
+
+      const session = new DevtoolsConnection(wall);
```
kmagiera: can this be called `connection`, we already have a mix of sessions and connections and it really confuses me

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -152,11 +126,126 @@ export class Devtools extends BaseInspectorBridge implements Disposable {
...
+      ws.on("close", () => {
+        session.disconnect();
+      });
+      ws.on("error", () => {
+        session.disconnect();
+      });
```
kmagiera: wonder whether this shouldn't be moved to `DevtoolsConnection` this way hiding the details of `disconnect` implementation that also calls into dispose. This code here relies on the dispose to be called but since we don't call it directly we need to know internal impl detail that this method is called there.

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -152,11 +126,126 @@ export class Devtools extends BaseInspectorBridge implements Disposable {
...
+}
 
-  protected send(message: any) {
-    this.socket?.send(JSON.stringify({ event: "RNIDE_message", payload: message }));
+export abstract class DevtoolsServer implements Disposable {
+  private readonly connectionEventEmitter: EventEmitter<DevtoolsConnection> = new EventEmitter();
+  private _connection: DevtoolsConnection | undefined;
+
+  protected setConnection(connection: DevtoolsConnection | undefined) {
+    this._connection?.dispose();
```
kmagiera: this is a bit inconsistent. We sometimes call `disconnect` on the connection instance and in other places we call `dispose`. Then `disconnect` actually calls into `dispose`, so maybe we shouldn't expose `disconnect` at all?

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -2,126 +2,100 @@ import http from "http";
...
+ * InspectorBridge implementation that uses the React DevTools frontend to receive messages from the application.
+ */
+export class DevtoolsInspectorBridge extends BaseInspectorBridge implements Disposable {
+  private devtoolsConnection: DevtoolsConnection | undefined;
+  private devtoolsServerListener: Disposable;
+  private devtoolsConnectionListeners: Disposable[] = [];
+
+  constructor(devtoolsServer: DevtoolsServer) {
+    super();
+    this.devtoolsConnection = devtoolsServer.connection;
```
kmagiera: `setupBridge` already sets `this.devtoolsConnection`

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -2,126 +2,100 @@ import http from "http";
...
-    const { resolve, promise } = Promise.withResolvers<void>();
-    const listener = this.onEvent("appReady", () => {
-      resolve();
-      listener.dispose();
-    });
-    return promise;
-  }
+  private setupBridge = (connection: DevtoolsConnection) => {
+    this.devtoolsConnection = connection;
+    disposeAll(this.devtoolsConnectionListeners);
```
kmagiera, line 50: just wondering whether we really need to dispose these. It adds a bit of boilerplate and in the end of the day those listeners are owned by the connection instance that is being disposed, so I'd expect it to dispose the listeners registered with it too, no?

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -125,14 +154,44 @@ export class ApplicationSession implements Disposable {
...
-    private readonly devtools: Devtools,
+    private readonly devtoolsServer: DevtoolsServer,
     private readonly packageNameOrBundleId: string,
     private readonly supportedOrientations: DeviceRotation[]
   ) {
-    this.registerDevtoolsListeners();
+    this.disposables.push(
+      this.devtoolsServer.onConnection((devtools) => {
+        this.devtools?.dispose();
+        this.devtools = devtools;
```
kmagiera, line 162: I know this won't be possible now but perhaps worth following the same flow here and to check `server.connection` here along with setting the listener?

packages/vscode-extension/src/project/devtools.ts
jwajgelt: `DevtoolsConnection` doesn't know what transport it uses.

> This code here relies on the dispose to be called but since we don't call it directly we need to know internal impl detail

In my book, it's ok to depend on the implementation here, since it's all the same module -- the classes/functions in this module don't need to hide their implementation from each other, only from callers importing them from outside.

packages/vscode-extension/src/project/devtools.ts
jwajgelt: This should probably call `disconnect`.
I'll take another look.

packages/vscode-extension/src/project/devtools.ts
jwajgelt: The connection can theoretically outlive the `InspectorBridge`, in which case the listeners won't get cleaned up and will retain the `Bridge` after it should have been cleaned up.

PR #1500, opened 2025-08-29
feat: move to the fusebox RDT over CDP implementation

jwajgelt:
Implements the React Devtools connection using the debugger rather than a Websocket connection, following the RN devtools implementation.

- uses the new RDT-over-CDP implementation for 

### How Has This Been Tested: 
- open app in Radon (test 0.75, >=0.76, an expo-go app)
- check that element inspector still works
- check that tool plugins work
- check that devtools work after JS reload
- check that setting `"useOldDevtools": false` in launch config for newer RN versions falls back to the websocket devtools server

Mostly internal changes.
We may want to document the new launch config option.

### DIFF
--- a/packages/vscode-extension/lib/react_devtools_agent.js
+++ b/packages/vscode-extension/lib/react_devtools_agent.js
@@ -17,10 +17,14 @@ const setDevtoolsAgent = (newDevtoolsAgent) => {
     return;
   }
   devtoolsAgent = newDevtoolsAgent;
+  // TODO: clean these up if the devtools agent changes
   devtoolsAgent._bridge.addListener("RNIDE_message", (message) => {
     if (agent.onmessage) {
       agent.onmessage(message);
     }
+  });
+  devtoolsAgent._bridge.addListener("shutdown", () => {
+    devtoolsAgent = undefined;
   });
   const messages = messageQueue;
   messageQueue = [];
@@ -32,9 +36,8 @@ const setDevtoolsAgent = (newDevtoolsAgent) => {
 const hook = window.__REACT_DEVTOOLS_GLOBAL_HOOK__;
 if (hook.reactDevtoolsAgent) {
   setDevtoolsAgent(hook.reactDevtoolsAgent);
-} else {
-  hook.on("react-devtools", setDevtoolsAgent);
 }
+hook.on("react-devtools", setDevtoolsAgent);
 
 globalThis.__radon_agent = agent;
 
--- a/packages/vscode-extension/lib/react_devtools_agent.js
+++ b/packages/vscode-extension/lib/react_devtools_agent.js
@@ -14,6 +14,7 @@ const agent = {
 
 const setDevtoolsAgent = (newDevtoolsAgent) => {
   if (!newDevtoolsAgent) {
+    devtoolsAgent = undefined;
     return;
   }
   devtoolsAgent = newDevtoolsAgent;
--- a/packages/vscode-extension/lib/react_devtools_agent.js
+++ b/packages/vscode-extension/lib/react_devtools_agent.js
@@ -30,8 +30,6 @@ const setDevtoolsAgent = (newDevtoolsAgent) => {
     if (devtoolsAgent === newDevtoolsAgent) {
       devtoolsAgent = undefined;
     }
-    bridge.removeListener("RNIDE_message", onIdeMessage);
-    bridge.removeListener("shutdown", onBridgeShutdown);
   }
 
   bridge.addListener("RNIDE_message", onIdeMessage);
--- a/packages/vscode-extension/lib/react_devtools_agent.js
+++ b/packages/vscode-extension/lib/react_devtools_agent.js
@@ -18,19 +18,29 @@ const setDevtoolsAgent = (newDevtoolsAgent) => {
     return;
   }
   devtoolsAgent = newDevtoolsAgent;
-  // TODO: clean these up if the devtools agent changes
-  devtoolsAgent._bridge.addListener("RNIDE_message", (message) => {
+  const bridge = devtoolsAgent._bridge;
+
+  function onIdeMessage(message) {
     if (agent.onmessage) {
       agent.onmessage(message);
     }
-  });
-  devtoolsAgent._bridge.addListener("shutdown", () => {
-    devtoolsAgent = undefined;
-  });
+  }
+
+  function onBridgeShutdown() {
+    if (devtoolsAgent === newDevtoolsAgent) {
+      devtoolsAgent = undefined;
+    }
+    bridge.removeListener("RNIDE_message", onIdeMessage);
+    bridge.removeListener("shutdown", onBridgeShutdown);
+  }
+
+  bridge.addListener("RNIDE_message", onIdeMessage);
+  bridge.addListener("shutdown", onBridgeShutdown);
+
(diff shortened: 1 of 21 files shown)


packages/vscode-extension/package.json
```diff
@@ -618,6 +618,10 @@
               "usePrebuild": {
                 "type": "boolean",
                 "description": "When set, Radon will run Expo Prebuild before building the application. Only applies to applications using Expo CLI."
+              },
+              "useOldDevtools": {
```
filip131311, line 622: this is not really clear for the user. I believe that the word "devtools" would bring to mind network inspector or redux tool, not the way radon internally comunicates with the application. 

This should be explaind clearly to the user, 
more over what are the setupas that require it? can we detect them? Even if don't know them can we have an easy way to add the detection in case we are made aware of them. Is there a specific type of crash that we might use to prompt users to try and change that setting?

packages/vscode-extension/package.json
```diff
@@ -840,6 +844,10 @@
...
```
filip131311, line 848: same

packages/vscode-extension/lib/react_devtools_agent.js
```diff
@@ -14,27 +14,39 @@ const agent = {
...
   messages.forEach((message) => {
-    devtoolsAgent._bridge.send("RNIDE_message", message);
+    bridge.send("RNIDE_message", message);
   });
 };
 
 const hook = window.__REACT_DEVTOOLS_GLOBAL_HOOK__;
 if (hook.reactDevtoolsAgent) {
   setDevtoolsAgent(hook.reactDevtoolsAgent);
-} else {
```
filip131311, line 35: why was it not needed befor but we need to listen for it all the time here? 

can you leave a comment in code?

packages/vscode-extension/src/common/Project.ts
```diff
@@ -155,8 +155,8 @@ export const StartupStageWeight = [
   { StartupMessage: StartupMessage.Building, weight: 7 },
   { StartupMessage: StartupMessage.Installing, weight: 1 },
   { StartupMessage: StartupMessage.Launching, weight: 1 },
-  { StartupMessage: StartupMessage.WaitingForAppToLoad, weight: 6 },
   { StartupMessage: StartupMessage.AttachingDebugger, weight: 1 },
+  { StartupMessage: StartupMessage.WaitingForAppToLoad, weight: 6 },
```
filip131311: hmmmm when the metro progress is shown now if you switch it around? does debugger not need to wait for it?

packages/vscode-extension/src/project/ApplicationContext.ts
```diff
@@ -23,8 +25,18 @@ export type ResolvedLaunchConfig = LaunchOptions & {
...
   usePrebuild: boolean;
+  useOldDevtools: boolean;
 };
 
+function checkFuseboxSupport(appRoot: string): boolean {
+  const reactNativePackage = requireNoCache("react-native/package.json", {
+    paths: [appRoot],
+  });
+  const reactNativeVersion = new SemVer(reactNativePackage.version);
+  const supportsFusebox = reactNativeVersion.compare("0.76.0") >= 0;
```
filip131311, line 36: is it not also disabled by some env var, or configuration in pod file? can we check for that?

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -309,12 +312,12 @@ export class DeviceSessionsManager implements Disposable {
     if (previousSession === session) {
       return;
     }
+    await previousSession?.deactivate();
     if (session === undefined) {
```
filip131311: if session is undefied won't the old one be dissposed anyway? why do we need to deactivate it

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -124,18 +132,16 @@ export class ApplicationSession implements Disposable {
...
-        if (!hasBundleError) {
-          onLaunchStage(StartupMessage.AttachingDebugger);
-          await cancelToken.adapt(activatePromise);
-        }
+        onLaunchStage(StartupMessage.AttachingDebugger);
+        await cancelToken.adapt(Promise.race([activatePromise, bundleErrorPromise]));
+      }
+
+      const hasBundleError = stateManager.getState().bundleError !== null;
+      if (!hasBundleError && getIsActive()) {
```
filip131311, line 142: this change makes metro bundling process unobsarvable for the user (I thinnk)

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -199,9 +185,47 @@ export class ApplicationSession implements Disposable {
...
+      }
+      if (this.stateManager.getState().inspectorBridgeStatus === InspectorBridgeStatus.Connected) {
+        this.stateManager.setState({
+          inspectorBridgeStatus: InspectorBridgeStatus.Disconnected,
+        });
+      }
+      this.devtools = undefined;
+      this._inspectorBridge.setDevtoolsConnection(undefined);
+    });
+    devtools.onProfilingChange((isProfiling) => {
```
filip131311, line 211: where is it handled otherwise?

jwajgelt: I don't think I understand the question.

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: It seems to work fine.

packages/vscode-extension/src/project/applicationSession.ts
filip131311: when dev tools do not exist

packages/vscode-extension/src/project/applicationSession.ts
filip131311: it does not it stays on "attaching debugger" forever and never shows the percentage bundling progres

packages/vscode-extension/lib/react_devtools_agent.js
jwajgelt: Previously the devtools would connect to the websocket once, on initial bundle execution.
Now, it can happen whenever the Extension-side implementation requests it, so we need to have the listener in case the extension wants to reconnect.
I think commenting in code is excessive -- it's clear what the code does, it updates the stored `reactDevtoolsAgent` whenever the hook emits the event.

jwajgelt: If you didn't see the diff, just the final code, you wouldn't bat an eye at the listener.

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: Then you can't start profiling.

packages/vscode-extension/lib/react_devtools_agent.js
filip131311: 👍

packages/vscode-extension/package.json
jwajgelt: I think this is only necessary if you mess with React Native's feature flags, which I _hope_ no one does.
On newer versions of RN, the flag doesn't even seem to exist anymore.
I can add to the description which flag it's related to.

PR #1502, opened 2025-09-01
feat: introduce test coverage for app interactions

KeyJayY:
### Description

This PR adds tests for interactions with applications running inside Radon. For now, these tests use a temporary app available here: [react-native-apps](https://github.com/KeyJayY/react-native-apps). In the future, the applications from the main test apps repository ([radon-ide-test-apps](https://github.com/software-mansion-labs/radon-ide-test-apps)) will be modified to work with these tests.  

### How Has This Been Tested

Tests were run both locally and on a virtual machine (Github actions runners). In most cases, all tests pass successfully. 

To Reproduce:

There are two ways to run the tests locally:

1. Directly on your machine (warning: this will remove devices added in Radon). To do this, run:
   `npm run prepare-and-run-tests`

2. On a local virtual machine:
   - First, set up the environment by running:
     `npm run set-up-VM-environment`
     (This downloads a preconfigured macOS system image, so you need about 80 GB of free storage)
   - Then, run the tests with:
     npm run run-tests-on-VM

To run the tests on a GitHub Actions runner:
- Go to the Actions tab in GitHub
- Select the workflow `Build VSIX and run tests (macOS)`
- Choose the branch `@KeyJayY/AppInteractions`

Not documented yet

### DIFF
--- a/.github/workflows/ui-tests.yaml
+++ b/.github/workflows/ui-tests.yaml
@@ -114,13 +114,13 @@ jobs:
       - name: Run setup-run-tests
         run: npm run setup-run-tests -- ${{ github.event.inputs.testsToRun }} --exclude ${{ github.event.inputs.testsToExclude }}
         working-directory: packages/vscode-extension-tester
-        timeout-minutes: 1
         env:
           PROJECT_NAME: ${{ github.event.inputs.testApp }}
           IS_RECORDING: ${{ github.event.inputs.isRecording }}
 
       - name: make video
-        if: ${{ !cancelled() && github.event.inputs.isRecording == 'true' }}
+        if: ${{ always() && github.event.inputs.isRecording == 'true' }}
+        timeout-minutes: 1
         run: ffmpeg -framerate 10 -i frame-%04d.png -vf "scale=trunc(iw/2)*2:trunc(ih/2)*2" -c:v libx264 -pix_fmt yuv420p vscode-ui-recording.mp4
         working-directory: packages/vscode-extension-tester/videos
 
@@ -134,6 +134,7 @@ jobs:
       - name: build output
         if: always()
         uses: actions/upload-artifact@v4
+        timeout-minutes: 1
         with:
           name: output
           path: packages/vscode-extension-tester/output.txt
@@ -141,6 +142,7 @@ jobs:
       - name: Upload video recording
         if: ${{ always() && github.event.inputs.isRecording == 'true' }}
         uses: actions/upload-artifact@v4
+        timeout-minutes: 1
         with:
           name: vscode-ui-recording
           path: packages/vscode-extension-tester/videos/vscode-ui-recording.mp4


.github/workflows/ui-tests.yaml
```diff
@@ -11,28 +11,28 @@ on:
...
+          # - "expo-51-dev-client"
+          # - "expo-51"
+          # - "expo-52-eas"
+          # - "expo-52-prebuild-with-plugins"
+          # - "expo-53"
+          # - "expo-54"
+          # - "expo-go"
+          # - "expo-router"
+          # - "nx-monorepo"
+          # - "pnpm-monorepo"
```
kacperkapusciak, line 35: i a perfect world the folders would be dynamically fetched from software-mansion-labs/radon-ide-test-apps as they this dropdown would get outdated really quickly. 

but this is way beyond this PR so it's okay

.github/workflows/ui-tests.yaml
```diff
@@ -112,8 +119,8 @@ jobs:
           IS_RECORDING: ${{ github.event.inputs.isRecording }}
 
       - name: make video
-        if: ${{ github.event.inputs.isRecording == 'true' }}
-        run: ffmpeg -framerate 10 -i frame-%04d.png -c:v libx264 -pix_fmt yuv420p vscode-ui-recording.mp4
+        if: ${{ always() && github.event.inputs.isRecording == 'true' }}
```
kacperkapusciak, line 122: hmmm i'm not sure if i'm a fan of `always()`

```
Avoid using **always** for any task that could suffer from a critical failure, for example: getting sources, otherwise the workflow may hang until it times out. If you want to run a job or step regardless of its success or failure, use the recommended alternative: if: ${{ !cancelled() }}
```

https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always

packages/vscode-extension-tester/scripts/downloadRepo.sh
```diff
@@ -4,13 +4,14 @@ set -e
...
 if [ -z "$1" ]; then
-  APP="react-native-77"
+  APP="react-native-81"
 fi
 
 
 FOLDER_NAME="$APP"
 TARGET_DIR="./data/react-native-app"
-REPO_URL="https://github.com/software-mansion-labs/radon-ide-test-apps.git"
+# temporary repo with react native apps (one app so far)
```
kacperkapusciak: there's nothing more permanent than a temporary solution. if i were you i'd add `FIXME:` at the beginning of this pr just to be sure we'll eventually change this ;) 
```suggestion
# FIXME: temporary repo with react native apps (one app so far)
```

packages/vscode-extension-tester/ui-tests/04-buildingApp.test.js
```diff
@@ -0,0 +1,166 @@
...
+    ({ driver, view, workbench } = get());
+
+    await deleteAllDevices(driver);
+    await addNewDevice(driver, "newDevice");
+    try {
+      await findAndClickElementByTag(driver, `modal-close-button`);
+    } catch {}
+
+    await waitForAppToLoad(driver);
+    await toggleShowTouches(driver);
```
kacperkapusciak: all the functions  `findAndWaitForElementByTag`, `deleteAllDevices`, `addNewDevice` beg to be grouped into a class where the class so you didn't have to pass the driver as argument each time you want to use it

```
const s = new Service(driver);

await s.deleteAllDevices()
await s.addNewDevice()
```

the name i took could be better tho

packages/vscode-extension-tester/utils/helpers.js
```diff
@@ -64,3 +65,38 @@ export async function waitUntilElementGone(
...
+  return (await editor.getCoordinates())[0];
+}
+
+export async function getFileNameInEditor(driver) {
+  const editor = new TextEditor();
+  const fullPath = await editor.getFilePath();
+  return path.basename(fullPath);
+}
+
+export async function getDebuggerStopLineNumber(driver) {
```
kacperkapusciak: the `driver` argument is unused in any of these 3 functions

packages/vscode-extension-tester/utils/helpers.js
kacperkapusciak, line 1: why no typescript? 😟

packages/vscode-extension/src/webview/components/Debugger.tsx
```diff
@@ -5,7 +5,7 @@ import IconButton from "./shared/IconButton";
 function Debugger() {
   const { project } = useProject();
   return (
-    <div className="debugger-container">
+    <div className="debugger-container" data-test="app-debugger-container">
```
kacperkapusciak:
```suggestion
    <div className="debugger-container" data-testid="app-debugger-container">
```

i believe there's a well established convention that this is called `data-testid` in the industry

https://testing-library.com/docs/queries/bytestid/

packages/vscode-extension-tester/utils/helpers.js
```diff
@@ -64,3 +65,38 @@ export async function waitUntilElementGone(
     message
   );
 }
+
+export async function getCursorLineInEditor(driver) {
+  const editor = new TextEditor();
+  return (await editor.getCoordinates())[0];
```
kacperkapusciak: i believe it would be easier to read when we split this into several lines

packages/vscode-extension-tester/ui-tests/interactions.js
```diff
@@ -163,17 +220,87 @@ export async function modifyDeviceName(driver, deviceName, modifiedDeviceName) {
...
-
+  deviceNameInput.click();
+  await deviceNameInput.sendKeys(Key.chord(Key.COMMAND, "a"), Key.BACK_SPACE);
+  deviceNameInput.clear();
   await driver.wait(async () => {
-    const value = await deviceNameInput.getAttribute("value");
-    return value === "";
-  }, 3000);
-  deviceNameInput.sendKeys(modifiedDeviceName);
+    return (await deviceNameInput.getAttribute("value")) === "";
```
kacperkapusciak: it used to be more readable when defined in 2 lines

packages/vscode-extension-tester/ui-tests/interactions.js
```diff
@@ -163,17 +220,87 @@ export async function modifyDeviceName(driver, deviceName, modifiedDeviceName) {
...
+  await driver.wait(
+    async () => {
+      try {
+        const el = await driver.findElement(
+          By.css('[data-test="phone-screen"]')
+        );
+        if (await el.isDisplayed()) {
+          return el;
+        }
+      } catch {}
```
kacperkapusciak: try catch pass? don't we want to terminate the test when this fails?

packages/vscode-extension-tester/ui-tests/interactions.js
```diff
@@ -163,17 +220,87 @@ export async function modifyDeviceName(driver, deviceName, modifiedDeviceName) {
...
+        );
+        if (await popup.isDisplayed()) {
+          findAndClickElementByTag(driver, "alert-open-logs-button");
+          return popup;
+        }
+      } catch {}
+
+      return false;
+    },
+    600000,
```
kacperkapusciak: 10 minutes to wait for the app to load on CI is a bit excessive i think

.github/workflows/ui-tests.yaml
KeyJayY: This is a known issue with GitHub workflow dispatch inputs. I found this discussion: [https://github.com/orgs/community/discussions/12029](https://github.com/orgs/community/discussions/12029?utm_source=chatgpt.com), but I couldn’t find any workaround.

packages/vscode-extension-tester/ui-tests/interactions.js
KeyJayY: In the previous code, a thrown error meant that the element was not found, which indicated that the app had not loaded yet. However, I agree that using a try-catch pass might be misleading, so I changed it to make the code more readable.

packages/vscode-extension-tester/ui-tests/interactions.js
KeyJayY: I'm aware of the problem, usually, it doesn’t take that long. The main slowdown is the first build of the application, which takes quite a long time. Right now, the whole test set takes about 11 mins to finish — that includes four app launches, one of them from scratch. I also stop waiting if an error popup shows up.

I don’t know a better solution right now, maybe you have any idea?

PR #1506, opened 2025-09-04
feat: network inspector - add context menu with copy, sort and filter options

### LINKED ISSUE #1237: copy curl, request and response of a request in network tab
**Description**
Is there any possibility of being able to copy curl of a request in network tab, I was using reactotron previously, but while using radon ide the requests are not visible in it, so it will be so helpf ful if somehow we can get an option to copy the curl of the request and json of response and request

**Environment**

- Version of currently used Radon IDE 1.7.1
- Version of React Native/Expo used in the project. React Native 0.77.1, Expo ~52.0.41

pFornagiel:
### Description

This PR introduces a right-click **Context Menu** in the Network Inspector with following functionalities:
- `Copy` - opens a submenu with different copy options for the request
  image

- `Sort` - opens a submenu with sorting options. The behaviour is the same as clicking a header - the sorting in one column will cycle upon subsequent clicks: `Ascending -> Descending -> None`
  Screenshot 2025-09-04 at 09 18 48

- `Filter` - opens if needed and forwards focus to the Filter input field

- When triggered on element of table other than row (for example, table header), the context menu will only show `Sort` and `Filter` options.
  image

  https://github.com/user-attachments/assets/568f2841-5644-427d-8d04-96ed8b9ba9ba

### Fixes #1237
The PR fixes  / adds features mentioned in #1237

### How Has This Been Tested: 

- **right click on logs, see how context menu behaves, use different `copy` options to make sure formatting works correctly**

WIP.


packages/vscode-extension/src/network/utils/sharedUtils.ts
kacperkapusciak, line 1: i feel like names like `utils`, `common`, `helper` are an antipattern. they can host from everything to nothing.  `sharedUtils` even more so.

https://lostechies.com/chrismissal/2009/06/01/anti-patterns-and-worst-practices-utils-class/

packages/simulator-server
kacperkapusciak, line 1: was it intentional?

packages/vscode-extension/src/network/components/ContextMenu/ContextMenuItems.tsx
```diff
@@ -0,0 +1,240 @@
...
+    await copyToClipboard(responseJson);
+  };
+
+  const handleCopyRequestJson = async () => {
+    if (!networkLog) {
+      return;
+    }
+    const requestJson = getRequestJson(networkLog);
+    await copyToClipboard(requestJson);
+  };
```
kacperkapusciak: it's basically the same code 6 times. it could be maybe simplified by creating a higher order function taking a callback as an argument 🤔

packages/vscode-extension/src/network/utils/contextMenuUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (Object.keys(headers).length > 0) {
+    fetchCode += `,\n  headers: ${JSON.stringify(headers, null, 4).replace(/^/gm, "  ")}`;
+  }
+
+  // body for requests
+  if (postData !== null && postData !== undefined) {
+    try {
+      // try to format as JSON if it's valid
+      const parsedData = JSON.parse(postData);
+      fetchCode += `,\n  body: JSON.stringify(${JSON.stringify(parsedData, null, 4).replace(/^/gm, "  ")})`;
```
kacperkapusciak: `JSON.stringify(XYZ, null, 4).replace(/^/gm, "  ")`
could be extracted to some one-liner function as it's repeated twice

packages/vscode-extension/src/network/utils/requestFormatUtils.ts
```diff
@@ -26,3 +28,78 @@ export function formatUrlParams(url: string): string {
...
+    headers: log.request.headers || {},
+  };
+
+  if (log.request.postData) {
+    try {
+      requestData.body = JSON.parse(log.request.postData);
+    } catch {
+      requestData.body = log.request.postData;
+    }
+  }
```
kacperkapusciak:
```suggestion
  const { method, url, headers = {}, postData } = log.request;

  const requestData: Record<string, unknown> = {
    method,
    url,
    headers,
  };

  if (postData) {
    try {
      requestData.body = JSON.parse(postData);
    } catch {
      requestData.body = postData;
    }
  }
```
that's a little less verbose i guess

packages/vscode-extension/src/network/utils/requestFormatUtils.ts
```diff
@@ -26,3 +28,78 @@ export function formatUrlParams(url: string): string {
...
+
+  return JSON.stringify(requestData, null, 2);
+}
+
+export function getResponseJson(log: NetworkLog): string {
+  if (!log.response) {
+    return JSON.stringify({ error: "No response data" }, null, 2);
+  }
+
+  const responseData: Record<string, unknown> = {
```
kacperkapusciak: same here

kacperkapusciak (commented):
in some of the themes there are still separator lines visible
image

pFornagiel:
> in some of the themes there are still separator lines visible image

Removed separators in 825c14f9.

Tough they would look alright, as we've actually used them in the url menu, but given the theme you provided they feel very out of place.

packages/vscode-extension/src/network/utils/sharedUtils.ts
pFornagiel: > i feel like names like `utils`, `common`, `helper` are an antipattern. they can host from everything to nothing. `sharedUtils` even more so.

> https://lostechies.com/chrismissal/2009/06/01/anti-patterns-and-worst-practices-utils-class/

> that's a little less verbose i guess

> same here

Restructured the logic in 3d19f6dc

pFornagiel:
> i feel like names like `utils`, `common`, `helper` are an antipattern. they can host from everything to nothing. `sharedUtils` even more so.

> https://lostechies.com/chrismissal/2009/06/01/anti-patterns-and-worst-practices-utils-class/

> that's a little less verbose i guess

> same here

Restructured the logic in 3d19f6dc

pFornagiel:
> basically the same code 6 times. it could be maybe simplified by creating a higher order function taking a callback as an argument 🤔

Introduced higher-order function in 08f66364, which avoids redundancy.

```ts
const parseAndCopy = async (parseLog: (networkLog: NetworkLog) => string) => {
  if (!networkLog) {
    return;
  }
  const clipboardText = parseLog(networkLog);
  await copyToClipboard(clipboardText);
};
```

packages/simulator-server
pFornagiel: It was unintentional, did not bump the simulator-server on my side and pushed by accident. Fixed in 7e1027ee.

packages/vscode-extension/src/network/utils/sharedUtils.ts
pFornagiel: Agreed, made bigger changes in https://github.com/software-mansion/radon-ide/commit/3d19f6dc5bfb3fab01707791c14898cbca15e8fc, c2acf7a4 and 2657beeb to `utils` folder:
- renamed `requestFormatUtls` to `requestFormatters`, as it gives a clearer idea of the file contents
- got rid of `contextMenuUtils` file, instead changed naming and moved the functions, which were parsing the requests into the `requestFormatters` file
- exctracted the clipboard-copy functionality to `clipboard` file, for the sake of separation of concerns
- split `NetworkLogUtils` into `NetworkLogsSort` and `networkLogParsers` for descriptive file names, extracted NETWORK_LOG_COLUMNS into `types` folder and added `as const` declaration.

I believe the changes made the structure more readable, thanks for pointing that out

packages/vscode-extension/src/network/components/ContextMenu/ContextMenuItems.tsx
pFornagiel: Answered below as well, extracted the common logic into higher-order function in https://github.com/software-mansion/radon-ide/commit/08f66364ce81b2f6c6997f70dccc3e7881a58641 to avoid this repetition. Did not pay much attention to the issue after it was working, so thanks for that.

```ts
const parseAndCopy = async (parseLog: (networkLog: NetworkLog) => string) => {
  if (!networkLog) {
    return;
  }
  const clipboardText = parseLog(networkLog);
  await copyToClipboard(clipboardText);
};

const handleCopyCurl = () => parseAndCopy(createCurlCommand);

const handleCopyFetch = () => parseAndCopy(createFetchCommand);

const handleCopyUrl = () => parseAndCopy(getUrl);

const handleCopyResponseJson = () => parseAndCopy(getResponseDetails);

const handleCopyRequestJson = () => parseAndCopy(getRequestDetails);

const handleCopyRequestPayload = () => parseAndCopy(getRequestPayload);

const getResponseBodyAsJsonString = () => {
  if (responseBody === null || typeof responseBody !== "string") {
    return "{}";
  }
  return formatRequestBody(responseBody);
};

  const handleCopyResponseBody = () => parseAndCopy(getResponseBodyAsJsonString);
```

packages/vscode-extension/src/network/utils/contextMenuUtils.ts
pFornagiel: Made changes to the code structure in https://github.com/software-mansion/radon-ide/commit/3d19f6dc5bfb3fab01707791c14898cbca15e8fc:
- extracted common logic and used existing functions, in places previously overlooked,
- renamed functions to be more descriptive
- extracted constants from object where needed to avoid unnecessary chaining for brevity, introduced some constants to not rely on magic strings

packages/vscode-extension/src/network/utils/requestFormatUtils.ts
pFornagiel: Made changes to the code structure in https://github.com/software-mansion/radon-ide/commit/3d19f6dc5bfb3fab01707791c14898cbca15e8fc:
- extracted common logic and used existing functions, in places previously overlooked,
- renamed functions to be more descriptive
- extracted constants from object where needed to avoid unnecessary chaining for brevity, introduced some constants to not rely on magic strings

packages/vscode-extension/src/network/utils/requestFormatUtils.ts
pFornagiel: Answered and talked about details in above comments, https://github.com/software-mansion/radon-ide/commit/3d19f6dc5bfb3fab01707791c14898cbca15e8fc

pFornagiel:
> I wonder if you'd considered using context menus from vscode-elements? They use the built-in vscode API for displaying context menus which transform into the macos native menus. Not sure if they support colored icons (or icons altogether) but the benefit is that they can show beyond the iframe viewport which may be particularly important for the network panel which would typically only occupy a small portion of the screen

Discussed in the office, vscode-elements internally uses [Lit](https://lit.dev) for its components, but it's still standard web API, without any native vscode functionality. Manipulating native context-menu was not something I considered possible and have yet found a way to introduce it, but in the future it is to be researched. 

Came to the conclusion together that current implementation is fine for the time being.

PR #1507, opened 2025-09-04
refactor: Make `OutputChannelRegistry` a global singleton

latekvo:
As discussed in #1386 and in-person, storing `OutputChannelRegistry` in `IDE` class is unnecessary while making things complicated. This PR makes the `OutputChannelRegistry` a global singleton.

### How Has This Been Tested: 

All logs and custom output channels store and display the logs correctly.


packages/vscode-extension/src/project/OutputChannelRegistry.ts
```diff
@@ -5,25 +5,46 @@ import { createReadableOutputChannel, ReadableLogOutputChannel } from "./Readabl
...
   }
 
-  dispose() {
-    this.channelByName.entries().forEach(([k, c]) => {
-      c.dispose();
-    });
+  public dispose() {
+    this.channelByName.values().forEach((channel) => channel.dispose());
+    this.channelByName.clear();
+    OutputChannelRegistry.instance = null;
```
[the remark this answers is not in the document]
latekvo: Invalid suggestion

kmagiera (changes requested):
I feel like it'd be better to keep the instance-based reference as much as possible. The registry needs to be accessible as singleton because we want to be able to use it as soon as the extension starts and in particular we want to use in the Logger module. Other than that, all the other components can still keep using instances. It'd provide a cleaner module boundaries for them. In the ideal world we'd have separate output channels for different device sesssion, but due to the way channels are handled by VSCode this isn't practical at the given moment. But we already have a good abstraction in place and I don't see that we need to expose the fact how output channels are implemented into the individual components.

To summarize:
1) we still want the singleton based interface
2) we will access the singleton interface only at the top layer and defer to session-specific instances
3) we want to have some nice abstration about handling dispose with channels. Maybe instead of having `getOrCreate` we could just have it being `created` at each time in which case the previous channel will be truncated.

packages/vscode-extension/src/connect/Connector.ts
```diff
@@ -228,7 +232,7 @@ export class Connector implements Disposable {
 
   public static getInstance(): Connector {
     if (!Connector.instance) {
-      Connector.instance = new Connector();
+      Connector.instance = new Connector(OutputChannelRegistry.getInstance());
```
filip131311, line 235: I am not sure what the original intention of this PR was but the final result seems to be mixing up to approaches

Why do you inject OotputChanel instance sometimes but call the static `getInstance` other times

Can we either always pas it down as a dependency or always get it from the static method? 
why the mixed approach?

latekvo: I initially used just the static getter, but from what I understood from  [this](https://github.com/software-mansion/radon-ide/pull/1507#pullrequestreview-3251016313) request, we want to be using the passed-down reference when possible, and the static getter when the other approach is not possible. In case of this Connector class, there wasn't a passed-down reference available, so I used the static getter.

I provided more details about this change [here](https://github.com/software-mansion/radon-ide/pull/1507#issuecomment-3351790565#:~:text=Just%20to%20note)

latekvo:
@kmagiera just to be clear before I make major changes, by "we discussed this" and "this change" are you referring to this PR, or to the method of accessing `OutputChannelRegistry`? 

If it's the latter, do you want me to revert back to the static-only approach, or try harder to get the pass-down approach working? The static approach is the one I prefer the most, but I'm not certain if that is what you said here.

latekvo:
This PR initially removed `IDE`'s dependance on `OutputChannelRegistry`, which did clear up quite a few things. This was however reverted in [f2aa6b4](https://github.com/software-mansion/radon-ide/pull/1507/commits/f2aa6b439d806597a370eba35879fa646a08e60c), so I agree that right now there isn't really a point to this PR.

PR #1511, opened 2025-09-08
refactor: Use direct communication between webviews

latekvo:
Replaced websocket-based comms with direct comms using `window.postMessage`, `vscode.postMessage`, and `window.addEventListener("message", ...)`.

This PR is a simplifies adding a `view_network_logs` tool to Radon AI.

### How Has This Been Tested: 

Every feature of the network tab should be working as it used to before this PR.

### DIFF
--- a/packages/vscode-extension/src/plugins/network/network-plugin.ts
+++ b/packages/vscode-extension/src/plugins/network/network-plugin.ts
@@ -31,6 +31,7 @@ async function initialize() {
 interface CDPMessage {
   method: string;
   params: unknown;
+  id?: string;
 }
 
 type BroadcastListener = (message: CDPMessage) => void;
@@ -57,6 +58,7 @@ export class NetworkPlugin implements ToolPlugin {
   handleWebviewMessage(event: WebviewCDPMessage) {
     if (event.command === WebviewCommand.CDPCall) {
       this.sendCDPMessage({
+        id: event.id,
         method: event.method,
         params: event.params ?? {},
       });
--- a/packages/vscode-extension/src/network/types/cdp.ts
+++ b/packages/vscode-extension/src/network/types/cdp.ts
@@ -13,4 +13,5 @@ export interface WebviewCDPMessage {
   command: WebviewCommand.CDPCall;
   method: string;
   params: Record<string, unknown>;
+  id: string;
 }


packages/vscode-extension/src/webview/utilities/communicationTypes.ts
pFornagiel, line 1: NIT: I would consider moving the types to `src/network/types` directory, as they are used directly in relation to the Network Panel communication, and then getting rid of `Types` in the name -> something like `communication.ts` or `panelMessageProtocol.ts`

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -101,44 +43,50 @@ export class NetworkPlugin implements ToolPlugin {
...
+    commands.executeCommand("setContext", `RNIDE.Tool.Network.available`, true);
+
+    this.devtoolsListeners.push(
+      this.inspectorBridge.onEvent("pluginMessage", (payload) => {
+        if (payload.pluginId === "network") {
+          this.messageListeners.forEach((cb) => cb(payload.data));
+        }
+      })
+    );
+
```
pFornagiel:
```suggestion

  private devtoolsListeners: Disposable[] = [];
  private broadcastListeners: BroadcastListener[] = [];

  constructor(private readonly inspectorBridge: RadonInspectorBridge) {
    initialize();
  }

  private sendCDPMessage(messageData: CDPMessage) {
    this.inspectorBridge.sendPluginMessage("network", "cdp-message", messageData);
  }

  public onPanelMessage(event: WebviewCDPMessage) {
    if (event.command === WebviewCommand.CDPCall) {
      this.sendCDPMessage({
        method: event.method,
        params: event.params ?? {},
      });
    }
  }

  addBroadcastListener(cb: BroadcastListener): Disposable {
    this.broadcastListeners.push(cb);
    return new Disposable(() => {
      let index = this.broadcastListeners.indexOf(cb);
      if (index !== -1) {
        this.broadcastListeners.splice(index, 1);
      }
    });
  }

```

Maybe we could direct the handling logic of the event caught in `NetworkDevtoolsWebviewProvider` to the `network-plugin` itself? The change in the provider could look like this:

```ts
//...
this.messageListenerDisposable = webview.onDidReceiveMessage((event: WebviewCDPMessage) => {
   networkPlugin.onPanelMessage(event);     
  });
```

Additionally, the changes in naming suggested is more declarative of logic imo.

packages/vscode-extension/src/webview/utilities/communicationTypes.ts
latekvo: Done in [a906c5e](https://github.com/software-mansion/radon-ide/pull/1511/commits/a906c5e04d4e74163c9be89a804c155b7a7bb58f) & [159e4ab](https://github.com/software-mansion/radon-ide/pull/1511/commits/159e4ab9392e2d55f80a28f831a659d087be1015). Placed it in `src/network/types/cdp.ts` (i didn't notice your `communication.ts` suggestion before committing).

packages/vscode-extension/src/plugins/network/network-plugin.ts
latekvo: Renamed `messageListeners` -> `broadcastListeners` in [df67e4a](https://github.com/software-mansion/radon-ide/pull/1511/commits/df67e4aa40b9f5d49b6147b6fada32489f40ce24).  

Correct me if i'm wrong, but i think we want to use `on*` (or `add*Listener`) functions to supply listeners, and something like `handle*` to supply the events to the listeners. So instead of `onPanelMessage(event)` we should have `handlePanelMessage(event)` (I actually like `handleWebviewMessage` even better, since we already have `WebviewCommand` and `WebviewCDPMessage`). 

As a side-note, it's not a strong opinion, but I prefer `on*` notation over `add*Listener` notation. The former is shorter, and i think more conventional in our repo. But I acknowledge that the latter is also a popular convention overall, so it's not a strong opinion.

>Maybe we could direct the handling logic of the event caught in NetworkDevtoolsWebviewProvider to the network-plugin itself? The change in the provider could look like this:

Done in [d80a74a](https://github.com/software-mansion/radon-ide/pull/1511/commits/d80a74aab99557d710c6814d82f3a3214e0886b7).

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -101,44 +43,59 @@ export class NetworkPlugin implements ToolPlugin {
...
-    this.inspectorBridge.sendPluginMessage("network", "cdp-message", messageData);
-  };
+  handleWebviewMessage(event: WebviewCDPMessage) {
+    if (event.command === WebviewCommand.CDPCall) {
+      this.sendCDPMessage({
+        method: event.method,
+        params: event.params ?? {},
+      });
+    }
+  }
```
pFornagiel:
```suggestion
  handleWebviewMessage(event: WebviewCDPMessage) {
    if (event.command === WebviewCommand.CDPCall) {
      this.sendCDPMessage({
        id: event.id,
        method: event.method,
        params: event.params ?? {},
      });
    }
  }
```

Right now, we do not send the `id` of the event via sendCDPMessage (which maps to `requestId` on the application side), which causes `getResponseBody` to never resolve the request. 

Pass the `id: event.id` to `sendCDPMessage` object and update types accordingly.

latekvo: Fixed in [f9922f9](https://github.com/software-mansion/radon-ide/pull/1511/commits/f9922f986fc4e5371e5cdeab9d27b99340b1f45c). Tested, and the response body is retrieved correctly now.

pFornagiel: Can confirm, it's now fixed, thanks.

PR #1514, opened 2025-09-09
feat: network inspector - update response body handling logic

pFornagiel:
### Description

This PR revamps the response body storing, sending and handling both on the application and extension side.
Main changes:
- In the `lib/network.js` replaced the `WeakRef` approach to taking care of unused request bodies with cache mechanism using new utility class - `AsyncBoundedResponseBuffer`.

  The buffer stores truncated response bodies, which will be sent to the frontend upon request. Truncation takes place during the `put` operation on the buffer and is meant to reduce the size of text responses sent to the frontend through debugger websocket (the frontend now allows for viewing the full response by refetching the request instead, see more below).  Number of characters left after truncation is defined in `TRUNCATED_LENGTH` constant.
  
  Additionally, to prevent using too much memory on the application side, the buffer  has an upper memory limit (defined by `REQUEST_BUFFER_MAX_SIZE_BYTES` constant), exceeding which will get rid of oldest messages.
  
  The buffer itself offers One-Time-Access to the request bodies, just like the previous `WeakRef` approach - upon ` get` operation, the entry gets removed, as it will be stored on the frontend from that point on.

- In the frontend, added an indicator informing that the response body was truncated due to its size, both in the `Response Tab` and `Context Menu`

image

- In the frontend, added a button allowing for refetching and opening the response in the native editor window and Context Menu. If the fetch or opening the editor fail, an error message is shown to the user.

image
 
- Refactored the folder structure of Network frontend components to group Details tabs together

- Fixed an issue, where URL's name would be shown as (pending) in the network inspector, if it ended  with slash `/`

### How Has This Been Tested: 

- **make fetches (preferably greater in size), see if the information shown in the inspector and accessed through the menu is correct, check whether different types of text files are formatted correctly when prompted to open in the editor.**

Not applicable.


packages/vscode-extension/lib/AsyncBoundedResponseBuffer.ts
```diff
@@ -0,0 +1,270 @@
...
+   * preview is available, or to `undefined` when the response should not be
+   * buffered.
+   *
+   * @param xhr The XMLHttpRequest instance to parse the response from.
+   * @returns A promise resolving to an InternalResponseBodyData when a text
+   * preview is available, or `undefined` for non-parsable or uncached responses.
+   */
+  private async parseResponseBody(
+    xhr: XMLHttpRequest
+  ): Promise<InternalResponseBodyData | undefined> {
```
jwajgelt:
```suggestion
  private async readResponseText(
    xhr: XMLHttpRequest
  ): Promise<InternalResponseBodyData | undefined> {
```
- I think that "parse" is a strong word for a function that just guards that the response type is correct and reads the data if it is.
- I'd highlight that this only handles textual payloads

packages/vscode-extension/lib/AsyncBoundedResponseBuffer.ts
```diff
@@ -0,0 +1,270 @@
...
+ * All operations are asynchronous as response body reading requires async
+ * operations (e.g., reading blob content as text).
+ *
+ * The Buffer is One-Time-Access - once a response is retrieved with `get`,
+ * it is removed from the buffer to free up space. The frontend needs to fetch
+ * response body only once per request.
+ *
+ * Inspired by the C++ BoundedRequestBuffer implementation from React Native.
+ */
+class AsyncBoundedResponseBuffer {
```
jwajgelt: Right now, this class serves two roles:
- it reads the responses from `XMLHttpRequest`,
- it keeps a cache of the response payloads, up to a certain size, possibly truncating them

I think it'd be better to split these two concerns -- have the `AsyncBoundedResponseBuffer.put` accept a `Promise<ResponseBodyData>`, and pass the result of `parseResponseBody` to it.

jwajgelt: Since this class is mostly pure logic, some unit tests would be nice.

packages/vscode-extension/lib/AsyncBoundedResponseBuffer.ts
```diff
@@ -0,0 +1,270 @@
...
+   * @param xhr The XMLHttpRequest object to read response from
+   * @returns Promise<boolean> True if the response body processing was initiated successfully
+   */
+  public async put(requestId: string, xhr: XMLHttpRequest): Promise<boolean> {
+    try {
+      // Remove existing request with the same ID, if any
+      // Done to rearrange the order when re-adding the same requestId
+      if (this.responseMap.has(requestId)) {
+        this.remove(requestId);
+      }
```
jwajgelt: How can this happen?

packages/vscode-extension/lib/AsyncBoundedResponseBuffer.ts
```diff
@@ -0,0 +1,270 @@
...
+          response ? { body: response.body, wasTruncated: response.wasTruncated } : undefined
+        )
+      );
+      this.order.push(requestId);
+
+      // Handle the response when it resolves
+      const response = await responseBodyPromise;
+
+      // Check if the request was removed while we were processing
+      if (!response || !this.responseMap.has(requestId)) {
```
jwajgelt:
```suggestion
      const storedPromise = responseBodyPromise.then((response) =>
        response ? { body: response.body, wasTruncated: response.wasTruncated } : undefined
      )
      this.responseMap.set(
        requestId,
        storedPromise
      );
      this.order.push(requestId);

      // Handle the response when it resolves
      const response = await responseBodyPromise;

      // Check if the request was removed while we were processing
      if (!response || this.responseMap.get(requestId) !== storedPromise) {
```
Currently, if you're waiting for two responses with same `requestId`, the one that _loaded_ later will overwrite the other, not the one that was `put()` later.

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -8,9 +8,86 @@ import { extensionContext } from "../../utilities/extensionContext";
...
+      }
+      return "xml";
+    } else if (trimmedBody.startsWith("{") || trimmedBody.startsWith("[")) {
+      return "json";
+    }
+
+    return "text";
+  };
+
+  return guessLanguageFromText();
```
jwajgelt: Why not just

```suggestion

    const trimmedBody = body.trim();
    if (trimmedBody.startsWith("<?xml") || trimmedBody.startsWith("<")) {
      if (trimmedBody.includes("<!DOCTYPE html") || trimmedBody.includes("<html")) {
        return "html";
      }
      return "xml";
    } else if (trimmedBody.startsWith("{") || trimmedBody.startsWith("[")) {
      return "json";
    }

    return "text";
```

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -31,27 +108,87 @@ class NetworkCDPWebsocketBackend implements Disposable {
...
+
+  private async handleWebSocketMessage(
+    message: WebSocketMessageData,
+    ws: WebSocket
+  ): Promise<void> {
+    try {
+      const payload: CDPMessage = JSON.parse(message.toString());
+
+      if (payload.method === "Network.fetchFullResponseBody") {
+        await this.handleFetchFullResponseBody(payload.params?.request);
```
jwajgelt: I'm not sure it's a good idea to "pretend" this is a CDP message.
I think I'd prefer if the communication between the Network Panel and the extension differentiated between CDP messages and IDE-related ones.
Something along the lines of:
```ts
type NetworkPanelMessage {
  type: "CDP" | "openResponseInEditor"
  payload: ...
}
```
and here you'd check the `message.type` and call appropriate function.

packages/vscode-extension/lib/AsyncBoundedResponseBuffer.ts
pFornagiel: Restructured and renamed in 4104321a, it does give clearer idea about function's purpose.

```ts
async function readResponseText(
  xhr: XMLHttpRequest
): Promise<InternalResponseBodyData | undefined> {
  //...
}
```

packages/vscode-extension/lib/AsyncBoundedResponseBuffer.ts
pFornagiel: Restructured in https://github.com/software-mansion/radon-ide/commit/4104321af64926eecfd0d23c0c41d9ed37d7cc3e.

As proposed, the `AsyncBoundedResponseBuffer` now accepts `Promise<ResponseBodyData|undefined>` in the `put` method and internally only stores promises, without additionally doing the parsing.

```ts
xhr.addEventListener("loadend", (event) => {
  //...
  const responsePromise = readResponseText(xhr);
  responseBuffer.put(requestId, responsePromise);
  //...
}
```

Additionally, instead of moving the `readResponseText` back to `network.js` decided to move utility functions handling the parsing to `networkRequestParsers.ts` file, in order to cleanup `network.js` and keep typing introduced when implementing the buffer. Also moved `mimeTypeFromResponseType` and `deserializeRequestData` from `network.js` apart from the methods from the buffer.

Moved the files `network.js`, `AsyncBoundedResponseBuffer.ts`, `networkRequestParsers.ts` into the network folder to make file navigation a little cleaner.

packages/vscode-extension/lib/AsyncBoundedResponseBuffer.ts
pFornagiel: Theoretically, it should not, because we only put after `loadend` event on `XHRInterceptor`, which is synonymous with `Network.loadingFinished` CDP event, after which the request under given ID should not change its state, but it is safer to guard against such situation, when order on the buffer matters in the case of removing the items in the future.

packages/vscode-extension/lib/AsyncBoundedResponseBuffer.ts
pFornagiel: Fixed in 0aaf6e67, thank you for the catch

packages/vscode-extension/src/plugins/network/network-plugin.ts
pFornagiel: Thought it would be more self-explanatory, without writing additional comment, but this looks more natural. Implemented in e9883fc0. Additionally, added missing `application/x-javascript` mime type to PARSABLE_APPLICATION_CONTENT_TYPES in `networkRequestParsers.ts`, which was present in `determineLanguage` function checks.

packages/vscode-extension/lib/network/networkRequestParsers.ts
```diff
@@ -0,0 +1,240 @@
...
+  };
+
+  const reconstructTypedArray = (serializedData: SerializedTypedArray) => {
+    const length = Object.keys(serializedData).length;
+    const uint8Array = new Uint8Array(length);
+    Object.keys(serializedData).forEach((key) => {
+      uint8Array[parseInt(key)] = serializedData[key];
+    });
+    return uint8Array;
+  };
```
jwajgelt: NIT: can these functions be lifted to the top level? They don't seem to capture anything from this scope, and the extra nesting makes the outer function harder to read.

packages/vscode-extension/src/network/components/ContextMenu/NetworkLogContextMenu.tsx
```diff
@@ -34,6 +37,22 @@ function NetworkLogContextMenu({ children, menuItems }: NetworkLogContextMenuPro
...
+  const handleOpenChange = async (open: boolean) => {
+    // In order to prevent fetching responseBody as soon as the request log is rendered
+    // (which has memory implications on the backend), we wait until the user
+    // opens the menu for the first time
+    if (!open || !networkLog) {
+      return;
+    }
+    // Prefetch response body for copy menu when context menu opens
+    const bodyData = await getResponseBody(networkLog);
+    setResponseBodyData(bodyData);
```
jwajgelt, line 55: Can this cause `getResponseBody` to be called more than once, if the menu is opened twice in short succession?

packages/vscode-extension/src/plugins/network/network-plugin.ts
pFornagiel: Introduced this communication protocol in #1521, as it required more typing revamps than expected

packages/vscode-extension/lib/AsyncBoundedResponseBuffer.ts
pFornagiel: Introduced unit tests in 293b4866dec7a509768c9fe35af5ce02e71168d0 `/lib/network/AsyncBoundedResponseBuffer.test.ts`

packages/vscode-extension/lib/network/networkRequestParsers.ts
pFornagiel: Agreed, after moving those utils to separate file instead of `network.js` taking care of all those things, it is cleaner to keep those in the outer scope.

Changed in 04bab49b

packages/vscode-extension/src/network/components/ContextMenu/NetworkLogContextMenu.tsx
pFornagiel: Yes, this can cause it to be called more than once before caching the response, therefore - making the calls to the app, but there is a `?` guard in the `/lib/network/network.js`:
```js
function listener(message) {
  try {
    if (message.method === "Network.disable") {
      networkProxy.removeMessageListener("cdp-message", listener);
    } else if (
      message.method === "Network.getResponseBody" &&
      message.params.requestId.startsWith(requestIdPrefix)
    ) {
      const requestId = message.params.requestId;
      const responsePromise = responseBuffer.get(requestId);

      // Upon initial launch, the message gets send twice in dev, because of
      // react Strict Mode and dependency on useEffect. To be fixed in next PR's.
      responsePromise
        ?.then((responseBodyData) => {   // ? guard here - if responsePromise is undefined after `responseBuffer.get`
          networkProxy.sendMessage(      // the message will not be sent through websocket
            "cdp-message",
            JSON.stringify({
              id: message.id,
              result: responseBodyData,
            })
          );
        })
        .catch(() => {});
    }
  } catch (error) {}
}
```

In order to not leave the listeners hanging, we could implement an interval after which the listener gets terminated if the ResponseBody was already set, but this would require `ref` introduction because of closure.

`getResponseBody` in `NetworkProvider.tsx`

```ts
const getResponseBody = (networkLog: NetworkLog): Promise<ResponseBodyData | undefined> => {
    const requestId = networkLog.requestId;
    const ws = networkTracker.ws;

    if (!requestId || !ws) {
      return Promise.resolve(undefined);
    }

    if (responseBodies[requestId]) {
      return Promise.resolve(responseBodies[requestId]);
    }

    const id = Math.random().toString(36).substring(7);

    // Send the message to the network-plugin backend
    ws.send(
      JSON.stringify({
        id,
        method: "Network.getResponseBody",
        params: {
          requestId: requestId,
        },
      })
    );

    // Add a listener to capture the response
    return new Promise((resolve) => {
      const listener = (message: MessageEvent) => {
        try {
          const parsedMsg = JSON.parse(message.data);
          if (parsedMsg.id !== id) {
            return;
          }

          const bodyData = parsedMsg.result;
          setResponseBodies((prev) => ({
            ...prev,
            [requestId]: bodyData,
          }));

          resolve(bodyData);

          ws.removeEventListener("message", listener);
        } catch (error) {
          console.error("Error parsing WebSocket message:", error);
        }
      };

      ws.addEventListener("message", listener);
    });
  };
```

The returned promise could possibly be changed to:

```ts
//...
const [responseBodies, setResponseBodies] = useState<Record<string, ResponseBodyData>>({});
const responseBodiesRef = useRef(responseBodies);

// ref in sync with state
responseBodiesRef.current = responseBodies;

//...
return new Promise((resolve) => {
  const resolvePromise = (interval: NodeJS.Timeout, data: ResponseBodyData | undefined) => {
    clearInterval(interval);
    ws.removeEventListener("message", listener);
    resolve(data);
  };

  // Periodically check if response was cached by another request
  const checkCacheInterval = setInterval(() => {
    if (responseBodiesRef.current[requestId]) {
      resolvePromise(checkCacheInterval, responseBodiesRef.current[requestId]);
    }
  }, 2000);

  const listener = (message: MessageEvent) => {
    setTimeout(() => {
      if (responseBodiesRef.current[requestId]) {
        resolvePromise(checkCacheInterval, responseBodiesRef.current[requestId]);
        return;
      }

      try {
        const parsedMsg = JSON.parse(message.data);
        if (parsedMsg.id !== id) {
          return;
        }

        const bodyData = parsedMsg.result;
        setResponseBodies((prev) => ({
          ...prev,
          [requestId]: bodyData,
        }));

        resolvePromise(checkCacheInterval, bodyData);
      } catch (error) {
        console.error("Error parsing WebSocket message:", error);
      }
    }, 1000);
  };

  ws.addEventListener("message", listener);
});
```

But I am not sure whether this is needed and it is definitely more obscure than the previous approach.

packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.test.ts
```diff
@@ -0,0 +1,314 @@
...
+      const stats = buffer.getStats();
+      assert.strictEqual(stats.entryCount, 0);
+      assert.strictEqual(stats.currentSizeBytes, 0);
+    });
+
+    it("should handle concurrent removal during promise resolution", async () => {
+      const requestId = "concurrent-request";
+      let resolvePromise: (value: InternalResponseBodyData) => void;
+
+      const responsePromise = new Promise<InternalResponseBodyData>((resolve) => {
```
jwajgelt, line 255: You can use `Promise.withResolvers` to get the promise and `resolve` function directly.

packages/vscode-extension/src/network/components/ContextMenu/NetworkLogContextMenu.tsx
pFornagiel: Discussed internally, instead of implementing approach with Interval, we send the `undefined` response from app (`\lib\network\network.js`) and in the `listener` check if response is undefined and return ref value if so.

Implemented in 99c72194

```ts
//...
const responseBodiesRef = useRef<Record<string, ResponseBodyData | undefined>>({});
//...
return new Promise((resolve) => {
    const listener = (message: MessageEvent) => {
      try {
        const parsedMsg = JSON.parse(message.data);
        if (parsedMsg.id !== id) {
          return;
        }
        const bodyData: ResponseBodyData | undefined = parsedMsg.result;

        if (bodyData === undefined) {
          ws.removeEventListener("message", listener);
          resolve(responseBodiesRef.current[requestId]);
          return;
        }

        responseBodiesRef.current[requestId] = bodyData;

        resolve(bodyData);
        ws.removeEventListener("message", listener);
      } catch (error) {
        console.error("Error parsing WebSocket message:", error);
      }
    };

    ws.addEventListener("message", listener);
  });
```

packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.test.ts
pFornagiel: Discussed internally, sticking to current implementation because of `es2024` typescript problems (current `es6` config does not support `withResolvers`)

packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.test.ts
```diff
@@ -79,15 +79,15 @@ describe("AsyncBoundedResponseBuffer", () => {
...
     const retrievedPromise1 = buffer.get(requestId1);
     const retrievedPromise2 = buffer.get(requestId2);
     assert(retrievedPromise1 !== undefined);
-    assert(retrievedPromise2 === undefined);
+    assert(retrievedPromise2 !== undefined);
 
-    assert.strictEqual(result1, buffer.get(requestId1));
-    assert.strictEqual(result2, buffer.get(requestId2));
+    assert.strictEqual(undefined, buffer.get(requestId1));
+    assert.strictEqual(undefined, buffer.get(requestId2));
```
jwajgelt: The convention is to write `assert.strictEqual(actual_value, expected_value)`

```suggestion
    assert.strictEqual(buffer.get(requestId1), undefined);
    assert.strictEqual(buffer.get(requestId2), undefined);
```

packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.test.ts
```diff
@@ -79,15 +79,15 @@ describe("AsyncBoundedResponseBuffer", () => {
...
     const result2 = await buffer.put(requestId2, Promise.resolve(responseData2));
     assert.strictEqual(result1, true);
-    assert.strictEqual(result2, false);
+    assert.strictEqual(result2, true);
 
     const retrievedPromise1 = buffer.get(requestId1);
     const retrievedPromise2 = buffer.get(requestId2);
     assert(retrievedPromise1 !== undefined);
-    assert(retrievedPromise2 === undefined);
+    assert(retrievedPromise2 !== undefined);
```
jwajgelt: You should use the method, it gives more context in case the assertion fails.
```suggestion
      assert.notStrictEqual(retrievedPromise2, undefined);
```

jwajgelt (approved):

packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.test.ts
pFornagiel: Changed as suggested in df824052

packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.test.ts
pFornagiel: Changed here and in all other places, where I've used triple-equals - df824052. Thanks for pointing that out.

PR #1521, opened 2025-09-11
feat: network inspector - revamp typing system, improve communication between plugin and panel

pFornagiel:
### Description

This PR revamps the network inspector typing system to make it more concise and remove redundant types and casts:

- split type definitions from `network.ts` to `network.d.ts`, `networkLog.d.ts`, `networkFilter.d.ts`, `contextMenu.d.ts`, `panelMessageProtocol.d.ts`

- Introduced new communication protocol for Network Panel messages. WebSocket message handling now differentiates between sent message types:
  - `CDP` - messages that follow the format defined for Chrome Devtools Protocol, which is used to communicate
     with devtools
  - `IDE` - messages, that are meant to communicate with IDE part of the extension

- Cleaned up unused code and duplicate type definitions, improved overall structure for better maintainability.

### How Has This Been Tested: 

- **behaviour should not change, check if everything works as expected making different types of network requests**

Not applicable.

### DIFF
--- a/packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.test.ts
+++ b/packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.test.ts
@@ -1,0 +1,314 @@
+import assert from "assert";
+import { describe, beforeEach, it } from "mocha";
+import { InternalResponseBodyData } from "./networkRequestParsers";
+import { AsyncBoundedResponseBuffer as AsyncBuffer } from "./AsyncBoundedResponseBuffer";
+
+const {
+  AsyncBoundedResponseBuffer,
+  REQUEST_BUFFER_MAX_SIZE_BYTES,
+} = require("./AsyncBoundedResponseBuffer");
+
+describe("AsyncBoundedResponseBuffer", () => {
+  let buffer: AsyncBuffer;
+
+  beforeEach(() => {
+    buffer = new AsyncBoundedResponseBuffer();
+  });
+
+  describe("constructor", () => {
+    it("should initialize with empty state", () => {
+      const stats = buffer.getStats();
+      assert.strictEqual(stats.entryCount, 0);
+      assert.strictEqual(stats.currentSizeBytes, 0);
+      assert.strictEqual(stats.maxSizeBytes, REQUEST_BUFFER_MAX_SIZE_BYTES);
+      assert.strictEqual(stats.utilization, "0.00%");
+    });
+  });
+
+  describe("put and get operations", () => {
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -10,61 +10,62 @@ import { NetworkDevtoolsWebviewProvider } from "./NetworkDevtoolsWebviewProvider
...
+    contentTypeLowerCase.includes("text/javascript") ||
+    contentTypeLowerCase.includes("application/javascript") ||
+    contentTypeLowerCase.includes("application/x-javascript")
   ) {
     return "javascript";
-  } else if (contentType.includes("text/plain")) {
+  }
+  if (contentTypeLowerCase.includes("text/plain")) {
     return "text";
   }
```
jwajgelt: VERY NIT: I feel like converting this into a mapping could be easier to read, something like
```ts
const LANGUAGE_BY_CONTENT_TYPE = {
  "text/javascript": "javascript",
  /* etc... */
}

/*
  ...
*/

for (const [contentType, language] of Object.entries(LAUNGUAGE_BY_CONTENT_TYPE)) {
  if (contentTypeLowerCase.includes(contentType)) return language;
}
```
it's fine as is though.

pFornagiel: Changed to this approach in 6a58cf04

```ts
const LANGUAGE_BY_CONTENT_TYPE = {
  "application/json": "json",
  "text/json": "json",
  "text/html": "html",
  "application/xhtml+xml": "html",
  "text/xml": "xml",
  "application/xml": "xml",
  "text/css": "css",
  "text/javascript": "javascript",
  "application/javascript": "javascript",
  "application/x-javascript": "javascript",
  "text/plain": "text",
};

function determineLanguage(contentType: string, body: string): string {
  const contentTypeLowerCase = contentType.toLowerCase();

  for (const [contentTypeKey, language] of Object.entries(LANGUAGE_BY_CONTENT_TYPE)) {
    if (contentTypeLowerCase.includes(contentTypeKey)) {
      return language;
    }
  }

  // Fallback: try to guess based on content structure
  const trimmedBody = body.trim();
  if (trimmedBody.startsWith("<?xml") || trimmedBody.startsWith("<")) {
    return trimmedBody.includes("<!DOCTYPE html") || trimmedBody.includes("<html") ? "html" : "xml";
  }
  if (trimmedBody.startsWith("{") || trimmedBody.startsWith("[")) {
    return "json";
  }

  return "text";
}
```

PR #1525, opened 2025-09-12
rebranding: add Feature page and feature section to Landing

spaleka:
This PR adds Feature page to Radon IDE web application and feature section to its Landing page.

Changes include:
- Creating Features page.
- Creating or changing Features and Landing pages mutual components to match designs of each page.

Additional change:
- Update of Mobile sidebar design

### How Has This Been Tested: 

Tested by using app manually.

Not applicable

### DIFF
--- a/packages/docs/src/components/Features/FeaturesGrid/index.tsx
+++ b/packages/docs/src/components/Features/FeaturesGrid/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState, useMemo, useCallback } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import FeaturesGridCard from "../FeaturesGridCard";
 import styles from "./styles.module.css";
 import useBaseUrl from "@docusaurus/useBaseUrl";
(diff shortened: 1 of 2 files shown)


packages/docs/src/components/Features/FeaturesGrid/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+      title: "Search through the logs easily",
+      content:
+        "Radon IDE uses the built-in VSCode console allowing you to filter through the logs. The links displayed in the console automatically link back to your source code.",
+      imageSrc: useBaseUrl("/img/features/feature_logs_dark.svg"),
+    },
+    {
+      label: "Previews",
+      title: "Develop components in isolation",
+      content:
+        "Radon IDE comes with a package allowing to preview components in full isolation. Develop your components individually without distractions.",
```
filip131311: does this section need to be in code? cant we pull this information from some mdx files? Is it standard practice through out the landing page or just this place (I'm sorry this is nee code base for me)

filip131311: even if it has to be in code for some reason it would be good to move it out of the components logic as it is something component consumes

packages/docs/src/components/Features/FeaturesGrid/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const [cardWidth, setCardWidth] = useState(0);
+  const [cardPosition, setCardPosition] = useState(0);
+  const [visibleCards, setVisibleCards] = useState(3);
+  const [isDragging, setIsDragging] = useState(false);
+
+  const windowRef = useRef(null);
+  const containerRef = useRef(null);
+  const cardRefs = useRef([]);
+
+  const calculatedPosition = () => {
```
filip131311, line 67: there is something wrong with the logic: 

https://github.com/user-attachments/assets/5ed8ac48-66e2-4e68-8c9a-781df65234c8

packages/docs/src/components/Features/FeaturesGrid/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+        "Missed a bug? You can rewatch what happened on the device anytime. No need to manually start the recording ever again.",
+      imageSrc: useBaseUrl("/img/features/feature_recording_dark.svg"),
+    },
+  ];
+
+  const { isLanding } = usePageType();
+
+  const [first, setFirst] = useState(0);
+  const [cardWidth, setCardWidth] = useState(0);
+  const [cardPosition, setCardPosition] = useState(0);
```
filip131311, line 59: I am not sure that it means from the name alone I belive this should be named `firstCardPosition` or something more decriptive like that

packages/docs/src/components/Features/FeaturesGrid/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+        </motion.div>
+      </div>
+      {isLanding && (
+        <div className={styles.navigationArrows}>
+          <div className={styles.arrow}>
+            <button
+              className={styles.arrowLeft}
+              disabled={first === 0 || (visibleCards === 3 && first === 1)}
+              onClick={handlePrevArrow}>
+              <ArrowRightSmallIcon />
```
filip131311, line 190: Arrow is not visible for me unless this is a diffrent component? 
Screenshot 2025-09-22 at 09 42 15

packages/docs/src/components/MultipleSWMIcon/index.tsx
```diff
@@ -0,0 +1,764 @@
+import React from "react";
+
+export default function MultipleSWMIcon(props) {
```
filip131311, line 3: does it have to be inlined like that? is it some kind of platform constraint? Why not just static `.svg`?

packages/docs/src/components/SWMIcon/index.tsx
```diff
@@ -0,0 +1,92 @@
+import React from "react";
+
+export default function SWMIcon(props) {
```
filip131311, line 3: same question

filip131311 (changes requested):
there is some thing wrong with the theming of svgs in the light mode
Screenshot 2025-09-22 at 10 05 01

in the designs those logos ar white

I also left some inline comments tho i think some of them are unrelated as i didn't look at the correct page at the beginning

packages/docs/src/components/Features/FeaturesGrid/index.tsx
spaleka: It is currently only visible in the dark mode. Light mode will be added later.

packages/docs/src/components/MultipleSWMIcon/index.tsx
spaleka: Yes, it’s a platform constraint, I was instructed to add the .svgs that way.

packages/docs/src/components/Features/FeaturesGrid/index.tsx
spaleka: Only data from the Docs page are in mdx files. Other data, if not very long or reusable, is kept in the component where it’s used. 
This data was inside a component because it uses the useBaseUrl hook, but I could, keep it outside the component logic and call the hook while mapping.

PR #1526, opened 2025-09-15
feat:D

KeyJayY:
This PR introduces **automated tests for Device Settings**, and extends the existing suite of **Radon tools tests**.
---

### How Has This Been Tested

Tests have been executed:  
- On **GitHub Actions** (iOS)  
- Locally on **Android** and **iOS**  

To run tests on GitHub CI:  
Start Workflow on branch `@KeyJayY/DeviceSettingsTests` from the Actions tab in GitHub.  

To run tests locally:  
`npm run prepare-and-run-tests -- react-native-80`  
(or any other supported test app).  

To run only specific tests:  
`npm run prepare-and-run-tests -- react-native-81 4 5 6`  
executes tests 4, 5 and 6. Any combination is supported.  

In `utils/configuration.js`, you can switch between Android and iOS test runs.  
Local VM can also be used.  

---

Not documented yet.


packages/vscode-extension-tester/utils/imageProcessing.js
```diff
@@ -0,0 +1,28 @@
+import { createCanvas } from "canvas";
+
+export function cropCanvas(canvas, position) {
+  // + 0.5 to translate origin to left top corner (originally it's in the center)
```
filip131311: I think we shold not do this translation on the application side to avoid this unnecesery logic

packages/vscode-extension-tester/services/radonInteractionsServices.js
```diff
@@ -165,13 +188,39 @@ export class RadonSettingsService {
     this.elementHelperService = new ElementHelperService(driver);
   }
 
-  async toggleShowTouches() {
+  async toggleShowTouches(value = true) {
```
filip131311: now that you added a value argument this is not a toggle function, 
please create separate methods for `enableShowTouches` and `DisableShowTouches`

packages/vscode-extension-tester/services/radonInteractionsServices.js
```diff
@@ -165,13 +188,39 @@ export class RadonSettingsService {
...
+    const switchElementState =
+      (await switchElement.getAttribute("data-state")) == "checked";
+
+    if (value !== switchElementState) {
+      switchElement.click();
+    }
+    this.driver.actions().sendKeys(Key.ESCAPE).perform();
+  }
+
+  async toggleEnableReplays(value = true) {
```
filip131311: same

packages/vscode-extension-tester/configuration.js
```diff
@@ -1,5 +1,5 @@
 export default {
-  isRecording: false,
+  isRecording: process.env.IS_RECORDING === "true",
```
filip131311: this is not what i meant, sorry for causing mix up. I Think having the file with default configurations is good and I don't think this file should perform any logic. 
On the other hand i think that the way we get "current configuration" should treat `process.env.IS_RECORDING` with bigger priority then this default config and it should do it explicitly (without using the fact that this default setting is false)

packages/vscode-extension-tester/configuration.js
```diff
@@ -8,5 +8,5 @@ export default {
   // https://github.com/ReactiveCircus/android-emulator-runner/issues/350
   // https://github.com/utmapp/UTM/issues/6821
   // android tests can be run on a local machine with android studio installed
-  isAndroid: false,
+  isAndroid: process.env.IS_ANDROID === "true",
```
filip131311: same

packages/vscode-extension-tester/services/radonInteractionsServices.js
```diff
@@ -9,7 +9,7 @@ import {
   OutputView,
 } from "vscode-extension-tester";
 import * as fs from "fs";
-import config from "../utils/configuration.js";
+import config from "../configuration.js";
```
filip131311: I think we should have some kind of function "get Configuration" or something similar that resolves the configuration

packages/vscode-extension-tester/utils/helpers.js
```diff
@@ -1,105 +1,8 @@
...
-
-  let line;
-
-  const match = tooltip.match(/line (\d+)/);
-  if (match) {
-    line = parseInt(match[1], 10);
-    console.log("Line number:", line);
-  }
-
-  return line;
```
filip131311, line 104: what is it why is it no longer needed?

KeyJayY: it was supposed to be deleted two PRs ago, when I refactored the code and moved it to the services

KeyJayY:
> I think the tests that were failing had problems with perimissions before because I remember clicking approve and I am not able to reproduce it now. Is it something that could happen on github actions?

It seems odd to me because previously failing tests didn’t require any special permissions beyond what other tests needed. These tests pass on GitHub actions.
- I modified the `should open preview` test logic slightly. My guess based on the output is that in your run, the test app read the text from the URL bar before it was updated after opening the preview (this is just a guess, as it didn’t happen in my runs). The change should improve test stability.
- The `should show breakpoint stop view on phone screen` test was added in some PRs ago and seemed to be stable, but your case shows that it can fail. I don’t currently know why it fails, but I’ll keep an eye on it in future runs.

packages/vscode-extension-tester/ui-tests/setupTest.js
```diff
@@ -14,8 +14,10 @@ import {
   closeServer,
 } from "../server/webSocketServer.js";
 import startRecording from "../utils/screenRecording.js";
+import getConfiguration from "../configuration.js";
 
-const IS_RECORDING = process.env.IS_RECORDING === "true";
+const config = getConfiguration();
+const IS_RECORDING = config.isRecording;
```
filip131311: you can do: 
```
const { IS_RECORDING } = getConfiguration();
```

PR #1532, opened 2025-09-15
fix: react profiler cannot start

jwajgelt:
Fixes a regression in #1497 where the React Profiler starting does not set the application state correctly, making it unusable.

### How Has This Been Tested: 
- open an app in Radon
- press "start react profiler" in the tools menu
- cause some app renders
- press "stop react profiler" in the tools menu
- the profile should open


packages/vscode-extension/src/project/devtools.ts
```diff
@@ -145,6 +148,7 @@ export class DevtoolsConnection implements Disposable {
     this.bridge.shutdown();
     this.disconnectedEventEmitter.dispose();
     this.ideMessageEventEmitter.dispose();
```
filip131311, line 150: nit: can we move those to `disposables` field?

jwajgelt: I'll have more changes here in #1500, I'd rather change it there to prevent more conflicts with that branch.

PR #1534, opened 2025-09-16
New style for alert dialogs

kmagiera:
This PR changes the visuals of different alerts that we show within the IDE.

The problem with the older alerts that we noticed:
1) they didn't look aligned visually with the rest of the UI after we started adapting vscode-elements
2) there were issues with longer text being displayed i.e. in case of some xcodebuild errors included in the descriptions
3) the actions were difficult to use because the icons weren't descriptive enough and one would need to check tooltips every time

What we are changing here:
1) We're implementing a new design that aligns with the vscode theme (background color, shadows, border colors)
2) We're changing the structure of the dialog such that it look closer to system notifications with title, description and buttons in separate rows
3) We're expanding the alert interface to take "closeable" flag which allows for the "x" close action to be shown naturally in the top right corner
4) We further expand the api to accept types other than "error", namely "warning" and "info" which use different icon and icon-color when displayed (similary to how error/info/warning notifications API work in VSCode)
5) We're updating the callsites to use the new "closeable" flag and to also migrate from using IconButton to just Button with text. We also reshuffle the actions such that there's one "primary" action button on every alert actions list.

The new look with different themes:
image
image
image

### How Has This Been Tested: 
1) Open project and simulate a bunch of errors in different places:
a) build error on Android by deleting "android" folder
b) add some invalid import to trigger bundle error dialog
c) update some color in app.json to see native dependencies change alert (closeable)
d) double-tap r to trigger reload which shows disconnected alert
e) create launch config, switch to it, delete it from json -> see the "Unknown launch config" alert
2) See how the above dialogs look under different themes


packages/vscode-extension/src/webview/providers/AlertProvider.tsx
```diff
@@ -31,14 +33,17 @@ export default function AlertProvider({ children }: { children: React.ReactNode
...
-  const openAlert = useCallback(({ id, title, description, actions, priority }: AlertType) => {
-    setAlerts((oldAlerts) => {
-      if (oldAlerts.some((alert) => alert.id === id)) {
-        return oldAlerts;
-      }
-      return [...oldAlerts, { id, title, description, actions, priority }];
-    });
-  }, []);
+  const openAlert = useCallback(
+    ({ id, title, description, actions, priority, closeable, type }: AlertType) => {
```
jwajgelt: NIT: Must we unpack this object? Why don't we just add the object we get in the argument to the `alerts` array?

PR #1538, opened 2025-09-17
fix: crashes due to missing fields in simctl output

### LINKED ISSUE #1517: iOS > Cannot read properties of undefined (reading 'includes')
**Description**
When trying to launch a react-native project created using npx @react-native-community/cli@latest init the radon device throws `[error] Couldn't start device session Cannot read properties of undefined (reading 'includes')`

Looking into it a bit I'm guessing it is from https://github.com/software-mansion/radon-ide/blob/cf43b602ec6d6552d33daa01227d0377199f020b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts#L751 

**Environment**

- Version of currently used Radon IDE (Extensions tab -> click on Radon IDE extension -> version is located next to the extension name): 1.11.0
- Version of React Native/Expo used in the project.: 0.81.1
- Any other versions of tools related to the problem.

I reproduced this by creating a new project and launching the simulator without any changes.
Log: 

Image

jwajgelt:
Fixes #1517

For older runtimes, the `deviceTypeIdentifier` field is missing in the output of `simctl`, causing crashes when trying to read the devices in the system-wide device set.
Previously this wasn't an issue, because we only accessed the devices created by us, which always included that information.
However, in #1471 we read the system-wide devices to choose a destination to pass to `xcodebuild`, resulting in build errors when trying to access the non-existent property.
This PR fixes that.

### DIFF
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -40,7 +40,7 @@ interface SimulatorInfo {
   type?: "simulator" | "device" | "catalyst";
   booted?: boolean;
   lastBootedAt?: string;
-  deviceTypeIdentifier: string;
+  deviceTypeIdentifier?: string;
 }
 
 interface SimulatorData {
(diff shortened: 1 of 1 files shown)


filip131311 (approved):
👍

PR #1539, opened 2025-09-17
feat: implement connect and new devtools for custom debugger

jwajgelt:
Implements `addBinding` support for the custom JS debugger to allow it to work with Radon Connect and the new RDT-over-CDP implementation when configured.

### How Has This Been Tested: 
- add `"useCustomJSDebugger": true` to a launch configuration for an app running RN >=0.76
- open the app in Radon
- verify the app runs and devtools connect

### DIFF
--- a/packages/vscode-extension/src/debugging/CDPSession.ts
+++ b/packages/vscode-extension/src/debugging/CDPSession.ts
@@ -1,5 +1,5 @@
 import WebSocket from "ws";
-import { Disposable, EventEmitter } from "vscode";
+import { EventEmitter } from "vscode";
 import { OutputEvent, Source, StackFrame } from "@vscode/debugadapter";
 import { DebugProtocol } from "@vscode/debugprotocol";
 import { Minimatch } from "minimatch";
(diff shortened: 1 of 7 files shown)


packages/vscode-extension/src/debugging/CDPSession.ts
```diff
@@ -313,10 +301,24 @@ export class CDPSession {
     return firstPosition!;
   }
 
+  private async waitForDebuggerReady(timeoutMs: number) {
+    const { resolve: resolveTimeout, promise: timeoutPromise } = Promise.withResolvers<void>();
+    const timeoutId = setTimeout(resolveTimeout, timeoutMs);
```
kmagiera: we can use `sleep` from `utiities/retry`

Then we'd only write:
```
await Promise.race([sleep(timeoutMs), readyPromise])
```

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -15,6 +15,8 @@ const VSCODE_JS_DEBUGGER_TYPE = "com.swmansion.proxy-debugger";
 export const DEBUG_CONSOLE_LOG = "RNIDE_consoleLog";
 export const DEBUG_PAUSED = "RNIDE_paused";
 export const DEBUG_RESUMED = "RNIDE_continued";
+export const SCRIPT_PARSED = "RNIDE_bundleParsed";
```
kmagiera: I saw this already in the previous PR but was already merged before I can comment. Would prefer to not use "bundle" for this event as it actually maps 1-to-1 to script parsed event. In particular this gets triggered for individual scripts that we load via debugger while bundle has a more concrete meaning. A script could be the main bundle but not everything that triggers scriptParsed is a bundle.

To further deepen the confusion here we actually use "bundle parsed" for the string while the constant is called "script parsed"

packages/vscode-extension/src/debugging/CDPSession.ts
jwajgelt: `sleep` cannot be cancelled before it finishes, but that should be ok.

packages/vscode-extension/src/debugging/DebugSession.ts
jwajgelt: I agree.

PR #1542, opened 2025-09-18
feat: Shortcuts test coverage

KeyJayY:
This PR focuses on automated tests for **Keyboard Shortcuts** and **Radon Settings**. It also extends device settings tests. 

---

### How Has This Been Tested

Tests have been executed:  
- On **GitHub Actions** (iOS)  
- Locally on **Android** and **iOS**  

To run tests on GitHub CI:  
Start Workflow on branch `@KeyJayY/KeyboardShortcutsTests` from the Actions tab in GitHub.  

To run tests locally:  
`npm run prepare-and-run-tests -- react-native-80`  
(or any other supported test app).  

To run only specific tests:  
`npm run prepare-and-run-tests -- react-native-81 3 6 8 9`  
executes tests 3, 6, 8 and 9 (tests that were modified). Any combination is supported.  

In `configuration.js`, you can switch between Android and iOS test runs.  

Local VM can also be used, to run tests on VM you can use command:  
`npm run run-tests-on-VM -- react-native-81 3 6 8 9`

> Note: `should change device appearance mode` test is expected to fail on iOS since this functionality doesn't work  

---

Not documented yet.


packages/vscode-extension-tester/ui-tests/03-manageDevices.test.js
```diff
@@ -82,4 +83,78 @@ describe("Adding device tests", () => {
...
+      "radon-bottom-bar-device-select-dropdown-trigger"
+    );
+    await elementHelperService.findAndWaitForElementByTag("device-select-menu");
+
+    await elementHelperService.findAndWaitForElementByTag(
+      `device-${deviceName}`
+    );
+
+    await elementHelperService.findAndClickElementByTag("device-running-badge");
+    await elementHelperService.waitUntilElementGone(
```
filip131311, line 124: it would be nice to have this test check if the actual emulator/simulator process stoped running (as the running badge might lie)

filip131311: if there are tests like that in other places you could change it there as well, and we could have a separete test for the running badge to check if it shows the correct information

packages/vscode-extension-tester/ui-tests/03-manageDevices.test.js
```diff
@@ -82,4 +83,78 @@ describe("Adding device tests", () => {
...
+    await driver
+      .actions()
+      .keyDown(Key.COMMAND)
+      .keyDown(Key.SHIFT)
+      .sendKeys("9")
+      .keyUp(Key.SHIFT)
+      .keyUp(Key.COMMAND)
+      .perform();
+
+    await driver.wait(async () => {
```
filip131311, line 232: this is a nice beginning, but we also need a test to check if device is acctually changed, so we need to check if debuger is displaing information about correct device and so on (other features as well) 

we should also have a separate test if the "Shut down devices when switching:" option is enabled and we should check if the previous device is actually killed

filip131311: you shoul make a separete tests out of it

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
```diff
@@ -80,47 +80,95 @@ describe("Device Settings", () => {
...
+      const orientation =
+        await appManipulationService.sendMessageAndWaitForResponse(
+          appWebsocket,
+          "getOrientation"
+        );
+      return orientation.value === "portrait";
+    }, 5000);
+  });
+
+  it("should rotate device using shortcuts", async () => {
```
filip131311, line 134: we should have a few test that change the rotation very rapidly (you can randomly generate the sequence of rotation), we had problems with rotation going into some king of infinite spiral after rapid changes.

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
```diff
@@ -184,4 +232,35 @@ describe("Device Settings", () => {
...
+      "open-app-switcher-button"
+    );
+
+    await driver.wait(async () => {
+      const appState =
+        await appManipulationService.sendMessageAndWaitForResponse(
+          getAppWebsocket(),
+          "getAppState"
+        );
+      // this test works on iOS only, Android app's state stays active in app switcher
```
filip131311: is there a value on the device that can be checked on android so we can know if the app swittecher is visable 
also if it works only on ios is it turned off for android? is there a way for as to decorate tests "andorid only" or some other elegant way of marking that?

filip131311 (changes requested):
left some inline comments

packages/vscode-extension-tester/ui-tests/03-manageDevices.test.js
KeyJayY: There is no such as clean method as decorators in python in js. My suggestion is to use wrapper function that takes condition.

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
KeyJayY: There is no as clean method as decorators in python in js. My suggestion is to use wrapper function that takes condition.

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
KeyJayY: I've added test that rotates rapidly

packages/vscode-extension-tester/ui-tests/03-manageDevices.test.js
KeyJayY: I added test that checks if debug process changes when device switch

packages/vscode-extension-tester/ui-tests/03-manageDevices.test.js
KeyJayY: I've added test that checks if process disappears after stopping device

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
```diff
@@ -171,6 +173,71 @@ describe("Device Settings", () => {
     }, 5000);
   });
 
+  it("should stay stable after rapid rotations", async () => {
+    function getRandomInt(min, max) {
```
filip131311: you misunderstood me, it shouldn't randomize it each time it would make tests not repleacable. You should generate a few seqences once (or pass a seed to the generator that would work even better)

filip131311: please extract the seed to a const at the top of the file when you do so

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
```diff
@@ -171,6 +173,71 @@ describe("Device Settings", () => {
...
+      return Math.floor(Math.random() * (max - min + 1)) + min;
+    }
+
+    const numberOfRotations = getRandomInt(20, 50);
```
filip131311: this should be constant there is no reason for it not to be

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
```diff
@@ -171,6 +173,71 @@ describe("Device Settings", () => {
...
+    const start = Date.now();
+    const firstOrientation =
+      await appManipulationService.sendMessageAndWaitForResponse(
+        appWebsocket,
+        "getOrientation"
+      );
+
+    let changes = 0;
+
+    // checks for 5 seconds if device orientation does not change
```
filip131311: 5 s is excessive 1 should be enough I think

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
```diff
@@ -171,6 +173,71 @@ describe("Device Settings", () => {
     }, 5000);
   });
 
+  it("should stay stable after rapid rotations", async () => {
```
filip131311, line 178: would you mind checking this test before this PR https://github.com/software-mansion/radon-ide/commit/8b5d3f7b296ec9e7ee4933df91fa9fbbb5040a1e this is when we had an issue (right before the fix) so it should fail

KeyJayY: There is a problem with running tests on older versions since extension in these versions has no data-testid assigned to its elements.

PR #1545, opened 2025-09-18
fix: device appearance on ios

### LINKED ISSUE #1528: Dark/Light mode doesn't work on iOS
**Description**  
Changing the device appearance (light/dark mode) does not update in real time on iOS.  
After rebooting, the device starts with the selected mode.  
On Android, this works correctly.  

**Environment**  
- Radon version: 1.11.1  
- OS: iOS 18.5

https://github.com/user-attachments/assets/94049605-ca56-4e79-b715-ab279e547b91

filip131311:
Fixes #1528

After #1475 we started noticing that light/dark mode change is not working on ios devices. This is because after `com.apple.cfprefsd.xpc.daemon` restart `com.apple.SpringBoard` stops working properly. To adres that we jst restart the SpringBoard process. 

### How Has This Been Tested: 

- run  `expo-54` test app and perform all kinds of reloads 
- test appearance settings after each one 
- turn of radon and do the same

internal

### DIFF
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -309,28 +309,25 @@ export class IosSimulatorDevice extends DeviceBase {
       `${bundleID}.plist`
     );
     Logger.debug(`Defaults location ${userDefaultsLocation}`);
-    try {
-      await exec(
-        "/usr/libexec/PlistBuddy",
-        [
-          "-c",
-          "Delete :RCT_jsLocation",
-          "-c",
-          `Add :RCT_jsLocation string localhost:${metroPort}`,
-          userDefaultsLocation,
-        ],
-        { allowNonZeroExit: true }
-      );
-    } catch (e) {
-      // Delete command fails if the key doesn't exists, but later commands run regardless,
-      // despite that process exits with non-zero code. We can ignore this error.
-    }
-    try {
-      // Simulator may keep the defaults in memory with cfprefsd, we restart the daemon to make sure
-      // it reads the latest values from disk.
-      // We'd normally try to use defaults command that would write the updates via the daemon, however
-      // for some reason that doesn't work with custom device sets.
-      await exec("xcrun", [
+    await exec(
+      "/usr/libexec/PlistBuddy",
+      [
+        "-c",
+        "Delete :RCT_jsLocation",
+        "-c",
+        `Add :RCT_jsLocation string localhost:${metroPort}`,
+        userDefaultsLocation,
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -343,6 +343,24 @@ export class IosSimulatorDevice extends DeviceBase {
     } catch (e) {
       // ignore errors here and hope for the best
     }
+    try {
+      // After performing the above hack we need to restart SpringBoard as well,
+      // because com.apple.cfprefsd.xpc.daemon restart breaks it's ability to read
+      // the preferences (e.g appearance).
```
filip131311:
```suggestion
      // the settings (e.g appearance).
```

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -343,6 +343,24 @@ export class IosSimulatorDevice extends DeviceBase {
     } catch (e) {
       // ignore errors here and hope for the best
     }
+    try {
+      // After performing the above hack we need to restart SpringBoard as well,
+      // because com.apple.cfprefsd.xpc.daemon restart breaks it's ability to read
+      // the settings (e.g appearance).
```
kmagiera:
```suggestion
      // Restarting cfprefsd breaks the SpringBoard process which in particular controls things like
      // the appearance settings. We need to restart it following the cfprefsd reset.
```

We should not use 'hack' here as the word itself has negative association and also it isn't really a hack in its most common usage.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -343,6 +343,24 @@ export class IosSimulatorDevice extends DeviceBase {
...
+        "--set",
+        deviceSetLocation,
+        "spawn",
+        this.deviceUDID,
+        "launchctl",
+        "kickstart",
+        "-k",
+        "system/com.apple.SpringBoard",
+      ]);
+    } catch (e) {
```
kmagiera: I recently found out that we can just pass `reject: false` as option to exec. This results in the possible errors not being thown while we'd still get the errors reported in the logs. In may be cleaner to use it here and in the above calls as well.

kmagiera (approved):
Looks good, left some small comments inline

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
filip131311: thanks, that's good to know

PR #1550, opened 2025-09-19
feat: allow running Radon with a single Metro server on a specific port

jwajgelt:
Adds the ability to configure Radon to use a single Metro server, shared by all devices, running on a specified port.
- adds a `metroPort` launch configuration entry which allows specifying the port
- when the `metroPort` config is unspecified, the current behavior should be preserved -- each running device gets its own Metro server, running on an ephemeral port
- when the `metroPort` config is specified, switches from the current behavior to instead running a single metro server on the specified port, shared by all running devices. This imposes some technical limits on concurrent devices:
  - when the CDP devtools transport is used, any number of iOS devices _with different display names_ can run at the same time, but only a single Android emulator can run concurrently -- this is due to the fact, that we use the `deviceName` inspector page property to differentiate debugger targets, and for all Android emulators, that field is the same.
  - when the "old" websocket devtools transport is used, only a single device can run at the same time. This is due to the devtools server port being shared by all running devices, making it hard to tell them apart.

### How Has This Been Tested: 
- try the different combinations of React Native (>= 0.76, <= 0.75) x `useOldDevtools` (true, false) x `metroPort` (8081, unspecified)
- verify the app launches and the debugger and devtools connect
- verify switching devices works
- verify appropriate device number limits are enforced:
  - none if `metroPort` is unspecified
  - at most 1 Android if it is
  - at most 1 device if both it and `useOldDevtools` are set

WIP


packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -790,6 +789,9 @@ export async function listSimulators(location: SimulatorDeviceSet): Promise<IOSD
         .filter((e) => e !== undefined);
     })
     .flat();
+  if (location === SimulatorDeviceSet.RN_IDE) {
+    await ensureUniqueDisplayNames(simulators);
```
filip131311, line 793: feels awkward to do while listing devices instead of when deviaces are created. Is there a reason to do it that way?

packages/vscode-extension/src/project/ApplicationContext.ts
```diff
@@ -125,9 +151,24 @@ export class ApplicationContext implements Disposable {
     this.launchConfig = resolveLaunchConfig(launchConfig);
     this.applicationDependencyManager.setLaunchConfiguration(this.launchConfig);
     await this.applicationDependencyManager.runAllDependencyChecks();
+
+    (await this._devtoolsServer)?.dispose();
```
filip131311: As we want to discard the old server do we need to await it can't it be: 
```suggestion
    (this._devtoolsServer)?.then(server => server.dispose())

File: packages/vscode-extension/src/project/DebuggerTarget.ts [hunk 3]
```diff
@@ -0,0 +1,187 @@
...
+      if (cancelToken.cancelled) {
+        return undefined;
+      }
+      throw e;
+    }
+  }
+  return undefined;
+}
+
+export async function getDebuggerTargetForDevice(
```
filip131311: nit: seems that it would be nice to consistently  pass objects as arguments

packages/vscode-extension/src/project/DeviceSessionsManager.ts
```diff
@@ -109,36 +119,44 @@ export class DeviceSessionsManager implements Disposable {
...
-    const stopPreviousDevices = selectDeviceOptions?.stopPreviousDevices;
 
     // if there's an existing session for the device, we use it instead of starting a new one
     const existingDeviceSession = this.deviceSessions.get(deviceInfo.id);
     if (existingDeviceSession) {
       this.updateSelectedSession(existingDeviceSession);
-      if (stopPreviousDevices) {
-        await this.terminatePreviousSessions();
-      }
+      await this.terminateSessionsOverLimit(deviceInfo);
```
filip131311, line 159: how can a change to existing device cause us to go over limit?

packages/vscode-extension/src/project/metro.ts
```diff
@@ -1,34 +1,123 @@
...
+  constructor(
+    private readonly launchConfiguration: ResolvedLaunchConfig,
+    private readonly devtoolsPort: Promise<number | undefined> = Promise.resolve(undefined)
+  ) {
+    this.port = this.launchConfiguration.metroPort;
+  }
+
+  public async getMetroSession({ resetCache }: { resetCache: boolean }) {
+    const session = await this.metroSession;
+    if (session === undefined || session.refCount <= 0 || resetCache) {
```
filip131311, line 72: can refCount get lower then 0?

packages/vscode-extension/src/project/metro.ts
```diff
@@ -442,172 +392,134 @@ export class MetroLauncher extends Metro implements Disposable {
...
 
-    return initPromise;
+    // NOTE: if the process exits before the "initialize_done" event, we reject the promise
+    bundlerProcess
+      .catch(() => {
+        // ignore the error, we are only interested in the process exit
+      })
+      .then(() => {
+        this.bundlerReady.reject(new Error("Metro bundler exited unexpectedly"));
+        this.onServerStoppedEventEmitter.fire();
```
filip131311: is it ever used?

packages/vscode-extension/src/project/DeviceSessionsManager.ts
jwajgelt: You can change the `stopPreviousDevices` workspace setting before switching, in which case we want to terminate the other devices.
This is the behaviour we have currently, this just preserves it.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
jwajgelt: People can have non-unique device names right now, we need to correct them after the update.
Also, in theory, they can use `simctl` directly to rename the devices, in which case they'd brake Radon as well in a hard-to-debug way.
This is a defensive way to do it.

packages/vscode-extension/src/project/ApplicationContext.ts
jwajgelt: We don't need to, but we might as well.
I can change it, it shouldn't matter much.

packages/vscode-extension/src/project/metro.ts
jwajgelt: Yes.
One consumer can call `disposeInner()` to force disposal, and later some other consumer may `dispose()` it normally, in which case the counter goes into negative.
There's also nothing stopping the callers from calling it incorrectly (i.e. `dispose`ing more times than `retain`ing).

packages/vscode-extension/src/project/metro.ts
jwajgelt: Removed it.

PR #1551, opened 2025-09-19
fix: rotation on iOS 

filip131311:
after #1545, we started experiencing issues with rotation not being applied to the device, this seems to be caused by `com.apple.backboardd` service not working properly after `SpringBoard` restart. To fix that we restart `backboardd` as well.

### How Has This Been Tested: 

- run ios device in the new setup and try rotation and appearance change 

internal

### DIFF
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -357,6 +357,24 @@ export class IosSimulatorDevice extends DeviceBase {
       ],
       { reject: false }
     );
+
+    // Restarting SpringBoard process breaks the backboardd which in particular controls things like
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -357,6 +357,24 @@ export class IosSimulatorDevice extends DeviceBase {
       ],
       { reject: false }
     );
+
+    // Restarting SpringBoard process breaks the backboardd which in particular controls things like
+    // the rotation settings. We need to restart it following the cfprefsd springboard reset.
```
jwajgelt, line 362:
```suggestion
    // the rotation settings. We need to restart it following the SpringBoard reset.
```

jwajgelt (approved):

PR #1552, opened 2025-09-19
feat: network inspector - add syntax highlighting

pFornagiel:
### Description

This PR introduces syntax highlighting to code displayed in `Response` and `Payload` tabs in network inspector. 

The syntax colouring is loaded from TextMate format json files provided by installed vscode extensions, which are read and sent to the Webview as objects from the extension's network plugin. The Webview itself uses [React Shiki](https://react-shiki.vercel.app) (wrapper for [Shiki](https://shiki.style)), to properly render the code based on the provided definitions.

### How Has This Been Tested: 

- **test whether colouring works as expected, there are no errors / missing responses where they should be present. Test whether themes work correctly (those downloaded from vscode extensions as well).**

Not applicable.


packages/vscode-extension/src/utilities/themeExtractor.ts
```diff
@@ -0,0 +1,121 @@
...
+  [ThemeVariant.HighContrastLight]: theme_hc_light,
+};
+
+const VARIANT_FALLBACK: ThemeVariant = ThemeVariant.Dark;
+
+// Cache for theme data to avoid repeated file reads
+const themeCache = new Map<string, ThemeData>();
+
+function getDefaultTheme(themeVariant: ThemeVariant | undefined): ThemeData {
+  return DEFAULT_THEME_MAPPING[themeVariant ?? VARIANT_FALLBACK];
```
jwajgelt: VERY NIT: this could be written with a default parameter instead, since they are used when you pass `undefined` explicitly as well
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#passing_undefined_vs._other_falsy_values

```suggestion
function getDefaultTheme(themeVariant: ThemeVariant = VARIANT_FALLBACK): ThemeData {
  return DEFAULT_THEME_MAPPING[themeVariant];
```

Not sure which you prefer, feel free to leave it as is if you think it's better.

packages/vscode-extension/src/network/hooks/useThemeExtractor.tsx
```diff
@@ -0,0 +1,63 @@
...
+      for (const mutation of mutations) {
+        if (
+          mutation.type !== "attributes" ||
+          (mutation.attributeName !== THEME_VARIANT_ATTRIBUTE &&
+            mutation.attributeName !== THEME_ID_ATTRIBUTE)
+        ) {
+          continue;
+        }
+
+        const themeVariant = body.getAttribute(THEME_VARIANT_ATTRIBUTE) as ThemeVariant;
```
jwajgelt: Are we sure that this cannot have any other values than one of `ThemeVariant`s?
I'd feel better if we validated this here (as well as when loading the initial theme).
Something along the lines of:

```ts
function isThemeVariant(s:string): s is ThemeVariant {
  return Object.values(ThemeVariant).includes(s);
}
```
or
```ts
function intoThemeVariant(s:string): ThemeVariant {
  return Object.values(ThemeVariant).includes(s) ? s : FALLBACK
}
```

packages/vscode-extension/src/utilities/themeExtractor.ts
```diff
@@ -0,0 +1,121 @@
...
+
+      // Add "include" paths to stack for processing
+      if (themeData.include) {
+        const includePath = path.join(path.dirname(currentPath), themeData.include);
+        themeStack.push(includePath);
+      }
+
+      // Exclude the include property to avoid circular references
+      const { include: _include, ...themeWithoutInclude } = themeData;
+      mergedTheme = { ...mergedTheme, ...themeWithoutInclude };
```
jwajgelt: This means that values set by included theme files will overwrite the values set by the currently read file, since they will be processed in a later iteration of the `while` loop. Is this the intended behaviour?

packages/vscode-extension/src/network/providers/NetworkProvider.tsx
```diff
@@ -26,6 +27,59 @@ interface NetworkProviderProps extends NetworkTracker {
...
+      const { command, payload }: WebviewMessage = message.data;
+      if (command !== WebviewCommand.CDPCall || payload.id !== messageId) {
+        return;
+      }
+
+      const bodyData = payload.result as ResponseBodyData | undefined;
+
+      if (bodyData === undefined) {
+        resolve(responseBodiesRef.current[requestId]);
+        window.removeEventListener("message", listener);
```
jwajgelt, line 51: Since this function is responsible for calling `window.removeEventListener`, I think it'd be better if it also registered the listener.

packages/vscode-extension/src/network/providers/NetworkProvider.tsx
```diff
@@ -26,6 +27,59 @@ interface NetworkProviderProps extends NetworkTracker {
   toggleTimelineVisible: () => void;
   fetchAndOpenResponseInEditor: (networkLog: NetworkLog) => Promise<void>;
   getResponseBody: (networkLog: NetworkLog) => Promise<ResponseBodyData | undefined>;
+  getThemeData: (themeDescriptor: ThemeDescriptor) => Promise<ThemeData>;
+}
+
+function responseBodyListener(
+  resolve: (value: ResponseBodyData | undefined) => void,
```
jwajgelt: Passing the `resolve` callback here feels awkward.
I feel it'd make more sense for this function to return the Promise which resolves when the event is received.
```ts
function getResponseBody(messageId, requestId, responseBodiesRef): Promise<ResponseBodyData | undefined>
```

packages/vscode-extension/src/network/providers/NetworkProvider.tsx
jwajgelt: Same goes for `themeListener`

packages/vscode-extension/src/network/providers/NetworkProvider.tsx
jwajgelt: Same goes for `themeListener`

packages/vscode-extension/package.json
```diff
@@ -1091,6 +1092,7 @@
     "@rollup/rollup-linux-x64-gnu": "4.40.0"
   },
   "dependencies": {
-    "@tanstack/react-virtual": "^3.13.12"
+    "@tanstack/react-virtual": "^3.13.12",
+    "react-shiki": "^0.8.0"
```
jwajgelt: We generally add all of the dependencies as `devDependencies`. Would be good to move these there as well.

packages/vscode-extension/src/utilities/themeExtractor.ts
pFornagiel: Thanks for letting me know, changed to proposed in 925382f2

packages/vscode-extension/src/network/hooks/useThemeExtractor.tsx
pFornagiel: Made changes in c47a60a6

Introduced isThemeVariant function as requested:
```ts
export function isThemeVariant(value: unknown): value is ThemeVariant {
  return Object.values(ThemeVariant).includes(value as ThemeVariant);
}
```

`useThemeExtractor.tsx`
```tsx
//... 
// usage in mutation observer
const variantAttributeValue = body.getAttribute(THEME_VARIANT_ATTRIBUTE);

const themeVariant = isThemeVariant(variantAttributeValue)
  ? variantAttributeValue
  : THEME_VARIANT_FALLBACK;
const themeId = body.getAttribute(THEME_ID_ATTRIBUTE);

//...

// initialisation
const initialVariantAttributeValue = body.getAttribute(THEME_VARIANT_ATTRIBUTE);
const initialThemeVariant = isThemeVariant(initialVariantAttributeValue)
  ? initialVariantAttributeValue
  : THEME_VARIANT_FALLBACK;

const initialThemeId = body.getAttribute(THEME_ID_ATTRIBUTE) || undefined;
previousThemeIdRef.current = initialThemeId;

const initialThemeDescriptor: ThemeDescriptor = {
  themeVariant: initialThemeVariant,
  themeId: initialThemeId,
};
getThemeData(initialThemeDescriptor).then((initialThemeData) => {
  setEditorThemeData(initialThemeData);
});

//...
```

Also made changes to themeExctarctor findThemePath to accept `undefined` and return early if provided so, instead of passing an empty string - feels more natural this way after thought.

```themeExtractor.ts
//...
const themePath = findThemePath(themeId ?? workspaceThemeId);
//...

function findThemePath(themeId: string | undefined): string | undefined {
  if (!themeId) {
    return undefined;
  }
  
  //...

```

packages/vscode-extension/src/utilities/themeExtractor.ts
pFornagiel: Thanks for the catch, my oversight.

Usually, when the themeFile references other file, it is due to the fact that it shares some of the colour values with the other (see [Dark+ theme](https://github.com/microsoft/vscode/blob/main/extensions/theme-defaults/themes/dark_plus.json) for vscode - it uses [Default Dark theme](https://github.com/microsoft/vscode/blob/main/extensions/theme-defaults/themes/dark_vs.json) as base and adds to it).

The logic here seems to be that the values _should_ be overwritten by the files that `includes` other file, if the file redefines them, so I messed up the object unpacking order.

Additionally, the objects themselves cannot be shallow copied (as done before), because theme definitions include arrays and object, as described in [vscode source](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/services/themes/common/colorThemeSchema.ts)

So, in order to handle this correctly, theme resolution system has been modified to allow for deep-merge of theme objects, with caveats discussed internally and explained in code. 

Implemented in 0d101522

packages/vscode-extension/src/network/providers/NetworkProvider.tsx
pFornagiel: Refactored and introduced `createBodyResponsePromise` and `createThemeResponsePromise` functions, which encapsulate listener management and promise creation - 5e721d53

```ts
function createBodyResponsePromise(
  messageId: string,
  requestId: string,
  responseBodiesRef: React.RefObject<Record<string, ResponseBodyData | undefined>>
) {
  const { promise, resolve } = Promise.withResolvers<ResponseBodyData | undefined>();

  const listener = (message: MessageEvent) => {
    try {
      const { command, payload }: WebviewMessage = message.data;
      if (command !== WebviewCommand.CDPCall || payload.id !== messageId) {
        return;
      }

      const bodyData = payload.result as ResponseBodyData | undefined;

      if (bodyData === undefined) {
        resolve(responseBodiesRef.current[requestId]);
        window.removeEventListener("message", listener);
        return;
      }

      responseBodiesRef.current[requestId] = bodyData;

      resolve(bodyData);
      window.removeEventListener("message", listener);
    } catch (error) {
      console.error("Error parsing Window message:", error);
    }
  };

  // Setup listener to capture the response
  window.addEventListener("message", listener);

  return promise;
}
``` 

```ts
function createThemeResponsePromise(messageId: string) {
  const { promise, resolve } = Promise.withResolvers<ThemeData>();

  const listener = (message: MessageEvent) => {
    try {
      const { payload }: WebviewMessage = message.data;
      if (payload.method !== "IDE.Theme" || payload.id !== messageId) {
        return;
      }

      const themeData = payload.result as ThemeData;

      resolve(themeData);
      window.removeEventListener("message", listener);
    } catch (error) {
      console.error("Error parsing Window message:", error);
    }
  };

  // Setup listener to capture the response
  window.addEventListener("message", listener);

  return promise;
}
```

Current usage:
```ts
const promise = createBodyResponsePromise(messageId, requestId, responseBodiesRef);

// and 

const promise = createThemeResponsePromise(messageId);
```

packages/vscode-extension/src/network/providers/NetworkProvider.tsx
pFornagiel: Explained in the comment above, refactor in 5e721d53

packages/vscode-extension/package.json
pFornagiel: Moved to `devDependencies` in 010f5cd9

PR #1553, opened 2025-09-22
feat: support for react native 82

filip131311:
This PR adds support for react native 82, by updating the radon specific renderer logic to include the v19.1.1 react patch that react native 82 also includes.

### How Has This Been Tested: 

- run new radon native 82 test app. 
- run tests with `vscode-extension-tester` useing command `npm run run-tests-on-VM -- react-native-82`

to be updated


jwajgelt (approved):
LGTM

PR #1554, opened 2025-09-22
Filter out internal expo-router params from url bar

kmagiera:
This PR prevents internal expo-router params from being presented in the URL bar.

In expo-router v6 with new native tabs, expo started appending some internal params to the paths in order to control things like screen animations. Those are typically very long and not useful from the user standpoint. After this change we no longer show these params in the URL bar.

Before:
image

After:
image

### How Has This Been Tested: 
1) Run expo router app with native tabs
2) Switch between tabs
3) Open URL bar and select some of the routes that belong to tabs
4) before this change you'll see params like `__internal_expo_router_no_animation` in the URL bar being appended to some paths, after this change those params are no longer showing there.


filip131311 (approved):
looks good

PR #1559, opened 2025-09-23
fix: prevent setting state with DerivedStateManager after the item is removed

jwajgelt:
Fixes stopping a selected device failing after #1540 

### How Has This Been Tested: 
- open radon and wait for the app to load
- stop the selected device

### DIFF
--- a/packages/vscode-extension/src/project/StateManager.ts
+++ b/packages/vscode-extension/src/project/StateManager.ts
@@ -1,7 +1,7 @@
 import _ from "lodash";
 import { Disposable, EventEmitter } from "vscode";
 import { disposeAll } from "../utilities/disposables";
-import { RecursivePartial } from "../common/State";
+import { RecursivePartial, REMOVE } from "../common/State";
 import { mergeAndCalculateChanges } from "../common/Merge";
 
 export abstract class StateManager<T extends object> implements Disposable {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/project/StateManager.ts
```diff
@@ -74,13 +74,29 @@ class RootStateManager<T extends object> extends StateManager<T> {
 }
 
 class DerivedStateManager<T extends object, K extends object> extends StateManager<T> {
+  // NOTE: after the derived state is removed from the parent, we want to ignore any further updates
+  // made through this derived state manager to prevent accidental re-adding of the removed state or updating
+  // an entry that is added after this one is removed.
+  private removed: boolean;
```
filip131311, line 80: I am okay, with it as it is but what do you think about: 

```suggestion
  private get removed(): boolean {
    this.parent.getState()[this.keyInParent] === undefined;
  };
```

packages/vscode-extension/src/project/StateManager.ts
```diff
@@ -74,13 +74,29 @@ class RootStateManager<T extends object> extends StateManager<T> {
...
     super();
+    this.removed = this.parent.getState()[this.keyInParent] === undefined;
     this.disposables.push(
       parent.onSetState((partialParentState: RecursivePartial<K>) => {
+        if (this.removed) {
+          return;
+        }
+
+        const parentState = this.parent.getState();
+        if (parentState[this.keyInParent] === undefined) {
```
filip131311: and then we could remove this

PR #1564, opened 2025-09-24
fix: the close button for modals appears in bottom-left corner

jwajgelt:
Changes in #1534 caused the close button for modals to move to the bottom-left corner.
This was caused by the css files being loaded in a different order than before, making the rules for `icon-button` class override those for `modal-close-button` class.

Before:
image

After:
image


pFornagiel (approved):
Love it, works fine

PR #1557, opened 2025-09-22
feat: network inspector - styling / UX changes and fixes

pFornagiel:
### Description

This PR introduces changes in styling / user experience as well as minor bug fixes related to network inspector.

Changes and fixes related to network inspector:
- replaced vscode-elements [Scrollable component](https://vscode-elements.github.io/components/scrollable/) used inside `NetworkLogDetails`, with [OverlayScrollbars React](https://github.com/KingSora/OverlayScrollbars/tree/master/packages/overlayscrollbars-react) due to the fact that `Scrollable` does not support horizontal scrolling and vanilla css does not allow for the scrollbar to overlay the content, which caused visual problems. 
 `OverlayScrollbarsComponent` has been styled to match the `Scrollable` component of vscode-elements for consistency.
 
Screenshot 2025-09-22 at 17 07 03

- made styling changes in `HeadersTab` - the content is now padded properly (padding caused visual issues in some of the themes), made the header names and values visually distinct, added coloured Status Code indicator to match the Chrome Devtools

Screenshot 2025-09-22 at 16 32 16

- bumped version of [vscode-elements](https://github.com/vscode-elements/elements) to introduce official fix for issue described in #1484

- rewritten the `VscodeTableInternalFix` to tackle another issue - due to the incorrect height calculation scrollbar fails to appear as soon as table rows begin to overflow and works only after much more rows are added. The issue and solution is described within component's comments. PR to fix the issue in the original repository is under development.

- fixed problem with the chosen Tab in `NetworkLogDetails` resetting to `Headers` whenever user changed the inspected log

Apart from changes made to network inspector, styling changes have been introduced to main extension window (Phone Preview), which make the `Dropdown Menus`, `Modals` and `Tooltips` visually consistent with `Context Menu` introduced in #1506 in terms of border and shadow details. The changes are also meant to make the menus distinct from the extension background. These changes have to be discussed due to their impact on visual style of the extension.

image

Screenshot 2025-09-22 at 17 15 51

image

image

image

### How Has This Been Tested: 

Features were tested in local development environment of extension as follows:
- open radon-ide/packages/vscode-extension
- run the extension locally using Run and Debug menu -> Run Extension
- open an application supporting Radon IDE in editor - example of app supporting  rotation: radon-ide-test-apps/react-native-80 
- run Radon IDE extension panel, launch app on selected device, open network inspector
- **use NetworkLogDetails tabs, see if everything behaves correctly**

### How Has This Change Been Documented:

Not applicable.

PR #1565, opened 2025-09-24
fix: dropdown submenu animation issue fix

pFornagiel:
### Description

This PR fixes issue with the submenu styling, causing it to appear in the wrong place when screen is too small and then snap to the correct position. The issue was introduced in #1557 and was present because animation did not take the media-query transform into account.

The fix corrects the transform, so the animation now plays correctly. Another solution could also bo to disable animation of submenus altogether, can be discussed here.

https://github.com/user-attachments/assets/4982cadc-87c5-4a0e-8caa-cf7357f98c5c

### How Has This Been Tested: 

- **open submenus in settings dropdowns, see if everything behaves as expected**

Not applicable.


filip131311 (approved):
thank you

PR #1566, opened 2025-09-24
fix: switching to a device with bundle error fails to connect devtools

jwajgelt:
Fixes an issue where the CDP devtools fail to connect when switching to a running device with broken code, which produces a bundle error.
The bundle error causes the source maps to not load correctly when reconnecting the debugger, which in turn means the `debugSession.onScriptParsed` event is never fired and the devtools server never establishes the connection with the application. 

### How Has This Been Tested: 
- launch `react-native-81` test app
- launch 2 devices at the same time
- introduce an error in the code to force a bundle error
- switch to the other running device
- check that the devtools connect
- check that fixing the bundle error closes the error alert
- check that after switching to the other running device, the error alert disappears as well ()


filip131311 (approved):
If you think it helps, Its difficult for me to judge that

PR #1567, opened 2025-09-24
refactor: device settings

filip131311:
This is almost an end in our quest to get rid off updating state using the old `eventLinstener` mechanic. 
This time we remove `deviceSettingsChanged` event and add deviceSettings to the new state management solution. 
At the same time we choose to locate them in `workspaceConfiguration` so the user might share the settings with team members or other devices. 

The move to `workspaceConfiguration` would make them bloated so we refactor them to have "subsections". 

### How Has This Been Tested: 

- run test applications and test if sttings are applied, and persisted. 

internal (?) there is a change to the settings structure but it is self documented

### DIFF
--- a/packages/vscode-extension/src/project/DeviceSessionsManager.ts
+++ b/packages/vscode-extension/src/project/DeviceSessionsManager.ts
@@ -90,7 +90,8 @@ export class DeviceSessionsManager implements Disposable {
   }
 
   private get deviceLimits() {
-    const stopPreviousDevices = this.applicationContext.workspaceConfiguration.stopPreviousDevices;
+    const stopPreviousDevices =
+      this.applicationContext.workspaceConfiguration.deviceControl.stopPreviousDevices;
     const launchConfig = this.applicationContext.launchConfig;
     const usesSingleMetro = !!launchConfig.metroPort;
     const usesOldDevtools = launchConfig.useOldDevtools;
--- a/packages/vscode-extension/src/webview/components/DeviceSelect.tsx
+++ b/packages/vscode-extension/src/webview/components/DeviceSelect.tsx
@@ -93,8 +93,6 @@ function DeviceSelect() {
   const store$ = useStore();
   const selectedDeviceSessionState = useSelectedDeviceSessionState();
 
-  const stopPreviousDevices = use$(store$.workspaceConfiguration.deviceControl.stopPreviousDevices);
-
   const { projectState, project } = useProject();
 
   const devices = use$(store$.devicesState.devices) ?? [];
--- a/packages/vscode-extension/src/webview/views/ManageDevicesView.tsx
+++ b/packages/vscode-extension/src/webview/views/ManageDevicesView.tsx
@@ -36,8 +36,6 @@ function DeviceRow({
   isRunning,
   dataTest,
 }: PropsWithDataTest<DeviceRowProps>) {
-  const store$ = useStore();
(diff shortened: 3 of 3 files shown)


packages/vscode-extension/src/common/Merge.ts
```diff
@@ -75,3 +75,42 @@ export function merge<T extends { [P in keyof T]: T[P] }>(
...
+        current[pathKeys[i]] = {};
+      }
+      current = current[pathKeys[i]];
+    }
+    current[pathKeys[pathKeys.length - 1]] = value;
+  }
+
+  traverse(partial);
+  return result;
+}
```
jwajgelt: We can avoid most of the unsafe typing outside of `traverse` here:

```suggestion
export function splitRecursivePartialToSingleLeaves<T>(
  partial: RecursivePartial<T>
): Array<RecursivePartial<T>> {
  const result: Array<RecursivePartial<T>> = [];

  function traverse(obj: NonNullable<object>, pathKeys: string[] = []) {
    for (const [key, value] of Object.entries(obj)) {
      if (value && typeof value === "object" && !Array.isArray(value)) {
        traverse(value, [...pathKeys, key]);
      } else {
        const leafObject = createPathObject([...pathKeys, key], value);
        result.push(leafObject as RecursivePartial<T>);
      }
    }
  }

  function createPathObject(pathKeys: string[], value: unknown) {
    let current = value;
    while (pathKeys.length !== 0) {
      const key = pathKeys.pop()!;
      current = { [key]: current };
    }
    return current;
  }

  traverse(partial);
  return result;
}
```

packages/vscode-extension/src/utilities/workspaceConfiguration.ts
```diff
@@ -0,0 +1,162 @@
...
+      stopPreviousDevices: config.get<boolean>(
+        WorkspaceConfigurationKeyMap.deviceControl.stopPreviousDevices
+      )!,
+    },
+    deviceSettings: {
+      deviceRotation: config.get<DeviceRotation>(
+        WorkspaceConfigurationKeyMap.deviceSettings.deviceRotation
+      )!,
+      appearance:
+        config.get<Appearance>(WorkspaceConfigurationKeyMap.deviceSettings.appearance)! ?? "dark",
```
jwajgelt:
```suggestion
        config.get<Appearance>(WorkspaceConfigurationKeyMap.deviceSettings.appearance) ?? "dark",
```
surely, the non-null assertion (postfix `!`) and null coalescing (`??`) aren't both needed?

packages/vscode-extension/src/utilities/workspaceConfiguration.ts
```diff
@@ -0,0 +1,162 @@
...
+      startDeviceOnLaunch:
+        config.get<boolean>(WorkspaceConfigurationKeyMap.deviceControl.startDeviceOnLaunch) ?? true,
+      stopPreviousDevices: config.get<boolean>(
+        WorkspaceConfigurationKeyMap.deviceControl.stopPreviousDevices
+      )!,
+    },
+    deviceSettings: {
+      deviceRotation: config.get<DeviceRotation>(
+        WorkspaceConfigurationKeyMap.deviceSettings.deviceRotation
+      )!,
```
jwajgelt: Why can we be sure this will return a value? Does vscode return the default configured in `package.json` if this is not set by the user?

filip131311: it is how it was I will check and do fallbacks anyway just in case

jwajgelt (approved):

PR #1570, opened 2025-09-25
fix: throw error when JS reload fails

jwajgelt:
Fixes hanging on "Waiting for app to load...." due to unhandled errors when reloading JS

### How Has This Been Tested: 
- open app
- introduce a bundle error
- trigger reload JS a couple times
- check radon doesn't get stuck, just shows error alerts

### DIFF
--- a/packages/vscode-extension/src/project/applicationSession.ts
+++ b/packages/vscode-extension/src/project/applicationSession.ts
@@ -64,6 +64,12 @@ function waitForAppReady(inspectorBridge: RadonInspectorBridge, cancelToken?: Ca
   return promise;
 }
 
+export class ReloadJSError extends Error {
+  constructor(message: string) {
+    super(message);
+  }
+}
+
 export class ApplicationSession implements Disposable {
   private disposables: Disposable[] = [];
   private debugSession?: DebugSession & Disposable;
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -308,6 +314,13 @@ export class ApplicationSession implements Disposable {
       debugSession.onDebuggerResumed(this.onDebuggerResumed),
       debugSession.onProfilingCPUStarted(this.onProfilingCPUStarted),
       debugSession.onProfilingCPUStopped(this.onProfilingCPUStopped),
+      // debugSession.onScriptParsed(({ isMainBundle }) => {
+      //   // NOTE: if the scriptParsed event is received for the main bundle,
+      //   // it must have loaded successfully, so we clear any existing bundle error
+      //   if (isMainBundle) {
+      //     this.stateManager.updateState({ bundleError: null });
+      //   }
+      // }),
```
jwajgelt:
```suggestion
```

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -289,6 +289,15 @@ export class DeviceSession implements Disposable {
           },
         });
         return;
+      } else if (e instanceof ReloadJSError) {
```
filip131311, line 292: can we use switch and `this.handle<ErrorType>Error` here it starts to be a bit much

filip131311 (commented):
This does not solve the problem of downgradeing user expirience form the version i which reload wa trigered by metro

PR #1576, opened 2025-09-26
feat: support prebuilding expo dev client applications

filip131311:
This PR adds automatic detection of expo dev client setups. 

additionally we use this opportunity to clean up the `/lib` section of our repository as it was a bit overblown for some time and we add a new file to it.

additionally additionally we fix the way we import expo cli in a path that was missed in https://github.com/software-mansion/radon-ide/pull/1522

### How Has This Been Tested: 

- create a new expo 54 test app and 

internall-ish?


packages/vscode-extension/src/builders/BuildManager.ts
```diff
@@ -92,6 +93,38 @@ export function createBuildConfig(
...
+        return {
+          appRoot,
+          platform: platform as DevicePlatform.Android,
+          env,
+          type: BuildType.DevClient,
+          productFlavor: android?.productFlavor,
+          buildType: android?.buildType,
+          fingerprintCommand,
+        };
+      }
```
jwajgelt, line 126: I'm not convinced this is better than returning `BuildType.Local` with `usePrebuild` from this function if `isDevClientProject` returns `true`.
Currently both options seem to be handled in the same way, just with checks (`Local`+`usePrebuild`) or (`DevClient`) everywhere.

jwajgelt (approved):
It's fine.

packages/vscode-extension/src/builders/BuildManager.ts
filip131311: I am hoping that we'll develop a better "should prebuild" system for local builds and the handling will be different if we will

PR #1577, opened 2025-09-26
fix: inspector bridge message retransmission

jwajgelt:
After #1500 moved the communication channel used for React DevTools to the debugger, we've encountered issues related to devtools messages being lost for backgrounded devices, which are disconnected from the debugger (mostly to avoid UI/UX issues with multiple active debug sessions in vscode).
This PR aims to fix these issues by introducing a message retransmission mechanism to our "Inspector Bridge" implementation, which allows the extension to ask the app for the messages it missed once the debugger is reattached.

The mechanism introduces:
- an incremental `id` field to inspector bridge messages
- an `"ack"` control message sent by the extension to the application to signify i successfully received messages up to a specified `id`
- a `"retransmit"` control message which asks the application to resend all (not-`"ack"`ed) messages with `id`s higher than a specified `id`

### How Has This Been Tested: 
- open app in Radon
- open two devices, A and B
- rotate device while device A is selected
- switch to device B
- check the element inspector works correctly on device B -- since it relies on messages sent during orientation change, if it works, the messages sent while the device was backgrounded were sent successfully

### DIFF
--- a/packages/vscode-extension/lib/inspector_bridge.js
+++ b/packages/vscode-extension/lib/inspector_bridge.js
@@ -6,10 +6,14 @@ if (!agent) {
 }
 
 const messageListeners = [];
+let nextMessageId = 1;
+const unacknowledgedMessages = [];
 
 const inspectorBridge = {
   sendMessage: (message) => {
-    agent.postMessage(message);
+    const messageWithId = { id: nextMessageId++, ...message };
+    unacknowledgedMessages.push(messageWithId);
+    agent.postMessage(messageWithId);
   },
   addMessageListener: (listener) => {
     messageListeners.push(listener);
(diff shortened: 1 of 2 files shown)


filip131311 (commented):
Radon never starts on this branch

packages/vscode-extension/lib/inspector_bridge.js
```diff
@@ -6,10 +6,14 @@ if (!agent) {
 }
 
 const messageListeners = [];
+let nextMessageId = 1;
```
filip131311, line 9: Is it possible to reset this state and not reset the one on extension side? therefore not making it ever retransmit anything?

filip131311 (commented):
in line

packages/vscode-extension/lib/inspector_bridge.js
jwajgelt: It happens on JS reloads, but it's fine, try it. :)
What happens is, when the first message from the app is received, the counter on the extension side is reset to the value sent by the app (since we don't ensure monotonicity there), and everything works fine.

This could cause issues with JS reloads though, since we'd ask for a retransmit with a high id, losing the messages sent during app initialization but before the devtools are set up. However, this shouldn't actually happen, since the devtools connection should survive JS reloads -- the debugger binding is not removed on reloads, and the app-side code checks if it exists and sets up the devtools bridge during initialization if it does.

filip131311 (approved):
thx

PR #1580, opened 2025-09-28
Remove the need to process source maps in extension code

kmagiera:
With some client's apps, we found that very large source maps may crash the wasm based source maps library that we currently use. After investigating, we found that we don't actually need to process source maps at all, and only rely on the vscode-js-debug implementation (with some modifications: https://github.com/software-mansion-labs/vscode-js-debug/pull/8)

1) we used source map to resume breakpoints on skipped files (i.e. in node_modules). This behavior work when skipFiles are provided but only after https://github.com/software-mansion-labs/vscode-js-debug/pull/8 the skipFiles can work reliably well
2) we also used source maps to annotate CPU profile with file location metadata. This is something that vscode-js-debug does already but we just needed to find a way to utilize the built-in profiling mechanism.

This PR, updates the `ProxyDebugAdapter` implementation, such that it no longer use source map manager and doesn't feed source maps into it. Instead, we pass `skipFiles` flag directly to the `vscode-js-debug` session (assuming the mentioned fix is in place).

For the profile annotation to work, we actually change a few more things:
1) we capture the child debug session that vscode-js-debug spawns. Only that session can receive custom start/stop profiler commands
2) we listen to custom profiler events to propagate the profiler state back to the UI
3) we delete all the code responsible for annotating, saving and opening the file, vscode-js-debug already does this.

The only side effect is that we no longer can control the location and file name of the profile file. However, vscode-js-debug already places the file in the same location, only prefixes the file with `vscode-profile-${date}` which is acceptable. If we need more control in the future we'd need to patch `vscode-js-debug` as it doesn't provide any way to customize the profile saving process.

### How Has This Been Tested: 
1) run bsky app w/o https://github.com/software-mansion-labs/vscode-js-debug/pull/8 but with this change
2) set some breakpoint before the app is loaded -> notice that it takes looong time for the breakpoint to turn "red" (activate). This is because the source map processing takes ages
3) with https://github.com/software-mansion-labs/vscode-js-debug/pull/8 this problem no longer occurs and the breakpoints load almost instantly
4) with all the changes in: test setting breakpoints, put some breakpoints under node_modules (for example in alter implementation for RN, then trigger alert), for the breakpoints see that the debugger wouldn't stop there. Also test console.log from that place, see that the log link will show place in app code that triggered the internal method (alert call in this example)
5) test CPU profiler.

### DIFF
--- a/packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
+++ b/packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
@@ -26,21 +26,6 @@ export class ProxyDebugSessionAdapterDescriptorFactory
   }
 }
 
-// strip the wildcard `*` from the sourceMapPathOverrides before passing them to the SourceMapsRegistry
-function sourceMapAliasesFromPathOverrides(
-  sourceMapPathOverrides: Record<string, string>
-): [string, string][] {
-  return Object.entries(sourceMapPathOverrides).map(([key, value]): [string, string] => {
-    if (key.endsWith("*")) {
-      key = key.slice(0, -1);
-    }
-    if (value.endsWith("*")) {
-      value = value.slice(0, -1);
-    }
-    return [key, value];
-  });
-}
-
 const CHILD_SESSION_TYPE = "radon-pwa-node";
 
 export class ProxyDebugAdapter extends DebugSession {
@@ -52,10 +37,6 @@ export class ProxyDebugAdapter extends DebugSession {
 
   constructor(private session: vscode.DebugSession) {
     super();
-
-    const sourceMapAliases = sourceMapAliasesFromPathOverrides(
-      this.session.configuration.sourceMapPathOverrides
-    );
 
     const proxyDelegate = new RadonCDPProxyDelegate();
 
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
```diff
@@ -195,8 +163,43 @@ export class ProxyDebugAdapter extends DebugSession {
...
+              this.sendEvent(new Event("RNIDE_profilingCPUStarted"));
+              break;
+            case "profilerStateUpdate":
+              if (event.body.running === false) {
+                this.sendEvent(new Event("RNIDE_profilingCPUStopped"));
+              }
+              break;
+          }
+        })
       );
```
jwajgelt: Since it uses `this.childDebugSession` and not anything from this scope, we should be able to register this in the `constructor`.

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
```diff
@@ -195,8 +163,43 @@ export class ProxyDebugAdapter extends DebugSession {
...
+        if (session.parentSession?.id === vscDebugSession.id) {
+          this.childDebugSession = session;
+          resolve();
         }
+      });
+      promise.finally(() => onDidStartDisposable.dispose());
+
+      this.disposables.push(
+        debug.onDidReceiveDebugSessionCustomEvent((event) => {
+          console.log("Custom event", event);
```
jwajgelt: Please remove the `console.log`s before merging.

PR #1582, opened 2025-09-29
fix: expo router integration

filip131311:
After #1497 we moved the navigation listener setup to "after" the communication with the application is established, which caused the expo router to integration to not capture information about paths existing in the application. 

This PR fixes it by moving the listeners initialization to application session 

### How Has This Been Tested: 

- run expo-53 test app
- check out the navigation list 
- navigate 
- go home 

internal

### DIFF
--- a/packages/vscode-extension/src/project/applicationSession.ts
+++ b/packages/vscode-extension/src/project/applicationSession.ts
@@ -28,6 +28,9 @@ import {
   DeviceType,
   InspectorAvailabilityStatus,
   InspectorBridgeStatus,
+  NavigationHistoryItem,
+  NavigationRoute,
+  NavigationState,
   StartupMessage,
 } from "../common/State";
 import { isAppSourceFile } from "../utilities/isAppSourceFile";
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -652,6 +678,32 @@ export class ApplicationSession implements Disposable {
     this.stateManager.updateState({ logCounter: 0 });
   }
 
+  public resetHomeTarget() {
+    this.navigationHomeTarget = undefined;
+  }
```
jwajgelt, line 683: I don't like this. I feel like this is `ApplicationSession`'s internal state, and shouldn't be controlled by something outside of it.

We only call this when first starting or reloading the app.
In either case we either:
- create a new `ApplicationSession` which will clear this anyway
- reload JS
Maybe we should simply clear this in `reloadJS`?

kmagiera:
Haven't noticed this before but now see it will have changes conflicting with #1595 

I checked the PR and the code changes seem to be doing much more than covered in the description. Would be good to make the description cleaner as I don't know how much of an overlap does this change have with #1595

packages/vscode-extension/src/project/applicationSession.ts
kmagiera: In #1595 I'm deleting the whole concept of keeping this home target altogether

PR #1584, opened 2025-09-29
fix: preview on expo 54

filip131311:
For some reason on expo-54, the old logic of replacing the `preview` function that is in node_modules is not working so we replace `index.js` of "radon-ide" and "react-native-ide" packages entirely.

### How Has This Been Tested: 

- run expo-54 test app and use preview

internal

### DIFF
--- a/packages/vscode-extension/lib/babel_transformer.js
+++ b/packages/vscode-extension/lib/babel_transformer.js
@@ -79,7 +79,7 @@ function transformWrapper({ filename, src, ...rest }) {
     isTransforming("node_modules/react-native-ide/index.js") || // using react-native-ide for compatibility with old NPM package name
     isTransforming("node_modules/radon-ide/index.js")
   ) {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/lib/babel_transformer.js
```diff
@@ -79,7 +79,7 @@ function transformWrapper({ filename, src, ...rest }) {
     isTransforming("node_modules/react-native-ide/index.js") || // using react-native-ide for compatibility with old NPM package name
     isTransforming("node_modules/radon-ide/index.js")
   ) {
-    src = `${src};preview = require("__RNIDE_lib__/preview.js").preview;`;
+    src = `module.exports = { preview: require("__RNIDE_lib__/preview.js").preview };`;
```
kmagiera: perhaps we could just do `module.exports = require("__RNIDE_lib__/preview.js")`

Issue #1586, opened 2025-09-29
Figure out typescript configuration issue that leads to no warnings when WebSockets are not imported


kmagiera:
**Description**
In #1550 we broke metro.ts by removing WebSocket imports. These are necessary and in extension envieonment and typescript should yell about the missing import.

PR #1587, opened 2025-09-29
Fix Websocket is not defined crash

kmagiera:
In #1550 we broke metro.ts by deleting Websocket import. The imports are needed because in extension environment, unlike in web, WebSocket is not defined globally. Apparently typescript seems to be misconfigured and didn't yield an error in this case (I haven't had time to investigate but created #1586)

This PR adds the missing import.

### How Has This Been Tested: 
1) use "Open dev menu" option -> get WebSocket is not defined error
2) with these changes in, the "Open dev menu" works correctly.

### DIFF
--- a/packages/vscode-extension/src/project/metro.ts
+++ b/packages/vscode-extension/src/project/metro.ts
@@ -1,6 +1,7 @@
 import path from "path";
 import fs from "fs";
 import stripAnsi from "strip-ansi";
+import WebSocket from "ws";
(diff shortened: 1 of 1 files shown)


jwajgelt (approved):
I couldn't reproduce the issue, but I don't think this change can hurt.

kmagiera:
Looks like it doesn't happen on VSCode but on Cursor it does. Could be related to some older node version Cursos (on VSCode 1.99 now) where WebSocket isn't available globally.

PR #1593, opened 2025-09-30
fix: network inspector - fix various syntax highlighting issues 

pFornagiel:
### Description

This PR fixes various issues with the Shiki syntax highlighter used in the network inspector, to properly render network payloads and responses.

Changes:
- Replaced imports from `react-shiki` in favour of `react-shiki/core`, which provides minimal needed functionality reducing the bundle size. [See bundle options](https://github.com/avgvstvs96/react-shiki). 
  In order for the highlighter to work, list of supported languages had to be provided explicitly. The list of supported language highlighting is based on [official RN Network Inspector Frontend](https://github.com/facebook/react-native-devtools-frontend/blob/main/front_end/ui/components/code_highlighter/CodeHighlighter.ts). The `lib/network/networkRequestParsers` and `src/network/utils/requestFormatters` files were also updated accordingly, to support the languages.
  NOTE: The `HighlightedCodeBlock` now imports all the language definition using static imports. The change did not hurt performance on my side, but it should be discussed whether we wish to introduce dynamic imports instead.

- updated the `HighlightedCodeBlock.tsx` component to not attempt to highlight the syntax if text is too large, as it causes performance issues. Relevant information is displayed if this case happens.

image

- remove Shiki itself from `package.json` as it is not needed when `react-shiki` is installed

- updated `vite.config.js` to properly bundle Shiki lang-definition files in separate directory

- updated compiler options in `tsconfig` with new `module` and `moduleResolution`, which helped resolving issues with imports from Shiki. The change did not cause issues when testing and as far as I am aware it should not cause any during development. It Is worth noting however, in case something breaks.
  ```ts
  "compilerOptions": {
    "module": "es2022",
    "moduleResolution": "bundler"
    }
  ```

### How Has This Been Tested: 

- **best test would be building the .vsix file and checking whether everything connected with syntax-highlighting in network-inspector works correctly on production**

Not applicable.

### DIFF
--- a/packages/vscode-extension/src/network/hooks/useThemeExtractor.tsx
+++ b/packages/vscode-extension/src/network/hooks/useThemeExtractor.tsx
@@ -72,7 +72,7 @@ export default function useThemeExtractor() {
     });
 
     return () => observer.disconnect();
-  }, [getThemeData]);
+  }, []);
 
   return editorThemeData;
 }


packages/vscode-extension/src/ai/mcp/models.ts
```diff
@@ -1,4 +1,4 @@
-import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp";
+import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
```
jwajgelt, line 1: Why is this change necessary?

packages/vscode-extension/src/network/components/Tabs/HighlightedCodeBlock.tsx
```diff
@@ -10,16 +46,136 @@ interface HighlightedCodeBlockProps {
...
+    langs: LANG_ARRAY,
+    engine: JS_REGEX_ENGINE,
+  }).then((instance) => {
+    highlighterInstance = instance;
+    return instance;
+  });
+
+  return highlighterPromise;
+};
+
```
jwajgelt: I think it'd be better to move this to a separate file that exports the `getHighlighter` function.

packages/vscode-extension/src/ai/mcp/models.ts
pFornagiel: Without the .js extension typescript interprets it as an error:
```
Cannot find module '@modelcontextprotocol/sdk/server/streamableHttp' or its corresponding type declarations. ts(2307)
```

This happens due to the setting `"moduleResolution": "bundler"` in the tsconfig file.

From what I could gather, with this option TypeScript expects explicit file extensions for ES module imports and the package is published as an ES module

packages/vscode-extension/src/network/components/Tabs/HighlightedCodeBlock.tsx
pFornagiel: Moved to `network/utils/highlighter.ts` in fe2ce340

jwajgelt (approved):

Issue #1579, opened 2025-09-27
Issue loading build with Expo 54 SDK and PNPM workspaces


rhinck:
## Description

When running Radon IDE with a PNPM workspace and Expo SDK 54, the IDE logs consistently show the following message after the app builds:

```
2025-09-26 17:53:57.718 [info] App is taking very long to boot up, it might be stuck. Device preview URL: http://127.0.0.1:54522/stream.mjpeg
```

If you open the provided URL from the log, the app is loading and functioning normally. However, within Radon IDE the **inspector** and **bottom-right settings icons** remain disabled. This behavior occurs on both iOS and Android emulators.

## Environment

- Radon IDE: 1.11.2  
- Expo: 54.0.6  
- macOS: 26.0  

## Steps to Reproduce

1. Clone the [expo-monorepo-example](https://github.com/byCedric/expo-monorepo-example) repo.  
2. Configure Radon IDE with the following `launch.json`:
   ```json
   {
     "configurations": [
       {
         "name": "Radon IDE",
         "type": "radon-ide",
         "request": "launch",
         "appRoot": "apps/example",
         "metroConfigPath": "apps/example/metro.config.js",
         "isExpo": true
       }
     ]
   }

3.	Start the app in Radon IDE.
4.	Wait for the build to complete and observe the IDE logs.


Expected Behavior
	•	The app should launch normally.
	•	Radon IDE should enable the inspector and settings icons, allowing full debugging capabilities.


Actual Behavior
	•	Radon IDE logs indicate the app may be stuck, despite it running fine at the preview URL.
	•	The inspector and settings icons remain disabled in Radon IDE.
	•	This occurs consistently across both iOS and Android emulators.

Notes
	•	Issue is reproducible in both our company’s private monorepo and the public expo-monorepo-example repo.
	•	Behavior seems specific to PNPM workspace + Expo 54 setup.

PR #1594, opened 2025-09-30
feat: support pnpm workspaces

filip131311:
This PR changes how we resolve/require/override modules in application dependencies. 

Previously we assumed that if something is a dependency of `react-native` it can be found in the applications `node_modules` as it is a how `yarn` and `npm` do it, unfortunately while using some monorepo setups (`pnpm workspace` in priticular) only the dependencied of the application are linked in its node modules. ( You can read more about it [here](https://pnpm.io/faq#pnpm-does-not-work-with-your-project-here. ) ) 

To solve this problem we require the dependencies from the dependency directory instead of the app directory. 

additionally after introducing this concept to "react-native" dependencies it felt silly to still do it by hand for "expo", as introduced in #1522, so we move those imports/overides to the new solution as well. 

relates to: #1579

### How Has This Been Tested: 

- Clone the [expo-monorepo-example](https://github.com/byCedric/expo-monorepo-example) repo.
- run it in radon

work in progress

### DIFF
--- a/packages/vscode-extension/lib/babel_transformer.js
+++ b/packages/vscode-extension/lib/babel_transformer.js
@@ -124,19 +124,23 @@ function transformWrapper({ filename, src, ...rest }) {
       version.startsWith("0.76") ||
       version.startsWith("0.77")
     ) {
-      const rendererAsString = fs.readFileSync(`${process.env.RADON_IDE_LIB_PATH}/rn-renderer/react-native-74-77/${rendererFileName}`, "utf-8");
+      const rendererFilePath = path.join(process.env.RADON_IDE_LIB_PATH, "rn-renderer", "react-native-74-77", rendererFileName);
+      const rendererAsString = fs.readFileSync(rendererFilePath, "utf-8");
       src = rendererAsString;
     }
     if (version.startsWith("0.78") || version.startsWith("0.79")) {
-      const rendererAsString = fs.readFileSync(`${process.env.RADON_IDE_LIB_PATH}/rn-renderer/react-native-78-79/${rendererFileName}`, "utf-8");
+      const rendererFilePath = path.join(process.env.RADON_IDE_LIB_PATH, "rn-renderer", "react-native-78-79", rendererFileName);
+      const rendererAsString = fs.readFileSync(rendererFilePath, "utf-8");
       src = rendererAsString;
     }
     if (version.startsWith("0.80") || version.startsWith("0.81")) {
-      const rendererAsString = fs.readFileSync(`${process.env.RADON_IDE_LIB_PATH}/rn-renderer/react-native-80-81/${rendererFileName}`, "utf-8");
+      const rendererFilePath = path.join(process.env.RADON_IDE_LIB_PATH, "rn-renderer", "react-native-80-81", rendererFileName);
+      const rendererAsString = fs.readFileSync(rendererFilePath, "utf-8");
       src = rendererAsString;
     }
     if (version.startsWith("0.82")) {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/lib/pnpm/resolve_pnpm_module.js
```diff
@@ -0,0 +1,124 @@
...
+          } catch (e) { }
+        }
+      }
+    }
+  } catch (e) {
+    console.error("[resolve-pnpm-module]:", e);
+  }
+}
+
+module.exports = { resolvePnpmPackageFromAppRoot };
```
filip131311:
```suggestion
module.exports = { resolvePnpmPackageFromAppRoot };

```

packages/vscode-extension/lib/metro_helpers.js
```diff
@@ -11,7 +13,16 @@ function requireFromAppDir(module, options) {
...
+  try {
+    return require.resolve(module, { paths });
+  } catch (e) {
+    // if there was an error, we try to resolve the module via pnpm resolution
+  }
+
+  // we try pnpm resolution here as it is our only "special-cased" resolution mechanism
+  // if we need to support other package managers, we can add more resolution mechanisms here
+  // and switch between them based on some criteria (e.g. presence of lock file)
+  return resolvePnpmPackageFromAppRoot(module);
```
jwajgelt: Is this really necessary?
Which dependency we try to import is causing problems?

packages/vscode-extension/lib/pnpm/resolve_pnpm_module.js
jwajgelt, line 1: I don't like this approach in principle.
Isn't there some way we could reuse pnpm code for module resolution, if we _need_ to go this route?
I'm weary of missing some special cases -- for example, does this work with [peer dependencies](https://pnpm.io/how-peers-are-resolved) as is?

packages/vscode-extension/lib/babel_transformer.js
```diff
@@ -116,16 +124,20 @@ function transformWrapper({ filename, src, ...rest }) {
       version.startsWith("0.76") ||
       version.startsWith("0.77")
     ) {
-      src = `module.exports = require("__RNIDE_lib__/rn-renderer/react-native-74-77/${rendererFileName}");`;
+      const rendererAsString = fs.readFileSync(`${process.env.RADON_IDE_LIB_PATH}/rn-renderer/react-native-74-77/${rendererFileName}`, "utf-8");
```
kmagiera: would prefer that we use `path.join` (maybe it will work on windows too)

packages/vscode-extension/lib/metro_helpers.js
```diff
@@ -11,6 +11,34 @@ function requireFromAppDir(module, options) {
 
 function resolveFromAppDir(module, options) {
   const paths = options && options.paths ? [...options.paths, appRoot] : [appRoot];
+
+  return require.resolve(module, { paths });
+}
+
+function requireFromAppDependency(dependency, module, options) {
```
jwajgelt, line 18: VERY NIT: I'd change the order of parameters in the `FromAppDependency` function to have the `dependency` last -- it feels more natural to me that the signature is "like `require`" (so it takes `(module, options)`) + an extra parameter at the end.

filip131311: I thought about it, but the options argument is optional, forcing the ordering (and (module, dependency, options) seems wrong

jwajgelt: It seems less wrong to me than the alternative...

PR #1600, opened 2025-10-02
feat: skip waiting for fingerprint on autoreload

jwajgelt:
Currently, when "Auto Reload" is requested, we calculate the project's fingerprint to check, if we need to rebuild the app. This can make the reloads costly on larger projects.
This PR changes the behavior to instead check the "isUsingStaleBuild" state, which is set on fingerprint mismatch, which allows us to skip the calculation and start reloading immediately.
We also immediately check if "isUsingStaleBuild" should be set, so that if the user auto-reloaded before some native project changes were picked up by Radon, they will be prompted to reload again as soon as the fingerprint calculation completes.

### How Has This Been Tested: 
- open an app which takes a long time to calculate the fingerprint, for example by setting the `fingerprintCommand` launch configuration to 
```
sleep 10; echo 'A';
```
- trigger autoreload
- the app should restart quickly, without waiting for the fingerprint to calculate


packages/vscode-extension/src/project/FingerprintProvider.ts
```diff
@@ -49,6 +49,7 @@ export async function runFingerprintCommand(
     const output = await command(fingerprintCommand, {
       env,
       cwd: appRoot,
+      shell: true,
```
kmagiera: this was done for testing only, do we actually want to keep it? If yes, should it be consistent with the way we launch the custom build command as well?

packages/vscode-extension/src/utilities/throttle.ts
```diff
@@ -58,5 +59,10 @@ export function throttleAsync<T extends AsyncFn>(
     }
   };
 
+  throttledFunction.flush = () => {
+    throttledFunction.cancel();
+    execute();
```
kmagiera: I'm a bit sceptical about implementing it this way. There are two problems:
1) minor one is that `flush` in lodash doesn't execute if there was no call scheduled. Now we always execute right before flush so this is ok, but just inconsistent
2) the second more serious issue is that this way of implementing it may lead to the fingerprint getting executed concurrently. In particular, when fingerprint is already running (async), the call for flush would execute the second concurrent call.

kmagiera (approved):
Looks mostly ok, just have concerns about the flush implementation I described in inline comments.

packages/vscode-extension/src/project/FingerprintProvider.ts
jwajgelt: I added it for tests initially, but I thought it made sense.
I can revert this change, I suppose it might break something for some users.

packages/vscode-extension/src/utilities/throttle.ts
jwajgelt: > minor one is that flush in lodash doesn't execute if there was no call scheduled. Now we always execute right before flush so this is ok, but just inconsistent

My idea is that it wouldn't, since the `recentArgs` variable would be `null` in that case... but now that I look closer I see that we only clear them once the promise resolves. I'll fix that.

> the second more serious issue is that this way of implementing it may lead to the fingerprint getting executed concurrently. In particular, when fingerprint is already running (async), the call for flush would execute the second concurrent call.

Yes, I thought about that. I don't know if that's a serious issue in this specific case, but we may want to use it elsewhere down the line and the person using it may not be aware of this. I'll fix that as well.

PR #1606, opened 2025-10-02
feat: network inspector - support new inspector (fuseboxNetworkInspection)

pFornagiel:
### Description

This PR introduces support for the new network inspector React Native implementation (`fuseboxNetworkInspection`). 

As of today, this feature is not enabled in React Native by default, in order to toggle it on consult `src/plugins/network/README.md`.

What has been changed:
- `bridge.ts` has been generalised - there is now an abstract `GenericBridge` class which allows for construction of communication-bridges in the application. Apart from the already used `RadonInspectorBridge`, the `NetworkBridge` has been introduced to allow for communication between `ApplicationCommand`s (which now streamline the network events) and `network-plugin.ts`
- Network plugin now includes support for the new inspector and legacy methods used so far for backward compatibility. In order to turn on the new functionalities, `ENABLE_NEW_INSPECTOR` flag has to be set to `true`
- In the old `lib/network/network.js` network-event-handling, the `xhr._aborted` and `xhr._error` is now correctly set and considered when sending network events, to prevent situation, where request fails, but after failure still sends `Network.loadingFinished` event, which caused issues.
- made styling changes and patches to network log table, which now properly formats, displays information about failed logs and makes them properly highlighted
image
 
### How Has This Been Tested: 

- open an application supporting Radon IDE in editor - example of app supporting  rotation: radon-ide-test-apps/react-native-82 **(new inspector works on 81 and 82)**
- **turn the new inspector support on as provided in `README.md`, make sure everything behaves as expected and there are no undefined behaviour**

Not applicable.

### DIFF
--- a/packages/vscode-extension/src/builders/buildIOS.ts
+++ b/packages/vscode-extension/src/builders/buildIOS.ts
@@ -94,7 +94,6 @@ function buildProject(
     "-configuration",
     configuration,
     "TARGETED_DEVICE_FAMILY=1,2",
-    'OTHER_CFLAGS="-DREACT_NATIVE_DEBUGGER_ENABLED=1"',
     "-scheme",
     scheme,
     "-destination",


packages/vscode-extension/src/builders/buildIOS.ts
```diff
@@ -94,6 +94,7 @@ function buildProject(
     "-configuration",
     configuration,
     "TARGETED_DEVICE_FAMILY=1,2",
+    'OTHER_CFLAGS="-DREACT_NATIVE_DEBUGGER_ENABLED=1"',
```
jwajgelt: Remember to remove this before merging.

packages/vscode-extension/src/network/hooks/useNetworkTracker.tsx
```diff
@@ -28,9 +31,95 @@ export const networkTrackerInitialState: NetworkTracker = {
...
+    }
+
+    if (loadingFinishedTimestamp !== undefined && requestWillBeSentTimestamp !== undefined) {
+      timingData.totalTime = this.calculateTimeDifference(
+        requestWillBeSentTimestamp,
+        loadingFinishedTimestamp
+      );
+    }
+
+    if (loadingFinishedTimestamp !== undefined && responseReceivedTimestamp !== undefined) {
```
jwajgelt: NIT:
You could hide this check inside `this.calculateTimeDifference`

```
private calculateTimeDifference(timeStart: number | undefined, timeEnd: number | undefined): number | undefined {
    if (timeStart === undefined || timeEnd === undefined) return undefined;
    return _.round((timeEnd - timeStart) * 1000, 2);
}
```
removing branching from this function and making it more straightforward read.
It's up to you which you find more readable though.

packages/vscode-extension/src/network/hooks/useNetworkTracker.tsx
```diff
@@ -41,7 +130,8 @@ const useNetworkTracker = (): NetworkTracker => {
         return null;
       }
 
-      const haveRequiredFields = payload.params?.timestamp && payload.params?.requestId;
+      // Timestamp check was removed because Network.requestExtraInfo does not have information about it
```
jwajgelt: This comment won't make sense outside of this PR, since when you read the code in the future, you won't know the timestamp check was ever here. I'd remove it.

packages/vscode-extension/src/network/utils/networkLogParsers.ts
```diff
@@ -23,39 +50,54 @@ const NetworkLogFormatters = {
...
-    }
-    const units = ["B", "KB", "MB", "GB", "TB"];
-    let unitIndex = 0;
-    let formattedSize = size;
-    while (formattedSize >= 1024 && unitIndex < units.length - 1) {
-      formattedSize /= 1024;
-      unitIndex++;
-    }
-    return `${parseFloat(formattedSize.toFixed(2) || "")} ${units[unitIndex]}`;
+    const exctractLogSize = () => {
```
jwajgelt:
```suggestion
    const extractSize = () => {
```

packages/vscode-extension/src/network/utils/networkLogParsers.ts
```diff
@@ -1,12 +1,39 @@
...
   remainingText: string;
 }
 
+/**
+ * Higher-order function that handles common state checking logic for network log formatters.
+ * Returns appropriate messages based on the loading state and value presence.
+ */
+function getLogValueWithChecks<T>(
+  log: NetworkLog,
+  extractLogValue: (log: NetworkLog) => T | undefined | null,
```
jwajgelt: It looks like you always pass a function which ignores the `log` argument. You could just us well just accept a `value: T | undefined | null`.

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -1,39 +1,32 @@
...
+/**
+ * Toggles usage of handling the new network inspector devtools communication.
+ * Disabled by default until new network inspector is integradated into react native.
+ * The instructions to enable the features are in plugins/network/README.md
+ */
+const ENABLE_NEW_INSPECTOR = false;
+
+export type BroadcastListener = (message: WebviewMessage) => void;
+
+export interface InspectorStrategy {
```
jwajgelt: `Inspector` is an overloaded term in Radon -- there's the "Network Inspector", "Element Inspector", "InspectorBridge" (which you even reference in this file)... I think it'd be better to be explicit in this case, and call this `NetworkInspectorStrategy`

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -56,182 +49,45 @@ async function initialize() {
...
+  private readonly strategy: InspectorStrategy;
 
-  constructor(private readonly inspectorBridge: RadonInspectorBridge) {
+  constructor(
+    readonly inspectorBridge: RadonInspectorBridge,
+    readonly networkBridge: NetworkBridge
+  ) {
+    this.strategy = ENABLE_NEW_INSPECTOR
+      ? new NewInspectorStrategy(this)
+      : new LegacyInspectorStrategy(this);
```
jwajgelt: I wouldn't really call it "The Strategy Pattern" when you delegate literally every call to the object...
I'd just call the interface `NetworkInspector`, with two implementations.

packages/vscode-extension/src/plugins/network/strategies/NewInspectorStrategy.ts
```diff
@@ -0,0 +1,171 @@
...
+
+import { NETWORK_EVENTS } from "../../../network/types/panelMessageProtocol";
+import { NETWORK_EVENT_MAP, NetworkBridge, RadonInspectorBridge } from "../../../project/bridge";
+import { ResponseBodyData } from "../../../network/types/network";
+
+// Truncation constants
+const MAX_MESSAGE_LENGTH = 1000000;
+const TRUNCATED_LENGTH = 1000000;
+
+export default class NewInspectorStrategy extends BaseInspectorStrategy {
```
jwajgelt: `New` and `Legacy` aren't very helpful for a person which doesn't have all the context of how the implementations came to be.
Maybe `DebuggerNetworkInspector` and `InspectorBridgeNetworkInspector` (based on the underlying communication channel)?

packages/vscode-extension/src/plugins/network/strategies/DebuggerNetworkInspector.ts
```diff
@@ -0,0 +1,171 @@
...
+      case NetworkMethod.GetResponseBody:
+        this.handleGetResponseBody(payload);
+        break;
+    }
+  }
+
+  public activate(): void {
+    if (!this.pluginAvailable) {
+      return;
+    }
```
jwajgelt, line 183: Isn't this an error state? Will we ever activate if we go into this branch?

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -311,13 +316,25 @@ export class ApplicationSession implements Disposable {
     }
   };
 
+  private onNetworkEvent = (event: DebugSessionCustomEvent): void => {
+    const method = event.body?.method;
+    if (!method || !isCDPMethod(method)) {
+      Logger.error("Unknown network event method:", method);
+      this.networkBridge.emitEvent("unknownEvent", event.body);
```
jwajgelt, line 327: Is there any reason to emit an event here? Why not just log the error and `return`?

packages/vscode-extension/src/project/bridge.ts
```diff
@@ -2,45 +2,34 @@ import { Disposable } from "vscode";
...
+ * - emitEvent method to invoke listeners,
+ * - abstract send method that subclasses must implement to forward messages externally
+ *
+ * @template E - Map of event names to listener-argument tuples (e.g. { foo: [number, string] }).
+ * @template K - Union of string literal event names (keys of E).
+ */
+abstract class GenericBridge<E extends BridgeEventsMap<K>, K extends string> {
+  private listeners = new Map<K, Array<(...payload: any) => void>>();
 
   protected abstract send(message: any): void;
```
jwajgelt: This method isn't public and isn't referenced anywhere in this class -- IMO, it shouldn't be here.
Once you remove it, this class becomes more-or-less just node's `EventEmitter` -- but which returns the `Disposable` from listener registration.
I'd change this class to just be an `EventEmitter`, possibly rewrite to use the `node` implementation internally (to avoid having to write the listeners management ourselves, like we do now) and have the same API as currently, with the `onEvent` returning `Disposable`.
Since this is intended to always be extended by the class actually firing the events, you may want to make `emitEvent` protected as well.
(the name `EventEmitter` is bad, since it conflicts with both `node` and `vscode`'s `EventEmitter`s, both of which are used in the project... but I don't have an alternative. Event Source maybe?)

jwajgelt (changes requested):

packages/vscode-extension/src/network/hooks/useNetworkTracker.tsx
pFornagiel: Changed in 03d72f67 as suggested

packages/vscode-extension/src/network/hooks/useNetworkTracker.tsx
pFornagiel: Removed in 0155f48a

packages/vscode-extension/src/network/utils/networkLogParsers.ts
pFornagiel: Applied suggestion in 61548173, refactored to pass the `value: T | undefined | null` into `getLogValueWithChecks` function and extracted `extractSize` outside the object for consistency

packages/vscode-extension/src/network/utils/networkLogParsers.ts
pFornagiel: Changed name in https://github.com/software-mansion/radon-ide/commit/61548173e28bd16a4e5f49ea26897e017f91c995, and moved the function outside the object definition.

packages/vscode-extension/src/plugins/network/network-plugin.ts
pFornagiel: Renamed to `NetworkInspector` in 61548173 for consistency with other naming decisions after refactor.

List of changes:
- gotten rid of strategy pattern convention in naming
- renamed interface `InspectorStrategy` to `NetworkInspector`
- renamed `NetworkPlugin` `strategy` private property to `networkInspector`
- renamed class `BaseInspectorStrategy` to `BaseNetworkInspector`
- renamed class `NewInspectorStrategy` to `DebuggerNetworkInspector`
- renamed class `LegacyInspectorStrategy` to `InspectorBridgeNetworkInspector`
- renamed flag `ENABLE_DEBUGGER_INSPECTOR` to `ENABLE_DEBUGGER_INSPECTOR`

packages/vscode-extension/src/plugins/network/network-plugin.ts
pFornagiel: Renamed to `networkInspector` in 61548173 for consistency with other naming decisions after refactor.

List of changes:
- gotten rid of strategy pattern convention in naming
- renamed interface `InspectorStrategy` to `NetworkInspector`
- renamed `NetworkPlugin` `strategy` private property to `networkInspector`
- renamed class `BaseInspectorStrategy` to `BaseNetworkInspector`
- renamed class `NewInspectorStrategy` to `DebuggerNetworkInspector`
- renamed class `LegacyInspectorStrategy` to `InspectorBridgeNetworkInspector`
- renamed flag `ENABLE_DEBUGGER_INSPECTOR` to `ENABLE_DEBUGGER_INSPECTOR`

packages/vscode-extension/src/plugins/network/strategies/NewInspectorStrategy.ts
pFornagiel: 61548173 , renames explained below:
List of changes:
- renamed class `BaseInspectorStrategy` to `BaseNetworkInspector`
- renamed class `NewInspectorStrategy` to `DebuggerNetworkInspector`
- renamed class `LegacyInspectorStrategy` to `InspectorBridgeNetworkInspector`

packages/vscode-extension/src/plugins/network/strategies/DebuggerNetworkInspector.ts
pFornagiel: Right, this could pose problems in the future if setting debug session happens after the network inspector is activated for some reason.

In efdd22d9 introduced better activationState handling in `DebuggerNetworkInspector`, as well as `bridgeAvailable` event in NetworkBridge

packages/vscode-extension/src/plugins/network/strategies/BaseNetworkInspector.ts
```diff
@@ -0,0 +1,160 @@
...
+  IDEMethod,
+} from "../../../network/types/panelMessageProtocol";
+import { RequestData, RequestOptions } from "../../../network/types/network";
+import { Logger } from "../../../Logger";
+import { determineLanguage } from "../../../network/utils/requestFormatters";
+import { openContentInEditor, showDismissableError } from "../../../utilities/editorOpeners";
+import { extractTheme } from "../../../utilities/themeExtractor";
+import { ContentTypeHeader } from "../../../network/types/network";
+
+export abstract class BaseInspectorStrategy implements NetworkInspector {
```
jwajgelt:
```suggestion
export abstract class BaseNetworkInspector implements NetworkInspector {
```
you probably used vsocde's rename in the wrong place, and renamed an import instead of this class.

packages/vscode-extension/src/project/bridge.ts
```diff
@@ -1,10 +1,59 @@
...
+ *
+ * This class manages event listeners keyed by event names and exposes:
+ * - onEvent method to register listeners,
+ * - emitEvent method to invoke listeners,
+ * - abstract send method that subclasses must implement to forward messages externally
+ *
+ * @template E - Map of event names to listener-argument tuples (e.g. { foo: [number, string] }).
+ * @template K - Union of string literal event names (keys of E).
+ */
+abstract class EventDispatcher<E extends EventMap<K>, K extends string> {
```
jwajgelt: Could you move this to its own file?

packages/vscode-extension/src/project/bridge.ts
```diff
@@ -120,3 +121,102 @@ export abstract class BaseInspectorBridge implements RadonInspectorBridge {
...
+  onEvent<K extends NetworkBridgeEventNames>(
+    event: K,
+    listener: (...payload: RadonNetworkBridgeEvents[K]) => void
+  ): Disposable;
+}
+
+export type NetworkBridgeSendMethodArgs = Record<string, unknown>;
+export type NetworkBridgeGetResponseBodyArgs = { requestId: string | number };
+
+export class NetworkBridge
```
jwajgelt: I'd move this to a separate file as well.

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -56,185 +49,45 @@ async function initialize() {
...
+  private readonly networkInspector: NetworkInspector;
 
-  constructor(private readonly inspectorBridge: RadonInspectorBridge) {
+  constructor(
+    readonly inspectorBridge: RadonInspectorBridge,
+    readonly networkBridge: NetworkBridge
+  ) {
+    this.networkInspector = ENABLE_DEBUGGER_INSPECTOR
+      ? new DebuggerNetworkInspector(this)
+      : new InspectorBridgeNetworkInspector(this);
```
jwajgelt: I think I'd prefer if you just passed `inspectorBridge` and `networkBridge` here explicitely.

Then they don't need to be properties of `NetworkPlugin` either, since I don't think they're referenced outside of the `Inspector` implementations.

packages/vscode-extension/src/plugins/network/network-plugin.ts
```diff
@@ -56,185 +49,45 @@ async function initialize() {
 export class NetworkPlugin implements ToolPlugin {
   public readonly id: ToolKey = NETWORK_PLUGIN_ID;
   public readonly label = "Network";
-  public readonly persist = true;
-
-  public pluginAvailable = true;
+  public readonly persist = false;
```
jwajgelt: Why was this changed to `false`?

packages/vscode-extension/src/project/eventDispatcher.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * This class manages event listeners keyed by event names and exposes:
+ * - onEvent method to register listeners,
+ * - emitEvent method to invoke listeners,
+ * - abstract send method that subclasses must implement to forward messages externally
+ *
+ * @template E - Map of event names to listener-argument tuples (e.g. { foo: [number, string] }).
+ * @template K - Union of string literal event names (keys of E).
+ */
+export default abstract class EventDispatcher<E extends EventMap<K>, K extends string> {
```
jwajgelt:
```suggestion
export abstract class EventDispatcher<E extends EventMap<K>, K extends string> {
```

packages/vscode-extension/src/project/eventDispatcher.ts
```diff
@@ -0,0 +1,39 @@
...
+
+type EventMap<K extends string> = Record<K, unknown[]>;
+
+/**
+ * Abstract base class that provides an event-based bridging mechanism.
+ *
+ * This class manages event listeners keyed by event names and exposes:
+ * - onEvent method to register listeners,
+ * - emitEvent method to invoke listeners,
+ * - abstract send method that subclasses must implement to forward messages externally
```
jwajgelt:
```suggestion
```

packages/vscode-extension/src/project/inspectorBridge.ts
```diff
@@ -0,0 +1,76 @@
...
+export interface RadonInspectorBridge {
+  sendPluginMessage(pluginId: string, type: string, data: any): void;
+  sendInspectRequest(x: number, y: number, id: number, requestStack: boolean): void;
+  sendOpenNavigationRequest(id: string): void;
+  sendOpenPreviewRequest(previewId: string): void;
+  sendShowStorybookStoryRequest(componentTitle: string, storyName: string): void;
+  onEvent<K extends RadonInspectorEventName>(
+    event: K,
+    listener: (...payload: RadonInspectorBridgeEvents[K]) => void
+  ): Disposable;
```
jwajgelt: Might be worth introducing a
```
interface EventDispatcher<E extends EventMap<K>, K extends string> {
  onEvent<L extends K>(event: L, listener: (...payload: E[L]) => void): Disposable
}
```
and have the `EventDispatcher` abstract class implement that interface (and rename it to something like `EventDispatcherImpl`/`ConcreteEventDispatcher`/whatever), so that we don't have to write out this signature in every interface we want to expose the events.

PR #1617, opened 2025-10-03
Avoid concurrent initialization of devtools connection over CDP binding

kmagiera:
This problem fixes an issue with concurrent initialization of the CDP-backed devtools connection.

Before this change, we would call the connection initialization code from both `onJSDebugSessionStarted` and `onScriptParsed` when the main bundle got loaded. We had a check for whether the connection was already set to avoid initializing it twice. However, since the initialization was async, it could happen that both of these calls would advance past the check. This eventually resulted in us emitting two connect events and shutting down a functioning connection.

While in practice we'd expect that re-creating the agent should work alright and we should just get a second instance of an agent. It appears however that interleaving the binding call and evaluate call may confuse the devtools backend code too much and results in the agent connection being dropped entirely. When testing this, if the second initialization call happens after some delay, this work correctly.

However, we still want to avoid two connections being initialized and therefore we need to prevent the concurrent initialization anyway.

The consequence of the above bug was that the devtools connection would shut down shortly after being initialize and as a result we wouldn't be able to receive communication over the devtools channel (i.e. the network inspector calls).

So one way we could reproduce the problem was on a simple expo 52 dev client app. After starting the app and opening the network panel, we woudn't be able to see network requests. However, after "reload JS" the request would start appearing as the connection was re-established.

### How Has This Been Tested: 
1. I could only reproduce the issue on small expo-52 dev-client setups
2. `npx create-expo-app@latest --template default@52`
3. `npx expo install expo-dev-client`
4. Add button that performs network fetch
5. before the change after launching the app and opening network inspector you couldn't see requests coming until you click fast refresh
6. after this change the requests are showing up after clean start

### DIFF
--- a/packages/vscode-extension/src/project/devtools.ts
+++ b/packages/vscode-extension/src/project/devtools.ts
@@ -200,27 +200,40 @@ const DEVTOOLS_DOMAIN_NAME = "react-devtools";
 
 export class CDPDevtoolsServer extends DevtoolsServer implements Disposable {
   private disposables: Disposable[] = [];
+  private initializeConnectionPromise: Promise<void> | undefined;
 
   constructor(private readonly debugSession: DebugSession) {
     super();
     this.disposables.push(
       debugSession.onJSDebugSessionStarted(() => {
-        this.createConnection();
+        this.maybeInitializeConnection();
       }),
       debugSession.onScriptParsed(({ isMainBundle }) => {
         if (isMainBundle) {
-          this.createConnection();
+          this.maybeInitializeConnection();
         }
       })
     );
   }
 
-  private async createConnection() {
-    const debugSession = this.debugSession;
+  private maybeInitializeConnection() {
     if (this.connection) {
       // NOTE: a single `DebugSession` only supports a single devtools connection at a time
       return;
+    } else if (this.initializeConnectionPromise) {
+      // devtools connections cannot be initialized concurrently, we only can establish
+      // one connection, so if we are already in a process of initializing one, we only schedule
+      // a new attempt if the previous one fails
+      this.initializeConnectionPromise = this.initializeConnectionPromise.catch(() => {
+        return this.initializeConnection();
+      });
+    } else {
+      this.initializeConnectionPromise = this.initializeConnection();
     }
+  }
+
+  private async initializeConnection() {
+    const debugSession = this.debugSession;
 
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/project/devtools.ts
```diff
@@ -200,27 +200,40 @@ const DEVTOOLS_DOMAIN_NAME = "react-devtools";
...
       return;
+    } else if (this.initializeConnectionPromise) {
+      // devtools connections cannot be initialized concurrently, we only can establish
+      // one connection, so if we are already in a process of initializing one, we only schedule
+      // a new attempt if the previous one fails
+      this.initializeConnectionPromise = this.initializeConnectionPromise.catch(() => {
+        return this.initializeConnection();
+      });
+    } else {
+      this.initializeConnectionPromise = this.initializeConnection();
```
jwajgelt: Should we catch the errors from this promise and log something?
Right now, if the second initialization attempt doesn't happen before the first throws (which would install the `catch` handler in the previous branch), we'll get an "Uncaught promise rejection" in the output logs, which I'd prefer to avoid (it looks like an unhandled / unexpected case, which this is not)
```suggestion
      this.initializeConnectionPromise = this.initializeConnection();
      this.initializeConnectionPromise.catch((e) => {
        Logger.error("Devtools connection could not be established.")
      })
```

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -261,7 +274,9 @@ export class CDPDevtoolsServer extends DevtoolsServer implements Disposable {
       shutdownListener.dispose();
     });
     connection.onDisconnected(() => {
-      this.setConnection(undefined);
+      if (this.connection === connection) {
+        this.setConnection(undefined);
+      }
```
jwajgelt, line 291: I don't think this check can ever fail?
It's still fine to leave it as is, of course, since we might change things around it in the future.

jwajgelt (approved):
Seems fine. Thank you for taking care of this 🙏

packages/vscode-extension/src/project/devtools.ts
kmagiera: wanted to add it here since session is actually controlled by the superclass. One thing I was wondering was whether the resetting connection logic shouldn't be a part of the superclass, but was too afraid to change this bit

PR #1624, opened 2025-10-06
Don't enable Radon AI on other VSCode forks

kmagiera:
Radon AI causes some additional friction for the users on Cursor (and other VSCode forks) as it requires that we update mcp.json file every time the editor starts. This doesn't happen on VSCode as since version 1.100 there's an API that allows our extension to register the MCP server dynamically without updating the configuration file.

As a consequence, Radon AI may get in a way of users who doesn't want to use it. 

On top of the mcp json file being generated in the user's workspace, Cursor recently started asking whether to enable MCP servers added there. Because the mcp.json record has a port number that changes with every start, Radon AI generates prompts with every start as there's a new, unique record added to mcp.json each time.

Below are highlights of what this PR is changing:
1. We're changing the setting name to RadonAi.enabled and use 'enum' based setting instead of boolean. Boolean setting doesn't allow us to tell when the setting is not set (default value). It is important as we use different defaults for VSCode and Cursor now. Users who had turned the setting off on VSCode will need to turn it off again, but since Radon Ai isn't intrusive on VSCode as it is on Cursor I don't think this should be an issue.
2. We restructure the code around initialization: we move functions and types for detecting VSCode vs Cursor to general utility file (was under "ai" before). We refactor RadonAi loading scheme such that it reads from the config and decides about the defaults afterwards.
3. We delege all the code responsible for defining radon ai settings in our configuration scheme. This wasn't used anywhere and it was just a lot of boilerplate
4. We update the registration logic such that it can respond to changes in the setting in runtime. This allows us to modify the setting description to no longer include the note about resetting.

### How Has This Been Tested: 
1. Run extension on Cursor
5. Don't get "enable MCP server" prompt
6. Change config to enable Radon AI
7. Get the prompt there and use Radon in the chat.
8. Do the same in cursor – verify Radon AI is available w/o changes to the setting

### DIFF
--- a/packages/vscode-extension/src/ai/mcp/index.ts
+++ b/packages/vscode-extension/src/ai/mcp/index.ts
@@ -18,6 +18,7 @@ import { LocalMcpServer } from "./LocalMcpServer";
 import { disposeAll } from "../../utilities/disposables";
 import { ConnectionListener } from "../shared/ConnectionListener";
 import { EditorType, getEditorType } from "../../utilities/editorType";
+import { RadonAIEnabledState } from "../../common/State";
 import { registerRadonChat } from "../chat";
 
 async function removeMcpConfig(location: ConfigLocation) {
@@ -110,8 +111,13 @@ function isEnabledInSettings() {
   // and hence we want the users to explicitely enable it
   const enableRadonAiByDefault = getEditorType() === EditorType.VSCODE;
 
-  const enabledVal = workspace.getConfiguration("RadonIDE").get<string>("radonAI.enabled");
-  return enabledVal === "enabled" || (enabledVal === "default" && enableRadonAiByDefault);
+  const enabledVal = workspace
+    .getConfiguration("RadonIDE")
+    .get<RadonAIEnabledState>("radonAI.enabled");
+  return (
+    enabledVal === RadonAIEnabledState.Enabled ||
+    (enabledVal === RadonAIEnabledState.Default && enableRadonAiByDefault)
+  );
 }
 
 export default function registerRadonAi(context: ExtensionContext): Disposable {
--- a/packages/vscode-extension/src/common/State.ts
+++ b/packages/vscode-extension/src/common/State.ts
@@ -420,6 +420,11 @@ export enum DeviceType {
   Tablet = "Tablet",
 }
 
+export enum RadonAIEnabledState {
+  Enabled = "enabled",
+  Default = "default",
+}
+
 export type DeviceInfo = AndroidDeviceInfo | IOSDeviceInfo;
 
 export type AndroidDeviceInfo = {


packages/vscode-extension/src/utilities/workspaceConfiguration.ts
```diff
@@ -45,6 +47,12 @@ const WorkspaceConfigurationKeyMap = {
 };
 
 export function getCurrentWorkspaceConfiguration(config: WorkspaceConfiguration) {
+  // Radon AI is enabled by default on VSCode only, where we can use the API to register the MCP server
+  // On other editors, we need to write to mcp.json file which introduces additional friction for the user
+  // and hence we want the users to explicitely enable it
+  const editorType = getEditorType();
+  const enableRadonAIByDefault = editorType === EditorType.VSCODE;
```
latekvo: Perhaps it'd be better to change the default write location of `mcp.json` to `~/.cursor/mcp.json` instead of `<project>/.cursor/mcp.json`? We already have this capability implemented, but it's turned off by default:

```diff
radonAI: {
      enableRadonAI:
        config.get<boolean>(WorkspaceConfigurationKeyMap.radonAI.enableRadonAI) ?? true,
      MCPConfigLocation:
        config.get<MCPConfigLocation>(WorkspaceConfigurationKeyMap.radonAI.MCPConfigLocation) ??
-        "Project",
+        "Global",
    },
```

latekvo (approved):

packages/vscode-extension/src/utilities/workspaceConfiguration.ts
kmagiera: This still wouldn't solve the issue with cursor prompting to enable MCP and will on top of that cause issues when different windows use different MCP instances

latekvo: You're right, my bad. Disabling it altogether is probably the best approach then.

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -77,19 +104,61 @@ function isDirectLoadingAvailable() {
...
+function isEnabledInSettings() {
+  // Radon AI is enabled by default on VSCode only, where we can use the API to register the MCP server
+  // On other editors, we need to write to mcp.json file which introduces additional friction for the user
+  // and hence we want the users to explicitely enable it
+  const enableRadonAiByDefault = getEditorType() === EditorType.VSCODE;
 
-  if (isDirectLoadingAvailable()) {
-    const disposables = directLoadRadonAI(server);
+  const enabledVal = workspace.getConfiguration("RadonIDE").get<string>("radonAI.enabled");
+  return enabledVal === "enabled" || (enabledVal === "default" && enableRadonAiByDefault);
```
latekvo: Would prefer to have a type for these string literals instead of just typing them as `string`, but that's just a nitpick.

packages/vscode-extension/src/utilities/editorType.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export enum EditorType {
+  CURSOR = "cursor",
+  VSCODE = "vscode",
+}
+
+export function getEditorType(): EditorType {
+  // Cursor features different settings than VSCode
+  const config = workspace.getConfiguration();
+  if (config.get("cursor") !== undefined) {
```
latekvo, line 11: nit: we don't need `!== undefined` here, do you prefer to keep it either way?

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -77,19 +104,61 @@ function isDirectLoadingAvailable() {
...
+    if (isDirectLoadingAvailable()) {
+      directLoadRadonAI(server, disposables);
+    } else {
+      disposables.push(
+        server.onReload(() => {
+          fsLoadRadonAI(server);
+        })
+      );
+    }
+  }
```
latekvo, line 144: Do we have to **_nest_** this function definition here? It makes the code really unreadable, and the only thing that is being used from the outer scope is `disposables`.

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -77,19 +104,61 @@ function isDirectLoadingAvailable() {
...
 
-    return new Disposable(() => disposeAll([onReloadDisposable, server, connectionListener]));
+  function unloadRadonAi() {
+    fsUnloadRadonAi();
```
latekvo: `unloadRadonAi` doesn't do anything besides wrapping `fsUnloadRadonAi`. Can't `fsUnloadRadonAi` be used directly?

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -77,19 +104,61 @@ function isDirectLoadingAvailable() {
...
+    }
+  });
+
+  if (radonAiEnabled) {
+    loadRadonAi();
+  } else {
+    unloadRadonAi();
+  }
+
+  return new Disposable(() => {
```
latekvo, line 163: I think we should also run `unloadRadonAi` on disposal, because otherwise we'll be leaving dead entries in the `mcp.json`. This way, on the next cursor launch, cursor won't try loading an entry that won't work anymore, but will instead wait for a valid entry.

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -77,19 +104,61 @@ function isDirectLoadingAvailable() {
...
+    disposeAll(disposables);
+    configChangeDisposable.dispose();
```
latekvo, line 168: We can push `configChangeDisposable` to `disposables` instead having it as a separate `const` .

packages/vscode-extension/src/ai/mcp/index.ts
```diff
@@ -18,39 +38,46 @@ async function updateMcpConfig(port: number) {
...
 
-  return new Disposable(() => disposeAll([onReloadDisposable, mcpServerEntry]));
+async function fsUnloadRadonAi() {
+  await Promise.all([
+    removeMcpConfig(ConfigLocation.Global),
+    removeMcpConfig(ConfigLocation.Project),
+  ]).catch((e) => {
+    Logger.error(
+      MCP_LOG,
+      `Failed removing MCP config: ${e instanceof Error ? e.message : String(e)}`
```
latekvo, line 79: nit: `e instanceof Error ? e.message : String(e)` pattern occurs over and over in our code, could we isolate it into a small util?

packages/vscode-extension/package.json
```diff
@@ -408,11 +408,12 @@
...
-        "RadonIDE.radonAI.enableRadonAI": {
-          "type": "boolean",
+        "RadonIDE.radonAI.enabled": {
+          "type": "string",
+          "enum": ["enabled", "disabled", "default"],
+          "default": "default",
           "scope": "window",
-          "default": true,
-          "description": "Enables the Radon AI functionality in the extension. Note: A full restart of the IDE is required for this setting to take effect."
+          "description": "Enables the Radon AI functionality in the extension."
```
latekvo, line 416: >Note: A full restart of the IDE is required for this setting to take effect.

This text was removed, is it not the case anymore? I can't see changes that'd suggest the code dynamically reacts to this setting now, but i could've missed something.

packages/vscode-extension/src/utilities/editorType.ts
kmagiera: thanks, that was a leftover from the previous iteration where this was a boolean

packages/vscode-extension/src/ai/mcp/index.ts
kmagiera: I fount this the most practical approach. The functions need to capture disposables array which I didn't want to make global.

packages/vscode-extension/src/ai/mcp/index.ts
kmagiera: IMO it is clearer and for the symmetry of the original method that calls it. I'll add some comments inside the call to explain why we only call a single method whereas the load function does other things too.

packages/vscode-extension/src/ai/mcp/index.ts
kmagiera: we weren't doing that before, also the extension disposal is very unreliable. It is only called effectively if the user starts a new window and closes the previous one

packages/vscode-extension/package.json
kmagiera: yes, I described in in the PR description. Restart isn't needed because we use on configuration change listener now.

packages/vscode-extension/src/ai/mcp/index.ts
kmagiera: no, the `disposables` are also disposed on configuration change. If we disposed also this one we'd stop receiving updates. This is ok

packages/vscode-extension/src/ai/mcp/index.ts
kmagiera: ok, I actually removed it entirely in favor of just calling `fsUnload` in dispose

packages/vscode-extension/src/ai/mcp/index.ts
kmagiera: ok, I added this to the disposable that gets added in the load method, so it will be cleaned up in case dispose is called for the extension

packages/vscode-extension/src/utilities/editorType.ts
kmagiera: this was actually a different code that I just moved from another place. Don't want to change it as part of this PR to avoid unnecessary breakages

packages/vscode-extension/src/ai/mcp/index.ts
latekvo: I think we can just pass `disposables` as an argument to `loadRadonAi`, the same way we're doing with `directLoadRadonAI`. This way we don't have to make them global, and we still avoid the nesting

packages/vscode-extension/src/ai/mcp/index.ts
latekvo: sorry, i missed the `disposeAll(disposables);` in `onDidChangeConfiguration`

packages/vscode-extension/package.json
latekvo: apologies, i somehow missed the `onDidChangeConfiguration` in the diff

PR #1625, opened 2025-10-06
fix: CPU Profiles aren't saved correctly

jwajgelt:
This PR fixes CPU profiles not saving after #1580 
- I (and other team members) could not get the CPU Profiler behaviour described in #1580 to happen, and `vscode-js-debug` doesn't save the profiles as described. 
By reading the code, it turned out the profile file is save to `tmpdir` and the path to that file is sent in the DAP custom event fired when profiling starts. We can use that path to restore the previous behaviour of asking the user where the file should be stored and automatically opening it
- After that, `vscode-js-debug` could not annotate the sources for the profile correctly.
The issue is caused by the `Profiler.Stop` result always referring to `scriptId: '0'`, while the debugger never emitted a `scriptParsed` event for that `scriptId`. This was resolved for now by replacing the `scriptId` field with the last loaded main bundle's id in the `ProxyDelegate`.

### How Has This Been Tested: 
- run an application
- start and stop the CPU profiler
- the file picker dialog should appear
- after saving the profile, the JS function entries in the profile should link to files in the filesystem instead of the Metro server (the profile also includes functions implemented within Hermes's native code as well as Host Functions, which aren't mapped by the source maps and still show `localhost` urls, but fixing that is outside the scope of this PR)

### DIFF
--- a/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
+++ b/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
@@ -4,6 +4,22 @@ import _ from "lodash";
 import { CDPProxyDelegate, ProxyTunnel } from "./CDPProxy";
 import { Logger } from "../Logger";
 
+function isProfilerStopResult(result: object): result is Cdp.Profiler.StopResult {
+  // StopResult has a "profile" property
+  if (!("profile" in result) || typeof result.profile !== "object") {
+    return false;
+  }
+
+  // The "profile" property has a list of "nodes"
+  if (!result.profile || !("nodes" in result.profile)) {
+    return false;
+  }
+
+  // if we're here, then either the result has the required type,
+  // or the result is malformed (which we don't care to handle)
+  return true;
+}
+
 export class RadonCDPProxyDelegate implements CDPProxyDelegate {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -4,6 +4,22 @@ import _ from "lodash";
...
 import { Logger } from "../Logger";
 
+function isProfilerStopResult(result: object): result is Cdp.Profiler.StopResult {
+  // StopResult has a "profile" property
+  if (!("profile" in result) || typeof result.profile !== "object") {
+    return false;
+  }
+
+  // The "profile" property has a list of "nodes"
+  if (!result.profile || !("nodes" in result.profile)) {
```
filip131311, line 14: why once you check `!result.profile`  z and in the other place `!("profile" in result)` can you just check once  `!result.profile`?

filip131311 (approved):
It works thank you

packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
jwajgelt: No, TS complains.

PR #1626, opened 2025-10-06
feat: Add JS-side logs to AI

latekvo:
Currently, Radon AI does not have access to the JS-side application logs. The "Device" logs (`iOS Simulator Logs`, `Android Emulator Logs` channels), which it does have access to, only capture the native logs produced by the monitored application, but do not capture the JS-side logs (such as those produced by `console.log` calls).

This PR monitors metro traffic for incoming `"client_log"`s, which are the JS-side logs produced by the monitored app. The newly added log output channel has no overlapping content with the other output channels as far as I see.

I did consider other entry points besides `metro.ts`, but the one used in this PR seemed like the simplest to tap into.

#### The app logs are hidden from the user, as the user already has access to them through the debug console.

device logs|app logs (hidden from the user)
---|---
image|image

#### Complete output of `view_application_logs` (the last section is new, multiple other sections are omitted)
image

### How Has This Been Tested: 

- Run `view_application_logs` by explicitly asking AI to call it.

### DIFF
--- a/packages/vscode-extension/src/common/OutputChannel.ts
+++ b/packages/vscode-extension/src/common/OutputChannel.ts
@@ -4,7 +4,7 @@ export enum Output {
   BuildAndroid = "Radon IDE (Android build)",
   AndroidDevice = "Radon IDE (Android Emulator Logs)",
   IosDevice = "Radon IDE (iOS Simulator Logs)",
-  JSApplication = "Radon IDE (JS App Logs)",
+  JSApplication = "Radon IDE (JS Application Logs)",
   MaestroAndroid = "Radon IDE (Android Maestro tests)",
   MaestroIos = "Radon IDE (iOS Maestro tests)",
   PackageManager = "Radon IDE (JS Package Manager)",


packages/vscode-extension/src/ai/mcp/toolExecutors.ts
```diff
@@ -77,7 +79,12 @@ export async function readLogsToolExec(): Promise<ToolResponse> {
   }
 
   if (!deviceLogs.isEmpty()) {
-    const rawLogs = ["=== APPLICATION LOGS ===\n\n", ...deviceLogs.readAll()];
+    const rawLogs = ["=== NATIVE-SIDE APP LOGS ===\n\n", ...deviceLogs.readAll()];
```
filip131311, line 144: is it a context that is ment to be understood by the model "NATIVE-SIDE APP LOGS"?

packages/vscode-extension/src/common/OutputChannel.ts
```diff
@@ -4,6 +4,7 @@ export enum Output {
   BuildAndroid = "Radon IDE (Android build)",
   AndroidDevice = "Radon IDE (Android Emulator Logs)",
   IosDevice = "Radon IDE (iOS Simulator Logs)",
+  Application = "Radon IDE (App Logs)",
```
filip131311: It would be good to name it in a way that contains information that those are js only logs like  `JSApplication`

packages/vscode-extension/src/project/metro.ts
```diff
@@ -511,6 +509,12 @@ class SubprocessMetroSession extends Metro implements Disposable {
         Logger.info(log);
         this.bundlerReady.resolve();
         break;
+      case "client_log":
+        const content = event.data[0];
+        if (content !== "__RNIDE_INTERNAL") {
+          this.appOutputChannel.appendLine(stripAnsi(content));
```
filip131311, line 515: If you looked into geting those logs form another entry (preferably the debugger) it would be good to do it like that.
The current method works but meta does plan to remove it (and already did once, but the state of devtools was not ready for the transition)

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
filip131311: shouldn't we attach more detailed explanation?

kmagiera (dismissed):
Have a few high level comments, haven't dived into the code too much:
1) would prefer that we align the naming with what chrome devtools MCP now has: https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/tool-reference.md#list_console_messages – specifically in this case those are called "console messages" not JS-App logs. In particular these are not 'logs' really as the console also displays warning and errors. Same goes for the "native-app logs" which
2) we should test (this in particular should be part of the test plan) scenario where app throws an error (red box). The LLM should be able to access the details of the redbox using this mechanims (AFAIK the errors are also posted into the console, just not sure if all the details from the redbox are present there)
3) How do we handle case where there are multiple sessions running? I.e. you have one Android and iOS device in the IDE – this scenario should also be tested and included in the test plan

packages/vscode-extension/src/ai/mcp/toolExecutors.ts
latekvo: I tested this, the model is completely aware that `Native-side` logs refer to these produced by `Obj-C`, `Kotlin` or `C++` parts of the app, and that the `JS-side` logs are these produced by `console.log` et cetera. 

Generally we only want to add descriptions to the parts that aren't oblivious to the model, as descriptions take up attention of the model, and have a chance of really biasing the answers it makes in some edge-cases.

packages/vscode-extension/src/common/OutputChannel.ts
latekvo: `JSApplication` sounds good, couldn't come up with anything better. Changed in [d7b84aa](https://github.com/software-mansion/radon-ide/pull/1626/commits/d7b84aa76d9cdf36bf1263e31566f15c2cb2912a)

kmagiera:
The fragment of the stack trace you posted has a link to generated source which isn't typically useful. The stack traces should be decoded into original files + line numbers. We should likely handle this scenario as the redbox that shows up in the app symbolicates the stacktrace.

My guess is that the stack trace that shows up in the cosole is just an instance of js Error posted to console and it just gets formatted into a string.

packages/vscode-extension/src/project/metro.ts
latekvo: @filip131311 I couldn't really find another spot (looked into `src/debugging` too) where we're receiving these JS-side logs (app `console.log`s). Are you sure we're capturing this anywhere?

latekvo:
@kmagiera I've merged the latest `main` branch into this PR and the stack traces are now replaced with small, symbolicated snippets:

```tsx
[Error: TransformError SyntaxError: /Users/ignacylatka/dev/react-native-shine/example/src/App.tsx: Expected corresponding JSX closing tag for <Viaew>. (77:4)

  75 |         touchPosition={touchPosition}
  76 |       />
> 77 |     </View>
     |     ^
  78 |   );
  79 | }
  80 |]
```

I tested this behaviour on 4 different apps with different versions of RN, metro, etc., and this behaviour occurred on all of them. While I couldn't trace the exact PR that made this change, I think this update removes the need for symbolication which you requested.

packages/vscode-extension/src/common/OutputChannel.ts
```diff
@@ -4,6 +4,7 @@ export enum Output {
   BuildAndroid = "Radon IDE (Android build)",
   AndroidDevice = "Radon IDE (Android Emulator Logs)",
   IosDevice = "Radon IDE (iOS Simulator Logs)",
+  JSApplication = "Radon IDE (JS App Logs)",
```
jwajgelt:
```suggestion
  JSApplication = "Radon IDE (JS Application Logs)",
```
or even

```suggestion
  JSApplication = "Radon IDE (Application Logs)",
```

packages/vscode-extension/src/project/metro.ts
jwajgelt: In `RadonCDPProxyDelegate` we handle the console logs in the `handleConsoleAPICalled` method.
You _could_ extract the arguments passed to `console.log` from the CDP command params, and add them to the event that we fire there, but I'm not 100% sure the juice is worth the squeeze. I'd say the current approach is ~fine.

latekvo: Right, I've just looked into capturing these events from `RadonCDPProxyDelegate`, and I agree it's not a simple change. It would be nice to merge this PR before release, so for now I agree with you it's better to leave it as it is.

packages/vscode-extension/src/common/OutputChannel.ts
latekvo: Changed to `"JS Application Logs"` in [9dbbbf9](https://github.com/software-mansion/radon-ide/pull/1626/commits/9dbbbf92e11bf2ba41510648eb65ca1a510fb1c4)

PR #1627, opened 2025-10-06
fix: unhandled exeptions

filip131311:
After recent changes to the boot up system we had a lot of unhandled exceptions flying on loads and reloads this PR solves the following: 
1) `devdools` initilization failed - we had a los promise that was unhandled 
2) sourceMap parseing failed - turns out source maps can come to us without the sources filed causing us to log unnecessary error.
3) wait for an app to load canceled - because of how the application is initialized we need to set up the `appReadyPromise` before performing actions like `reloadWithDebugger` those actions can fail tho before we would normally await `appReadyPromise` making it reject outside of the try catch scope on the next reload when it was cancelled. 

### How Has This Been Tested: 

- run expo 54 test app 
- perform reloads bundle errors and reloads during bundelle errors 

internal

### DIFF
--- a/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
+++ b/packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
@@ -250,6 +250,9 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
 
     try {
       const sourceMapData = await this.getSourceMapData(sourceMapURL);
+      if (sourceMapData.sources === undefined) {
+        return command;
+      }
       const isMainBundle = sourceMapData.sources.some((source: string) =>
         source.includes("__prelude__")
       );
--- a/packages/vscode-extension/src/project/applicationSession.ts
+++ b/packages/vscode-extension/src/project/applicationSession.ts
@@ -57,16 +57,24 @@ interface LaunchApplicationSessionDeps {
   devtoolsPort?: number;
 }
 
-function waitForAppReady(inspectorBridge: RadonInspectorBridge, cancelToken?: CancelToken) {
+function waitForAppReady(
+  inspectorBridge: RadonInspectorBridge,
+  cancelToken?: CancelToken,
+  message?: string
+) {
   // set up `appReady` promise
   const { promise, resolve, reject } = Promise.withResolvers<void>();
   cancelToken?.onCancel(() => {
-    reject(new CancelError("Cancelled while waiting for the app to be ready."));
+    reject(new CancelError("Cancelled while waiting for the app to be ready." + message));
   });
   const appReadyListener = inspectorBridge.onEvent("appReady", resolve);
-  promise.finally(() => {
-    appReadyListener.dispose();
-  });
+  promise
+    .finally(() => {
+      appReadyListener.dispose();
+    })
+    .catch(() => {
+      // we ignore cancellation rejections as this is another surfaces for it to bubble up
+    });
   return promise;
 }
 
@@ -147,9 +155,8 @@ export class ApplicationSession implements Disposable {
         await cancelToken.adapt(Promise.race([activatePromise, bundleErrorPromise]));
       }
 
-      const hasBundleError = stateManager.getState().bundleError !== null;
-      if (!hasBundleError && getIsActive()) {
-        await cancelToken.adapt(appReadyPromise);
+      if (getIsActive()) {
+        await cancelToken.adapt(Promise.race([appReadyPromise, bundleErrorPromise]));
       }
 
       return session;
--- a/packages/vscode-extension/src/project/applicationSession.ts
+++ b/packages/vscode-extension/src/project/applicationSession.ts
@@ -57,15 +57,11 @@ interface LaunchApplicationSessionDeps {
   devtoolsPort?: number;
 }
 
-function waitForAppReady(
-  inspectorBridge: RadonInspectorBridge,
-  cancelToken?: CancelToken,
-  message?: string
-) {
+function waitForAppReady(inspectorBridge: RadonInspectorBridge, cancelToken?: CancelToken) {
   // set up `appReady` promise
   const { promise, resolve, reject } = Promise.withResolvers<void>();
   cancelToken?.onCancel(() => {
-    reject(new CancelError("Cancelled while waiting for the app to be ready." + message));
+    reject(new CancelError("Cancelled while waiting for the app to be ready."));
   });
   const appReadyListener = inspectorBridge.onEvent("appReady", resolve);
   promise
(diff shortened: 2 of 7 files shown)


packages/vscode-extension/src/debugging/RadonCDPProxyDelegate.ts
```diff
@@ -250,9 +250,9 @@ export class RadonCDPProxyDelegate implements CDPProxyDelegate {
 
     try {
       const sourceMapData = await this.getSourceMapData(sourceMapURL);
-      const isMainBundle = sourceMapData.sources.some((source: string) =>
-        source.includes("__prelude__")
-      );
+      const isMainBundle =
+        sourceMapData.sources &&
+        sourceMapData.sources.some((source: string) => source.includes("__prelude__"));
```
jwajgelt:
```suggestion
      const isMainBundle =
        sourceMapData.sources !== undefined ?
        sourceMapData.sources.some((source: string) => source.includes("__prelude__")) : false;
```
or

```suggestion
      const isMainBundle =
        sourceMapData.sources?.some((source: string) => source.includes("__prelude__")) ?? false;
```
or even

```suggestion
      if (sourceMapData.sources === undefined) {return command;}
      const isMainBundle =
        sourceMapData.sources.some((source: string) => source.includes("__prelude__"));
```
since we wouldn't fire the event before, and maybe we don't want to in this case.

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -57,19 +57,6 @@ interface LaunchApplicationSessionDeps {
...
-  const { promise, resolve, reject } = Promise.withResolvers<void>();
-  cancelToken?.onCancel(() => {
-    reject(new CancelError("Cancelled while waiting for the app to be ready."));
-  });
-  const appReadyListener = inspectorBridge.onEvent("appReady", resolve);
-  promise.finally(() => {
-    appReadyListener.dispose();
-  });
-  return promise;
-}
```
jwajgelt, line 71: You should be able to keep this helper function by doing
```
finally {
  appReadyPromise.catch(noop)
  /* ... */
}
```
instead of 
```
finally {
  resolveAppReady();
  /* ... */
}
```
And actually, if you just want to suppress the "Uncaught promise rejection", I think you could just do `promise.catch(noop)` in this function and return `promise` (without `.catch`) -- then there's some rejection handler, but we still get the error outside of this function.

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -219,16 +233,20 @@ export class CDPDevtoolsServer extends DevtoolsServer implements Disposable {
   private maybeInitializeConnection() {
     if (this.connection) {
       // NOTE: a single `DebugSession` only supports a single devtools connection at a time
-      return;
+      return Promise.resolve();
```
jwajgelt:
```suggestion
      return Promise.reject(/* ... */);
```
to avoid logging the `"[Devtools] Devtools connection initialized..."` message in this case.

jwajgelt (changes requested):

packages/vscode-extension/src/project/applicationSession.ts
filip131311:
```
finally {
  appReadyPromise.catch(noop)
  /* ... */
}
```

this was my original solution but it wouldn necessery solve the problem of hanging promise just suppers it firing I wanted to explicite say. "If we are past this block this promise is resolved/inrelevant.

jwajgelt: Then you could pass a fresh cancel token to the function, and cancel it in the `finally`. It sucks that you have to remember to cancel it, but right now, you also have to remember to resolve the promise, so it's not really worse in that regard.

I don't like the way the code looks right now.
The non-trivial code for awaiting "app ready" is repeated in two places. This _will_ cause issues as soon as we try to touch it, things will be missed in one of these places.

packages/vscode-extension/src/project/devtools.ts
```diff
@@ -219,16 +233,20 @@ export class CDPDevtoolsServer extends DevtoolsServer implements Disposable {
   private maybeInitializeConnection() {
     if (this.connection) {
       // NOTE: a single `DebugSession` only supports a single devtools connection at a time
-      return;
+      return Promise.reject("[Devtools] Connection already established.");
```
jwajgelt:
```suggestion
      return Promise.reject(new Error("[Devtools] Connection already established."));
```
Let's stick to throwing / rejecting with `Error`s.
In many places we _assume_ the caught thing is an instance of `Error`, and while right now this shouldn't be an issue, after some refactors this rejection may be handled someplace else in which case we'll start getting type errors.

packages/vscode-extension/src/project/applicationSession.ts
filip131311: Okay so the escaped rejection was actuallythis one:
promise.finally(() => {
    appReadyListener.dispose();
  });

and i just missed it for some reason. 

Adding catch to it is the least invasive change so I will do it tho I don't like it that the promise is possibly hanging

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -57,16 +57,24 @@ interface LaunchApplicationSessionDeps {
   devtoolsPort?: number;
 }
 
-function waitForAppReady(inspectorBridge: RadonInspectorBridge, cancelToken?: CancelToken) {
+function waitForAppReady(
+  inspectorBridge: RadonInspectorBridge,
+  cancelToken?: CancelToken,
+  message?: string
```
jwajgelt: This doesn't seem to be passed into this function anywhere?

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -57,16 +57,24 @@ interface LaunchApplicationSessionDeps {
...
-  promise.finally(() => {
-    appReadyListener.dispose();
-  });
+  promise
+    .finally(() => {
+      appReadyListener.dispose();
+    })
+    .catch(() => {
+      // we ignore cancellation rejections as this is another surfaces for it to bubble up
+    });
```
jwajgelt, line 73: Having `finally` before `catch` is trippy to me, but it's ultimately fine.

jwajgelt (approved):
Thank you for your service.

packages/vscode-extension/src/project/applicationSession.ts
jwajgelt: Yeah, it leaks the listener on the InspectorBridge.
Since this is called each time you trigger "Reload JS", this can lead to having a bunch of these hanging listeners if you keep reloading while in the `bundleError` state.
This is assuming you go out of your way to do "Reload JS", since auto reload would restart the app, disposing the `InspectorBridge` and the listener with it -- so the risk of it actually causing problems is limited.

packages/vscode-extension/src/project/applicationSession.ts
filip131311: 😈

PR #1632, opened 2025-10-07
fix: network inspector - android header case convention problems

pFornagiel:
### Description

This PR fixes problems with Network Inspector elements dependent on header fields (i.e. syntax highlighting not working properly on android).

As it turns out, due to XHRInterceptor implementation differences, headers on IOS are written capitalised (`Content-Type`) and on android in lower case (`content-type`). This caused problems with Syntax Highlighting on android, no other issues have been found.

### How Has This Been Tested: 

- **open network inspector on android and iOS, check whether every functionality works as expected (especially syntax highlighting)** 

Not applicable.

### DIFF
--- a/packages/vscode-extension/src/network/types/network.ts
+++ b/packages/vscode-extension/src/network/types/network.ts
@@ -39,3 +39,8 @@ export interface TimelineEvent {
   durationMs?: number;
   ttfb?: number;
 }
+
+export enum ContentTypeHeader {
+  IOS = "Content-Type",
+  ANDROID = "content-type",
+}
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/network/types/network.ts
```diff
@@ -39,3 +39,8 @@ export interface TimelineEvent {
   durationMs?: number;
   ttfb?: number;
 }
+
+export enum ContentTypeHeader {
```
filip131311, line 46: you left a comment in `lib` that this is duplicated and I trust this is necessary, but please left a comment here as well to avoid missalignment

filip131311 (approved):
left one comment about comments

packages/vscode-extension/src/network/types/network.ts
pFornagiel: Added comment in acd4c2c9, thanks for pointing that out

PR #1633, opened 2025-10-07
fix: network inspector - filters being case sensitive

pFornagiel:
### Description

This PR fixes problem with network log filters being case sensitive, which is unwanted behaviour.

### How Has This Been Tested: 

- **apply filters, see if they are not case sensitive as expected** 

Not applicable.

### DIFF
--- a/packages/vscode-extension/src/network/providers/NetworkFilterProvider.tsx
+++ b/packages/vscode-extension/src/network/providers/NetworkFilterProvider.tsx
@@ -113,7 +113,7 @@ export function NetworkFilterProvider({ children }: PropsWithChildren) {
 
     // AND between columns, OR within column values
     return NETWORK_LOG_COLUMNS.every((columnName) => {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/network/providers/NetworkFilterProvider.tsx
```diff
@@ -151,6 +153,7 @@ export function NetworkFilterProvider({ children }: PropsWithChildren) {
       NETWORK_LOG_COLUMNS.some((column) =>
         getNetworkLogValue(log, column).toLowerCase().includes(filterTextValue.toLowerCase())
       );
+    console.log(globalMatches);
```
filip131311: seems a debugging left over

pFornagiel: Removed in f493aebf

PR #1635, opened 2025-10-07
fix: radon connect 

filip131311:
This PR fixes a regression introduced in #1550 which removed an option to pass `watchFolders` to the metro class directly. Normally we really on metro process handing as this information during the set up process, but when we use radon connect we do not have an option of capturing metro config like in a normal case, so we need to add the  `watchFolders` information by hand. 

### How Has This Been Tested: 

- run expo 54 test app from terminal 
- use radon connect 

internal

### DIFF
--- a/packages/vscode-extension/src/project/metro.ts
+++ b/packages/vscode-extension/src/project/metro.ts
@@ -258,7 +258,7 @@ async function launchMetro({
 
 export class Metro implements MetroSession, Disposable {
   protected _expoPreludeLineCount = 0;
-  protected _watchFolders: string[] | undefined = undefined;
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/project/metro.ts
```diff
@@ -274,8 +274,11 @@ export class Metro implements MetroSession, Disposable {
 
   constructor(
     public readonly port: number,
-    protected readonly appRoot: string
+    protected readonly appRoot: string,
+    watchFolders: string[] | undefined = undefined
```
jwajgelt, line 278:
```suggestion
    watchFolders?: string[]
```

PR #1637, opened 2025-10-08
fix: Radon crashes when opening project before installing packages

jwajgelt:
Fixes Radon crashing if opened before installing application's dependencies.
The crash was caused by an unguarded `require` of `react-native` package from application's directory failing before the packages are installed.
The require was used to check RN version used by the project to verify fusebox support.
This PR changes the behavior to read the version from app's dependencies instead, and fall back (rather than crashing) when the require fails.

### How Has This Been Tested: 
- open Radon before installing the app's `node_modules`
- the app should open as normal


filip131311 (approved):
thank you

PR #1638, opened 2025-10-08
feat: make `usePrebuild` option more important then automatic heuristic 

filip131311:
In our recent changes to the prebuild system, we added an automatic detection to check if the project uses `dev-client` and force a prebuild if it does. This however made us ignore the old `usePrebuild` option in such a scenarion not allowing a user to opt in into a manual prebuild by default. 

### How Has This Been Tested: 

- run `social-app` 
- set the `usePrebuild` option to false
- after it works reload the whole of radon ide

not different from the 1.11.2

### DIFF
--- a/packages/vscode-extension/src/builders/BuildManager.ts
+++ b/packages/vscode-extension/src/builders/BuildManager.ts
@@ -112,6 +112,7 @@ export function createBuildConfig(
           configuration: ios?.configuration,
           fingerprintCommand,
           runtimeId: runtime.identifier,
+          usePrebuild,
         };
       } else {
         return {
@@ -122,6 +123,7 @@ export function createBuildConfig(
           productFlavor: android?.productFlavor,
           buildType: android?.buildType,
           fingerprintCommand,
+          usePrebuild,
         };
       }
     }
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/builders/BuildManager.ts
```diff
@@ -112,6 +112,7 @@ export function createBuildConfig(
           configuration: ios?.configuration,
           fingerprintCommand,
           runtimeId: runtime.identifier,
+          usePrebuild,
```
jwajgelt: I'd prefer if this property was non-optional, and instead in this function you had

```suggestion
          usePrebuild: usePrebuild ?? true,
```
for the DevClient configs, and

```suggestion
          usePrebuild: usePrebuild ?? false,
```
for the local ones.

packages/vscode-extension/src/dependency/ApplicationDependencyManager.ts
```diff
@@ -110,7 +110,10 @@ export class ApplicationDependencyManager implements Disposable {
 
   public async ensureDependenciesForBuild(buildConfig: BuildConfig, buildOptions: BuildOptions) {
     if (buildConfig.type === BuildType.Local || buildConfig.type === BuildType.DevClient) {
-      if (buildConfig.type === BuildType.DevClient || buildConfig.usePrebuild) {
+      if (
+        (buildConfig.type === BuildType.DevClient && buildConfig.usePrebuild !== false) ||
+        buildConfig.usePrebuild
```
jwajgelt: Then you can also simplify this to just 

```suggestion
        buildConfig.usePrebuild
```

jwajgelt (approved):

PR #1643, opened 2025-10-08
docs: Note Radon AI is disabled in Cursor by defualt

latekvo:
In the latest release (`1.12.0` via #1624) Radon AI has been disabled in Cursor by default.
This PR updates the documentation to reflect that.

This PR should be merged immediately after approval, since the related changes have already been released.

### DIFF
--- a/packages/docs/docs/features/radon-ai.md
+++ b/packages/docs/docs/features/radon-ai.md
@@ -115,7 +115,7 @@ To disable Radon AI assistant navigate to the editor settings.
 
 You can type "Preferences: Open User Settings" in the command palette (Ctrl+Shift+P or Cmd+Shift+P).
 
-Within editor settings, type "Radon AI: Enabled" into the search bar, press the first result and select `Disabled`.
+Within editor settings, type "Radon AI: Enabled", press the first result and select `Disabled`.
 
 ## Privacy
 


packages/docs/docs/features/radon-ai.md
```diff
@@ -58,6 +58,11 @@ There you can adjust which tools are enabled or disable the Radon AI MCP server
 
 ## Usage in Cursor
 
+:::tip
+**Radon AI is disabled in Cursor by default.**  
+To enable it, go to the `Settings -> VS Code Settings` tab (or press `Cmd + ,`), type `Radon AI` into the search bar, press the first result and select `Enabled`.
+:::
```
latekvo, line 64: Do we want to use `:::tip` here? This notation is only used in one other place in our docs, so it's not very conventional, but I feel like highlighting this fact is important. (c.c. @kacperkapusciak )

packages/docs/docs/features/radon-ai.md
```diff
@@ -110,7 +115,7 @@ To disable Radon AI assistant navigate to the editor settings.
 
 You can type "Preferences: Open User Settings" in the command palette (Ctrl+Shift+P or Cmd+Shift+P).
 
-Within editor settings, type "Radon IDE: Enable Radon AI" and uncheck the checkbox.
+Within editor settings, type "Radon AI" into the search bar, press the first result and select `Disabled`.
```
latekvo: `Radon IDE: Enable Radon AI` setting has been modified, the checkbox has been replaced with a dropdown list. Thus this change was made.

packages/docs/docs/features/radon-ai.md
kacperkapusciak: Yes, it's okay.

Normally I'm not a fan of these as it oftentimes leads to overuse (if everything is important then nothing really is) but here I think it's fitting ✅

PR #1644, opened 2025-10-08
Update custom build path parser to fix issues with dot directory parsing

kmagiera:
This PR fixes the logic behind the custom build file path parsing.

The previous regex that we used was falling short when there was a dot (".") in the path followed by an empty space, for example: `/Users/krzys/.builds/My App.app`

This PR updates the parsing regex to be more resilient and also consider the cases where quotes are put around the paths (which is often the case when paths include spaces).

On top of that the package contributes a number of unit tests for the new parser. 

### How Has This Been Tested: 
1. Run tests

### DIFF
--- a/packages/vscode-extension/src/builders/customBuild.test.ts
+++ b/packages/vscode-extension/src/builders/customBuild.test.ts
@@ -5,112 +5,101 @@ import { extractFilePath } from "./customBuild";
 
 describe("customBuild", () => {
   describe("extractFilePath", () => {
-    it("should extract file paths from various formats and edge cases", () => {
-      const testCases = [
-        {
-          input: `Path: "/Users/me/My App.app"`,
-          expected: "/Users/me/My App.app",
-          description: "quoted path with spaces",
-        },
-        {
-          input: `Installing 'C:\\Program Files\\AppFolder\\App.apk'...`,
-          expected: "C:\\Program Files\\AppFolder\\App.apk",
-          description: "Windows path with single quotes",
-        },
-        {
-          input: `Archive created at ./builds/output.tar.gz`,
-          expected: "./builds/output.tar.gz",
-          description: "relative path with tar.gz",
-        },
-        {
-          input: "/Users/johndoe/.builds/builds/app-3aabf66a76gggg/My App.app",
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/builders/customBuild.test.ts
```diff
@@ -0,0 +1,116 @@
+import assert from "assert";
+import { describe, it } from "mocha";
+
+import { extractFilePath } from "./customBuild";
+
+describe("customBuild", () => {
+  describe("extractFilePath", () => {
+    it("should extract file paths from various formats and edge cases", () => {
+      const testCases = [
```
jwajgelt: I'd rather each test case was a separate mocha test

```suggestion
      const testCases = [/* ... */]
    testCases.forEach(({input, expected, description}) => it(`handles ${description}`, () => {

```

packages/vscode-extension/src/builders/customBuild.ts
```diff
@@ -11,8 +11,18 @@ import { DevicePlatform } from "../common/State";
 
 type Env = Record<string, string> | undefined;
 
-// Extracts all paths from the last line, both Unix and Windows format
-const BUILD_PATH_REGEX = /(\/.*?\.\S*)|([a-zA-Z]:\\.*?\.\S*)/g;
+export function extractFilePath(line: string): string | null {
+  // Fixed regex pattern to properly capture only the path
+  const regex =
```
jwajgelt: I'd prefer if we did something like:
```suggestion
  const quotedStartsWithDot = "["']([^"']*\.(?:tar\.gz|app|apk))["']"
  const startsWithEscapedDot = "([a-zA-Z]:\\[^"']*\.(?:tar\.gz|app|apk))"
  const startsWithSlash = "(\/[^"']*\.(?:tar\.gz|app|apk))"
  const startsWithDotSlash = "(\.\/[^"']*\.(?:tar\.gz|app|apk))"
  const regex =
    new RegExp(`${quotedStartsWithDot}|${startsWithEscapedDot}|${startsWithSlash}|${startsWithDotSlash}`, "i");
```
or even splitting into 4 separate `RegExp` objects and checking each separately (though I understand that degrades performance -- not sure how severely though).
It would help with readability to see which alternative does what, instead of seeing one giant regex with no comments.

PR #1651, opened 2025-10-10
feat: network inspector - ignore internal React Native messages

pFornagiel:
### Description

This PR changes the network inspector behaviour, so that the the internal React Native network messages are ignored and not shown to the user. The filtering is done on the extension-site in the `BaseNetworkInspector` abstract class, at the point where the message is broadcast to the webviews.

### How Has This Been Tested: 

- *test whether the internal messages (such as the ones sent by opening the `LogBox`, for example caused by unexpected error in the React Native app) are being properly ignored in both implementations of network inspectors*

Not applicable.

### DIFF
--- a/packages/vscode-extension/src/plugins/network/strategies/BaseNetworkInspector.ts
+++ b/packages/vscode-extension/src/plugins/network/strategies/BaseNetworkInspector.ts
@@ -63,7 +63,7 @@ export abstract class BaseNetworkInspector implements NetworkInspector {
 
   protected broadcastMessage(message: Parameters<BroadcastListener>[0]): void {
     if (this.isInternalRequest(message)) {
-      Logger.info("Internal React Native network event, ignoring");
+      Logger.info(`Http request to metro filtered out: ${message.payload.params?.request?.url}`);
       return;
     }
 


packages/vscode-extension/src/plugins/network/strategies/BaseNetworkInspector.ts
```diff
@@ -40,7 +42,30 @@ export abstract class BaseNetworkInspector implements NetworkInspector {
     });
   }
 
+  private isInternalRequest(message: WebviewMessage, port: number): boolean {
```
jwajgelt: Can't you just use `this.metroPort`?
```suggestion
  private isInternalRequest(message: WebviewMessage): boolean {
```

pFornagiel: Made the methods use `this.metroPort` and added an android `10.0.2.2` chceck as well. Thank you for bringing it up, I did not notice that this could be the case.

Additionally, changed regex expressions to string comparison, as there is no need for them in such a plain comparison.

Done in 4ffe85be

packages/vscode-extension/src/plugins/network/strategies/BaseNetworkInspector.ts
```diff
@@ -40,7 +42,31 @@ export abstract class BaseNetworkInspector implements NetworkInspector {
...
+
+      return (isLocalhost || isAndroidHost) && isPortMatch;
+    } catch (error) {
+      return false;
+    }
+  }
+
   protected broadcastMessage(message: Parameters<BroadcastListener>[0]): void {
+    if (this.isInternalRequest(message)) {
+      Logger.info("Internal React Native network event, ignoring");
```
jwajgelt: I'm not sure there's any point in logging it.
```suggestion
```
If we expect the requests to be filtered _incorrectly_, and want to see them in the logs for troubleshooting, we should add more info to them -- at least the URL -- to be able to see what request was filtered out. Otherwise, we won't be able to tell from the logs if we got here due to an error.
```suggestion
      Logger.info(`Http request to metro filtered out: ${message.payload.params.request.url}`);
```

pFornagiel: Applied suggestion in 07730f99

PR #1652, opened 2025-10-10
fix: network inspector - optimise syntax highlighting on the frontend

pFornagiel:
### Description

This PR optimises syntax highlighting on the frontend in order to reduce freezes when using network inspector

Optimisations:
- introduced `isActive` predicate to `HighlightedCodeBlock`, to perform syntax highlighting only when the proper tab is active, which was not the case so far - even if the user was looking at the `Headers` tab, the code highlighting was still done on the main thread
- introduced `HighlighterContext`, whose main purpose is to keep cache of highlighted code at the top level of the application. Based on hash of code (dependent on requestId, code itself, theme and language) we keep the HTML of the highlighted code in memory.
- reduced the `MAX_HIGHLIGHT_LENGTH` from `100_000` to `65_000` based on tests. To be discussed.
- removed the `react-shiki` package in favour of [`shiki`](https://github.com/shikijs/shiki), because caching with `react-shiki` was not possible
- 
### How Has This Been Tested: 

- run Radon IDE extension panel and open network inspector
- *fetch different large files, which can be highlighted, see if caching and colouring works as expected without causing freezes*

Not applicable.

### DIFF
--- a/packages/vscode-extension/src/network/providers/HighlighterProvider.tsx
+++ b/packages/vscode-extension/src/network/providers/HighlighterProvider.tsx
@@ -63,17 +63,14 @@ export default function HighlighterProvider({ children }: { children: ReactNode 
     const cache = cacheRef.current;
     const entries = Array.from(cache.entries()).sort(([, a], [, b]) => a.accessTime - b.accessTime);
 
-    let freedSpace = 0;
     for (const [key, entry] of entries) {
-      if (cacheSizeRef.current - freedSpace + requiredSpace <= MAX_CACHE_SIZE) {
+      if (cacheSizeRef.current + requiredSpace <= MAX_CACHE_SIZE) {
         break;
       }
 
       cache.delete(key);
-      freedSpace += entry.size;
+      cacheSizeRef.current -= entry.size;
     }
-
-    cacheSizeRef.current -= freedSpace;
   };
 
   const generateCacheKey = async (


packages/vscode-extension/src/network/providers/HighlighterProvider.tsx
```diff
@@ -0,0 +1,148 @@
...
+    hash = (hash << 5) - hash + char;
+    hash = hash & hash; // Convert to 32-bit integer
+  }
+  return hash.toString(36);
+}
+
+const HighlighterContext = createContext<HighlighterContextValue | null>(null);
+
+export default function HighlighterProvider({ children }: { children: ReactNode }) {
+  const cacheRef = useRef<Map<string, CacheEntry>>(new Map());
```
jwajgelt, line 55: We never remove entries from the map, right?
I don't think it's a good idea to store them forever, I feel like it could quickly add up after running the app for a bit without restarting.
- I'd throw in some sort of LRU cache here, to make sure we don't store old entries unnecessarily
- the highlighter should be fast enough for small payloads -- you've looked at it longer than me, so you probably have a better idea when problems start happening, but I would guess text under a couple kb should be quick to highlight. Maybe we could only use the cache for large payloads, and always recalculate for small ones, so that the cache doesn't fill up with small entries which there may be a lot of.

packages/vscode-extension/src/network/providers/HighlighterProvider.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+  const generateCacheKey = (
+    language: string,
+    theme: ThemeData | undefined,
+    content: string,
+    requestId: string | number
+  ) => {
+    const contentHash = getHash(content);
+    const themeName = theme?.name || "none";
+    return `${language}:${themeName}:${requestId}:${contentHash}`;
```
jwajgelt:
```suggestion
    const contentHash = getHash(content);
    const themeName = theme?.name || "none";
    return `${language}:${themeName}:${contentHash}`;
```
do we need `requestId` here? If we have two requests with the same payload, why not reuse the result?

packages/vscode-extension/src/network/providers/HighlighterProvider.tsx
pFornagiel: Introduced LRU cache and conditional highlighting based on the content length in [5f55692](https://github.com/software-mansion/radon-ide/pull/1652/commits/5f556929f524119e70ca2d2a0bf9a830c3188613) and [1d6fcfa](https://github.com/software-mansion/radon-ide/pull/1652/commits/1d6fcfa6d4389e6d465335000b8f4d0c4aba3ff2).

After exceeding defined cache max size, the oldest entires (based on access time) are now evicted from the map.

The cache is limited by the total stored sum of characters in the html output of the highlighter, while the decision to highlight code fragment is based on the length of the code itself. The values were tested empirically, up to a discussion.

```ts
/**
 * Cache size limit in characters (after the html colouring is applied)
 */
const MAX_CACHE_SIZE = 5_000_000;
/**
 * Minimum size for a text block to be cached (before the html colouring is applied)
 */
const MINIMUM_SIZE_TO_CACHE = 20_000;
```

packages/vscode-extension/src/network/providers/HighlighterProvider.tsx
pFornagiel: In order to get rid of the `requestId`, the hash function has been rewritten in [5f55692](https://github.com/software-mansion/radon-ide/pull/1652/commits/5f556929f524119e70ca2d2a0bf9a830c3188613) and [1d6fcfa](https://github.com/software-mansion/radon-ide/pull/1652/commits/1d6fcfa6d4389e6d465335000b8f4d0c4aba3ff2).

`requestId` was provided to the hash function due to the fact, that collisions were possible, because of simplicity of hash function implementation. Using the old approach, it was thus important to differentiate between the requests, to not accidentally display incorrect message to the client because of the collision.

Now, instead of using the old method, `SubtleCrypto API` has been used to utilise `SHA-1` hashing, thus making the collisions nearly impossible for given string, eliminating the need to differentiate between requests.

```ts
async function getHash(str: string): Promise<string> {
  const data = encoder.encode(str);
  const hashBuffer = await crypto.subtle.digest("SHA-1", data);
  const hashArray = new Uint8Array(hashBuffer);
  return hashArray.reduce((res, byte) => res + byte.toString(16).padStart(2, "0"), "");
}
```

packages/vscode-extension/src/network/providers/HighlighterProvider.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const highlighterPromiseRef = useRef<Promise<HighlighterCore>>(getHighlighter());
+
+  /**
+   * Evict oldest entries from cache until size is under the limit
+   */
+  const evictOldestEntries = (requiredSpace: number) => {
+    const cache = cacheRef.current;
+    const entries = Array.from(cache.entries()).sort(([, a], [, b]) => a.accessTime - b.accessTime);
+
+    let freedSpace = 0;
```
jwajgelt: You could just subtract `cacheSizeRef.current -= entry.size` inside the loop, it would get you the same result.

pFornagiel: Applied change in 6969398a

PR #1659, opened 2025-10-14
feat: paywall enforced for pro features

filip131311:
This PR introduces paywall enforcement


packages/vscode-extension/src/webview/components/DeviceSelect.tsx
```diff
@@ -120,7 +126,18 @@ function DeviceSelect() {
     if (selectedDevice?.id !== value) {
       const deviceInfo = (devices ?? []).find((d) => d.id === value);
       if (deviceInfo) {
-        project.startOrActivateSessionForDevice(deviceInfo);
+        if (deviceInfo.deviceType === DeviceType.Tablet && hasAccessToProFeatures(licensedStatus)) {
```
jwajgelt: Surely if you have access, you don't want the paywall?
```suggestion
        if (deviceInfo.deviceType === DeviceType.Tablet && !hasAccessToProFeatures(licensedStatus)) {
```

packages/vscode-extension/src/common/License.ts
```diff
@@ -0,0 +1,27 @@
+// #region License Helpers
+
+export function hasAccessToProFeatures(licenseStatus: LicenseStatus) {
+  return (
+    licenseStatus === LicenseStatus.Pro ||
+    licenseStatus === LicenseStatus.Team ||
+    licenseStatus === LicenseStatus.Enterprise
+  );
+}
```
jwajgelt: _I'd prefer_ if this function instead took:

```
export function hasAccessToFeature(licenseStatus: LicenseStatus, feature: Feature) {
```
for some 
```
enum Feature {
  DeviceRotation,
  TabletEmulators,
  /* etc... */
}
```
and then have a map from `LicenseStatus` to `Set<Feature>` that you use to check feature support.

- it makes changing which features are in which tier easier, since you don't have to look for each guard for that feature, you just change one entry in the map.
- it would allow us to easily change where we store the information. I'd imagine at some point we may want to move the feature sets to the customer portal, so that we can make changes without pushing Radon updates. It would also make it possible to allow enterprise admins to block features in their orgs.

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -351,14 +388,25 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
 }
 
 const LocalizationItem = () => {
+  const store$ = useStore();
+  const licenseStatus = use$(store$.license.status);
+  const hasProAccess = hasAccessToProFeatures(licenseStatus);
```
jwajgelt: I'd turn this into a `useProAccess` hook, since this repeats in multiple places.

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -97,12 +100,49 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
...
+      return;
+    }
+
+    try {
+      await project.rotateDevices(direction);
+    } catch (error) {
+      if (error instanceof RestrictedFunctionalityError) {
+        openPaywall();
+      }
+    }
```
jwajgelt: I wonder -- could we put this inside the `project` proxy returned from `useProject`?
Basically, wrap each method on the `project` object in this try/catch, so that when you _call_ the functions, you cannot forget to handle the paywall?

packages/vscode-extension/src/webview/views/PreviewView.tsx
```diff
@@ -170,9 +176,18 @@ function PreviewView() {
   }, []);
 
   function toggleRecording() {
+    if (!hasAccessToProFeatures(licenseStatus)) {
+      openPaywall();
+      return;
+    }
```
jwajgelt: This feels like a common-enough pattern to lift it into a higher-order function. Something like
```
function withPaywallGuard(fn: Function, licenseStatus) {
  return (...args) => {
    if (!hasAccessToProFeatures(licenseStatus)) {
      openPaywall();
      return;
    }
    return fn(...args);
  }
}
```
Or even in hook form:
```
function usePaywalledCallback(fn: Function, dependencies: unknown[]) {
  const store$ = useStore();
  const licenseStatus = use$(store$.license.status);
  return useCallback(withPaywallGuard(fn, license), [license, ...dependencies])
}
```

packages/vscode-extension/src/common/License.ts
```diff
@@ -1,16 +1,70 @@
...
+  feature: Feature
+): FeatureAvailabilityStatus {
+  switch (licenseStatus) {
+    case LicenseStatus.Inactive:
+    case LicenseStatus.Free:
+      return FreeFeatures.has(feature)
+        ? FeatureAvailabilityStatus.Available
+        : FeatureAvailabilityStatus.InsufficientLicense;
+    case LicenseStatus.Pro:
+      return ProFeatures.union(FreeFeatures).has(feature)
```
jwajgelt: I know we're in JS land, and performance is a pipe dream, but creating a set union each time this check is performed (and doing it _twice_ in the `Team` branch and _three times_ in the `Enterprise`) sounds really excessive.
I'm also not 100% sure how I feel about splitting the features between maps -- in my mind, it'd make more sense for `ProFeatures` to simply include all of the `FreeFeatures` as well. I guess it might be useful to check if a feature is `Pro` but not `Free`, but I also feel like that's the case that can take a slower path (of checking both maps and returning `Pro.has && !Free.has`), rather than this one.

packages/vscode-extension/src/common/License.ts
```diff
@@ -25,3 +79,90 @@ export enum LicenseStatus {
...
+  ReactQueryDevTools,
+  RadonConnect,
+  RadonAI,
+  ReduxDevTools,
+  ScreenRecording,
+  ScreenReplay,
+  Screenshot,
+  SendFile,
+  StorybookIntegration,
+  VolumeButtons,
```
jwajgelt: NIT:
This is a change that's easy to do later if we actually need it, but:
```suggestion
  AndroidSmartphoneEmulators = "AndroidSmartphoneEmulators",
  AndroidTabletEmulators = "AndroidTabletEmulators",
  AppSwitcherButton = ...,
```
1. having string values here would make it easier to debug if you come across a `Feature` value in the debugger.
2. we might want to move the `Free`/`Pro`/etc. sets to sim-server, in which case this will make it harder to make a mistake in serializing.

packages/vscode-extension/src/common/License.ts
```diff
@@ -1,16 +1,70 @@
...
+  ) {
+    licenses.push(LicenseStatus.Team);
+  }
+  if (
+    getFeatureAvailabilityStatus(LicenseStatus.Enterprise, feature) ===
+    FeatureAvailabilityStatus.Available
+  ) {
+    licenses.push(LicenseStatus.Enterprise);
+  }
+  return licenses;
```
jwajgelt:
```suggestion
  const hasAccess = (license) => getFeatureAvailabilityStatus(license, feature) === FeatureAvailabilityStatus.Available;
  const licenses = [LicenseStatus.Free, LicenseStatus.Pro, LicenseStatus.TeamLicenseStatus.Enterprise];
  return licenses.filter(hasAccess);
```

packages/vscode-extension/src/project/project.ts
```diff
@@ -59,13 +59,59 @@ import { EnvironmentDependencyManager } from "../dependency/EnvironmentDependenc
...
+ */
+function guardFeatureAccess(
+  feature: Feature,
+  shouldSkipLicenseCheck?: (...args: any[]) => boolean
+) {
+  return function (target: any, propertyKey: string, descriptor: PropertyDescriptor) {
+    const originalMethod = descriptor.value;
+    descriptor.value = function (this: Project, ...args: any[]) {
+      if (shouldSkipLicenseCheck) {
+        const shouldSkip = shouldSkipLicenseCheck(args);
```
jwajgelt: You probably wanted to do
```suggestion
        const shouldSkip = shouldSkipLicenseCheck(...args);
```
Then the callbacks can accept the same arguments as the wrapped function, rather than the array.

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
```diff
@@ -272,7 +268,9 @@ function DeviceSettingsDropdown({ children, disabled }: DeviceSettingsDropdownPr
                     className="dropdown-menu-item"
                     data-testid={`device-settings-set-orientation-${option.label.trim().toLowerCase().replace(/\s+/g, "-")}`}
                     key={index}
-                    disabled={!hasProAccess && option.value !== DeviceRotation.Portrait}
+                    disabled={
+                      !isDeviceRotationAvailable && option.value !== DeviceRotation.Portrait
+                    }
```
jwajgelt, line 279: Should it be disabled? We don't do it with the other options, just with rotation.

packages/vscode-extension/src/webview/hooks/usePaywalledCallback.ts
```diff
@@ -0,0 +1,52 @@
...
+      await fn(...args);
+    } catch (e) {
+      if (e instanceof RestrictedFunctionalityError) {
+        openPaywall();
+        return;
+      }
+      throw e;
+    }
+  };
+}
```
jwajgelt: Add some type safety here
Something like:
```suggestion
function withPaywallGuard<F extends (...args: unknown) => void | Promise<void>>(
  fn: F,
  feature: Feature,
  licenseStatus: LicenseStatus
) {
  const { openPaywall } = usePaywall();

  return async (...args: Parameters<F>): Promise<void> => {
    const featureAvailability = getFeatureAvailabilityStatus(licenseStatus, feature);
    switch (featureAvailability) {
      case FeatureAvailabilityStatus.Available:
        break;
      case FeatureAvailabilityStatus.InsufficientLicense:
        openPaywall();
        return;
    }
    try {
      await fn(...args);
    } catch (e) {
      if (e instanceof RestrictedFunctionalityError) {
        openPaywall();
        return;
      }
      throw e;
    }
  };
}
```

Likewise for `usePaywalledCallback`

packages/vscode-extension/src/extension.ts
```diff
@@ -41,6 +41,20 @@ import { RestrictedFunctionalityError } from "./common/Errors";
 
 const CHAT_ONBOARDING_COMPLETED = "chat_onboarding_completed";
 
+function wrapPaywalledFunction(fn: (...args: any[]) => any, messageOnRestricted: string) {
+  return async (...args: any[]) => {
```
jwajgelt: Types plz.

```suggestion
function wrapPaywalledFunction<F extends (...args: unknown[]) => void>(fn: F, messageOnRestricted: string) {
  return async (...args: Parameters<F>) => {
```

packages/vscode-extension/src/project/project.ts
```diff
@@ -58,11 +59,59 @@ import { EnvironmentDependencyManager } from "../dependency/EnvironmentDependenc
...
```
jwajgelt: It would be great if we could add some types to this.

I think I got it to work like this:

```suggestion
type ProjectMethodParameters<K extends keyof Project> = Project[K] extends (
  ...args: infer P
) => unknown
  ? P
  : never;

/**
 * Checks if the current license status allows access to the specified feature.
 * @argument feature - The feature to check access for.
 * @throws {RestrictedFunctionalityError} if the user does not have access to Pro features
 */
function guardFeatureAccess<K extends keyof Project>(
  feature: Feature,
  shouldSkipLicenseCheck?: (...args: ProjectMethodParameters<K>) => boolean
) {
  return function (target: Project, propertyKey: K, descriptor: PropertyDescriptor) {
    const originalMethod = descriptor.value;
    descriptor.value = function (this: Project, ...args: ProjectMethodParameters<K>) {
      if (shouldSkipLicenseCheck) {
        const shouldSkip = shouldSkipLicenseCheck(...args);

```

packages/vscode-extension/src/common/License.ts
filip131311: I can create a helper object storing this information something like:
```js
const TeamProFreeFeatures = FreeFeatures.union(ProFeatures).union(TeamFeatures);
```

and then use it in the function but i don't want to create sets having duplicate positions, this is very hard to maintain with any change and the probability of someone deleting the position form pro but leaving it in other sets is high.

I guess we could right tests for it as well... 

anyway what do you think?

jwajgelt: > this is very hard to maintain with any change and the probability of someone deleting the position form pro but leaving it in other sets is high.

I had something like this in mind:
```ts
const ProFeatures = new Set([/* ...features... */, ...FreeFeatures])
```
rather than typing them out explicitely.

packages/vscode-extension/src/webview/components/DeviceSettingsDropdown.tsx
filip131311: yes i did that on purpose to allow people that down grade the plan form pro to free to return to portret and have visual indication in the ui that this is whats going on.

packages/vscode-extension/src/common/License.ts
filip131311: fair

packages/vscode-extension/src/webview/hooks/usePaywalledCallback.ts
filip131311: i can't use unknown as an argument because it would clash with the types of callbacks passed to the function but i did improve types

packages/vscode-extension/src/project/project.ts
```diff
@@ -58,11 +59,65 @@ import { EnvironmentDependencyManager } from "../dependency/EnvironmentDependenc
...
+/**
+ * Checks if the current license status allows access to the specified feature.
+ * @argument feature - The feature to check access for.
+ * @throws {RestrictedFunctionalityError} if the user does not have access to the feature.
+ */
+function guardFeatureAccess<K extends keyof Project>(
+  feature: Feature,
+  shouldSkipLicenseCheck?: (...args: ProjectMethodParameters<K>) => boolean
+) {
+  return function (target: Project, propertyKey: string, descriptor: PropertyDescriptor) {
```
jwajgelt:
```suggestion
  return function (target: Project, propertyKey: K, descriptor: PropertyDescriptor) {
```
should work, and let you omit parameter types in `shouldSkipLicenseCheck` callbacks.
Otherwise, I don't think TS would even check that `args` match the parameters of the decorated method...

packages/vscode-extension/src/project/project.ts
```diff
@@ -58,11 +59,65 @@ import { EnvironmentDependencyManager } from "../dependency/EnvironmentDependenc
...
+    const originalMethod = descriptor.value;
```
jwajgelt, line 92:
```suggestion
    const originalMethod = descriptor.value as Project[K];
```
should work, and get you some extra type checking in here.

filip131311: this is not a true cast as methods in originalDescriptor.value take additional argument this

jwajgelt: `this` is not an "additional argument" to the function, it's special typescript syntax that let's you type the `this` binding in the scope of the function.
https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function
so it _should_ work.

PR #1660, opened 2025-10-15
Use owner stacks for inspector functionality instead of overriding react renderer

kmagiera:
Since the removal of JSX transform's `__source` metadata, we were relying on modification to react renderer that'd bring that metadata back to modern versions of React. This approach required additional work with every RN release and was quite brittle (i.e. we couldn't support the cases where users would use a different version of React).

In recent versions the concept of owner stack got introduced which allows us to access the needed location of JSX tags in runtime without the metadata passed from the transform. In some early versions of the owner stack work, we'd only be getting the pointer to the sources where the component is defined and not where it is actually used which wasn't ideal and made us give up on this idea.

We recently tried owner stack based approach and since its "official" release in 19.1 it is stable and works ok for our usecases.

This PR adapts the owner-stack based logic for the inspector functionality for React Native 0.80 and above.

Highlights of this PR:
1) in wrapper, we detect whether `_debugStack` field is available. If it is, we extract the component stack based on it and pass that back as the inspector request callback.
2) since the new debug stack is calculated in runtime, it points to the bundled source. We add additional logic that detects the case when stack points to the bundled file. When this happens, we use the debugger to perform the source map lookup to give us the correct source of the element.
3) the source map lookup uses a custom request `RNIDE_findOriginalPosition` that is implemented by our debug session proxy (both for the vscode-js-debug based implementation and the custom CDP based implementation). Under the hood, it uses `getPreferredUILocation` custom request that is exposed by `vscode-js-debug`. We use a different name to make it easier to follow the custom request call path and be explicit that the request with `RNIDE` is handled within out code.
4) we update the wrapper logic to use `line0Based` and `column0Based` naming for line and column fields to make it clear when we handle different components of the system that use different numbering scheme for line and column numbers. For example, `_debugStack` uses 1-based numbering, but the source maps from `vscode-js-debug` use 0-based numbering.
5) we delete the overwrites in `babel_transform` for the render for versions 0.80 and above. We also comment out all the JSX-transformer fixes in case metro starts for version 0.80+ since we no longer rely on the transform-backed mechanism there
6) in order to skip JSX transformer fixes in specific versions, we add a new environment variable to metro `RADON_IDE_RN_VERSION` that indicates the version of React Native.
6) we delete all the custom renderer code for RN versions that no longer need it

### How Has This Been Tested: 
1) Run this on react native 0.81, 0.80 and expo-54 apps – check the inspector (right click)
2) Run this on older version (i.e. 0.78) and also check the inspector to verify it works there
3) I also checked 0.81 with old architecture to make sure it doesn't impact the debug owner stack capabilities.
4) finally, tested this on RN 82 app with "customJSDebug" option enabled in order to verify the new custom command works via `CDPDebugAdapter` too.

### DIFF
--- a/packages/vscode-extension/src/debugging/CDPDebugAdapter.ts
+++ b/packages/vscode-extension/src/debugging/CDPDebugAdapter.ts
@@ -11,6 +11,7 @@ import {
   TerminatedEvent,
   ThreadEvent,
   StackFrame,
+  ErrorDestination,
 } from "@vscode/debugadapter";
 import { DebugProtocol } from "@vscode/debugprotocol";
 import { Logger } from "../Logger";
@@ -443,10 +444,16 @@ export class CDPDebugAdapter extends DebugSession implements CDPSessionDelegate 
       this.sendResponse(response);
     } catch (e) {
       Logger.error("Error executing custom debugger request command:", command, e);
-      this.sendErrorResponse(response, {
-        format: (e as Error).message,
-        id: 1,
-      });
+      this.sendErrorResponse(
+        response,
+        {
+          format: (e as Error).message,
+          id: 1,
+        },
+        undefined,
+        undefined,
+        ErrorDestination.User
+      );
     }
   }
 


packages/vscode-extension/lib/babel_transformer.js
```diff
@@ -1,60 +1,87 @@
...
-overrideModuleFromAppDependency("react-native", "@babel/plugin-transform-react-jsx-self", {
-  name: "rnide-disabled-jsx-self-transform",
-  visitor: {},
-});
+// The access to component stack in versions prior to 0.80 we rely on the JSX transform plugin. Apparently
+// in many setup the transform is misconfigured and does not produce the desirable output that we can later access.
+// In addition, on versions between 0.74 and 0.79 we patch the React renderer in order to make the source information
+// being added and passed down to the fiber nodes (more on that below).
+// Radon IDE only supports RN 0.74+ so it is ok to only check for 0.7 prefix (hopefully we will never reach 0.700)
+if (RN_VERSION.startsWith("0.7")) {
```
filip131311, line 18: Why is this new env variable needed? 
previously we would use just use: 
`const { version } = requireFromAppDir("react-native/package.json");`

And if it does not work how does :
```js
 const jsxDevTransformer = requireFromAppDependency(
    "react-native",
    "@babel/plugin-transform-react-jsx/lib/development"
  );
```

work?

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -71,6 +72,9 @@ export interface DebugSession {
   startProfilingCPU(): Promise<void>;
   stopProfilingCPU(): Promise<void>;
 
+  // source map related
```
filip131311, line 75: Nit (?) I am not sure if this comment was ment to make it to review but: 
If this comment is for cutting out a section of the interface for all source map related functionality, please use 

```
// #region <region_name>

// some code

// #endregion <region_name>
```

and if it a comment for a specific function 

```
/**
* comment here 
*/

File: packages/vscode-extension/src/debugging/DebugSession.ts [hunk 3]
```diff
@@ -168,6 +172,23 @@ export class DebugSessionImpl implements DebugSession, Disposable {
     );
   }
 
+  public async findOriginalPosition(sourceInfo: SourceInfo): Promise<SourceInfo> {
+    try {
+      const response = await this.jsDebugSession?.customRequest(
+        "RNIDE_findOriginalPosition",
+        sourceInfo
+      );
+      return {
```
filip131311, line 182: why not just return the response?

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -690,6 +690,16 @@ export class ApplicationSession implements Disposable {
     let stack = undefined;
     if (requestStack && inspectData?.stack) {
       stack = inspectData.stack;
+      // for stack entries with source names that start with http, we need to decipher original source positions
+      // using source maps via the debugger
+      await Promise.all(
```
filip131311, line 695: I think even if we get a rejection here we should still serve something to the frontend here

packages/vscode-extension/src/project/metro.ts
```diff
@@ -223,6 +224,10 @@ async function launchMetro({
     RCT_METRO_PORT: `${port}`,
     RADON_IDE_LIB_PATH: libPath,
     RADON_IDE_VERSION: extensionContext.extension.packageJSON.version,
+    // we pass the RN version as environment variable as it is used in some of the transformer initialization code.
+    // it is ok to set this particular version for the metro process as RN updates ship with different versions of metro
+    // and the metro process would need to be restarted anyway.
+    RADON_IDE_RN_VERSION: findReactNativeVersion(appRoot),
```
filip131311: I believe it is unnecessary, I mention it in a comment above

packages/vscode-extension/lib/babel_transformer.js
```diff
@@ -1,60 +1,87 @@
...
+    name: "rnide-disabled-jsx-self-transform",
+    visitor: {},
+  });
 
-overrideModuleFromAppDependency("react-native", "@react-native/babel-preset/src/plugin-warn-on-deep-imports.js", buildPluginWarnOnDeeImports(process.env.RADON_IDE_LIB_PATH))
+  overrideModuleFromAppDependency(
+    "react-native",
+    "@react-native/babel-preset/src/plugin-warn-on-deep-imports.js",
+    buildPluginWarnOnDeeImports(process.env.RADON_IDE_LIB_PATH)
+  );
```
filip131311: this is not related to the jsx-transform this is our work around for a warning about react native internal imports we perform

kmagiera: ah, missed that

packages/vscode-extension/lib/babel_transformer.js
kmagiera: decided to go with the variable as I remember this setup was quite sensitive to the order of different modules being loaded. As such I didn't want to trigger any loading before the whole module override thing is over. Maybe I'm paranoic and perhaps it'd be safe to import the package.json file from react-native, at least this wouldn't trigger more modules getting loaded as the json file can't have dependencies.

packages/vscode-extension/src/debugging/DebugSession.ts
kmagiera: this is just to match the style of other comments that are above this line. Won't be changing this to regions

packages/vscode-extension/src/debugging/DebugSession.ts
kmagiera: customRequest returns `any` I could also use `as SourceInfo` but I don't like using `as` because it can hide certain categories of bugs. In this particular example it would make no difference since we implicitely have "as" because the return value is typed. Just avoiding "as" for the purpose of not popularizing it across the codebase.

packages/vscode-extension/src/project/applicationSession.ts
kmagiera: makes sense, we can just skip updating the source in this case. In the current setup this will effectively hide that record from the UI, but it could be that we'd get other stack items resolved properly

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
```diff
@@ -330,6 +331,20 @@ export class ProxyDebugAdapter extends DebugSession {
...
+    try {
+      const res = await this.childDebugSession?.customRequest("getPreferredUILocation", {
+        originalUrl: sourceInfo.fileName,
+        line: sourceInfo.line0Based,
+        column: sourceInfo.column0Based,
+      });
+      return { fileName: res.source.path, line0Based: res.line, column0Based: res.column };
+    } catch (e) {
+      Logger.error("Error calling getPreferredLocation from the JS debugger", e);
+      return { error: e };
```
jwajgelt: Is this enough for `jsDebugSession.customRequest("RNIDE_findOriginalPosition",...)` to throw an error when this command is sent? Don't we need to send an error back with `this.sendErrorResponse` instead?

packages/vscode-extension/src/debugging/DebugSession.ts
```diff
@@ -168,6 +172,23 @@ export class DebugSessionImpl implements DebugSession, Disposable {
     );
   }
 
+  public async findOriginalPosition(sourceInfo: SourceInfo): Promise<SourceInfo> {
+    try {
+      const response = await this.jsDebugSession?.customRequest(
```
jwajgelt: We should have a guard for `jsDebugSession`

```suggestion
    if (this.jsDebugSession === undefined) {
      throw new Error("Could not get original source position, JS Debugger is not connected.");
    }
    try {
      const response = await this.jsDebugSession.customRequest(
```

packages/vscode-extension/src/debugging/ProxyDebugAdapter.ts
```diff
@@ -338,6 +353,9 @@ export class ProxyDebugAdapter extends DebugSession {
   ) {
     response.body = response.body || {};
     switch (command) {
+      case "RNIDE_findOriginalPosition":
+        response.body = await this.findOriginalPosition(args);
+        break;
```
jwajgelt: We should implement this in the `CDPDebugAdapter` as well, since we provide a flag to turn it on for the newer RN versions.

jwajgelt (changes requested):

filip131311 (approved):
unblocking

PR #1661, opened 2025-10-15
Add new launch config option to disable runtime fingerprint checks

kmagiera:
We noticed that in some setups the fingerprint mechanism may trigger a number of weird issues specifically around fast refresh. This is due to the fact that some setups (and specifically when using a custom fingerprint script) may write files that then in turn trigger both metro watcher and also the fs watcher that we use for triggering the fingerprint checks.

To provide an escape hatch for troubleshooting those issues, we're adding new "hidden" option. This option will not be documented yet as we generally don't want people to use it unless instructed to do so.

### How Has This Been Tested: 
1. Run RN app project, place breakpoint in the fingeprint code, change something, see it stops
2. Now add launch config with `disableNativeBuildStaleChecks` option set to true and repeat the above, the bp will never pause


filip131311 (approved):
👍

PR #1662, opened 2025-10-15
Use initialUrl option to launch expo go and devclient projects on iOS

kmagiera:
A while ago (in SDK 52) expo added an option that allows for Expo Go and Dev Client to pass the link as the process arguments rather than using open URL: https://github.com/expo/expo/pull/33182

This has two main benefits:
1) we no longer need to fix the permission setting for the URL to open automatically without a dialog
2) we can monitor the app process output (when launching via deeplink we don't get the stdout from the app process)

This PR adapts this mechanism by first detecting the SDK version and using it conditionally.

1) we pass appRoot to launchApp because it is needed in order to detect the version. While this new parameter doesn't fit super well it seems to be the option.
2) we use appRoot to detect expo version using the package discovery (similar to how we check for min version)
3) we append `--initialUrl` with the deeplink as launch arguments for Expo SDK 52+
4) we're adding additional logs in that place to make it easier to detect issues in the launch process

### How Has This Been Tested: 
1) run a couple of different configurations of projects: expo devclient on SDK 52, 53 and 51, expo go on 53 and 51.
2) Verify the project launches and that it prints that the deeplink is used on SDK 51.
3) On SDK 53 you should see "Radon IDE (Simulator Logs)" among output channels

### DIFF
--- a/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
+++ b/packages/vscode-extension/src/devices/IosSimulatorDevice.ts
@@ -26,6 +26,7 @@ import {
   IOSRuntimeInfo,
   Locale,
 } from "../common/State";
+import assert from "assert";
 
 interface SimulatorInfo {
   availability?: string;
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -493,16 +493,38 @@ export class IosSimulatorDevice extends DeviceBase {
...
+          "Launching the app",
+          build.bundleID,
+          "using --initialUrl parameter",
+          expoDeeplink
+        );
+        // for Expo SDK 52+ we can use the --initialUrl parameter to pass to the launched process on iOS
+        // this option allows us to avoid launching via deeplink and this way to avoid linking permission
+        // overrides as well as allows us to capture the process logs into the output channel.
+        await this.configureMetroPort(build.bundleID, metroPort);
+        const launchArgsWithInitialUrl = [...launchArguments, "--initialUrl", expoDeeplink];
```
filip131311, line 524: If `launchArguments` is customizable is it possible that the user might set it themselves if so should we override such a setting here?

filip131311 (approved):
👍

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
kmagiera: the launch config can be customized, but in this setup the args are captured by expo go and dev client code. I don't think there's any way to access them anyway. I'll add an assert to fail in case `--initialUrl` is among the provided arguments

PR #1663, opened 2025-10-16
Use alias to reach RN internals while avoiding internal access warnings

kmagiera:
In #1249 we added an override for warn-on-deep-imports babel plugin to prevent the internal RN imports that Radon runtime is making from showing up as warnings. The approach works by monkey patching the node's module system and replacing the loaded module with a code that we provide.

This approach while effective, is very brittle. Going forward we should try to minimize the number of places where we rely on this mechanism. For example, in #1660 we remove all the other places where we use this technique (at least for modern enough versions of React Native that introduces better mechanisms for implementing different types of features that Radon IDE requires).

This PR deletes the overwrite entirely along with the plugin code that was meant to be a substitute for the warn-on-deep-imports plugin. Instead, we use metro's alias for the root of the react native package. This way, the plugin doesn't detect that we use deep imports, as it checks for whether the import include `react-native/` (with a trailing slash).

The aliasing technique is something we already use for other things (i.e. for importing files from the Radon's lib directory, or for importing files from the app root directory), so the mechanism is proven to be reliable under different project settings.

Along with this change, we also update the `instrumentation.js` that was doing direct `react-native` imports for FabricUIManager. We change it such that it uses common RNInternals module and add the FabricUIManager as one of the exported modules to rn-internals*.

### How Has This Been Tested: 
1) Run RN-81 app -> see no warnings on the debug console
2) I sampled a couple of older RN test apps to check whether I get eny errors (rn-74, rn-73, rn-75, rn-76, rn-77, rn-78, rn-82, rn-80)
3) Test outline rerenders on each of those apps

### DIFF
--- a/packages/vscode-extension/lib/instrumentation.js
+++ b/packages/vscode-extension/lib/instrumentation.js
@@ -8,12 +8,10 @@ const {
   isCompositeFiber,
   traverseRenderedFibers,
 } = require("__RNIDE_lib__/bippy");
-const {
-  getFabricUIManager,
-} = require("__REACT_NATIVE_INTERNALS__/Libraries/ReactNative/FabricUIManager.js");
+const RNInternals = require("./rn-internals/rn-internals");
 const { StatusBar, UIManager, Platform } = require("react-native");
 const DimensionsObserver = require("./dimensions_observer");
-const FabricUIManager = getFabricUIManager();
+const FabricUIManager = RNInternals.FabricUIManager.getFabricUIManager();
 
 const CORE_COMPONENT_NAMES = new Set([
   "ActivityIndicator",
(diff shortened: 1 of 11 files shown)


jwajgelt (approved):
👍

packages/vscode-extension/lib/instrumentation.js
```diff
@@ -8,7 +8,9 @@ const {
   isCompositeFiber,
   traverseRenderedFibers,
 } = require("__RNIDE_lib__/bippy");
-const { getFabricUIManager } = require("react-native/Libraries/ReactNative/FabricUIManager.js");
+const {
```
filip131311: Can we use `RNInternals` as an import here as well, I missed it in the past because is is not located in the wrapper.js nor in runntime.js, but  `RNInternals`  lets us keep track of those unstable imports

filip131311 (approved):
Left a question inline

packages/vscode-extension/lib/instrumentation.js
kmagiera: yes, makes sense. I can change it here because I'm already updating all the "internals" across all the RN versions

PR #1665, opened 2025-10-16
Enable Radon AI by default after redoing MCP server handling logic to use newly available APIs

kmagiera:
This PR switches Radon AI on by default. We previously had to disable Radon AI on Cursor, as we couldn't find a good way to register the server without triggering some annoying modals or modifying files that belong to the users' workspace. 

Recently Cursor added MCP integration API for extensions and this PR adapts it along with implementing a couple of improvements over the current mechanism: https://cursor.com/docs/context/mcp-extension-api#api-reference

While working on this update we discovered several issues with the way session/error handling was implemented in the local MCP server. This PR updates these along the way as they turned out to be quite coupled with the rest of the changes.

The key highlights of changes made here:
1) We update the code resposible for managing the mcp.json files. We no longer add new entries to these files and the lefover code is here to cleanup entries that the previous versions of Radon may have written. We make sure only http entries for Radon are deleted as now people can also add stdio entry to integrate Radon with other MCP clients or older versions of Cursor/VSCode.
2) The local http server now supports session properly. We use the template from https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#with-session-management for proper session id generation and for handling lost sessions. This helps with the scenarios where you have multiple agents connected (i.e. VSCode and Claude Code)
3) We delete all the reload handling logic that'd result in server being restarted. All MCP clients like to keep mcp sessions active and when restarting they need to go over initialization again. In most cases when MCP server gets disconnected, the client wouldn't retry the connection again. This is now changed and we avoid restarting the server altogether unless Radon AI is disabled in the settings.
4) Instead of restarting we now rely on tools-changed notification. This is a required part of MCP protocol. Apparently some clients (specifically Cursor) doesn't support this notification. However, we noticed that Cursor has a hidden "command" called "mcp.toolListChanged" that allows us to trigger the tool reload manually.
5) Finally, we try to handle the remote Radon AI server in a non-blocking fashion. The commands to the local MCP server are now never blocked by the remote calls. Before, when initializing, we'd need to fetch tools list from the remote server first. Now, instead of doing that, we allow for the Radon AI server communication to happen asynchronously. We introduce a concept of "local" and "remote" tools. The remote tools can be loaded at any time. When an error occurs when executing remote tool, we check if remote Radon AI server is available and if it is not, we reset the list of remote tools (triggering tool update notification), then retry connecting with Radon AI in the background.
6) The above mechanism replaces the need for `ConnectionLIstener`. Now instead of repeatedly  trying to connect with the remote Radon AI server (even when we have successfully communicated with it before), we only check for connection issues when the remote tool is being used. We only retry connection when the tool failed and we can no longer use MCP's ping method to ping the server.

## Test plan
1) For most tests we use Cursor with MCP Settings page opened. It shows the status of server and how many tools are available. Ideally we expect Radon AI to be always shown as available, there are 4 tools in total out of which 2 are local and 2 remote.
2) Open Radon in Cursor and notice it shows up as "extension-RadonAI" in the MCP settings list. After initializing it briefly shows 2 tools and then 4 tools after a few seconds.
3) We tested remote Radon AI errors by triggering Auth exceptions or delaying the requests with timeouts. In case of auth errors we'd only expect 2 tools to show. In case of delaying we would see 2 tools initially and then 4 tools.
4) We tested Radon AI with wifi turned off. The server would connect and show 2 tools although this test isn't super practical because we wouldn't be able to use the model anyways.
5) Test turning Radon AI off in the settings. This should result in Radon getting removed from the MCP Server page. This was tested both in Cursor and VSCode where the logic for handling this usecase has also been updated.
6) In VSCode after disabling and enabling RadonAI setting, the server shows up on the list but isn't automatically started. When Copilot wants to interact with it it starts it. But the behavior still differs compared to the case when the extension launches on an opened project (when project gets opened with setting ON the server starts automatically)

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -413,16 +413,6 @@
           "default": true,
           "scope": "window",
           "description": "Enables the Radon AI functionality in the extension."
-        },
-        "RadonIDE.radonAI.MCPConfigLocation": {
-          "type": "string",
-          "scope": "window",
-          "default": "Project",
-          "enum": [
-            "Project",
-            "Global"
-          ],
-          "markdownDescription": "Sets the write location for `mcp.json` when using Cursor or a legacy VSCode version (`<1.101.0`)."
         }
       }
     }],


packages/vscode-extension/src/ai/mcp/LocalMcpServer.ts
```diff
@@ -1,141 +1,258 @@
...
+
   private licenseTokenSubscription: Disposable;
 
-  constructor(connectionListener: ConnectionListener) {
-    const { promise, resolve } = Promise.withResolvers<number>();
+  protected onToolListChangedEmitter = new EventEmitter<void>();
 
+  constructor() {
+    this.mcpServer = new McpServer({
+      name: "RadonAI",
```
latekvo, line 36:
```suggestion
      name: ENTRY_KEY,
```

packages/vscode-extension/src/ai/shared/api.ts
```diff
@@ -5,22 +5,36 @@ import { getTelemetryReporter } from "../../utilities/telemetry";
...
 
 export async function invokeToolCall(
   toolName: string,
   args: unknown,
-  id: string,
-  connectionListener?: ConnectionListener
+  callId: string = PLACEHOLDER_ID
 ): Promise<ToolResponse> {
   getTelemetryReporter().sendTelemetryEvent("radon-ai:tool-called", { toolName });
+  const url = new URL("/api/tool_calls/", BACKEND_URL);
```
latekvo, line 33: The `BACKEND_URL` already has `/api/`, adding `/api/` to the endpoint path is redundant. 
This works, but is confusing and looks like it shouldn't work without knowing the [documentation](https://nodejs.org/api/url.html#new-urlinput-base).
I know this issue was already present in the codebase, but it'd be good to get rid of it either way.

image

packages/vscode-extension/src/ai/mcp/configCreator.ts
```diff
@@ -1,47 +1,76 @@
...
+  // In order to test the development version of the MCP server:
+  // 1) Run  `npm run build` in the `packages/radon-mcp` directory
+  // 2) update the `radonMcpEntry` to the following:
+  // {
+  //   command: "node",
+  //   args: [
+  //     path.join(extensionContext.extensionPath, "../packages/radon-mcp/dist/index.js"),
+  //     "${workspaceFolder}",
+  //   ],
+  // };
```
latekvo: This explanation is already present in the `README.md` of `radon-mcp` (#1670), so i think this block is redundant. Also not a big fan of storing such documentation in the source code. Instead, could we comment something like this:

```ts
// Development setup instructions for `radon-mcp` are available in README of the `radon-mcp` package 
```

?

packages/vscode-extension/src/ai/mcp/configCreator.ts
```diff
@@ -1,47 +1,76 @@
...
 
   try {
     const edits = modify(config, [rootKey, ENTRY_KEY], undefined, {
       formattingOptions: { insertSpaces: true, tabSize: 2 },
     });
+    if (edits.length === 0) {
+      return undefined;
+    } else {
+      return applyEdits(config, edits);
+    }
```
latekvo: Simplify to:

```suggestion
    if (edits.length) {
      return applyEdits(config, edits);
    }
```

?

packages/vscode-extension/src/ai/mcp/configCreator.ts
```diff
@@ -1,47 +1,76 @@
...
       formattingOptions: { insertSpaces: true, tabSize: 2 },
     });
-    const config = applyEdits(incompleteConfig, edits);
-    return config;
+    const newConfig = applyEdits(config, edits);
+    if (newConfig === config) {
+      return undefined;
+    } else {
+      return newConfig;
+    }
```
latekvo: Simplify to:

```suggestion
    if (newConfig !== config) {
      return newConfig;
    }
```

?

packages/vscode-extension/src/ai/mcp/LocalMcpServer.ts
```diff
@@ -1,141 +1,258 @@
...
-    return this.serverReloadEmitter.event(cb);
+  private async handleSessionRequest(req: express.Request, res: express.Response) {
+    const sessionId = req.headers["mcp-session-id"] as string | undefined;
+    if (!sessionId) {
+      res.status(400).send("Session ID is missing");
+      return;
+    }
+    const transport = this.transports.get(sessionId);
+    if (!transport) {
+      res.status(400).send("Invalid session ID");
```
latekvo: As far as I see in the documentation ([link](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management)), `400` should only be used for when `mcp-session-id` is missing, which is done correctly a few lines higher. When the session is reset, I think it'd be safer to use `404`, since it explicitly tells the client that its `mcp-session-id` has been discarded and has to be re-initialized.

packages/vscode-extension/src/ai/shared/api.ts
kmagiera: this was just moved a few lines up to be extracted out of the try catch scope

packages/vscode-extension/src/ai/mcp/configCreator.ts
kmagiera: we don't document it there and we shouldn't have it there. The README in radon-mcp will be more or less user-facing as people may want to use that package for setups with claude code or potentially other mcp clients.

We could have a separate readme file in the source directory structure under the extension code. However while this is a pattern I've seen in some projects I'd easily miss such files as I never use the file directory structure to navigate to files but rely on code search / fuzzy file search. If I ever needed to modify radon-mcp and test it from Cursor, this is the exact place I'd found first and this comment would turn out super handy.

packages/vscode-extension/src/ai/mcp/configCreator.ts
kmagiera: I'd rather be explicit with returning undefined. This makes it clear what the intent was and what undefined return value means. Otherwise we'd rely on opaque return which always leaves a doubt whether it was intentional or not

packages/vscode-extension/src/ai/mcp/LocalMcpServer.ts
kmagiera:
```suggestion
      res.status(404).send("Invalid session ID");
```

packages/vscode-extension/src/ai/shared/api.ts
kmagiera:
```suggestion
  const url = new URL("/tool_calls/", BACKEND_URL);
```

PR #1668, opened 2025-10-17
feat: network inspector - add image preview, change webview messaging protocol

pFornagiel:
### Description

This PR introduces image preview feature for network responses of `image/` type and makes needed changes to the extension <-> webview communication protocol.

The new `Preview` panel has been added to `NetworkLogDetails` and becomes available when the request is of appropriate image type and employs loading and error handling. It allows the user to view the image and displays its metadata in the bar below it. To properly handle bar nesting and content-setting, the `TabBarProvider` with appropriate context has been introduced.

Image preview functionality is available both using `DebuggerNetworkInspector` and `InspectorBridgeNetworkInspector` using the fact that image data is passed as base64 string, which can then be parsed and embedded in the `img` html component. In order to fully support it, changes have been made to both new implementation of inspectors and the `lib/network` scripts. 

Due to the fact that the cache used in `InspectorBridgeNetworkInspector` has limited size, when using this strategy, some of the larger photos may become unpreviewable due to response body truncation. 

image

Screenshot 2025-10-17 at 18 05 31

In order to properly handle images, it was handy to introduce additional information to the `Network.getResponseBody` method requests. Thus, to remain consistent with `CDP` protocol, the `IDE.getResponseBodyData` method was introduced, in order to be able to pass more information about the response body to the webview panel, without altering the original standard. 

```ts
export interface ResponseBody {
  body: string | undefined;
  base64Encoded: boolean;
}

export interface ResponseBodyData extends ResponseBody {
  type: ResponseBodyDataType;
  fullBody?: string;
  wasTruncated?: boolean;
}
```

Now upon requesting the response body, the webview uses `IDE.getResponseBodyData` method in the message. Upon receiving a message with this method, network inspector implementations use `Network.getResponseBody` internally to acquire response body using `CDP` standard and then wrap it with additional information in order to send it back to the webview.

### How Has This Been Tested: 

- run Radon IDE extension panel, open network inspector
- *make image fetches, see if the preview tab behaves as expected*

WIP.

### DIFF
--- a/packages/vscode-extension/lib/network/networkRequestParsers.ts
+++ b/packages/vscode-extension/lib/network/networkRequestParsers.ts
@@ -225,7 +225,7 @@ async function readResponseText(
     const responseType = xhr.responseType;
 
     if (responseType === "" || responseType === "text") {
-      return parseResponseBodyData(xhr.responseText);
+      return await parseResponseBodyData(xhr.responseText);
     }
 
     if (responseType === "blob") {
@@ -237,23 +237,18 @@ async function readResponseText(
       const isImageType = contentType.startsWith("image/");
       const isOctetStream = contentType === "application/octet-stream";
 
-      let promise: Promise<InternalResponseBodyData> | undefined = undefined;
-
       if (isImageType || isOctetStream) {
-        promise = readBlobAsBase64(xhr.response, isImageType);
+        return await readBlobAsBase64(xhr.response, isImageType);
       }
 
       if (isTextType || isParsableApplicationType) {
-        promise = readBlobAsText(xhr.response);
+        return await readBlobAsText(xhr.response);
       }
-
-      return promise?.catch((error) => handleReadResponseError(error));
     }
 
     // don't read binary data
     return undefined;
   } catch (error) {
-    // in case synchronous errors occur
     return handleReadResponseError(error);
   }
 }


packages/vscode-extension/lib/network/networkRequestParsers.ts
```diff
@@ -86,23 +110,84 @@ const TRUNCATED_LENGTH = 1000; // 1000 characters
...
+
+    reader.onload = () => {
+      const result = reader.result;
+      const textResult = typeof result === "string" ? result : undefined;
+      resolve(parseResponseBodyData(textResult, false, ResponseBodyDataType.Other));
+    };
+
+    reader.onerror = (error) => {
+      resolve(handleReadError(error));
+    };
```
```suggestion
  return new Promise((resolve, reject) => {
    const reader = new FileReader();

    reader.onload = () => {
      const result = reader.result;
      const textResult = typeof result === "string" ? result : undefined;
      resolve(parseResponseBodyData(textResult, false, ResponseBodyDataType.Other));
    };

    reader.onerror = (error) => {
      reject(error);
    };
```
(and likewise for `readBlobAsBase64`) would let you handle all your errors in one `catch`, and feels more natural.

packages/vscode-extension/src/network/components/Tabs/TabBar.css
```diff
@@ -0,0 +1,43 @@
...
+}
+
+.preview-bar{
+
+  display: flex;
+  width: 100%;
+  height: 100%;
+  div {
+    /* padding: 8px auto 0 auto; */
+    /* margin: 0 auto 0 auto; */
```
jwajgelt: Are these left in deliberately?

packages/vscode-extension/src/network/components/Tabs/TabBar.tsx
```diff
@@ -0,0 +1,20 @@
+import { useTabBar } from "../../providers/TabBarProvider";
+import "./TabBar.css";
+
+// const TAB_BAR_HEIGHT = 32;
```
jwajgelt: Remove the comments before merging.

jwajgelt (approved):
Left a couple nits, but overall looks good and seems to work fine.

packages/vscode-extension/lib/network/networkRequestParsers.ts
pFornagiel: Changed in 2b9ba016. The promises are now rejected and handled inside `readResponseText`.

```ts
async function readResponseText(
  xhr: XMLHttpRequest
): Promise<InternalResponseBodyData | undefined> {
  try {
    // @ts-ignore - RN-specific property // @ts-ignore - RN-specific property
    if (!xhr || !xhr._cachedResponse) {
      // if response was not accessed, it's not cached and we don't want to read it
      // here to avoid potential side effects
      return undefined;
    }

    const responseType = xhr.responseType;

    if (responseType === "" || responseType === "text") {
      return parseResponseBodyData(xhr.responseText);
    }

    if (responseType === "blob") {
      const contentType = getContentTypeHeader(xhr);
      const isTextType = contentType.startsWith("text/");
      const isParsableApplicationType = Array.from(PARSABLE_APPLICATION_CONTENT_TYPES).some(
        (type) => contentType.startsWith(type)
      );
      const isImageType = contentType.startsWith("image/");
      const isOctetStream = contentType === "application/octet-stream";

      let promise: Promise<InternalResponseBodyData> | undefined = undefined;

      if (isImageType || isOctetStream) {
        promise = readBlobAsBase64(xhr.response, isImageType);
      }

      if (isTextType || isParsableApplicationType) {
        promise = readBlobAsText(xhr.response);
      }

      return promise?.catch((error) => handleReadResponseError(error));
    }

    // don't read binary data
    return undefined;
  } catch (error) {
    // in case synchronous errors occur
    return handleReadResponseError(error);
  }
}
```

packages/vscode-extension/src/network/components/Tabs/TabBar.css
pFornagiel: An oversight. Removed in fa9389f3

packages/vscode-extension/src/network/components/Tabs/TabBar.tsx
pFornagiel: An oversight. Removed in fa9389f3

packages/vscode-extension/lib/network/networkRequestParsers.ts
```diff
@@ -237,19 +237,24 @@ async function readResponseText(
...
+
+      return promise?.catch((error) => handleReadResponseError(error));
     }
 
     // don't read binary data
     return undefined;
   } catch (error) {
-    return handleReadError(error);
+    // in case synchronous errors occur
+    return handleReadResponseError(error);
```
```suggestion
      if (isImageType || isOctetStream) {
          return await readBlobAsBase64(xhr.response, isImageType);
      }
  
      if (isTextType || isParsableApplicationType) {
          return await readBlobAsText(xhr.response);
      }
    }

    // don't read binary data
    return undefined;
  } catch (error) {
    // in case synchronous errors occur
    return handleReadResponseError(error);
```

should work as well, and reads better.
If an `await`ed promise rejects in a `try`/`catch` block, the `catch` will handle the error.
It doesn't work if you return without `await`ing, since in that case the function simply returns the promise synchronously rather than doing the `async`/`await` magic.

pFornagiel: Thanks, refactored in c4aafe09

PR #1675, opened 2025-10-21
fix: network inspector - response body not updating after the request status changed

pFornagiel:
### Description

This PR resolves a bug where opening the details tab during a `(pending)` request resulted in the response body not updating even after the request status changed and a response was received. The issue resulted from the useEffect hook only reacting to changes in networkLog.requestId.

### How Has This Been Tested: 

- run Radon IDE extension panel, open network inspector
- *simulate delay, open request when its pending, see if response body updates correctly*
- 

Not applicable.

### DIFF
--- a/packages/vscode-extension/src/network/components/NetworkLogDetails.tsx
+++ b/packages/vscode-extension/src/network/components/NetworkLogDetails.tsx
@@ -36,6 +36,8 @@ interface Tab {
   Tab: React.FC<TabProps>;
 }
 
+let prevRequestId: string | null = null;
+
 const NetworkLogDetails = ({ networkLog, handleClose, parentHeight }: NetworkLogDetailsProps) => {
   const headerRef = useRef<VscodeTabHeaderElement>(null);
   const [selectedTabIndex, setSelectedTabIndex] = useState(0);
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/network/components/NetworkLogDetails.tsx
```diff
@@ -35,6 +36,8 @@ interface Tab {
   Tab: React.FC<TabProps>;
 }
 
+let prevRequestId: string | null = null;
```
jwajgelt, line 39: This should be in a `useRef` in the component.

packages/vscode-extension/src/network/components/NetworkLogDetails.tsx
```diff
@@ -45,11 +48,23 @@ const NetworkLogDetails = ({ networkLog, handleClose, parentHeight }: NetworkLog
...
+    const shouldUpdate =
+      prevRequestId === networkLog.requestId &&
+      networkLog.currentState !== NetworkEvent.LoadingFinished &&
+      networkLog.currentState !== NetworkEvent.LoadingFailed;
+
+    if (!shouldUpdate) {
+      return;
+    }
+
+    prevRequestId = networkLog.requestId;
```
jwajgelt, line 63: I don't think this is a good fix.
I would expect the `getResponseBody` promise to resolve with the body even if the function is called while the request is pending (which, as I understand, it doesn't now, which is causing the issue).

jwajgelt (changes requested):

PR #1677, opened 2025-10-21
fix: preview with debug stack

filip131311:
after #1660 we no longer always have access to react components source location. 

To avoid problems with components source entirely we no longer use it to identify user previews instead we use source location of where the preview function was called retrieved using `new Error().stack` instead. Unfortunately the source retrieved in this way points to the function call made in the bundle, so we translate it to original source using `symbolicateStackTrace` from react native internals. 

### How Has This Been Tested: 

- run expo-54 (so react native 81) test app and check the previe functionality. 
- run every react native test app in our test app directory and check previe functionality, we have to check all of them as  internal imports change their location in react native module between versions

internal

### DIFF
--- a/packages/vscode-extension/lib/preview.js
+++ b/packages/vscode-extension/lib/preview.js
@@ -47,14 +47,13 @@ function getComponentName({ type }) {
 
 async function getCallSourceFromStack(stack) {
   const parsedStack = RNInternals.parseErrorStack(stack);
-  const symbolicatedStack = (await RNInternals.symbolicateStackTrace(parsedStack)).stack;
-
-  const callerFrame = symbolicatedStack[1];
+  const callerFrame = parsedStack[1];
+  const symbolicatedCallerFrame = (await RNInternals.symbolicateStackTrace([callerFrame])).stack[0];
 
   return {
-    fileName: callerFrame.file,
-    lineNumber: callerFrame.lineNumber,
-    columnNumber: callerFrame.column,
+    fileName: symbolicatedCallerFrame.file,
+    lineNumber: symbolicatedCallerFrame.lineNumber,
+    columnNumber: symbolicatedCallerFrame.column,
   }
 }
 


packages/vscode-extension/.vscode/settings.json
```diff
@@ -42,10 +42,13 @@
     "sdcard",
     "sdkmanager",
     "sharedpreferences",
+    "Simbolicate",
```
filip131311:
```suggestion
```

packages/vscode-extension/lib/preview.js
```diff
@@ -44,25 +45,66 @@ function getComponentName({ type }) {
   return "(unnamed)";
 }
 
+async function getSourceFromComponent(component) {
+  if (component._source) {
```
kmagiera: can we use `new Error().stack` instead of extracting the debug stack here. Then in the `PreviewCodeLensProvider` we'd just directly link the line where preview is called rather than where component is defined

packages/vscode-extension/lib/preview.js
```diff
@@ -44,25 +45,66 @@ function getComponentName({ type }) {
...
+
+  const debugStack = component._debugStack;
+  const parsedStack = RNInternals.parseErrorStack(debugStack.stack);
+
+  const { file } = parsedStack[1];
+
+  const url = new URL(file);
+  const metroAddress = url.origin;
+
+  const metroSymbolicateResponse = await fetch(
```
kmagiera: maybe we can use `react-native/Libraries/Core/Devtools/symbolicateStackTrace` via RNInternals here?

kmagiera: we'd need to use `parseErrorStack` for the above to work, but should be ok

filip131311: yes thats much better I didn't found it thank you

packages/vscode-extension/lib/preview.js
```diff
@@ -44,25 +45,40 @@ function getComponentName({ type }) {
...
+  const symbolicatedStack = (await RNInternals.symbolicateStackTrace(parsedStack)).stack;
+
+  const callerFrame = symbolicatedStack[1];
+
+  return {
+    fileName: callerFrame.file,
+    lineNumber: callerFrame.lineNumber,
+    columnNumber: callerFrame.column,
+  }
+}
```
jwajgelt, line 58: It'd make more sense if it created the error itself.

```suggestion
async function getCallSourceFromStack() {
  const stack = new Error().stack;
  const parsedStack = RNInternals.parseErrorStack(stack);
  const symbolicatedStack = (await RNInternals.symbolicateStackTrace(parsedStack)).stack;

  const callerFrame = symbolicatedStack[2];

  return {
    fileName: callerFrame.file,
    lineNumber: callerFrame.lineNumber,
    columnNumber: callerFrame.column,
  }
}
```

packages/vscode-extension/lib/preview.js
```diff
@@ -44,25 +45,40 @@ function getComponentName({ type }) {
   return "(unnamed)";
 }
 
+async function getCallSourceFromStack(stack) {
+  const parsedStack = RNInternals.parseErrorStack(stack);
+  const symbolicatedStack = (await RNInternals.symbolicateStackTrace(parsedStack)).stack;
+
+  const callerFrame = symbolicatedStack[1];
```
jwajgelt: Is it possible to send only a single frame to symbolicate, rather than the whole call stack? If so, it would save us some work and network traffic.

packages/vscode-extension/src/providers/PreviewCodeLensProvider.ts
```diff
@@ -84,17 +84,18 @@ export class PreviewCodeLensProvider implements CodeLensProvider {
     // Elliminate lines that contain double slashes indicating a comment.
     const previewRegex = /^(?:(?!\/\/) )*preview\(\s*<\s*/gm;
     for (const match of text.matchAll(previewRegex)) {
-      // get the line number where the last character (<) of the matched regex is located
-      const jsxOpeningTagPosition = document.positionAt(match.index + match[0].length - 1);
-      const jsxOpeningTagLine0Based = jsxOpeningTagPosition.line;
+      // get the line number where the opening parenthesis "(" in "preview(" is located
+      const openingParenIndex = match.index + match[0].indexOf("(");
```
jwajgelt, line 88: I'm not sure I understand why we look for the parenthesis, instead of the `preview` symbol, or event just the RE match.

packages/vscode-extension/lib/preview.js
filip131311: thats the unfortunate thing, I tried that but it behaves unpredictably if i do so. 

Screenshot 2025-10-21 at 16 15 50

jwajgelt: Ah, it's async.

jwajgelt: You can change it to be a regular function, and use `.then()` on the promise you get from `symbolicateStackTrace`, it should work then.

jwajgelt: And if not, let's keep it as is.

filip131311: anyway i would have to perform _some_ logic outside of this function so ill keep it as is bu i did change the logic so it sends only one frame

packages/vscode-extension/src/providers/PreviewCodeLensProvider.ts
filip131311: it should be preview symbol you are right

filip131311: this is how `new Error().stack` sees it anyway

filip131311: altho the whole regex should be fixed

filip131311: as something like that 
```
preview



(
  <Button
    title="Button"
    onPress={() => {
      console.log("console.log()");
    }}
  />
);
```
does not register the code lens

filip131311: i will fix it in a fallow up after 1.12.1 to not tempt the regex gods it is not a regression

PR #1679, opened 2025-10-22
fix: do not replace react renderer if react version does not match 

filip131311:
In order to make inspector work on react native 78-79 we replace the react renderer. Do detect if the replacement is necessary we relied on react native version found in the project, but as it turns out there are projects using `reactCanary` flag in `appConfig.js` that allow for upgrading react version independently from react native version. To avoid mismatch renderer versions we add additional check for react version. 

### How Has This Been Tested: 

- run react-native-79 test app and see if inspector works
- do the same for expo-53 and expo-53 with canary builds
- and the same for react-native-78

internal


kmagiera (approved):
looks good, you should also verify this works with RN 78 and with the setups with canary builds (this should be a part of the test plan)

PR #1682, opened 2025-10-22
fix: react query in projects with non flat dependency structures 

### LINKED ISSUE #1671: React Query devtools does not work since expo 53
**Description**
React Query is not detected, by the extension since expo 53

filip131311:
This PR improves multiple aspects of the react-querry dev tool:

1) in order to resolve issues with non flat dependency structures we opt-out of the module re-export mechanism previously used to add required logic the `QueryClient` object. Instead we inject our code as string. This way when we import `QueryClient` from react query core library we do so from `react-query` module that has it as a dependency. 

2) During testing we found that sometimes react query plugin is not registered until the first query is performed. This should not be a problem for most applications but we try to fix it anyway. We modify the dev plugin registration logic so we can safely register the same module multiple times, then we use it to try and register react query plug in as soon as `react-query-devtool.js` logic has ended.

3) During testing we found that sometimes the files of react query might have `.mjs` extension preventing us from detecting them in `babel_transormer.js`, so we include them in a list of possible `react-query` entrypoints.  

fixes #1671

note: this PR has no effect on react query v3 and older as it started using `@tanstack`  module namespace since v4

### How Has This Been Tested: 

- test pnpm monorepo with expo 54 and react-qerry v5 
- test pnpm monorepo with expo 54 and react-qerry v4
- test expo-53-prebuild-with-plugins with react-qerry v5 
- test expo-53-prebuild-with-plugins with react-qerry v4

internal

### DIFF
--- a/packages/vscode-extension/lib/plugins/react-query-devtools.js
+++ b/packages/vscode-extension/lib/plugins/react-query-devtools.js
@@ -1,6 +1,9 @@
 // IMPORTANT: this file is injected at the beginning of React Query index file 
 // so we need to use variable names that won't conflict with React Query's own variables
 // hence the "RadonIDE" suffixes on everything.
+
+// We import QueryClient from @tanstack/query-core directly
+// as in v4 version of React Query it is unavailable from the main package
 import { QueryClient as QueryClientRadonIDE } from "@tanstack/query-core";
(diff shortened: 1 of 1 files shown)


filip131311:
close in favor of #1687

PR #1684, opened 2025-10-23
Update launch config in reload if stale config was picked

kmagiera:
When changing the launch configuration we get "save and restart" button to apply the modifications. However, I often edit launch.json settings directly. When I update the setting, I get the `UnknownConfigurationAlert` displayed. 

When iterating on the launch config, I'd like the restart buttons to pick up the changes. Now if there's a build error due to misconfigured launch file and I update the settings, I get the `UnknownConfigurationAlert` that I dismiss. Underneath I get "restart" button that I click. In that scenario I end up restarting the process with the old configuration instead of the new one.

This PR fixes this flow by checking inside the reload command whether the configuration is still active or not. If it isn't, we either select configuration with the matching name or display an error.

This solution doesn't provide an ideal flow, as we still restart the whole device session because of triggering the launch configuration change. However, we are limited by the current design and it would be too risky to introduce a separate flow for updating the launch configuration w/o triggering the whole device session manager reboot.

### How Has This Been Tested: 
1) Open EAS local project and change the profile to invalid name, then build -> see an error
2) Open launch.json and modify the name to something else -> use rebuild button from the UI
3) Before, we'd get the exact same build error indicating the build process didn't pick up the launch configuration change. Now, when restarting it will switch the launch config to the new version.

### DIFF
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -593,15 +593,20 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
   public reloadCurrentSession(type: ReloadAction): Promise<void> {
     // upon reload, we verify the selected launch configuration is still present:
     // 1) if it is, we proceed to reloading the session.
-    // 2) if the launch configuration was modified, we update the launch configuration as
-    // part of the reload process.
-    // 3) if the launch configuration was deleted, we show an error message and throw an error.
-    const matchingLaunchConfig = this.launchConfigsManager.launchConfigurations.find(
-      (config) => config.name === this.selectedLaunchConfiguration.name
+    // 2) if the launch configuration was modified, we search for a matching launch configuration and select it.
+    // 3) if no matching launch configuration is found, we show an error message and throw an error.
+    const exactConfigExists = this.launchConfigsManager.launchConfigurations.some((config) =>
+      _.isEqual(config, this.selectedLaunchConfiguration)
     );
-    if (!_.isEqual(matchingLaunchConfig, this.selectedLaunchConfiguration)) {
-      if (matchingLaunchConfig) {
-        return this.selectLaunchConfiguration(matchingLaunchConfig);
+    if (!exactConfigExists) {
+      // find config with the same name and app root as the best candidate
+      const matchingConfigCandidate = this.launchConfigsManager.launchConfigurations.find(
+        (config) =>
+          config.name === this.selectedLaunchConfiguration.name &&
+          config.appRoot === this.selectedLaunchConfiguration.appRoot
+      );
+      if (matchingConfigCandidate) {
+        return this.selectLaunchConfiguration(matchingConfigCandidate);
       } else {
         window.showErrorMessage(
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/project/project.ts
```diff
@@ -591,6 +591,25 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
 
   public reloadCurrentSession(type: ReloadAction): Promise<void> {
+    // upon reload, we verify the selected launch configuration is still present:
+    // 1) if it is, we proceed to reloading the session.
+    // 2) if the launch configuration was modified, we update the launch configuration as
+    // part of the reload process.
+    // 3) if the launch configuration was deleted, we show an error message and throw an error.
+    const matchingLaunchConfig = this.launchConfigsManager.launchConfigurations.find(
+      (config) => config.name === this.selectedLaunchConfiguration.name
+    );
```
jwajgelt: Can't you have multiple launch configs with the same name? This check will break when picking non-first config with the same name in that case.

kmagiera: you technically can but this wouldn't be practical. You wouldn't be able to tell the difference from our UI in that case anyway

jwajgelt: That's not true. You can have configs with the same name, pointing to different app roots -- they're both shown in the UI.
Specifically, the configs created by vscode from the "initialConfigurations" template we provide in the `package.json` all have the same name ("Radon IDE panel"), so if users used that method to create them, they may have issues with this change.
You should be able to adapt your approach to handle multiple configs as well, by checking all the configs with the same name and picking _some_ config with the same name if the actual config was changed (since in that case we're in best-effort land anyway)

```suggestion
    const matchingConfigCandidates = this.launchConfigsManager.launchConfigurations.filter(
      (config) => config.name === this.selectedLaunchConfiguration.name
    );
    const matchingLaunchConfig = matchingConfigCandidates.find(config => _.isEqual(config, this.selectedLaunchConfiguration)) ?? matchingConfigCandidates[0];
```

jwajgelt: You might also want to have the check compare both `name` and `appRoot`, since in the case we're trying to improve here, you're likely not trying to run a different application (by changing the approot in the config), just tweak some settings for a specific app.

packages/vscode-extension/src/project/project.ts
```diff
@@ -591,6 +591,30 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
+          config.appRoot === this.selectedLaunchConfiguration.appRoot
+      );
+      if (matchingConfigCandidate) {
+        return this.selectLaunchConfiguration(matchingConfigCandidate);
+      } else {
+        window.showErrorMessage(
+          "The selected launch configuration was deleted or changed. Please select a new launch configuration from the list of available ones.",
+          "Dismiss"
+        );
+        throw new Error("Selected launch configuration is stale");
```
jwajgelt, line 615: I don't like us using vscode's native error dialog in this case, but let's not overcomplicate things for this specific case for now. We can come back to it later if we want.

jwajgelt (approved):

PR #1685, opened 2025-10-23
Add check for eas-cli before running local build

kmagiera:
We warn users of missing eas-cli when they have EAS launch config selected and attempt to build (see https://github.com/software-mansion/radon-ide/blob/main/packages/vscode-extension/src/builders/eas.ts#L30) but similar check was missing for the case of local EAS builds allowing the command to eventually run. On top of that the eas local build command uses `quietErrorsOnExit` option preventing the ENOENT type of errors to even be printed. This PR adds the check to match the flow for EAS non-local builds.

image

Along with this one, we also add a small change to the launch config and show "not a simulator profile" for the disabled profiles that don't have "simulator" option set. This makes it easier to understand why they are disabled.
image

### How Has This Been Tested: 
1) Uninstall eas-cli and open eas local project
2) Open launch config to see the labels next to disabled profiles.
3) Build the project to see the error

### DIFF
--- a/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
+++ b/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
@@ -575,15 +575,17 @@ function EasBuildConfiguration({
 
   const availableEasBuildProfiles = Object.entries(easBuildProfiles).map(
     ([buildProfile, buildProfileConfig]) => {
-      const canRunInSimulator =
-        buildProfileConfig.distribution === "internal" &&
-        (platform !== "ios" || buildProfileConfig.ios?.simulator === true);
-      return {
-        value: buildProfile,
-        label: buildProfile,
-        disabled: !canRunInSimulator,
-        reason: canRunInSimulator ? undefined : "not a simulator profile",
-      };
+      let disabled = false;
+      let reason = undefined;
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -614,7 +614,7 @@ function EasBuildConfiguration({
           name={`eas.${platform}.profile`}>
           {availableEasBuildProfiles.map((profile) => (
             <Option key={profile.value} value={profile.value} disabled={profile.disabled}>
-              {profile.label}
+              {profile.label} {profile.disabled ? "(not a simulator profile)" : ""}
```
jwajgelt:
```suggestion
              {profile.label}{profile.disabled ? " (not a simulator profile)" : ""}
```

jwajgelt: Should we show this for Android as well? Maybe the message should be tweaked to match the platform?

kmagiera: yeah, you're right. I think android doesn't have this restriction and only checks for debug build profiles

PR #1687, opened 2025-10-23
fix: react query in projects with non flat dependency structures

### LINKED ISSUE #1671: React Query devtools does not work since expo 53

filip131311:
This PR improves multiple aspects of the react-query dev tool:

1) in order to resolve issues with non flat dependency structures we import `QueryClient` from `react-query` main package instead of `query-core` that is only available as a dependency od `react-query` to do do we need to move our re-export of `react-query-devtool.js`  to the end of transformed file to allow for proper initialization of `QueryClient`

2) During testing we found that sometimes react query plugin is not registered until the first query is performed. This should not be a problem for most applications but we try to fix it anyway. We modify the dev plugin registration logic so we can safely register the same module multiple times, then we use it to try and register react query plug in as soon as `react-query-devtool.js` logic has ended.

3) During testing we found that sometimes the files of react query might have `.mjs` extension preventing us from detecting them in `babel_transormer.js`, so we include them in a list of possible `react-query` entrypoints.  

fixes #1671

note: this PR has no effect on react query v3 and older as it started using `@tanstack`  module namespace since v4

### How Has This Been Tested: 

- test pnpm monorepo with expo 54 and react-qerry v5 (`5.90.5`)
- test pnpm monorepo with expo 54 and react-qerry v4 (`4.42.0`)
- test expo-52-prebuild-with-plugins with react-qerry v5 (`5.90.5`)
- test expo-52-prebuild-with-plugins with react-qerry v4 (`4.42.0`)

internal

### DIFF
--- a/packages/vscode-extension/lib/wrapper.js
+++ b/packages/vscode-extension/lib/wrapper.js
@@ -454,7 +454,7 @@ export function AppWrapper({ children, initialProps, fabric }) {
         // 2. when the fullscreen LogBox is minimized, new errors won't bring it up unless we clear the old ones
         RNInternals.LogBoxData.clear();
         originalErrorHandler(error, isFatal);
-      } catch { }
+      } catch {}
     }
 
     global.ErrorUtils.setGlobalHandler(wrappedGlobalErrorHandler);


packages/vscode-extension/lib/wrapper.js
```diff
@@ -451,7 +454,7 @@ export function AppWrapper({ children, initialProps, fabric }) {
         // 2. when the fullscreen LogBox is minimized, new errors won't bring it up unless we clear the old ones
         RNInternals.LogBoxData.clear();
         originalErrorHandler(error, isFatal);
-      } catch {}
+      } catch { }
```
kmagiera:
```suggestion
      } catch {}
```

PR #1688, opened 2025-10-24
feat: support for physical Android devices

jwajgelt:
# Description:
Allows using a physical Android device inside Radon instead of an emulator.
Unlike "Radon Connect", this lets Radon fully control the process of building and running the app in the same way as when running on an emulator, making (almost) all Radon features available in this mode, including:
- device preview and simulated input in the Radon panel
- automatic project configuration, building and installation on the device
- JS debugger, developer tools, element inspector
- etc.

### Changes introduced in this PR:
- updating simulator-server to a version which supports physical Android device streaming and control
- DeviceManager refactor:
  - introduce a `DevicesProvider` abstraction which is responsible for discovering available devices and storing them in the extension state
  - move (some of) the platform-specific code from `DeviceManager` into `DevicesProvider`s
- introduce a base `AndroidDevice` abstract class for functionality shared between Emulators and Physical Android Devices, implement `AndroidEmulatorDevice` and `AndroidPhysicalDevice` on top of it
- make `Preview` react to `simulator-server` process ending before disposal, to handle a case where the sim-server disconnects from the physical device (most likely due to the device physically disconnecting from the development machine running Radon)
- adapt Preview Panel to display physical devices correctly -- including handling devices with no predefined `frame`

### KNOWN ISSUES:
- starting the app while the screen is locked hangs on "waiting for app to load"

### TODO / Not supported yet:
- clipboard sharing -- pasting to device works, but copying from device to Mac does not
- native device logs
- human-readable device names -- it shouldn't be hard to map the model number to a recognizable device name (https://storage.googleapis.com/play_public/supported_devices.html https://github.com/bsthen/device-models)
- handle device reconnects better -- we'd need the app to reconnect to metro and relaunch the sim-server. It sounds possible, but I consider it out-of-scope for this (already very large) PR

# How Has This Been Tested: 
- make sure the sim-server submodule is checked out and build it using `build:sim-server-debug` npm script
- connect a physical Android device to your mac and enable USB debugging
- Radon should automatically add the device to the devices list under "Connected Android Devices" in the "Device Select" drop-down and under "Physical Android Devices" in the "Manage Devices" view.
- disconnecting the device should hide it in the "Device Select" drop-down, and display it as disconnected in the "Manage Devices" view. The disconnected device should not be selectable.
- selecting the device should build, install, and launch the app
- check that Radon features work (touches, rotations, devtools, debugger, screenshots/recording/replays, etc.)
- check that reload actions work
- check that disconnecting the device shows an error dialog "The device was disconnected from Radon"
- check that reconnecting the device and reloading works

# How Has This Change Been Documented:
WIP

### DIFF
--- a/packages/vscode-extension/src/devices/DeviceManager.ts
+++ b/packages/vscode-extension/src/devices/DeviceManager.ts
@@ -34,7 +34,8 @@ export class DeviceManager implements Disposable {
     private readonly stateManager: StateManager<DevicesState>,
     private readonly devicesProviders: DevicesProvider[]
   ) {
-    this.loadDevicesIntoState();
+    this.syncDeviceCacheWithState();
+    this.loadDevices();
     this.loadInstalledImages();
 
     this.disposables.push(this.stateManager);
@@ -52,25 +53,19 @@ export class DeviceManager implements Disposable {
 
   private loadDevicesPromise: Promise<void> | undefined;
 
-  public async loadDevices(forceReload = false) {
+  private async loadDevices(forceReload = false) {
     if (forceReload) {
       // Clear the cache when force reload is requested
       extensionContext.globalState.update(DEVICE_LIST_CACHE_KEY, undefined);
     }
     if (!this.loadDevicesPromise || forceReload) {
-      this.loadDevicesPromise = this.loadDevicesInternal().then(() => {
-        const devices = this.stateManager.getState().devicesByType;
-        this.loadDevicesPromise = undefined;
-        // NOTE: only remember virtual devices,
-        // since available physical devices are likely to change
-        extensionContext.globalState.update(DEVICE_LIST_CACHE_KEY, devices);
-      });
+      this.loadDevicesPromise = Promise.all(this.devicesProviders.map((p) => p.listDevices())).then(
+        () => {
+          this.loadDevicesPromise = undefined;
+        }
+      );
     }
     await this.loadDevicesPromise;
-  }
-
-  private async loadDevicesInternal() {
-    await Promise.all(this.devicesProviders.map((p) => p.listDevices()));
   }
 
   private async listInstalledAndroidImages() {
@@ -81,18 +76,21 @@ export class DeviceManager implements Disposable {
     return getAvailableIosRuntimes();
   }
 
-  private async loadDevicesIntoState() {
+  private async syncDeviceCacheWithState() {
     const devicesByType = extensionContext.globalState.get<DevicesByType>(DEVICE_LIST_CACHE_KEY);
     if (devicesByType) {
-      // we still want to perform load here in case anything changes, just won't wait for it
+      // initially store the cached devices into the state manager
       this.stateManager.updateState({ devicesByType });
-      this.loadDevices();
-    } else {
-      await this.loadDevices();
     }
-    // after the initial load, we want to keep the cache updated
-    this.stateManager.onSetState((newState) => {
-      extensionContext.globalState.update(DEVICE_LIST_CACHE_KEY, newState.devicesByType);
+    // subscribe to state changes and update the cache accordingly
+    this.stateManager.onSetState(() => {
+      const { iosSimulators, androidEmulators } = this.stateManager.getState().devicesByType;
+      // NOTE: we only cache the emulated devices, since we expect the physical devices to change more frequently
+      // between session launches
+      extensionContext.globalState.update(DEVICE_LIST_CACHE_KEY, {
+        iosSimulators,
+        androidEmulators,
+      });
     });
   }
 


jwajgelt:
> the system version is not detected on samsung tab s 8
> Screenshot 2025-10-24 at 11 26 13

It's not detected for physical devices, at least for now.
If we decide to include that information (I'm not sure how useful it really is, but we might as well) it would a part of a later PR which also prettifies the device names.

packages/vscode-extension/src/builders/expoGo.ts
```diff
@@ -71,7 +71,12 @@ export function fetchExpoLaunchDeeplink(
       (res) => {
         if (res.statusCode === 307) {
           // we want to retrieve redirect location
-          resolve(res.headers.location);
+          const location = new URL(res.headers.location!);
+          // NOTE: for physical Android devices, the address for the host machine is different
```
filip131311, line 75: what is an adres we get in redirect? 

if it is different then on emulator because it is for remote device it should work 
and if it is the same we get for emulators it should work on remote devices for the same reason `localhost` works should it not?

packages/vscode-extension/src/webview/hooks/useFatalErrorAlert.tsx
```diff
@@ -177,6 +179,25 @@ function createMetroErrorAlert(metroErrorDescriptor: MetroErrorDescriptor) {
   };
 }
 
+function createPreviewErrorAlert(previewErrorDescriptor: PreviewErrorDescriptor) {
```
filip131311, line 182: Lets move it out of the webview code in a fallow up PR I know it is how it was done before but with growing number of possible errors it starts to bother me that the description mappings are berried in code like that, we should have them is a separate file and create coupled `comonne` utility "getErrorDescription" maybe we could also define possible actions in a file like that?

packages/vscode-extension/src/webview/hooks/useResizableProps.tsx
```diff
@@ -24,7 +24,7 @@ export function useResizableProps({
   setZoomLevel,
   device,
 }: UseResizableProps) {
-  const frame = useDeviceFrame(device);
+  const frame = useDeviceFrame(device) ?? { height: 800, width: 400 };
```
filip131311: this seems arbitrary? can you move the default to a constant and if there is an explanation please include it as a comment

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -0,0 +1,185 @@
...
+      // orientation is set to PortraitUpsideDown. We rotate the stream incorrectly in one of these cases,
+      // which _also_ breaks touch input.
+      Logger.warn("PortraitUpsideDown rotation is not supported on physical Android devices.");
+      return;
+    }
+    super.sendRotate(rotation);
+  }
+
+  protected override mirrorNativeLogs(_build: AndroidBuildResult): void {
+    // TODO:
```
filip131311, line 63: is it lef out on perpouse in this PR?

packages/vscode-extension/src/devices/AndroidDevice.ts
```diff
@@ -0,0 +1,344 @@
...
+    await this.terminateApp(build.packageName);
+
+    this.mirrorNativeLogs(build);
+
+    const deepLinkChoice =
+      build.packageName === EXPO_GO_PACKAGE_NAME ? "expo-go" : "expo-dev-client";
+    const expoDeeplink = await fetchExpoLaunchDeeplink(metroPort, "android", deepLinkChoice);
+    if (expoDeeplink) {
+      await this.configureExpoDevMenu(build.packageName);
+      await this.launchWithExpoDeeplink(metroPort, devtoolsPort, expoDeeplink);
```
filip131311, line 142: haven't @kmagiera  change it depending on a RN veriosn? it might have been missed since you branched out a long time ago

packages/vscode-extension/src/devices/AndroidDevice.ts
```diff
@@ -0,0 +1,344 @@
...
+    }
+  }
+
+  public async terminateApp(packageName: string) {
+    await exec(ADB_PATH, ["-s", this.serial!, "shell", "am", "force-stop", packageName]);
+  }
+
+  public async sendDeepLink(link: string, build: BuildResult) {
+    if (build.platform !== DevicePlatform.Android) {
+      throw new Error("Invalid platform");
```
filip131311, line 155: Dont we have invalid platform Error type?

packages/vscode-extension/src/devices/AndroidDevice.ts
```diff
@@ -0,0 +1,344 @@
...
+      "-a",
+      "android.intent.action.VIEW",
+      "-d",
+      link,
+      build.packageName,
+    ]);
+  }
+
+  public setUpKeyboard() {
+    // Keyboard setup is not required on Android Emulator devices.
```
filip131311: outdated comment now that we are in another abstraction above emulator devices

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -417,125 +394,20 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
-      this.serial!,
-      "shell",
-      "monkey",
-      "-p",
-      build.packageName,
-      "-c",
-      "android.intent.category.LAUNCHER",
-      "1",
-    ]);
+  async sendBiometricAuthorization(_isMatch: boolean) {
```
filip131311: this is already in the same form in `androidDevice` I think we can remove it here

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -417,125 +394,20 @@ export class AndroidEmulatorDevice extends DeviceBase {
...
-        this.serial!,
-        "shell",
-        `run-as ${packageName} sh -c 'mkdir -p /data/data/${packageName}/shared_prefs && cat > /data/data/${packageName}/shared_prefs/${packageName}_preferences.xml'`,
-      ],
-      {
-        // pass serialized prefs as input:
-        input: prefsXML,
-      }
-    );
+  makePreview(): Preview {
```
filip131311: some time ago we decided to always write `public` explicitly we can use this oportunity to fix it up here as well

packages/vscode-extension/src/devices/AndroidDevice.ts
```diff
@@ -0,0 +1,344 @@
...
+          1000
+        );
+      } catch (e) {
+        Logger.error("Error while uninstalling will be ignored", e);
+      }
+    };
+
+    // adb install sometimes fails because we call it too early after the device is initialized.
+    // we haven't found a better way to test if device is ready and already wait for boot_completed
+    // flag in waitForEmulatorOnline. But even after that even is delivered, adb install also sometimes
```
filip131311, line 81: i don't think this comment makes sense now

packages/vscode-extension/src/builders/expoGo.ts
jwajgelt: running locally I get
```
exp://192.168.92.80:8081
```
which is probably the address in the private LAN network. It's likely the issue was caused by the tablet I tested on not being connected to the same network as the development machine. Either way, `localhost` seems safer to me -- it should be correct in every case where the `adb` proxy is running, and it it's not, something else is likely to break anyway.

packages/vscode-extension/src/devices/AndroidDevice.ts
jwajgelt: It's an "`InstallationError`" reason, here it was thrown like this. I wouldn't touch it here.

packages/vscode-extension/src/webview/hooks/useResizableProps.tsx
jwajgelt: This was left out by mistake.

packages/vscode-extension/src/project/ide.ts
```diff
@@ -46,9 +49,25 @@ export class IDE implements Disposable {
...
+      ),
+      new AndroidEmulatorProvider(
+        devicesStateManager.getDerived("devicesByType"),
+        this.outputChannelRegistry
+      ),
+      new PhysicalAndroidDeviceProvider(
+        devicesStateManager.getDerived("devicesByType"),
+        this.outputChannelRegistry
+      ),
+    ];
```
filip131311: why is it exposed like that and not hidden in devicesManager?

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -0,0 +1,185 @@
...
+  ) {
+    super(deviceSettings, outputChannelRegistry);
+    this.serial = deviceInfo.id;
+  }
+
+  get lockFilePath(): string {
+    return path.join(extensionContext.extensionPath, `android_device_${this.serial}.lock`);
+  }
+  async bootDevice(): Promise<void> {
+    // NOOP
```
filip131311, line 37: whe should log something here this is a bad situation in which bootDevice is called on phisical device

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -0,0 +1,185 @@
...
+    public readonly deviceInfo: AndroidPhysicalDeviceInfo,
+    deviceSettings: DeviceSettings,
+    outputChannelRegistry: OutputChannelRegistry
+  ) {
+    super(deviceSettings, outputChannelRegistry);
+    this.serial = deviceInfo.id;
+  }
+
+  get lockFilePath(): string {
+    return path.join(extensionContext.extensionPath, `android_device_${this.serial}.lock`);
```
filip131311: I dont like keeping it in the main extension dir, we have cashes dir with android path for keeping info about devices

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+  async bootDevice(): Promise<void> {
+    // NOOP
+  }
+  protected changeSettings(_settings: DeviceSettings): Promise<boolean> {
+    return Promise.resolve(false);
+  }
+  public getClipboard(): Promise<string | void> {
+    // TODO:
```
filip131311, line 43: is it left on purpose?

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -0,0 +1,185 @@
...
+    return new Preview(["android_device", "--id", this.serial!]);
+  }
+
+  public override sendRotate(rotation: DeviceRotation): void {
+    if (rotation === DeviceRotation.PortraitUpsideDown) {
+      // NOTE: the issue here is that the screen sharing agent does not seem to distinguish between
+      // an application running in PortraitUpsideDown and one running in Portrait mode while the device
+      // orientation is set to PortraitUpsideDown. We rotate the stream incorrectly in one of these cases,
+      // which _also_ breaks touch input.
+      Logger.warn("PortraitUpsideDown rotation is not supported on physical Android devices.");
```
filip131311, line 56: is it true for all devices? not just phones?

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  protected override mirrorNativeLogs(_build: AndroidBuildResult): void {
+    // TODO:
+    return;
+  }
+}
+
+const ADB_ENTRY_REGEX = /^([a-zA-Z0-9\-]+)\s+device\s+((\w+:[\w\-]+\s?)*)$/;
+
+async function getPhysicalSize(
```
filip131311: this requires more context

filip131311: at least rename it to getPhisicalDevicesDimensions but comment would be better

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -0,0 +1,185 @@
...
+          }
+          const props = result[2].split(/\s+/).reduce(
+            (acc, entry) => {
+              const [key, value] = entry.trim().split(":");
+              acc[key] = value;
+              return acc;
+            },
+            {} as Record<string, string>
+          );
+          const physicalSize = await getPhysicalSize(result[1]);
```
filip131311: please create 
const serial = result[1] 

it is very difficult to parse otherwise

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -0,0 +1,185 @@
...
+export class PhysicalAndroidDeviceProvider
+  implements DevicesProvider<AndroidPhysicalDeviceInfo>, Disposable
+{
+  private intervalId: NodeJS.Timeout;
+
+  constructor(
+    private stateManager: StateManager<DevicesByType>,
+    private outputChannelRegistry: OutputChannelRegistry
+  ) {
+    this.intervalId = setInterval(() => {
```
filip131311: very soft suggestion: make a disposables array in this class and add clearInterval there

soft suggestion: can we extract the scanning logic to some abstraction and store it as a disposable dependency of something more top level like `DevicesManager` we could make it a comone thing for all devices types as in the future iosRemotes will have the same problem and Emulators can get out of date as well. 


hard suggestion (prefered): We should have Scanners/watchers Registration  Class in IDE, that holds all of them as disposable objects to control for duplication and conflicts between them.

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -52,95 +41,36 @@ export class DeviceManager implements Disposable {
...
-        device.dispose();
+    for (const provider of this.devicesProviders) {
+      const maybeDevice = await provider.acquireDevice(deviceInfo, deviceSettings);
+      if (maybeDevice !== undefined) {
+        return maybeDevice;
       }
     }
-
-    throw new DeviceAlreadyUsedError();
+    throw new Error(`Device ${deviceInfo.displayName} is not available`);
```
filip131311, line 51: why  not deviceAlreadyUsed?

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -52,95 +41,36 @@ export class DeviceManager implements Disposable {
...
-      | undefined;
+  public async loadDevices(forceReload = false) {
     if (forceReload) {
       // Clear the cache when force reload is requested
       extensionContext.globalState.update(DEVICE_LIST_CACHE_KEY, undefined);
     }
     if (!this.loadDevicesPromise || forceReload) {
-      this.loadDevicesPromise = this.loadDevicesInternal().then((devices) => {
+      this.loadDevicesPromise = this.loadDevicesInternal().then(() => {
+        const devices = this.stateManager.getState().devicesByType;
```
filip131311: this relies on a fact that loadDevicesInternal has updated the state, but we have no way of knowing it here (expect reading the underling method, can't we stick to returning devices to make it more understandable?

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -52,95 +41,36 @@ export class DeviceManager implements Disposable {
...
-    const simulators = shouldLoadSimulators
-      ? listSimulators(SimulatorDeviceSet.RN_IDE).catch((e) => {
-          Logger.error("Error fetching simulators", e);
-          return [];
-        })
-      : Promise.resolve([]);
-    const [androidDevices, iosDevices] = await Promise.all([emulators, simulators]);
-    const devices = [...androidDevices, ...iosDevices];
-    return devices;
+    await Promise.all(this.devicesProviders.map((p) => p.listDevices()));
```
filip131311: ad my previous comment: 
we could return lists from individual `loadDevices`

filip131311: otherwise we could rename all methods to something like `loadDevicesIntoState` but I prefer returning

packages/vscode-extension/src/devices/DeviceManager.ts
```diff
@@ -152,16 +82,18 @@ export class DeviceManager implements Disposable {
   }
 
   private async loadDevicesIntoState() {
```
filip131311: I see now that method named like that exists which makes it even more confusing, because if i have a method `loadDevicesIntoState` and `loadDevices` I would assume that only one loads in to state

packages/vscode-extension/src/builders/expoGo.ts
filip131311: can we ping the local network adres first and only if the ping fails change the host?

packages/vscode-extension/src/devices/AndroidDevice.ts
filip131311: 👍

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -949,3 +956,68 @@ async function ensureUniqueDisplayNames(devices: IOSDeviceInfo[]) {
...
+      simulatorInfo,
+      this.outputChannelRegistry
+    );
+
+    if (await device.acquire()) {
+      return device;
+    }
+
+    device.dispose();
+    throw new DeviceAlreadyUsedError();
```
filip131311, line 1017: we throw it in a bunch of places but I can't seem to find a catch

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
jwajgelt: The TBD pieces are in the description.

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
jwajgelt: I'd say it's fine? This method is called in the normal flow of `DeviceSession`, and since the device _must_ already be booted to be available here, I'd say making it a noop is ok -- I'd say it's better than digging up `DeviceSession` right now...

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
jwajgelt: It's broken on both devices I tried, a phone and a tablet.

packages/vscode-extension/src/devices/AndroidDevice.ts
jwajgelt: How so?

packages/vscode-extension/src/devices/DeviceManager.ts
jwajgelt: It's thrown in `Provider`s instead. This is a case where none of the `Provider` handles the device.

packages/vscode-extension/src/devices/DeviceManager.ts
jwajgelt: Then you don't have any guarantee that the returned devices are the same as in the state...

> this relies on a fact that loadDevicesInternal has updated the state

We (or rather I) want to keep the devices in the state. There are components which update the state (`Provider`s), and components which read the state (`DeviceManager`). If the `Provider` communicates the devices available by setting the state, it's normal that you expect to retrieve them from it.

> we have no way of knowing it here (expect reading the underling method

`DeviceManager` consumes the state. It should expect it to be set.

packages/vscode-extension/src/devices/DeviceManager.ts
jwajgelt: Yeah, I didn't change it here, though I agree the name is confusing.
This is the place where we load the cached devices, store them in the state, and initialize loading through the providers.
...which doesn't work that well with the fact, that I want `Provider`s to be the one setting the state, and `DeviceManager` to only consume it -- but I'm also not the fan of doing it in `Provider`s either, since now we can just grab all of the devices from cache and set the state just once, but providers would have to handle them one device type at a time.
I feel like (naming aside), it's ok as is.

packages/vscode-extension/src/devices/DeviceManager.ts
jwajgelt: The reason for this change is that some `Provider`s (`PhysicalAndroid` in our case) may want to update the list without having `load` called on them. Then, you have two ways `Provider`s give you the devices -- either in the result of `loadDevices` or in the state -- which is bound to cause issues. This way, you just have a single way to retrieve them.

packages/vscode-extension/src/devices/AndroidDevice.ts
filip131311: we don't always wait for the emulator online now, as this code is not exclusive to emulatores

packages/vscode-extension/src/devices/DeviceManager.ts
filip131311: okay I actually only have the problem with potentially inconsistant information in   `extensionContext.globalState.update(DEVICE_LIST_CACHE_KEY, devices); `

can you just move this line to `onSteteUpdated` listener?

packages/vscode-extension/src/devices/AndroidDevice.ts
jwajgelt: Yes, but this comment in whole only applies to emulators, as does the retry mechanism. 
On physical androids it succeeds on the first try, since we don't have the problem of the device still starting up there, but the comment itself is still relevant to the code next to it.

packages/vscode-extension/src/builders/expoGo.ts
jwajgelt: No.

jwajgelt: If you use the wrong address, the failure is on the device, with the Expo Go/Dev Client app failing to fetch the bundle with a network error.
I'm not sure _how_ we could even access the error on the device, but even if we did, I don't see the point when simply using `localhost` is more-or-less _guaranteed_ to work here, since the port is forwarded to the development machine.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
jwajgelt: It was previously thrown in `DeviceManager.acquireDevice`, where this method is now called -- so it should work the same as before.

filip131311 (approved):
thank you <3

PR #1696, opened 2025-10-28
feat: network inspector - preserve logs between session states

pFornagiel:
### Description

This PR introduces logs preservation between session states - for given application session, as long as the js debugger remains connected and session is not reset, the network logs should remain intact, unlike previous behaviour, where they would simply disappear.

The above was achieved by introducing following changes:
- in `lib/network` modified `AsyncBoundedResponseBuffer` to support enable/disable buffering, removed "one-time-access" restriction - responses now persist in buffer until cleared and changed the initialisation to be single-per-app. `Note: AsyncBoundedResponseBuffer.test.ts` is known to be broken, to be fixed in next PRs`
- Introduced `IDE.getSessionData` IDE method message, which is used in Network Inspector webview to synchronise network messages history and tracking status. Now, frontend (webview) requests session data on mount to restore previous state and backend stores network messages and tracking state, providing them on request, with bidirectional synchronisation.
- Added `suspend()` method to all plugins, which gets called by applicationSession on session deactivation, to differentiate between need to suspend and get rid of network inspector state. The implementation of `suspend` in all plugins, expect for network inspector, simply calls deactivate, as there was no need to change the logic in other places for now.
- On app restart (appReady event), network plugin new clears the stored network messages in order to prevent use from getting undefined network responses (as the response buffer resets on application js reloads)

### How Has This Been Tested: 

- *open network inspector, check if plugins work correctly, check if network logs are properly preserved between, for example, device switches or other events causing change of applicationSession. Check if the network requests are being recorded before the panel has been opened.*

Not applicable.


packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.ts
```diff
@@ -164,6 +157,31 @@ export class AsyncBoundedResponseBuffer {
...
+  /**
+   * Enable response buffering. When enabled, new responses will be stored in the buffer.
+   */
+  public enableBuffering(): void {
+    this.enabled = true;
+  }
+
+  /**
+   * Disable response buffering. When disabled, new responses will not be stored in the buffer.
+   * Existing buffered responses remain accessible until retrieved or cleared.
```
jwajgelt:
```suggestion
   * Existing buffered responses remain accessible until cleared.
```
From what I can see, we don't seem to remove entries on retrieval anywhere. Should we?

packages/vscode-extension/src/network/hooks/useNetworkTracker.tsx
```diff
@@ -241,13 +288,25 @@ const useNetworkTracker = (): NetworkTracker => {
...
+    setIsTracking((prev) => {
+      sendNetworkTrackingUpdate(!prev);
+      return !prev;
+    });
+  };
+
+  const setTracking = (shouldTrack: boolean) => {
+    setIsTracking(shouldTrack);
+    sendNetworkTrackingUpdate(shouldTrack);
+  };
```
jwajgelt: NIT:
Wouldn't you get the same effect if you used a `useEffect`?
```ts
useEffect(() => {
  sendNetworkTrackingUpdate(isTracking);
}, [isTracking])
```
Then you also don't need to duplicate the logic between `toggleTracking` and `setTracking`

packages/vscode-extension/src/network/utils/panelMessages.ts
```diff
@@ -0,0 +1,44 @@
+import { IDEMethod, WebviewMessage } from "../types/panelMessageProtocol";
+
+export function generateId(): string {
+  return Math.random().toString(36).substring(7);
+}
```
jwajgelt, line 7: I know we did it like this already, but: why do we use random `id`s rather than sequential?
```suggestion
let nextId = 0;
export function generateId(): string {
  return `${nextId++}`;
}
```

packages/vscode-extension/src/project/tools.ts
```diff
@@ -42,6 +42,7 @@ export interface ToolPlugin extends Disposable {
   pluginUnavailableTooltip?: string;
   activate(): void;
   deactivate(): void;
+  suspend(): void;
```
jwajgelt: Currently we have a concept of `activate` and `deactivate` for `ApplicationSession`. Maybe it'd make more sense to keep the same naming here, so:
- change current `activate`/`deactivate` to something like `enable`/`disable` (we already use `enabled` to signify it in the `ToolState` as well)
- instead of `suspend` have `activate`/`deactivate` methods which are called when the `ApplicationSession` is `activate`d or `deactivate`d.

What do you think?

jwajgelt (commented):
Looks fine, no real issues, but I'd like to discuss the few question I left before we merge this.

packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.ts
pFornagiel: Thanks, that was a leftover from the earlier documentation, when the buffer was still one-time-access, which I've overlooked. 

Changed in bfd71957

packages/vscode-extension/src/network/hooks/useNetworkTracker.tsx
pFornagiel: The problem with this approach is that, we do not wish to send the update right after the component mounts, due to issues with synchronisation. For example:
- The tracking is set to `true` by default in BaseNetworkInspector
- The user opens Network Inspector for the first time, it mounts and sends update with `true` to backend (no issue here)
- The user toggles the tracking in the component to `false` it updates the BaseNetworkInspector to false
- The user changes device, then switches back to the original
- Because device was changed, the Network panel has been made unavailable during the transition and gets enabled back once the user switches to the original device
- Because of above, the entire component was dismounted and now, when user opens the panel, mounts again
- Even though, the backend remembers that previous tracking state is `false` the webview sends message with `true` upon mounting, because of `useEffect` and default value provided in `useState`
- the component tries to synchronise its state with BaseNetworkInspector afterwards, but this happens in the next render cycle - BaseNetworkInspector was already set to `true` thus forgetting the state

We could, for example, introduce `firstMountRef` ref to imply whether the we should send update to the BaseNetworkInspector or leave the current logic.

```ts
  useEffect(() => {
    if (!isFirstMountRef.current) {
      sendNetworkTrackingUpdate(isTracking);
    }
    isFirstMountRef.current = false;
  }, [isTracking]);
```

packages/vscode-extension/lib/network/AsyncBoundedResponseBuffer.ts
pFornagiel: Thanks, that was a leftover from the earlier documentation, when the buffer was still one-time-access, which I've overlooked.

Changed in https://github.com/software-mansion/radon-ide/commit/bfd719571eece51fe622397f70ae4691d00c1ad2###

packages/vscode-extension/src/network/utils/panelMessages.ts
pFornagiel: This seems like a good change, as we are not using it in a single place anymore and it is a part of the panel's "legacy code". 

Changed in e4e2e306

packages/vscode-extension/src/project/tools.ts
pFornagiel: Agreed, it is more consistent.

In addition, decided to get rid of `cleanup` method definition in both of the inspectors and replace it with `deactivate(shouldDisableNetworkInspector: boolean = false)` with default value, as it seems to make more sense to reuse it. 

`InspectorBridgeNetworkInspector` example below:

```ts
  /**
   * "Soft" disable by default, deactivates without clearing messages to preserve state across reactivation
   */
  public deactivate(shouldDisableNetworkInspector: boolean = false): void {
    disposeAll(this.devtoolsListeners);
    commands.executeCommand("setContext", `RNIDE.Tool.Network.available`, false);
    if (shouldDisableNetworkInspector) {
      this.sendCDPMessage({ method: NetworkMethod.Disable, params: {} });
      this.clearNetworkMessages();
    }
  }

  public disable(): void {
    this.deactivate(true);
  }
```

Changed in ba2c1875.

packages/vscode-extension/src/network/hooks/useNetworkTracker.tsx
jwajgelt: You could probably get away with passing the initial value in the prop, and setting it as the initial state.
```ts
const [isTracking, setIsTracking] = useState(tracking)
```
I'm not sure how hard it would be to change that.
As it is now, I feel like it's easy to confuse `setTracking` with `setIsTracking` somewhere down the line, call the wrong one, and have a subtle bug hiding in the code base. I'd very much prefer if we could have just one setter, which produces the side effect.
The ref solution should also be fine.

packages/vscode-extension/src/plugins/network/strategies/DebuggerNetworkInspector.ts
```diff
@@ -249,15 +252,8 @@ export default class DebuggerNetworkInspector extends BaseNetworkInspector {
...
-    this.cleanup(true);
-  }
-
-  /**
-   * Suspends without clearing messages to preserve state across reactivation
-   */
-  public suspend(): void {
-    this.cleanup(false);
+  public disable(): void {
+    this.deactivate(true);
```
jwajgelt: Isn't this the only place you pass `true` here? I think it'd be cleaner to have the `disable`-specific login inside the `disable` method instead of bolting an extra parameter to `deactivate`
```suggestion
    this.deactivate();
    this.networkBridge.disableNetworkInspector();
    this.clearNetworkMessages();
```
Same goes for the other `NetworkInspector` impl.

packages/vscode-extension/src/project/tools.ts
```diff
@@ -40,9 +40,9 @@ export interface ToolPlugin extends Disposable {
   persist: boolean;
   pluginAvailable: boolean;
   pluginUnavailableTooltip?: string;
-  activate(): void;
+  enable(): void;
```
jwajgelt, line 43: I think we should have both `activate()` and `enable()` methods for symmetry -- it seems weird to me that you call `deactivate()` to turn it off, but `enable()` to turn it back on.

jwajgelt (approved):
Left a few more suggestions, but other than that, looks very nice.

packages/vscode-extension/src/network/hooks/useNetworkTracker.tsx
pFornagiel: Made changes in bce5f004

Decided to get rid of `setTracking` and `toggleTracking` whatsoever - the only state setter that remains is `setIsTracking` and original `isTracking` state.

Setting the initial state as proposed would be challenging, because we are synchronising the frontend only after we mount the component - using useEffect. Having information about current state on mount would require passing this information during the webview creation, which I wanted to avoid to not over-complicate the information flow.

Instead, created th `isSynchronisedRef` ref, which is toggled to `true` after the synchronisation is performed and acts as a guard to not allow for unwanted state updates.

```ts

const isSynchronisedRef = useRef(false);

// ...

const synchronizeSessionData = async () => {
    const sessionData = await getSessionData();
    const { networkMessages, shouldTrackNetwork } = sessionData || {};

    networkMessages.forEach(updateCDPMessages);

    setIsTracking(shouldTrackNetwork);
    isSynchronisedRef.current = true; // toggle the guard here
  };

// ...

  useEffect(() => {
    if (!isSynchronisedRef.current) {
      return;
    }
    sendNetworkTrackingUpdate(isTracking);
  }, [isTracking]);

  useEffect(() => {
    synchronizeSessionData();
    window.addEventListener("message", handleWindowMessage);

    return () => {
      window.removeEventListener("message", handleWindowMessage);
    };
  }, []);

```

packages/vscode-extension/src/plugins/network/strategies/DebuggerNetworkInspector.ts
pFornagiel: Right, that's cleaner. Introduced in 80a31432

packages/vscode-extension/src/project/tools.ts
pFornagiel: Introduced in 48e0542e.

The ToolPlugin interface now looks as follows:
```ts
export interface ToolPlugin extends Disposable {
  id: ToolKey;
  label: string;
  toolInstalled: boolean;
  persist: boolean;
  pluginAvailable: boolean;
  pluginUnavailableTooltip?: string;
  enable(): void;
  disable(): void;
  deactivate(): void;
  activate(): void;
  openTool?(): void;
}
``` 

Each plugin implements the `activate` method. It was discussed internally, but it is mainly change for consistency reasons, the functionality of `activate` is the same as `enable` method in all places as of today.

PR #1699, opened 2025-11-03
fix: fingerprint not taking platform into account

jwajgelt:
Passes the `platforms` argument to `@expo/fingerprint` when calculating native project changes.
This _should_ prevent changes to Android native builds from triggering "Native dependencies changed" warning from appearing on iOS and vice-versa.

### How Has This Been Tested: 
- run `radon-demo-app` (or another application running expo "54.0.15")
- the "Native Dependencies Changed" warning should not appear on iOS after running the app for some time
- after switching to Android, the warning should appear (due to a bug in `expo-modules-autolinking` changing the fingerprint each time it is generated)

Internal


filip131311 (approved):
let's hope this is enough this time

PR #1704, opened 2025-11-04
fix: network inspector - fix issue with axis not showing the response body

### LINKED ISSUE #1654: Responses not displayed in the Network tab
**Description**

When opening the Network tab in the latest version of Radon IDE, the response data for network requests is not displayed. The requests appear in the list, but their response body/content does not show up as expected.

**Environment**

- Radon IDE version: 1.12.0
- ReactNative version: 0.76.6
- Expo version: 52.0.28
- OS: macOS Sequoia 15.6.1

**Steps to Reproduce:**

	1.	Open Radon IDE (latest version).
	2.	Go to the Network tab.
	3.	Execute any API call.
	4.	Click on the network request to inspect details.

**Expected Result:**

The response for the selected network request should be displayed in the response panel.

**Actual Result:**

No response data is shown — the response panel is empty.

**Additional Notes:**

	•	Occurs consistently across different projects.
	•	Restarting the IDE or clearing cache does not resolve the issue.

pFornagiel:
### Description

This PR fixes an issue with Network Inspector, which caused the requests made by `axios` library to not show the response body. 

This behaviour was caused by `readResponseText function  defined in`lib/network/networkRequestParsers.ts` which is used as part of the response parsing in `AsyncBoundedResponseBuffer` and `network.js`. The problem stemmed from the check:
```ts
if (!xhr || !xhr._cachedResponse) {
    // if response was not accessed, it's not cached and we don't want to read it
    // here to avoid potential side effects
    return undefined;
  }
```

In react-native implementation xhr._cachedResponse is only set when `xhr.responseType` is set to some known defined value (see [implementation](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Network/XMLHttpRequest.js)).

React-native fetch implementation polyfill, the `whatwg-fetch ` explicitly sets that value to either `blob` or `arraybuffer` (see [implementation](https://github.com/JakeChampion/fetch/blob/main/fetch.js)), unlike `axios`, which only sets it when it is defined in the config and is not `json` (see [implementation](https://github.com/axios/axios/blob/v1.x/lib/adapters/xhr.js)).

Because of that, for axios requests, `responseType` was equal to an empty string`\`\`` and thus `xhr._cachedResponse` was equal to `null`. The check in our file resolved the response to `undefined`.

The solution is simply to omit the `xhr._cachedResponse`, which does not seem to cause any issues, but was part of the "legacy" `network.js` code.

### How Has This Been Tested: 

- *check if fetches using axios are now properly handled and responseBody can be seen*

Not applicable.

### DIFF
--- a/packages/vscode-extension/lib/network/networkRequestParsers.ts
+++ b/packages/vscode-extension/lib/network/networkRequestParsers.ts
@@ -223,8 +223,6 @@ async function readResponseText(
 
     const responseType = xhr.responseType;
 
-    console.log("REPONSETYPE", responseType);
-
     if (responseType === "" || responseType === "text") {
       return await parseResponseBodyData(xhr.responseText);
     }


packages/vscode-extension/lib/network/networkRequestParsers.ts
```diff
@@ -215,15 +215,16 @@ async function readResponseText(
...
-    if (!xhr || !xhr._cachedResponse) {
+    if (!xhr) {
       // if response was not accessed, it's not cached and we don't want to read it
       // here to avoid potential side effects
       return undefined;
     }
 
     const responseType = xhr.responseType;
 
+    console.log("REPONSETYPE", responseType);
```
```suggestion
```
remove the log before merging.

pFornagiel: Thanks, done - 75bb590e

PR #1705, opened 2025-11-04
feat: paywall deep integration

filip131311:
This PR connects the work we were doing on the simulator server and customer portal side into one paywall solution, we:
- check the license plan through sim-server binary 
- start the sim server with a token if possible 
-  block the android remote devices if license is insufficient 
- introduce multiple paywall messages depending on if the license is not activated at all or is free 
- show paywall if sim server stops because license was not activated 
- allow for custom title in the above case 
- add a activate license option in settings dropdown menu 

Screenshot 2025-11-05 at 11 14 02
Screenshot 2025-11-05 at 11 15 04

### How Has This Been Tested: 

- run radon with the liked version of sim server it will stop after only 15 s to allow for testing 
- check if the preview shutting down is handled correctly 
- restart radon and try using pro feature 
- do the same for remote device select it from the devices list 
- do this in `PaywallView.tsx` : `  const isLicenseInactive = false //licenseState === LicenseStatus.Inactive;` 
- check if the paywall for free licenses works correctly


packages/vscode-extension/src/common/License.ts
```diff
@@ -59,13 +59,24 @@ export enum LicenseStatus {
...
+}
+
 // #endregion License State
 
 // #region Features
 
 export enum Feature {
   AndroidSmartphoneEmulators = "AndroidSmartphoneEmulators",
   AndroidTabletEmulators = "AndroidTabletEmulators",
+  AndroidRemoteDevice = "AndroidRemoteDevice",
```
jwajgelt: We use "Android Physical Device" everywhere else for this, I'd prefer to keep it consistent.

packages/vscode-extension/src/devices/AndroidDevice.ts
```diff
@@ -150,6 +151,14 @@ export abstract class AndroidDevice extends DeviceBase implements Disposable {
     await exec(ADB_PATH, ["-s", this.serial!, "shell", "am", "force-stop", packageName]);
   }
 
+  protected makePreview(licenseToken?: string): Preview {
+    const args = ["android_device", "--id", this.serial!];
```
jwajgelt: It's `"android"` for emulators and `"android_device"` for physical devices. This is a bug.

packages/vscode-extension/src/devices/AndroidEmulatorDevice.ts
```diff
@@ -394,10 +393,6 @@ export class AndroidEmulatorDevice extends AndroidDevice {
     this.serial = await initPromise;
   }
 
-  public makePreview(): Preview {
-    return new Preview(["android", "--id", this.serial!]);
-  }
```
jwajgelt, line 399: Please revert this change.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -662,14 +662,18 @@ export class IosSimulatorDevice extends DeviceBase {
...
       "ios",
       "--id",
       this.deviceUDID,
       "--device-set",
       getOrCreateDeviceSet(this.deviceUDID),
-    ]);
+    ];
+    if (licenseToken !== undefined) {
+      args.push("-t", licenseToken);
+    }
```
jwajgelt, line 675: You can avoid doing this everywhere with a small refactor of `DeviceBase`:
- don't override `makePreview`  in subclasses
- make a separate `getPreviewArgs` method that you override in the subclasses
- call it in `makePreview` and either append the token here, or have `DeviceBase.getPreviewArgs` return it and call `super()` in the subclasses

packages/vscode-extension/src/project/project.ts
```diff
@@ -279,6 +271,12 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
   @guardFeatureAccess(Feature.IOSTabletSimulators, (deviceInfo: DeviceInfo) => {
     return deviceInfo.deviceType !== DeviceType.Tablet;
   })
+  @guardFeatureAccess(Feature.AndroidRemoteDevice, (deviceInfo: DeviceInfo) => {
+    if (deviceInfo.platform === DevicePlatform.IOS) {
+      return true;
+    }
+    return deviceInfo.emulator !== false;
```
jwajgelt, line 278: aka
```suggestion
    return deviceInfo.emulator;
```

packages/vscode-extension/src/utilities/license.ts
```diff
@@ -25,13 +26,23 @@ export enum ServerResponseStatusCode {
...
 }
 
+export type SimServerLicenseValidationResult =
+  | {
+      status: SimServerLicenseValidationStatus.Success;
+      licensePlan: LicenseStatus;
+    }
+  | {
+      status: Exclude<SimServerLicenseValidationStatus, SimServerLicenseValidationStatus.Success>;
+      licensePlan?: LicenseStatus;
```
jwajgelt: Is this ever set in that case?

packages/vscode-extension/src/webview/components/DeviceSelect.tsx
```diff
@@ -138,14 +138,22 @@ function DeviceSelect() {
...
-  const handleStartOrActivateSessionForAndroidSmartphoneDevice = usePaywalledCallback(
+  const handleStartOrActivateSessionForAndroidSmartphoneEmulator = usePaywalledCallback(
     async (deviceInfo: DeviceInfo) => {
       await project.startOrActivateSessionForDevice(deviceInfo);
     },
     Feature.AndroidSmartphoneEmulators,
     []
   );
 
+  const handleStartOrActivateSessionForAndroidRemoteDevice = usePaywalledCallback(
```
```suggestion
  const handleStartOrActivateSessionForAndroidPhysicalDevice = usePaywalledCallback(
```

packages/vscode-extension/src/project/project.ts
filip131311: i like checking explicitly here if the emulator filled would be missing for some reason it could mean that we added a 4th option like something fro amazon for example that does not include this filed but it would not skip this check if we opt-out the explicit condition

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
filip131311: make preview seems to have no other functionality then doing that I'll just revert to having 3 of them

packages/vscode-extension/src/project/project.ts
jwajgelt: You'd get a type error in that case, that's what typescript's for.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
jwajgelt: Yeah, but you _could_ forget to add the token here, when changing the implementation of one of the `Device` classes. I'd rather the token was added in a single place.

PR #1706, opened 2025-11-05
feat: add apollo client devtools support

jwajgelt:
Integrates the `apollo-client-devtools` extension into a panel inside Radon and connects it to the React Native application.

It uses the same approach outlined in #1698 to reuse the web extension code inside Radon.

image

### How Has This Been Tested: 
- open an app which uses apollo client (e.g. [`expo/dev-plugins` example](https://github.com/expo/dev-plugins/tree/main/apps/example))
- verify the "Apollo Client" tool appears in the tools drop-down
- verify opening the panel connects to the application and lets you view the queries, cached data etc.

 #1715

### DIFF
--- a/packages/vscode-extension/lib/babel_transformer.js
+++ b/packages/vscode-extension/lib/babel_transformer.js
@@ -185,9 +185,9 @@ function transformWrapper({ filename, src, ...rest }) {
       src = `module.exports = require("__RNIDE_lib__/JSXRuntime/react-native-78-79/${jsxRuntimeFileName}");`;
     }
   } else if (
-    isTransforming("node_modules/@apollo/client/index.js") ||
-    isTransforming("node_modules/@apollo/client/apollo-client.cjs") ||
-    isTransforming("node_modules/@apollo/client/main.cjs")
+    isTransforming("node_modules/@apollo/client/index") ||
+    isTransforming("node_modules/@apollo/client/main") ||
+    isTransforming("node_modules/@apollo/client/apollo-client")
   ) {
     src = `require("__RNIDE_lib__/plugins/apollo-client-devtools.js");\n${src}`;
   } else if (


packages/vscode-extension/lib/babel_transformer.js
```diff
@@ -184,6 +184,12 @@ function transformWrapper({ filename, src, ...rest }) {
     ) {
       src = `module.exports = require("__RNIDE_lib__/JSXRuntime/react-native-78-79/${jsxRuntimeFileName}");`;
     }
+  } else if (
+    isTransforming("node_modules/@apollo/client/index.js") ||
```
filip131311: can we use `"node_modules/@apollo/client/index"` instead? like we did for react-query? It might save as from some rundom problems in the future?

packages/vscode-extension/lib/runtime.js
```diff
@@ -101,3 +98,6 @@ AppRegistry.setWrapperComponentProvider = (provider) => {
     return require("__RNIDE_lib__/wrapper.js").createNestedAppWrapper(CustomWrapper);
   });
 };
+
```
filip131311, line 101: does the change have a reason?

filip131311: I intuit that it is best to move it as far as possible, but I wonder if you had any specific issues

packages/vscode-extension/package.json
```diff
@@ -432,9 +432,9 @@
           "icon": "assets/mmkv.svg"
         },
         {
-          "id": "RadonIDEToolExpoDevPluginReduxDevTools",
-          "title": "Redux",
-          "icon": "assets/redux.svg"
+          "id": "RadonIDEToolApolloClient",
```
filip131311, line 435: I don't understand why this PR would remove a tool unrelated to it, is it something left over from the redux PR or is it something important for apollo to work?

packages/vscode-extension/lib/babel_transformer.js
jwajgelt: It's 3 different entry points.

filip131311: I know i ment it as a pattern

filip131311: same for "node_modules/@apollo/client/apollo-client" and "node_modules/@apollo/client/main"

jwajgelt: As in, they compile it into 2 differently-named entry points:
`index.js` for `js`
`main.cjs` for `cjs`
So omitting the extension shouldn't change anything.

jwajgelt: Eh, it shouldn't hurt. I can change it.

packages/vscode-extension/lib/runtime.js
jwajgelt: I first tried installing `apollo-client-devtools` in the same way.
Apollo Client tries to immediately connect to the extension service worker (which in our case runs inside the IDE extension process), which made it run into a problem with `global.__RNIDE_register_dev_plugin` not yet being set up -- so I moved both to the end of this module, to have both set up in the same place in code.
I then removed the `apollo-client-devtools` setup from here, since the behaviour I just described _also_ meant that we would always activate the Radon tool, even if `apollo-client` wasn't used in the app.
So -- this change probably isn't needed, but it still makes sense to me to init the devtools after the globals they potentially use.

packages/vscode-extension/package.json
jwajgelt: It was left-over. I removed the Expo Dev Plugin version of Redux DevTools from the codebase, but not from the package.json

PR #1708, opened 2025-11-05
feat: network inspector - enhance XHR responseType support

pFornagiel:
### Description

This PR fixes introduces changes to `lib/network/networkRequestParsers.ts`, to properly handle response types, which can be set using, for example, `axios` instead of fetch.

So far, the implementation of react native `fetch` api, based on `XMLHttpRequest`,  always set the `responseType` attribute of `XMLHttpRequest`'s response to `blob` if it was supported (see https://github.com/JakeChampion/fetch/blob/main/fetch.js):

From the `fetch` implementation provided above:
```ts
if ('responseType' in xhr) {
  if (support.blob) {
    xhr.responseType = 'blob'
  } else if (
    support.arrayBuffer
  ) {
    xhr.responseType = 'arraybuffer'
  }
}
```

Using `XMLHttpRequest`directly, or using libraries which do so such as `axios`, enables to change this `responseType`, for which handling has been introduced.

### How Has This Been Tested: 

- *use, for example, `axios` or XMLHttpRequest instead of fetch and see whether setting different response types gives desired result and does not end up causing undefined behaviour*

Not applicable.

### DIFF
--- a/packages/vscode-extension/lib/network/networkRequestParsers.ts
+++ b/packages/vscode-extension/lib/network/networkRequestParsers.ts
@@ -288,7 +288,7 @@ async function readResponseText(
       }
     }
 
-    // don't read binary data
+    // fallback
     return undefined;
   } catch (error) {
     return handleReadResponseError(error);


jwajgelt (approved):
Looks good, no notes :+1:

PR #1709, opened 2025-11-05
feat: Launch Maestro tests inside Radon

HexHyperion:
This PR introduces the ability to run Maestro test files (flows) on Radon simulators, either directly from the YAML files, or by selecting the file to run from the IDE panel.

Screenshot 2025-11-05 at 11 58 29
Screenshot 2025-11-05 at 12 00 14

Upon the start of the test, a new output channel with Maestro logs will appear, along with an indicator button on the top bar of the IDE panel. A test runner bundle will be installed on the device, and all test actions will be performed automatically. The button can be clicked to abort the test, and the underlying Maestro process will be killed, allowing to launch another one on that device.

The tests are separate between devices, meaning that multiple tests can be run simultaneously, one on each device.

https://github.com/user-attachments/assets/df59927a-0968-44ed-b813-a07a2cdf8295

The `maestro` command that's being called by Radon uses `adb` and `xcodebuild`/`xcrun` under the hood, which means that:
- physical Android devices introduced in #1688 are fully supported and can be tested without any additional steps
- we have to use shim scripts replacing the XCode commands to point Maestro to our custom iOS device set; the `xcodebuild test-without-building` command doesn't support custom sets, so we omit it and install a prebuilt XCTest runner provided by Maestro through `xcrun simctl --set <path/to/our/set>`.

### How Has This Been Tested: 

I successfully ran numerous test files with different Maestro commands on Expo 52-54 and bare RN 79-82 apps, both on Android and iOS simulators, and, after updating the simulator server, on a physical Android 13 phone.

Processes of tests with failed assertions close properly with exit code `1`, manually aborted processes also close properly with different codes.

Attempts of launching a second test on a device where one is already running or aborting result in no action and a warning message.

Invalid instructions inside test files are handled by Maestro, and the process closes properly.

Switching devices mid-testing doesn't break the running tests.

If the selected device or app are unavailable, Maestro will close after a while.

WIP in #1721

### DIFF
--- a/packages/vscode-extension/src/webview/components/Preview.tsx
+++ b/packages/vscode-extension/src/webview/components/Preview.tsx
@@ -149,7 +149,7 @@ function Preview({
 
   // Maestro usually closes and opens the app during a test, so we prevent
   // the disconnected alert when testing not to confuse the user
-  const maestroTestStatus = use$(selectedDeviceSessionState.maestroTestState);
+  const maestroTestStatus = use$(selectedDeviceSessionState.applicationSession.maestroTestState);
   const isAppDisconnected =
     isRunning &&
     maestroTestStatus === "stopped" &&
--- a/packages/vscode-extension/src/webview/views/PreviewView.tsx
+++ b/packages/vscode-extension/src/webview/views/PreviewView.tsx
@@ -339,7 +339,9 @@ function PreviewView() {
   );
 
   const maestroTestState = use$(() =>
-    isRunning ? (selectedDeviceSessionState.maestroTestState.get() ?? "stopped") : "stopped"
+    isRunning
+      ? (selectedDeviceSessionState.applicationSession.maestroTestState.get() ?? "stopped")
+      : "stopped"
   );
 
   return (


packages/vscode-extension/scripts/shims/xcrun
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+#
+# Wrapper for xcrun that forces a custom device set for simctl commands
+
+CUSTOM_DEVICE_SET="$HOME/Library/Caches/com.swmansion.radon-ide/Devices/iOS"
```
jwajgelt: Can we set this in the extension instead, when launching the `maestro` binary?

packages/vscode-extension/src/devices/AndroidDevice.ts
```diff
@@ -217,6 +222,76 @@ export abstract class AndroidDevice extends DeviceBase implements Disposable {
...
+    const maestroProcess = exec(
+      "maestro",
+      ["--device", this.serial || "device", "test", ...fileNames],
+      { buffer: false, stdin: "ignore" }
+    );
+    this.maestroProcess = maestroProcess;
+
+    lineReader(maestroProcess).onLineRead(this.maestroLogsOutputChannel.appendLine);
+
+    const resultOrError = await maestroProcess.catch((e) => e);
```
jwajgelt: There's an option you can pass to `exec` (I think it's called `reject`) to make it resolve instead of throwing on error.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -706,6 +709,87 @@ export class IosSimulatorDevice extends DeviceBase {
...
     return { canSafelyRemove: true };
   }
+
+  protected async runMaestroTest(fileNames: string[]) {
+    this.maestroLogsOutputChannel.show(true);
+    this.maestroLogsOutputChannel.appendLine("");
+
+    if (fileNames.length === 1) {
+      const fileName = fileNames[0];
+      const isFile = fs.lstatSync(fileName).isFile();
```
jwajgelt: Since we're in `async` context anyway, it would be better to use the `fs.promises` version of the API instead, to avoid blocking JS execution while this runs.

packages/vscode-extension/src/devices/IosSimulatorDevice.ts
```diff
@@ -706,6 +709,87 @@ export class IosSimulatorDevice extends DeviceBase {
     await exec("xcrun", args);
     return { canSafelyRemove: true };
   }
+
+  protected async runMaestroTest(fileNames: string[]) {
```
jwajgelt: Isn't this code basically the same between IOS and Android versions? Can't we reuse it between them?

The only difference I can see at a glance is that here we set up the shims, and for Android we don't -- but I feel like we might as well do it on Android as well (since the shims simply won't be called then), and share this code between both device platforms.

Otherwise, I'd extract the common parts of this function to a separate function / method that takes the `device` argument to be passed to `maestro` and the `env` to be set, and call that with appropriate arguments here and in `AndroidDevice`

packages/vscode-extension/src/devices/DeviceBase.ts
```diff
@@ -244,4 +248,12 @@ export abstract class DeviceBase implements Disposable {
...
     return this.previewStartPromise;
   }
+
+  public async startMaestroTest(fileNames: string[]) {
+    await this.runMaestroTest(fileNames);
+  }
+
+  public async stopMaestroTest() {
+    await this.abortMaestroTest();
+  }
```
jwajgelt: I don't like that we're adding more functionality to this class. It's already pretty bloated (imo), and launching `maestro` doesn't seem like something that the abstraction of a "Device" should be doing anyway.
Instead, I guess I'd like to see a `MaestroTestRunner` (or something) class, which:
- has the `start`/`stopMaestroTest` methods.
- updates the `MaestroTestState` state.
- in the constructor, gets passed the device id and other parameters needed to launch the test
- is probably initialized and managed by `DeviceSession`, since you wouldn't want to run the test on a non-running device (presumably)

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -909,6 +909,30 @@ export class DeviceSession implements Disposable {
     this.inspectorBridge?.sendShowStorybookStoryRequest(componentTitle, storyName);
   }
 
+  public async startMaestroTest(fileNames: string[]) {
+    if (this.stateManager.getState().maestroTestState !== "stopped") {
+      window.showWarningMessage(
+        "A Maestro test is already running on this device. Abort it before starting a new one."
+      );
+      return;
```
jwajgelt: Maybe this should call `stop`, instead of returning?

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
```diff
@@ -0,0 +1,58 @@
...
+    // include valid steps beginning with "-" that are not comments
+    const text = document.getText();
+    if (
+      document.fileName.endsWith("config.yaml") ||
+      document.fileName.endsWith("config.yml") ||
+      !text.includes("appId:")
+    ) {
+      return false;
+    }
+    const splitText = text.split("---");
```
jwajgelt: You could probably look for a _line that starts with_ the three dashes, something like
```suggestion
    const splitText = text.split(/^\s*---/m);
```
to avoid matching on files that have the dashes somewhere in the text.

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    return codeLenses;
+  }
+
+  checkMaestroFile(document: TextDocument): boolean {
+    // To fairly certainly identify a Maestro test file that we can run, it must:
+    // not be named config.yaml or .yml
+    // include "appId" (url on web, but we only do mobile)
+    // include "---" splitting configuration and steps
+    // include valid steps beginning with "-" that are not comments
```
jwajgelt, line 37: I wonder if we'd rather want to parse the yaml file entirely, instead of just matching some substrings.

packages/vscode-extension/src/webview/components/ToolsDropdown.tsx
```diff
@@ -126,6 +126,22 @@ function ToolsDropdown({ children, disabled }: { children: React.ReactNode; disa
...
+              const fileDialogPromise = project.openSelectMaestroFileDialog();
+              fileDialogPromise.then((fileNames) => {
+                if (fileNames) {
+                  project.startMaestroTest(fileNames);
+                }
+              });
+            }}>
+            <span className="codicon codicon-github-action" />
+            Start Maestro test(s)...
+          </DropdownMenu.Item>
```
jwajgelt: I don't like the idea of showing these unconditionally.
Most people don't use / don't care about maestro testing, and this will just clutter their UI with a feature they'll never use.
We can discuss this offline, but I'd prefer if we had some detection step that we run before deciding to show this option to the user.

packages/vscode-extension/src/project/deviceSession.ts
HexHyperion: Hm, I don't know, I figured that it could lead to users accidentally interrupting their tests, some of which can take quite some time to relaunch. Not that it's likely to happen, but (IMO) it's a kind of unclear behavior. But on the other hand, the requirement to always stop the test via the IDE panel button can be irritating as well...

Maybe it can `showWarningMessage` with a choice to `stop` or return?

packages/vscode-extension/src/webview/components/ToolsDropdown.tsx
HexHyperion: How about showing this and the YAML CodeLens only when Maestro is found by the dependency check? It'd make sense that the user could be interested in the integration if they have the tool installed.

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
HexHyperion: That would be ideal, but it seems we'd have to use an external lib for that, no?

[Maestro's flow parser](https://github.com/mobile-dev-inc/Maestro/blob/2c82177b3b30c9a939b3b3fce722747e2d8500c4/maestro-orchestra/src/main/java/maestro/orchestra/yaml/MaestroFlowParser.kt) is huge, and since it doesn't make any sense to replicate it all, checking the files for general YAML correctness wouldn't change much - they still could be invalid flow files.

Obviously there were [suggestions for adding such a "dry run" command](https://github.com/mobile-dev-inc/Maestro/issues/1783) in their repo, but they got either dismissed or forgotten...

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -67,11 +71,13 @@ export class DeviceSession implements Disposable {
   private applicationSession: ApplicationSession | undefined;
   private metro: (MetroSession & Disposable) | undefined;
   private maybeBuildResult: BuildResult | undefined;
+  private previousDeviceInfo: DeviceInfo | undefined;
   private buildManager: BuildManager;
   private cancelToken: CancelToken = new CancelToken();
   private deviceSettingsStateManager: StateManager<DeviceSettings>;
   private frameReporter: FrameReporter;
   private navigationStateManager: StateManager<NavigationState>;
+  private maestroTestRunner: MaestroTestRunner | undefined;
```
jwajgelt:
```suggestion
  private maestroTestRunner: MaestroTestRunner;
```
You set it in the constructor, and I don't see you resetting it to `undefined` anywhere, so it shouldn't be necessary.

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
```diff
@@ -42,7 +53,7 @@ export class MaestroCodeLensProvider implements CodeLensProvider {
     ) {
       return false;
     }
-    const splitText = text.split("---");
+    const splitText = text.split(/^---$/m);
```
jwajgelt:
```suggestion
    const splitText = text.split(/^---.*$/m);
```
While I think you need to start a line with "`---`" for it to work, I've seen examples where a comment is put on the same line:
```
- some 
- preample
---    # i will start the document now
document
```

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
```diff
@@ -6,22 +6,33 @@ import {
...
+    if (!ide) {
+      return [];
+    }
+    try {
+      const state = await ide.getState();
+      const maestroStatus =
+        state.projectState?.applicationContext?.applicationDependencies?.maestro?.status;
+      if (maestroStatus !== "installed") {
+        return [];
+      }
```
jwajgelt: I feel like the codelens is fine to show even when the user doesn't have `maestro` installed, as long as the error in that case is good enough.
The difference between this and the tools drop-down is that in this case you're already looking at a maestro test file, so it's less likely that you _don't_ want the feature to be present.

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
```diff
@@ -6,22 +6,33 @@ import {
...
+    } catch (e) {
+      return [];
+    }
 
     const command: Command = {
       title: "Run Maestro test",
```
jwajgelt: I'm not sure what the convention for these is, but I feel like including "Radon" somewhere in the title might be good for users to know where this comes from.
From what I see, the ones added by the Maestro extension shouldn't get confused with ours
https://github.com/nexlabstudio/maestro-workbench
so maybe it's fine as is.

packages/vscode-extension/src/project/deviceSession.ts
HexHyperion: Right, looks like a leftover from when its initialization was async...

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
HexHyperion: Comments, of course...

And how about `/^---\s*#.*$/m` to prevent everything except them?

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
HexHyperion: Now the error just says "Maestro is not installed." to match the Storybook one, but if that's not enough then sure, we can expand it

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
HexHyperion: Sure, looks _way_ better now:

Screenshot 2025-11-19 at 15 40 32

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
jwajgelt: Should be fine.
I haven't read the yaml parser to check how they parse the document separator, and [the reference](https://tutorialreference.com/yaml/yaml-structure) is not very helpful either, so it's possible we're still missing something, but it's probably good enough.

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
jwajgelt: I'd change it to something that prompts the user to install or otherwise troubleshoot the issue, like:
"Failed to run Maestro tests: could not find the executable. Make sure Maestro is installed and accessible in your PATH environment variable."
A link to either maestro installation guide or our own troubleshooting docs would be good as well, but we don't need to go overboard.

packages/vscode-extension/src/providers/MaestroCodeLensProvider.ts
```diff
@@ -0,0 +1,59 @@
...
+    // include valid steps beginning with "-" that are not comments
+    const text = document.getText();
+    if (
+      document.fileName.endsWith("config.yaml") ||
+      document.fileName.endsWith("config.yml") ||
+      !text.includes("appId:")
+    ) {
+      return false;
+    }
+    const splitText = text.split(/^---\s*#.*$/m);
```
jwajgelt:
```suggestion
    const splitText = text.split(/^---(?:\s*#.*)?$/m);
```
Now you accidentally require the `#` comment start.
You can pack the extra stuff into a non-capturing group and make it optional.

PR #1714, opened 2025-11-06
fix: network inspector - fix message cleaning on LogBox trigger

pFornagiel:
### Description

This PR fixes network messages disappearing when LogBox was triggered, because of the `appReady` event generated. The mechanism was introduced to clean the messages on app reload. Instead, now whenever network plugin is initialised on application end, it sends `IDE.clearStoredMessages` message to the extension, to achieve desired effect.

The mechanism has to be in place in order to avoid situation where, because we are now keeping the message information on the extension end, the app reloads and loses information about response bodies (kept in buffer in the app), while the messages are still available to inspect on the panel end.

### How Has This Been Tested: 

- *see if logbox no longer triggers the message clearing, see if the messages are being correctly disposed on application reload*

Not applicable.

### DIFF
--- a/packages/vscode-extension/src/plugins/network/strategies/InspectorBridgeNetworkInspector.ts
+++ b/packages/vscode-extension/src/plugins/network/strategies/InspectorBridgeNetworkInspector.ts
@@ -82,9 +82,12 @@ export default class InspectorBridgeNetworkInspector extends BaseNetworkInspecto
         if (payload.pluginId === "network") {
           try {
             const payloadData = JSON.parse(payload.data);
-
             if (payload.type === WebviewMessageDescriptor.IDEMessage) {
               this.broadcastMessage(payloadData, WebviewCommand.IDECall);
+
+              if (payloadData.method === IDEMethod.ClearStoredMessages) {
+                this.clearNetworkMessages();
+              }
             } else {
               this.broadcastMessage(payloadData, WebviewCommand.CDPCall);
             }


packages/vscode-extension/lib/network/network.js
```diff
@@ -19,6 +19,9 @@ export function setup() {
   const messageBridge = new PluginMessageBridge("network");
   const responseBuffer = new AsyncBoundedResponseBuffer();
 
+  // Clear any stored messages on the extension end on setup
+  messageBridge.sendMessage("ide-message", JSON.stringify({ method: "IDE.clearStoredMessages", params: {} }));
```
jwajgelt, line 23: Doesn't this mean we no longer clear logs on JS reload?

pFornagiel: We do clear messages on JS reload, as it causes the setups on application end to be rerun and broadcasts the `IDE.clearStoredMessages` message to the inspector webview panel, clearing the logs in the panel itself. 

In order to prevent synchronisation issues, also introduced message clearing on the extension end - 1858f9e2

```ts
private setupListeners() {
    // Broadcast network messages from the app to Network Panel Webview
    this.devtoolsListeners.push(
      this.inspectorBridge.onEvent("pluginMessage", (payload) => {
        if (payload.pluginId === "network") {
          try {
            const payloadData = JSON.parse(payload.data);
            if (payload.type === WebviewMessageDescriptor.IDEMessage) {
              this.broadcastMessage(payloadData, WebviewCommand.IDECall);
    
              // clearing the message buffer in the extension
              if (payloadData.method === IDEMethod.ClearStoredMessages) {
                this.clearNetworkMessages();
              }
            } else {
              this.broadcastMessage(payloadData, WebviewCommand.CDPCall);
            }
          } catch (error) {
            console.error("Failed to parse Webview message:", payload.data);
          }
        }
      })
    );
```

pFornagiel:
> can't we determine if app ready comes from log box by using appKey filled?
> 
> ```js
> // wrapper.js
> const appKey = getCurrentScene();
>       inspectorBridge.sendMessage({
>         type: "appReady",
>         data: {
>           appKey,
>           navigationPlugins: navigationPlugins.map((plugin) => plugin.name),
>         },
>       });
> ```

Discussed in the office - we can do that, but it is matter of fixing things beyond the issue with dissapearing network logs, while the implemented fix is more reliable in that case imo. So this is a matter for another fix PR.

jwajgelt (approved):

filip131311 (approved):

PR #1719, opened 2025-11-07
docs: add physical Android devices entry

jwajgelt:
Adds documentation about running on physical Android Devices

### DIFF
--- a/packages/docs/docs/guides/physical-devices.md
+++ b/packages/docs/docs/guides/physical-devices.md
@@ -24,5 +24,6 @@ If you want to debug an application on a device running iOS, an [out-of-tree pla
      To avoid this, make sure the device is unlocked.
    - on newer Android versions, you may be prompted to send the application for verification.
      This will block launching the app until an action is taken.
+     For more information about this, see the [Google Play Protect documentation](https://developers.google.com/android/play-protect/warning-dev-guidance#send_app_for_security_check).
 
    <img height="400" src="/img/docs/ide_physical_play_protect.png" className="shadow-image" />


packages/docs/docs/getting-started/panel-mode.md
```diff
@@ -39,6 +39,11 @@ You will be able to add or remove simulators later using the device menu in the
 In case the IDE cannot locate system images to use for the device, you will see an empty list when creating new emulator or simulator.
 Please follow the [simulators](/docs/guides/simulators) section to learn how to manage system versions of Android emulators or iOS simulators that the IDE can use.
 
+### Run on a physical device
+
+You may also run your application on a [physical Android device](/docs/guides/physical-devices.md)
```
kacperkapusciak, line 44:
```suggestion
You may run your application on a [physical Android device](/docs/guides/physical-devices.md)
```
no one reads documentation from deska to deska

packages/docs/docs/guides/physical-devices.md
```diff
@@ -0,0 +1,28 @@
...
+
+   <img width="700" src="/img/docs/ide_physical_device_select.png" className="shadow-image" />
+
+3. Selecting the device should automatically build the application, install it on your device, and launch it.
+
+   - the application might fail to launch if the device is locked during the launching process,
+     causing the device preview to be stuck on the "Waiting for app to load..." step.
+     To avoid this, make sure the device is unlocked.
+   - on newer Android versions, you may be prompted to send the application for verification.
+     This will block launching the app until an action is taken.
```
kacperkapusciak, line 26: please explain what will happen when i push these buttons (`Send this time`, `Don't send`). for example how long is the verification proces? days? minutes?

jwajgelt: I have no idea.
I'm not sure we should care, I would expect the same thing to happen if you run it from CLI, the difference being that with Radon the user may not expect that they need to do something with their phone, and when running with CLI, everything happens on the phone.

packages/docs/docs/getting-started/panel-mode.md
jwajgelt: Sure, but I feel it still flows better -- even if you land here from a link, it makes sense for a subsection to start like this, and you're more likely to land on the "Running on a physical device" page if you're actively looking for this feature anyway.
I prefer it the way it is.

PR #1720, opened 2025-11-07
fix: Display human-readable names, Android and SDK versions for physical Android devices

HexHyperion:
This small PR fixes the problem with physical devices reporting confusing brand/model names and no Android/SDK versions on the connected devices list.

We fetch (and cache, as the object is _big_) a list of model IDs mapped to readable make/model data, which we then display on the list. Slight variations are accepted, and on conflict the model most similar to the reported name is displayed.

We also get the Android/SDK version by calling `adb shell getprop ...` with the device serial.

Screenshot 2025-11-07 at 12 23 53

### How Has This Been Tested: 

- launch Radon IDE with any project
- connect an Android device and enable USB debugging
- open the device list on the bottom toolbar or the Manage Devices screen
- the connected device now displays with a readable name, Android and SDK version

N/A

### DIFF
--- a/packages/vscode-extension/src/devices/DeviceNameCache.ts
+++ b/packages/vscode-extension/src/devices/DeviceNameCache.ts
@@ -15,7 +15,6 @@ type CacheFile = {
 
 const DEVICE_MODELS_URL = "https://cdn.jsdelivr.net/gh/bsthen/device-models/devices.json";
 const DEVICE_MODELS_FILENAME = "RNIDE_device_models.json";
-const CACHE_DURATION_MS = 5 * 24 * 60 * 60 * 1000;
 
 let fetchModelsPromise: Promise<DeviceModels | null> | null;
 
@@ -59,10 +58,12 @@ async function fetchDeviceModels(): Promise<DeviceModels | null> {
   }
 }
 
-async function getDeviceModels(noRefetchIfPresent?: string): Promise<DeviceModels | null> {
+async function getDeviceModels(refetchIfNotOnList?: string): Promise<DeviceModels | null> {
   const file = await readCacheFile();
-  const now = Date.now();
-  if (file && now - file.savedAt < CACHE_DURATION_MS && file.data[noRefetchIfPresent ?? ""]) {
+  if (
+    file &&
+    (!refetchIfNotOnList || findDeviceModelWithTolerance(refetchIfNotOnList, file.data).length > 0)
+  ) {
     return file.data;
   }
   if (fetchModelsPromise) {
@@ -82,55 +83,15 @@ export async function getClosestDeviceModel(modelId: string): Promise<DeviceMode
   if (!devices) {
     return null;
   }
-  if (devices[modelId]) {
-    return devices[modelId];
+  const matchingModels = findDeviceModelWithTolerance(modelId, devices);
+  if (matchingModels.length === 1) {
+    return matchingModels[0];
   }
-
-  // Some devices report model names with different (or no) characters in between, e.g. "LM-F100" vs "LM_F100" vs "LMF100"
-  const matchingDevices = Object.entries(devices).filter(([key, _]) => {
-    const cleanDeviceId = key.toLowerCase().replace(/[^a-z0-9]/g, "");
-    const cleanPropsModel = modelId.toLowerCase().replace(/[^a-z0-9]/g, "");
-    return cleanDeviceId === cleanPropsModel;
-  });
-  if (matchingDevices.length === 1) {
-    return matchingDevices[0][1];
-  }
-
   // If more than one ID matches, we pick the closest one
-  else if (matchingDevices.length > 1) {
+  else if (matchingModels.length > 1) {
     let bestMatch: DeviceModel | null = null;
     let smallestDistance = Infinity;
-
-    const levenshteinDistance = (a: string, b: string): number => {
-      if (a === b) {
-        return 0;
-      }
-      const m = a.length;
-      const n = b.length;
-      if (m === 0) {
-        return n;
-      }
-      if (n === 0) {
-        return m;
-      }
-
-      let previous = Array.from({ length: n + 1 }, (_, j) => j);
-      let current = new Array<number>(n + 1);
-
-      for (let i = 1; i <= m; i++) {
-        current[0] = i;
-        const ai = a.charAt(i - 1);
-        for (let j = 1; j <= n; j++) {
-          const cost = ai === b.charAt(j - 1) ? 0 : 1;
-          current[j] = Math.min(previous[j] + 1, current[j - 1] + 1, previous[j - 1] + cost);
-        }
-        [previous, current] = [current, previous];
-      }
-
-      return previous[n];
-    };
-
-    for (const [_, device] of matchingDevices) {
+    for (const device of matchingModels) {
       const distance = levenshteinDistance(device.name, modelId);
       if (distance < smallestDistance) {
         smallestDistance = distance;
@@ -141,6 +102,46 @@ export async function getClosestDeviceModel(modelId: string): Promise<DeviceMode
       return bestMatch;
     }
   }
-
   return null;
+}
+
+// Some devices report model names with different (or no) characters in between, e.g. "LM-F100" vs "LM_F100" vs "LMF100"
+function findDeviceModelWithTolerance(modelId: string, deviceModels: DeviceModels): DeviceModel[] {
+  if (deviceModels[modelId]) {
+    return [deviceModels[modelId]];
+  }
+  const cleanId = modelId.toLowerCase().replace(/[^a-z0-9]/g, "");
+  const matchingModels = Object.keys(deviceModels)
+    .filter((key) => {
+      const cleanModel = key.toLowerCase().replace(/[^a-z0-9]/g, "");
+      return cleanModel === cleanId;
+    })
+    .map((key) => deviceModels[key]);
+  return matchingModels;
+}
+
+function levenshteinDistance(a: string, b: string): number {
+  if (a === b) {
+    return 0;
+  }
+  const m = a.length;
+  const n = b.length;
+  if (m === 0) {
+    return n;
+  }
+  if (n === 0) {
+    return m;
+  }
+  let previous = Array.from({ length: n + 1 }, (_, j) => j);
+  let current = new Array<number>(n + 1);
+  for (let i = 1; i <= m; i++) {
+    current[0] = i;
+    const ai = a.charAt(i - 1);
+    for (let j = 1; j <= n; j++) {
+      const cost = ai === b.charAt(j - 1) ? 0 : 1;
+      current[j] = Math.min(previous[j] + 1, current[j - 1] + 1, previous[j - 1] + cost);
+    }
+    [previous, current] = [current, previous];
+  }
+  return previous[n];
 }


packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -0,0 +1,100 @@
+import { Logger } from "../Logger";
+
+type DeviceModel = {
+  brand: string;
+  name: string;
+};
+type DeviceModels = Record<string, DeviceModel>;
+
+const DEVICE_MODELS_URL = "https://cdn.jsdelivr.net/gh/bsthen/device-models/devices.json";
```
jwajgelt, line 16: I'd prefer if we either:
- just shipped this with the extension, as an asset
- fetched this once, and then wrote it to a file, so that we can just read it the next time we need it (with maybe occasional re-fetches to account for updates)

packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -0,0 +1,100 @@
...
+let cachedAt = 0;
+
+export async function getDeviceModels(): Promise<DeviceModels> {
+  const now = Date.now();
+  if (cachedPromise && now - cachedAt < CACHE_DURATION_MS) {
+    return cachedPromise;
+  }
+
+  try {
+    const response = await fetch(DEVICE_MODELS_URL);
```
jwajgelt, line 52: Correct me if I'm wrong, but right now, if I call `getDeviceModels` twice, both will get here, `await` on this `fetch`, and only after one of them succeeds `cachedPromise` will be written (and only subsequent calls will skip the fetch).

packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -0,0 +1,100 @@
...
+
+const DEVICE_MODELS_URL = "https://cdn.jsdelivr.net/gh/bsthen/device-models/devices.json";
+const CACHE_DURATION_MS = 24 * 60 * 60 * 1000;
+
+let cachedPromise: Promise<DeviceModels> | null = null;
+let cachedAt = 0;
+
+export async function getDeviceModels(): Promise<DeviceModels> {
+  const now = Date.now();
+  if (cachedPromise && now - cachedAt < CACHE_DURATION_MS) {
```
jwajgelt: I'd only refetch if the entry is not present in the cached json.
I wouldn't expect the names already on the list to change, only new entries being added as new devices come out, so if you've already got a version which contains your device, there's no need to refetch.

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -115,12 +116,40 @@ export async function listConnectedDevices(): Promise<AndroidPhysicalDeviceInfo[
...
+            "ro.build.version.release",
+          ]).then((res) => res.stdout.trim());
+
+          const apiLevel = await exec(ADB_PATH, [
+            "-s",
+            serial,
+            "shell",
+            "getprop",
+            "ro.build.version.sdk",
+          ]).then((res) => res.stdout.trim());
```
jwajgelt: I'd extract this to a helper function (`getSystemName` or whatever)

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -115,12 +116,40 @@ export async function listConnectedDevices(): Promise<AndroidPhysicalDeviceInfo[
...
+
+          const deviceInfo = await getClosestDeviceModel(props["model"]);
+          const deviceBrand = deviceInfo ? deviceInfo.brand : props["device"];
+          const deviceModel = deviceInfo ? deviceInfo.name : props["model"];
+
+          // To avoid things like "Samsung Samsung Galaxy..."
+          const brandRegex = new RegExp(`^${deviceBrand}$`, "m");
+          const isBrandInModel = brandRegex.test(deviceModel.split(" ")[0]);
+          const deviceName = `${!isBrandInModel ? deviceBrand + " " : ""}${deviceModel}`;
```
jwajgelt: I'd extract this to a helper function `getDeviceDisplayName()` or similar.

packages/vscode-extension/src/devices/DeviceNameCache.ts
jwajgelt: You could also store the object in `extensionContext.globalState` https://code.visualstudio.com/api/references/vscode-api#ExtensionContext if the size doesn't create problems for vscode.

HexHyperion: Yeah, that makes a lot of sense

Looks like the `extensionContext` also has a `.globalStoragePath`/`Uri` made for these exact purposes, so I guess it'd be reasonable to put the cache file there?

packages/vscode-extension/src/devices/DeviceNameCache.ts
HexHyperion: Yes, that would be right... And given that the method is currently being called every second, that could actually happen...

I now got rid of the explicit promise anyway, so that's resolved I guess...?

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -90,6 +91,41 @@ async function getPhysicalScreenDimensions(
...
+}
+
+async function getDisplayName(props: Record<string, string>): Promise<string> {
+  const deviceData = await getClosestDeviceModel(props["model"]);
+  const deviceBrand = deviceData ? deviceData.brand : props["device"];
+  const deviceModel = deviceData ? deviceData.name : props["model"];
+
+  // To avoid things like "Samsung Samsung Galaxy..."
+  const brandRegex = /^${deviceBrand}$/m;
+  const isBrandInModel = brandRegex.test(deviceModel.split(" ")[0]);
```
jwajgelt: Why split at space instead of having the regex detect it?
```suggestion
  const brandRegex = /^${deviceBrand} /;
  const isBrandInModel = brandRegex.test(deviceModel);
```

jwajgelt: Also, this is just
`deviceModel.startsWith(deviceBrand + " ")`

packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -0,0 +1,141 @@
...
+};
+
+const DEVICE_MODELS_URL = "https://cdn.jsdelivr.net/gh/bsthen/device-models/devices.json";
+const DEVICE_MODELS_FILENAME = "RNIDE_device_models.json";
+const CACHE_DURATION_MS = 5 * 24 * 60 * 60 * 1000;
+
+async function readCacheFile(): Promise<CacheFile | null> {
+  try {
+    const filePath = path.join(extensionContext.globalStorageUri.fsPath, DEVICE_MODELS_FILENAME);
+    const fileUri = Uri.file(filePath);
```
jwajgelt: You can use [`Uri.joinPath`](https://code.visualstudio.com/api/references/vscode-api#Uri) instead of extracting the path and creating a new URI.

packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+async function writeCacheFile(data: DeviceModels): Promise<void> {
+  try {
+    const filePath = path.join(extensionContext.globalStorageUri.fsPath, DEVICE_MODELS_FILENAME);
+    const contents = {
+      savedAt: Date.now(),
+      data: data,
+    };
+    await workspace.fs.createDirectory(Uri.file(path.dirname(filePath)));
```
jwajgelt: Surely?...
```suggestion
    await workspace.fs.createDirectory(extensionContext.globalStorageUri);
```

packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -0,0 +1,141 @@
...
+      return null;
+    }
+    Logger.error(`Error reading device models cache file: ${(error as Error).message}`);
+    return null;
+  }
+}
+
+async function writeCacheFile(data: DeviceModels): Promise<void> {
+  try {
+    const filePath = path.join(extensionContext.globalStorageUri.fsPath, DEVICE_MODELS_FILENAME);
```
jwajgelt: Same as in the other function, you can use `Uri.joinPath` instead of extracting the path and creating a new URI.

packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -0,0 +1,141 @@
...
+
+async function fetchDeviceModels(): Promise<DeviceModels> {
+  try {
+    const response = await fetch(DEVICE_MODELS_URL);
+    const json = await response.json();
+    return json;
+  } catch (error) {
+    Logger.error(`Error fetching device models: ${(error as Error).message}`);
+    return {};
+  }
```
jwajgelt, line 58: I'd either `throw` or return `undefined` here, so that we don't write the empty object to cache (and try to refetch instead)

packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+export async function getDeviceModels(noRefetchIfPresent?: string): Promise<DeviceModels> {
```
jwajgelt:
```suggestion
async function getDeviceModels(noRefetchIfPresent?: string): Promise<DeviceModels> {
```

packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -0,0 +1,141 @@
...
+  const file = await readCacheFile();
+  const now = Date.now();
+  if (file && now - file.savedAt < CACHE_DURATION_MS && file.data[noRefetchIfPresent ?? ""]) {
+    return file.data;
+  }
+  const json = await fetchDeviceModels();
+  await writeCacheFile(json);
+  return json;
+}
+
```
jwajgelt:
```suggestion
  if (this.fetchModelsPromise) {
    return await this.fetchModelsPromise;
  }
  this.fetchModelsPromise = fetchDeviceModels();
  const json = await this.fetchModelsPromise;
  await writeCacheFile(json);
  this.fetchModelsPromise = undefined;
  return json;
}

private fetchModelsPromise: Promise<Record<string, string>> | undefined;
```

We should ensure we don't fetch multiple times concurrently.
The previous approach had the problem that the promise was only assigned _after_ the fetch was already over -- if you do it this way, each time this function is called, only one caller can pass the `if` (since if you pass it, you will immediately write the `this.fetchModelsPromise` member -- there's no `await` between the check and assignment -- and the other callers will always succeed on the `if`).

jwajgelt (changes requested):

packages/vscode-extension/src/devices/DeviceNameCache.ts
HexHyperion: Surely...

packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -18,65 +17,71 @@ const DEVICE_MODELS_URL = "https://cdn.jsdelivr.net/gh/bsthen/device-models/devi
 const DEVICE_MODELS_FILENAME = "RNIDE_device_models.json";
 const CACHE_DURATION_MS = 5 * 24 * 60 * 60 * 1000;
```
jwajgelt: I think that since we check if the `modelId` is already in the cached map, I'd get rid of this completely, I don' think it's necessary to ever refetch it in that case.

packages/vscode-extension/src/devices/DeviceNameCache.ts
```diff
@@ -0,0 +1,146 @@
...
+  } catch (error) {
+    Logger.error(`Error fetching device models: ${(error as Error).message}`);
+    return null;
+  }
+}
+
+async function getDeviceModels(noRefetchIfPresent?: string): Promise<DeviceModels | null> {
+  const file = await readCacheFile();
+  const now = Date.now();
+  if (file && now - file.savedAt < CACHE_DURATION_MS && file.data[noRefetchIfPresent ?? ""]) {
```
jwajgelt: I think that rather than looking up `noRefetchIfPresent`, you should use the same logic that you use in `getClosesDeviceModel` of looking for an entry with low Levenshtein distance from the query.

HexHyperion: Ouch, that's right, with that condition it would practically always refetch...

Fixed now.

PR #1732, opened 2025-11-12
feat: Add element references to chatbot through element inspector

latekvo:
This PR adds a way to add element references to the chatbot window by selecting them within the element inspector, and pressing `"Reference in chat"`.

As a side-effect, this PR also alters the look of the `"Show all"` button in the element inspector.

before|after
---|---
image|image

### Notes:

- This feature is disabled when Radon AI is disabled
- This feature is paywalled

### Themes support

Dark Modern|Light Modern|Dark (Legacy)|Light (Legacy)|High contrast dark
---|---|---|---|---
image|image|image|image|image

#### Comparison to VSCode popups:

Dark Modern|Light Modern|Dark|Light|High contrast dark
---|---|---|---|---
image|image|image|image|image

### How Has This Been Tested: 

- Right-press element in simulator
- Press `Reference in chat` in element inspector
- See how chat now has the relevant files attached

https://github.com/user-attachments/assets/44935ba1-7326-4b8b-a1a5-a927017b1105

PR #1754

### DIFF
--- a/packages/vscode-extension/src/webview/utilities/vscode.js
+++ b/packages/vscode-extension/src/webview/utilities/vscode.js
@@ -21,8 +21,7 @@ class VSCodeAPIWrapper {
    * @remarks When running webview code inside a web browser, postMessage will instead
    * log the given message to the console.
    *
-   * @template { {command: string} } TMessage
-   * @param {TMessage} message - Arbitrary data (must be JSON serializable) to send to the extension context.
+   * @param {{command: string} & Record<string, unknown>} message - Arbitrary data (must be JSON serializable) to send to the extension context.
    */
   postMessage(message) {
     if (this.vsCodeApi) {


packages/vscode-extension/src/webview/components/InspectDataMenu.css
```diff
@@ -77,3 +84,63 @@
   justify-content: center;
   flex-wrap: nowrap;
 }
+
+.inspect-data-menu-item.inspector-button {
```
filip131311, line 88: this is hardcoded and does not match the editor theme

packages/vscode-extension/src/ai/mcp/loadAISettingsToState.ts
```diff
@@ -0,0 +1,28 @@
+import { Disposable, workspace } from "vscode";
+import { StateManager } from "../../project/StateManager";
+import { State } from "../../common/State";
+import { isRadonEnabledInSettings } from "./utils";
+
+export default function loadAISettingsToState(stateManager: StateManager<State>): Disposable {
```
filip131311: i don't like it I would much rather you implemented it in `WorkspaceConfigController` where the whole sync with the workspace settings is happening and then derived something like `AISettingsManager` from the workspace configuration to be passed wherever you need it. 

The approach here may cause different versions of workspace configuration management to go out of sync

packages/vscode-extension/src/utilities/workspaceConfiguration.ts
```diff
@@ -16,6 +16,7 @@ const WorkspaceConfigurationKeyMap = {
   general: {
     defaultMultimediaSavingLocation: "general.defaultMultimediaSavingLocation",
     enableExperimentalElementInspector: "general.enableExperimentalElementInspector",
+    enableRadonAI: "general.enableRadonAI",
```
filip131311: this is not a general setting It should be in AI section a specially that this way you can have `DerivedStateManager` for just AI related settings

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
```diff
@@ -80,6 +85,42 @@ export function InspectDataMenu({
...
+
+      const message = {
+        command: "RNIDE_add_to_chat_context",
+        filePaths: [childFilename, parentFilename],
+      };
+
+      vscode.postMessage(message);
+      e.preventDefault(); // prevents the dropdown from closing
+    },
+    Feature.RadonAI,
```
filip131311, line 114: nice <3

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
```diff
@@ -80,6 +85,42 @@ export function InspectDataMenu({
...
+      for (const item of inspectItems) {
+        const filename = item.source.fileName;
+        if (filename !== childFilename) {
+          parentFilename = filename;
+          break;
+        }
+      }
+
+      const message = {
+        command: "RNIDE_add_to_chat_context",
```
filip131311: this is what will happen under the hood but this is an abstraction leek

latekvo: fixed in [eed8b89](https://github.com/software-mansion/radon-ide/pull/1732/commits/eed8b89970b13f9462990e34c5d26ce5254dc31e)

packages/vscode-extension/src/utilities/workspaceConfiguration.ts
latekvo: Done in [9d00a9b](https://github.com/software-mansion/radon-ide/pull/1732/commits/9d00a9b2fd9ab6935db0459c13187f01bc1c89c7)

packages/vscode-extension/src/webview/components/InspectDataMenu.css
latekvo: fixed in [131579e](https://github.com/software-mansion/radon-ide/pull/1732/commits/131579e5836798bae805342b9d395e939cf091ae) (this took me way too long)

Dark Modern|Light Modern|Dark (Legacy)|Light (Legacy)
---|---|---|---
image|image|image|image

### Comparison to VSCode popups:

Dark Modern|Light Modern|Dark|Light
---|---|---|---
image|image|image|image

You may notice that our popup is off-colour. I think it is avoidable, we could be using the same coloring as VSCode does without it blending in with the background.

This is a matter for a separate PR, I will look into it as soon as i have some spare time.
On the screenshot, you see that our popup is 2 tones higher than it's background, while making it one tone higher (same as chat or the inner header) would be enough to make it pop out. 

image

packages/vscode-extension/src/ai/mcp/loadAISettingsToState.ts
latekvo: I removed it altogether in [79a1941](https://github.com/software-mansion/radon-ide/pull/1732/commits/79a1941b6699c839ebe9b03f41ead6de9e6852f1), I only had to fix ([99357ef](https://github.com/software-mansion/radon-ide/pull/1732/commits/99357ef38c6c3e2aaff05397385ad4611b1c21f1)) a typo within `WorkspaceConfigurationKeyMap` for `loadAISettingsToState` to be redundant.

packages/vscode-extension/src/webview/components/InspectDataMenu.tsx
```diff
@@ -80,6 +86,38 @@ export function InspectDataMenu({
...
+
+      for (const item of inspectItems) {
+        const filename = item.source.fileName;
+        if (filename !== childFilename) {
+          parentFilename = filename;
+          break;
+        }
+      }
+
+      project.addToChatContext(childFilename, parentFilename);
```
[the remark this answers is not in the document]
latekvo: Duplicate context addition gets ignored by VSCode.

packages/vscode-extension/src/ai/mcp/RadonMcpController.ts
```diff
@@ -82,6 +82,10 @@ class RadonMcpController implements Disposable {
     this.registerRadonMcpServerProviderInVSCode();
     cleanupOldMcpConfigEntries();
 
+    if (radonAiEnabled) {
+      this.enableServer();
+    }
```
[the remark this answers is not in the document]
latekvo: `radonAiEnabled` is stale in the function you mentioned. `isEnabledInSettings()` has to be called there.

packages/vscode-extension/src/common/Project.ts
```diff
@@ -178,6 +178,8 @@ export interface ProjectInterface {
   startOrActivateSessionForDevice(deviceInfo: DeviceInfo): Promise<void>;
   terminateSession(deviceId: DeviceId): Promise<void>;
 
+  addToChatContext(...filePaths: string[]): void;
```
jwajgelt: NIT: I'd prefer if we passed the array here explicitly, rather than having a viariadic argument
```suggestion
  addToChatContext(filePaths: string[]): void;
```
The way it is now, it's not obvious when reading the code that calls this that the positions of the argumets don't matter.

PR #1733, opened 2025-11-12
feat: Automatically restore the connection with un- and reattached physical devices

HexHyperion:
This PR enhances the behavior of the extension when a device got dis- and reconnected while running an app session. 

Before, when something interrupted the connection, Radon wasn't able to restore it and a full rebuild cycle was required to continue developing. Now, the extension "remembers" the disconnected device, and automatically connects it to Metro, DevTools and relaunches the preview when it becomes available again.

If the app hasn't been closed in the meantime, no refresh is needed to restore full functionality. If anything prevented it from connecting to DevTools, the "app disconnected from Radon" modal appears, allowing the user to relaunch the app.

https://github.com/user-attachments/assets/5a84bae5-99cc-4cbc-83d5-5e3a6f91579f

### How Has This Been Tested: 
- Open any RN app with Radon
- Connect a physical Android device and launch the app on it
- Wait for the app to load, and then disconnect the phone/tablet
- A "device disconnected" modal should appear
- Connect the device again
- The error should disappear, and the device screen should become visible inside Radon
- All IDE functionality should be available after a short while

N/A

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -70,6 +70,7 @@ export class DeviceSession implements Disposable {
   private applicationSession: ApplicationSession | undefined;
   private metro: (MetroSession & Disposable) | undefined;
   private maybeBuildResult: BuildResult | undefined;
+  private previousDeviceInfo: DeviceInfo | undefined;
   private buildManager: BuildManager;
   private cancelToken: CancelToken = new CancelToken();
   private deviceSettingsStateManager: StateManager<DeviceSettings>;
@@ -199,31 +200,10 @@ export class DeviceSession implements Disposable {
 
     // We observe the global devices state to detect when this device
     // gets reconnected (available changes from false to true)
-    const flattenDevices = (devicesByType: DevicesByType) =>
-      Object.keys(devicesByType).flatMap<DeviceInfo>(
-        (k) => (devicesByType?.[k as keyof DevicesByType] ?? []) as DeviceInfo[]
-      );
-    let previousDevices: DeviceInfo[] = flattenDevices(
-      this.devicesStateManager.getState().devicesByType
-    );
+    // as state.deviceInfo isn't updated in that case
+    this.previousDeviceInfo = this.state.deviceInfo;
     this.disposables.push(
-      this.devicesStateManager.onSetState((partialState) => {
-        try {
-          const devicesByType = partialState.devicesByType;
-          if (devicesByType === REMOVE || devicesByType === undefined) {
-            return;
-          }
-          const currentDevices = flattenDevices(devicesByType as DevicesByType);
-          const previousState = previousDevices.find((d) => d.id === this.state.deviceInfo.id);
-          const currentState = currentDevices.find((d) => d.id === this.state.deviceInfo.id);
-          if (previousState?.available === false && currentState?.available === true) {
-            void this.handleDeviceReconnection();
-          }
-          previousDevices = currentDevices;
-        } catch (e) {
-          Logger.warn("Error while handling devices state change", e);
-        }
-      })
+      this.devicesStateManager.onSetState(this.checkDeviceReconnected.bind(this))
     );
   }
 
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -71,6 +74,33 @@ export class AndroidPhysicalDevice extends AndroidDevice {
     // TODO:
     return;
   }
+
+  public notifyDeviceReconnected() {
+    this.deviceReconnectedEventEmitter.fire();
+  }
+
+  public async recoverConnectionAfterReconnect(metroPort: number, devtoolsPort?: number) {
```
jwajgelt: I'd rather we just exposed `async forwardPort(port:number)` method in `DeviceBase` (with iOS version being a noop, and Android version calling `adb reverse`) and in `DeviceSession` called
```ts
device.forwardPort(metroPort);
if (devtoolsPort) device.forwardPort(devtoolsPort);
```

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
```diff
@@ -178,6 +210,21 @@ export class PhysicalAndroidDeviceProvider
...
+      .filter((d) => {
+        const prevDevice = previousDevices.find((pd) => pd.id === d.id);
+        return prevDevice && !prevDevice.available;
+      })
+      .map((d) => d.id);
+
+    for (const deviceId of reconnectedDeviceIds) {
+      const activeDevice = this.activeDevices.get(deviceId);
+      if (activeDevice) {
+        activeDevice.notifyDeviceReconnected();
```
jwajgelt: I don't like this flow.
The way you wrote it, you have:
`DevicesProvider` -(calls)-> `AndroidPhysicalDevice` -(emits event)-> `DeviceSession` -(calls)->`AndroidPhysicalDevice`
The only reason `AndroidPhysicalDevice` is there (and is aware of the whole reconnection mechanism) is so that `DeviceSession` has something to listen on.
Instead, I think it'd be better for the `DeviceSession` to use `DevicesState` -- you can have a state change listener on `DevicesState` that checks if `deviceInfo.available` changed from `false` to `true` (thus, the device reconnected) and trigger the whole flow from there instead.

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
jwajgelt: This (together with the suggestion re. event emitter below) would also allow you to remove the downcasts (`device instanceof AndroidPhysicalDevice`) in `DeviceSession`, meaning if we ever support other kinds of devices, we get the flow "for free" without needing to fiddle with `DeviceSession`

HexHyperion: Okay, that's a lot better. Named it `forwardDevicePort` as `forwardPort` sounded a bit ambiguous (forward _to_ or _from_ the device?), is that OK?

packages/vscode-extension/src/devices/AndroidPhysicalDevice.ts
HexHyperion: How about now? Everything is now handled in `deviceSession`, but passing the whole `devicesStateManager` there feels wrong. Did I miss a better solution?

I tried listening to all variations of `state`, `stateManager` and `deviceInfo` already provided in `deviceSession`, but those don't seem to ever be updated after creating the session.

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -192,6 +196,35 @@ export class DeviceSession implements Disposable {
...
+          const currentDevices = flattenDevices(devicesByType as DevicesByType);
+          const previousState = previousDevices.find((d) => d.id === this.state.deviceInfo.id);
+          const currentState = currentDevices.find((d) => d.id === this.state.deviceInfo.id);
+          if (previousState?.available === false && currentState?.available === true) {
+            void this.handleDeviceReconnection();
+          }
+          previousDevices = currentDevices;
+        } catch (e) {
+          Logger.warn("Error while handling devices state change", e);
+        }
```
jwajgelt: I'd move this to a separate method.

packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -192,6 +196,35 @@ export class DeviceSession implements Disposable {
...
+
+    // We observe the global devices state to detect when this device
+    // gets reconnected (available changes from false to true)
+    const flattenDevices = (devicesByType: DevicesByType) =>
+      Object.keys(devicesByType).flatMap<DeviceInfo>(
+        (k) => (devicesByType?.[k as keyof DevicesByType] ?? []) as DeviceInfo[]
+      );
+    let previousDevices: DeviceInfo[] = flattenDevices(
+      this.devicesStateManager.getState().devicesByType
+    );
```
jwajgelt: Why not just store the previous `deviceInfo` for the current device?

jwajgelt:
> Hmm, interesting, as that's a thing I've never had problems with - both on the emulator and device I can completely close the app, open it again from the app list and it connects immediately... Peak of "works on my machine" :/

Completely closing the app is fine, the issue was with moving the App to the background, on the Android Emulator specifically.
But I also misremembered -- the issue was with React Devtools connection specifically, metro and debugger should work fine.
The problem occurred before we changed the React Devtools to communicate through the debugger connection (rather than a separate websocket) though, so it shouldn't be a problem on newer RN versions anymore either -- so it was my bad, it always should've worked.

jwajgelt (approved):

PR #1738, opened 2025-11-14
feat: network inspector - support `react-native-fetch-api` network requests

pFornagiel:
### Description

This PR introduces support for intercepting network requests made using the `react-native-fetch-api` polyfill package in React Native applications.

Previously, the Network Inspector could only intercept requests made via React Native's `XMLHttpRequest` implementation. However, applications using the [`react-native-fetch-api`](https://github.com/react-native-community/fetch) polyfill bypass XHR entirely and use React Native's `Networking` native module directly, making them invisible to the existing `XHRInterceptor`.

This PR adds a `PolyfillFetchInterceptor` that works alongside the existing `XHRInterceptor` mechanics to provide network request coverage.

Changes introduced:
- `PolyfillFetchInterceptor` (`lib/network/interceptors/PolyfillFetchInterceptor.ts`)
   Intercepts the internal lifecycle methods of the `Fetch.js` class from `react-native-fetch-api`, by overwriting its prototype methods to capture network activity. Those overrides should be compliant with all the package versions available. The functionality was tested on most recent [npm package](https://www.npmjs.com/package/react-native-fetch-api) and RN version: `react-native-82`, `react-native-80`, `react-native-73`, `expo-54`. The `PolyfillFetchInterceptor` has been additionally documented in `polyfill_readme.md` file.

- ~~refactored `lib/network/network.js` - as of now the module now is meant to setup the network interceptors, which have been moved to separate files - `XHRNetworkInterceptor` and `PolyfillFetchInterceptor`. XHR interception has been rewritten to be a Typescript file, changes have been made to `networkRequestParsers` to introduce support for new fetch interceptor, common types have been extracted to `lin/network/types.ts`~~
  
  This change will be moved to another PR, as it goes beyond the scope of current one.
  
- Changes to `networkRequestParsers.ts` have been made: 
    - type definitions were moved to `types.ts` file
    - `readResponseText` has been split to  separate functions to handle XHR and Fetch-polyfill data - `getXHRResponseDataPromise` and `getFetchDataPromise`. Functions that process the responses have been extracted from the `readResponseText` as well to account for reusability.
    - functions now, apart from being exported vie `module.exports` are also using the ES6 `export` to account for typescript, but it should be noted, that they still **have to be exported in module.exports as well**
    - made `trimContentType` to properly showcase the Type as mime type to make it more useful and mimic the chrome devtools functionality to some extent
- Created new `lib/polyfills/babel_transform/FetchPolyfill.js` and made changes to `babel_transform.js` to allow for requiring `react-native-fetch-api` without `try {} catch {}` clauses when it is not present in the project. Instead, if the babel does not find the module, it exports `undefined` by default.
- made changes to `useNetworkTracker` and `network.js` connected with time tracking. We used to divide the timestamps by 1000 every time we took one just to later multiply them by 1000 in NetworkTracker. In order to avoid doing so in the new interceptor, this has been changed to just taking the timestamp without dividing everywhere we have done so.

### How Has This Been Tested: 

Features were tested in local development environment of extension, the following steps have been performed:
- gotten the [fetch-api package](https://www.npmjs.com/package/react-native-fetch-api) and [polyfills-globals package](https://www.npmjs.com/package/react-native-polyfill-globals) by running `npm install` in correct application folder:
  ```bash
  npm i react-native-polyfill-globals react-native-fetch-api
  ```
- added the following code to chosen application to enable the fetch-polypill (in upper-most app section, or wherever used):
  ```tsx
  import { ReadableStream as ReadableStreamPolyfill } from 'web-streams-polyfill/dist/ponyfill';
  import { polyfill as polyfillFetch } from 'react-native-polyfill-globals/src/fetch';
  globalThis.ReadableStream = ReadableStreamPolyfill;
  polyfillFetch();
  ```
- tested whether the following network requests can be made:
  - POST request to JSONPlaceholder: `https://jsonplaceholder.typicode.com/posts` - method: POST, headers: Content-Type: application/json, body: {title, body, userId}
  - Fetch with Readable Stream: `https://reddit.com` - reactNative: {textStreaming: true}
  - Fetch HTML response: `https://www.reddit.com`
  - Fetch request visible in network panel: `https://pokeapi.co/api/v2/pokemon/ditto`
  - Fetch XML response: `https://www.w3schools.com/xml/note.xml`
  - Fetch JS response: `https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.30.1/moment.min.js`
  - Fetch CSS response: `https://cdnjs.cloudflare.com/ajax/libs/normalize/8.0.1/normalize.css`
  - Fetch Github Code response: `https://raw.githubusercontent.com/python/cpython/main/setup.py`
  - Fetch 404: `https://raw.githubusercontent.com/simple-icons/simple-icons/develop/icons/github.svg`
  - Fetch with no response body: `https://httpbin.org/status/204`
  - Fetch image: `https://cataas.com/cat`
  - Fetch delayed response: `http://localhost:3000/slow-response` (response delayed by 10 seconds, local server)

  Fetch with streaming enabled was performed like the following:
  ```ts
  const response = await fetch(url,  {reactNative: {textStreaming: true}});
  ```
  All the endpoints resulted in expected behaviour in network inspector.
  The streaming fetch also resulted in proper text streaming with stream consumption (using `getReader` method on response body) without causing errors. This was done in `react-native-82`, `react-native-73`, `expo-54` to confirm that everything works corerctly.
  
  Tested the scenario where user would use axios / manual XMLHttpRequest and scenario where the fetch polyfill is not installed - also did not get any issues.
  
Also tested compatibility with `react-native-73` which does not have `TextEncoder` built in by default (used in the PolyfillInterceptor) to check whether this will cause any issues. The tests were made with and without activating polyfills from the library itself, both successful. 

**NOTE**

Right now, the response body of streamed response is only shown when the request completes - before that, while data is being streamed, the response body will tab will yield `Loading` message. To be discussed whether this is desired behaviour.

Not applicable.

### DIFF
--- a/packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
+++ b/packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
@@ -284,7 +284,6 @@ class PolyfillFetchInterceptor {
 
     if (isIncrementalResponse) {
       const incrementalResponseQueue = this.incrementalResponseQueue.getQueue(requestId);
-
       const responseDataPromise = getIncrementalResponseData(incrementalResponseQueue, response);
       this.responseBuffer.put(requestId, responseDataPromise);
 
@@ -317,8 +316,7 @@ class PolyfillFetchInterceptor {
   private sendLoadingFinished(
     requestId: string,
     response: FetchResponse,
-    fetchInstance: PolyfillFetch,
-    shouldBufferResponse: boolean
+    fetchInstance: PolyfillFetch
   ) {
     if (this.completedRequestTracker.isCompleted(requestId)) {
       return;
@@ -344,9 +342,7 @@ class PolyfillFetchInterceptor {
     });
 
     this.completedRequestTracker.markCompleted(requestId);
-    if (shouldBufferResponse) {
-      this.bufferResponseBody(requestId, response, fetchInstance._nativeResponseType);
-    }
+    this.bufferResponseBody(requestId, response, fetchInstance._nativeResponseType);
   }
 
   private sendLoadingFailed(
@@ -455,7 +451,7 @@ class PolyfillFetchInterceptor {
         .then(() => {
           restoreReleaseLock();
           if (fetchInstance._response) {
-            interceptorInstance.sendLoadingFinished(requestIdStr, _response, fetchInstance, true);
+            interceptorInstance.sendLoadingFinished(requestIdStr, _response, fetchInstance);
           }
         })
         .catch((e: Error) => {
@@ -625,12 +621,7 @@ class PolyfillFetchInterceptor {
           true // shouldClearQueue
         );
       }
-      self.sendLoadingFinished(
-        requestIdStr,
-        _response,
-        this,
-        true // shouldBufferResponse
-      );
+      self.sendLoadingFinished(requestIdStr, _response, this);
     };
   }
 
--- a/packages/vscode-extension/lib/network/networkRequestParsers.ts
+++ b/packages/vscode-extension/lib/network/networkRequestParsers.ts
@@ -467,5 +467,6 @@ module.exports = {
   mimeTypeFromResponseType,
   getContentTypeHeader,
   getFetchResponseDataPromise,
+  getIncrementalResponseData,
   trimContentType,
 };


packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
```diff
@@ -0,0 +1,433 @@
...
+
+  private original__didCreateRequest: DidCreateRequestFn = () => {};
+  private original__didReceiveNetworkResponse: DidReceiveNetworkResponseFn = () => {};
+  private original__didReceiveNetworkIncrementalData: DidReceiveNetworkIncrementalDataFn = () => {};
+  private original__didReceiveNetworkData: DidReceiveNetworkDataFn = () => {};
+  private original__didCompleteNetworkResponse: DidCompleteNetworkResponseFn = () => {};
+  private original__abort: () => void = () => {};
+
+  private checkCompatibility() {
+    return !!Fetch;
```
kmagiera: we should likely check if the internal methods from the prototype that we override exists here too

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
```diff
@@ -0,0 +1,433 @@
...
+import {
+  getFetchResponseDataPromise,
+  deserializeRequestData,
+  trimContentType,
+} from "../networkRequestParsers";
+import type { NetworkProxy, NativeResponseType, BlobLikeResponse } from "../types";
+
+let Fetch: any = undefined;
+try {
+  Fetch = require("react-native-fetch-api/src/Fetch")?.default;
```
kmagiera: historically, we had issues with require calls inside try/catch. They'd cause bundle issues on projects where the package isn't installed. 
If this turns out to be the case here we could use the babel transformer as a way of "injecting" the overrides. 

Maybe @filip131311 remembers how did those require calls break

packages/vscode-extension/lib/network/networkRequestParsers.ts
```diff
@@ -255,52 +328,64 @@ async function readResponseText(
...
+  const responseBody = fetchResponse._body._bodyInit;
+
+  if (nativeResponseType === "blob") {
+    return await processBlobResponse(responseBody, contentTypeAnalysis);
+  }
+
+  // Text-streaming case
+  if (nativeResponseType === "text") {
+    const stream = fetchResponse.body;
+    const typedArray = await drainStream(stream);
```
kmagiera: I wonder whether this can result in some side-effects, in particular we may be reading more data than the client would in this case. Wonder if there's a way for us to only load the data read be the client. Also, it could be the case that the fetch library doesn't allow for data blocks to be read by multiple sources as it'd actually need to keep the whole response in the buffer. 

Ultimately, we also don't have infinite buffer, so in any case we likely only want to keep as much data as our buffers can store.

kmagiera (changes requested):
I left some inline comments. 

It's sad that we don't have a good way of covering fetch-compatible APIs and need to rely on the internals of the react-native-fetch-api, but that's still better than not supporting this use case at all. Good job on investigating the implementation details to distill the methods we need to override.

I have also a couple comments to the PR description:
1) there's a note that XHRNetworkInterceptor was rewritten to typescript but it isn't clear what changes have been made there apart from the rewrite. You mention networkRequestParsers there but this is in a separate file already. We should either be very very strict with documenting changes there (because it isn't easy to compare JS and Typescript file side-by-side) or have it extracted to a separate PR to avoid cluttering this PR that introduces a bunch of new functionality.
2) The test plan doesn't need to contain basic instructions on the repository setup / simulator-server details. The purpose of that section is to list things you've done to verify this change work as opposed to providing a full instructions on how to verify this on your own. There are many details in the first few steps but the important part of the test plan is very vague. I'd expect that you document that you've tested the cases where server returns blobs, images, and text. Then you should also test that the regular projects using standard fetch continue to work as expected. There are perhaps a couple more cases that should be tested and documented in that section given the scope of this changes.

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
pFornagiel: Added the check in cc828c4e.

```ts
private checkCompatibility() {
  if (!Fetch?.prototype) {
    return false;
  }

  const allMethodsExist = [
    Fetch.prototype.__didCreateRequest,
    Fetch.prototype.__didReceiveNetworkResponse,
    Fetch.prototype.__didReceiveNetworkIncrementalData,
    Fetch.prototype.__didReceiveNetworkData,
    Fetch.prototype.__didCompleteNetworkResponse,
    Fetch.prototype.__abort,
  ].every((method) => typeof method === "function");

  return allMethodsExist;
}
```

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
pFornagiel: Discussed with @filip131311, made according changes to the module resolution using `babel_transformer.js`, introduced in 20d74c517bca79834e840990169dd6c3a8d4eb91 and acc61d47b65c7f1f4d5985df562763ddb777499e. 

The module is now resolved from `lib/polyfills/babel_transform/FetchPolyfill.js`, which when transformed by babel, yields `undefined` if the module does not exist in `node_modules`.

```ts
if (isTransforming("/lib/polyfills/babel_transform/FetchPolyfill.js")) {
  const polyfillModule = "react-native-fetch-api/src/Fetch";
  const polyfillPath = doesModuleExist(polyfillModule) ? resolveFromAppDir(polyfillModule) : null;
  src = polyfillPath
    ? `module.exports = require("${polyfillPath}");`
    : `module.exports = undefined;`;
}
```

packages/vscode-extension/lib/network/networkRequestParsers.ts
pFornagiel: This functionality did cause issues. Implemented changes in 63f1759a78380f1efce952942aab9758f61d6ece and 89fbb83d3ad815b60fc477cd337eeec9b0ad91c5, which now handle incremental responses differently, which does not block the user from reading the received data from stream - to be discussed whether we wish to go further with this approach. 

It still does read the data chunks as they are received from the network module - keeping / displaying only the data consumed from stream by the user would require plugging into the streamReader implementation. Additionally - if request is made to the network and we are receiving incremental data, without consuming it, in my view it is desired behaviour to record the data received, not necessarily the one consumed / used, if I understood that remark correctly.

pFornagiel:
In 90642ee24000d72623c76ad06e2aa70b13680ddf and 39e76dc2e01729c339c5515b03a870d95ee0e4b4 added handling of situations when the stream is cancelled or errored before the end of the incoming data.

The details are explained in `polypill_readme.md` and implementation comments but in short - we stub into the `getReader` method of `_stream` fetch property to register listeners to the closure of the stream.

packages/vscode-extension/lib/babel_transformer.js
```diff
@@ -204,6 +206,12 @@ function transformWrapper({ filename, src, ...rest }) {
     const { version } = requireFromAppDir("react-native/package.json");
     const majorMinorVersion = version.split(".").slice(0, 2).join(".");
     src = `module.exports = require("__RNIDE_lib__/rn-internals/rn-internals-${majorMinorVersion}.js");`;
+  } else if (isTransforming("/lib/polyfills/babel_transform/FetchPolyfill.js")) {
```
kmagiera: why can't we scan for imports of `lib/network/interceptors/PolyfillFetchInterceptor` and then replace if with empty module when fetch-api is not found? We wouldn't need to have the placeholder module in this scenario

packages/vscode-extension/lib/polyfills/babel_transform/FetchPolyfill.js
```diff
@@ -0,0 +1,2 @@
+// This is a placeholder file that is modified by babel_transformer.js to
```
kmagiera: I don't like that this lives under babel_transform but also my next comment says why we may not need this at all

packages/vscode-extension/lib/network/networkRequestParsers.ts
```diff
@@ -383,20 +461,33 @@ function deserializeRequestData(data: RequestData, contentType: string | undefin
...
+  }
+  const semicolonIndex = contentType.indexOf(";");
+  if (semicolonIndex !== -1) {
+    return contentType.substring(0, semicolonIndex);
+  }
+  return contentType;
+}
+
+export function getContentTypeHeader(xhr: ExtendedXMLHttpRequest): string {
+  let headerValue =
```
kmagiera, line 454: header names in HTTP standard are case-insensitive so any casing can actually be correct here, not just lower case and default. However, most http stacks would just store them as lowercase regardless of the format they come in, so you likely only need to check that

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
```diff
@@ -0,0 +1,600 @@
...
+      this.responseBuffer.put(requestId, responseDataPromise);
+
+      this.incrementalResponseQueue.clearQueue(requestId);
+    } else {
+      const responseDataPromise = getFetchResponseDataPromise(response, nativeResponseType);
+      this.responseBuffer.put(requestId, responseDataPromise);
+    }
+  }
+
+  private handleDidCreateRequest() {
```
kmagiera: I don't like the "handleSth" naming here. This method doesn't handle the request as the name would suggest but instead it sets up the handler method overwrite

kmagiera (approved):
Good job, looks good overall. left some minor comments inline.

As for the babel transform approach I think that we may want to scope the substitution to some range of the fetch-api library versions since this approach is sensitive to changes to the lib internals.

packages/vscode-extension/lib/network/networkRequestParsers.ts
pFornagiel: The function was introduced to deal with the fact that we were using the inner `xhr._headers` attribute to extract headers in the past, which resulted in them having different casing on android and iPhone, which caused issues with getting the proper header values in the frontend ui. The at the time fix was to introduce this function, which would try both of the cases. 

After examining the [default `fetch` implementation](https://github.com/JakeChampion/fetch/blob/main/fetch.js) of RN, XHR and polyfill used here, the request headers should always be properly normalised to lowercase and the XHR object has the`getReponseHeaders` method defined, that normalises the response headers anyways.

This issue will be tackled in another PR as bug fix, as it needs changes on the frontend as well. Thank you for bringing this up.

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
pFornagiel: Changed the naming to `override__X` convention in 1ebae58c, also added a check in `override__getReader` to make sure the method exists first.

packages/vscode-extension/lib/babel_transformer.js
pFornagiel: Thanks for the proposal, does seem to be cleaner.

Implemented the lookup for `const Fetch` declaration in the `PolyfillFetchInterceptor.ts` in `babel_transformer.js`

```js
if (isTransforming("/lib/network/interceptors/PolyfillFetchInterceptor.ts")) {
    const polyfillModule = "react-native-fetch-api/src/Fetch";
    const polyfillPath = doesModuleExist(polyfillModule) ? resolveFromAppDir(polyfillModule) : null;
    const fetchValue = polyfillPath ? `require("${polyfillPath}")?.default` : "undefined";
    const fetchDeclaration = `const Fetch = ${fetchValue};`;

    src = src
      .split("\n")
      .map((line) => (line.trim().startsWith("const Fetch") ? fetchDeclaration : line))
      .join("\n");
  }
```

See d822a39f

packages/vscode-extension/lib/polyfills/babel_transform/FetchPolyfill.js
pFornagiel: Resolved in comment above. See https://github.com/software-mansion/radon-ide/commit/d822a39f75abe4f0fd92e50358caa899fa049215

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
```diff
@@ -0,0 +1,664 @@
+import { AsyncBoundedResponseBuffer } from "../AsyncBoundedResponseBuffer";
+import {
+  getFetchResponseDataPromise,
+  deserializeRequestData,
+  trimContentType,
+  // getIncrementalFetchResponseDataPromise,
```
jwajgelt:
```suggestion
```

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
```diff
@@ -0,0 +1,664 @@
...
+    fetchInstance: PolyfillFetch,
+    interceptorInstance: PolyfillFetchInterceptor
+  ) {
+    const original__getReader = stream.getReader;
+    if (typeof original__getReader !== "function") {
+      return;
+    }
+
+    // @ts-ignore - Overriding getReader to intercept stream closure
+    stream.getReader = function (this: FetchReadableStream, options?: { mode?: string }) {
```
jwajgelt, line 422: Does this mean we'll never see the request if the application doesn't call `getReader` on the `Response` object?

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
```diff
@@ -0,0 +1,664 @@
...
+
+        interceptorInstance.bufferResponseBody(
+          requestIdStr,
+          _response,
+          fetchInstance._nativeResponseType
+        );
+
+        const mimeType = trimContentType(_response._body._mimeType);
+
+        interceptorInstance.sendCDPMessage("Network.loadingFinished", {
```
jwajgelt: Won't this get fired for each time `getReader` was called? Doesn't this cause any problems?

jwajgelt: We also send this in `didCompleteNetworkResponse`. Is it not called for this case?

packages/vscode-extension/lib/babel_transformer.js
```diff
@@ -204,6 +206,26 @@ function transformWrapper({ filename, src, ...rest }) {
...
+    const isCompatible = moduleExists && compatibleVersions.includes(version);
+
+    const polyfillPath = isCompatible ? resolveFromAppDir(polyfillModule) : null;
+    const fetchValue = polyfillPath ? `require("${polyfillPath}")?.default` : "undefined";
+    const fetchDeclaration = `const Fetch = ${fetchValue};`;
+
+    src = src
+      .split("\n")
+      .map((line) => (line.trim().startsWith("const Fetch") ? fetchDeclaration : line))
+      .join("\n");
```
jwajgelt: NIT:
You should be able to use [`String.replace`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) with a regex (like `/^\s*const Fetch.*$/m`) to find just the line you're looking for and replace it.
It's _fine_ as is, but:
- there's a lot of unnecessary splitting and joining when we're only looking for a single line
- it will actually replace _each_ line starting with `const Fetch`, not just the first one. Maybe it's not a real issue, but I could see it causing problems.

packages/vscode-extension/lib/babel_transformer.js
```diff
@@ -204,6 +206,26 @@ function transformWrapper({ filename, src, ...rest }) {
     const { version } = requireFromAppDir("react-native/package.json");
     const majorMinorVersion = version.split(".").slice(0, 2).join(".");
     src = `module.exports = require("__RNIDE_lib__/rn-internals/rn-internals-${majorMinorVersion}.js");`;
+  } else if (isTransforming("/lib/network/interceptors/PolyfillFetchInterceptor.ts")) {
```
jwajgelt, line 209: Is it _really_ necessary? The `require` should just work in the `PolyfillFetchInterceptor.ts`.
The only justification I can see in the PR description is to avoid `try`/`catch` around the `require`, but I don't feel it justifies the complexity added by having this babel transformer.

jwajgelt (commented):
Left a couple comments and questions.

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
pFornagiel: The incremental reception of the actual request is based on the argument provided to the Networking module function:

```ts
doFetch() {
this.subscribeToNetworkEvents();

    Networking.sendRequest(
        this._request.method,
        this._request.url.toString(), // request tracking name
        this._request.url.toString(),
        this._nativeRequestHeaders,
        this._request._body._bodyInit ?? null,
        this._nativeResponseType,
        this._nativeResponseType === "text", // send incremental events only when response type is text <- thats where the streaming option comes in
        this._nativeRequestTimeout,
        this.__didCreateRequest.bind(this),
        ["include", "same-origin"].includes(this._request.credentials) // with credentials
    );
}

```

[https://github.com/react-native-community/fetch/blob/master/src/Fetch.js#L111-L121](https://github.com/react-native-community/fetch/blob/master/src/Fetch.js#L111-L121)

So, the reception of the response is independent from reading the stream, it actually happens in `__didReceiveIncrementalData` , the `getStream` override is there to handle the situation when the stream gets closed by the user, becuase then - the event listeners in the polyfill get detached. See below:

[https://github.com/react-native-community/fetch/blob/master/src/Fetch.js#L142-L145](https://github.com/react-native-community/fetch/blob/master/src/Fetch.js#L142-L145)

When the listeners get detached, we have to properly send either `loadingFinished` or `loadingFailed` based on whether the stream got closed gracefully or because of an error.

I will have to document that further, as this is not obvious from just reading the code.

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
pFornagiel: When `getReader` gets called, the stream is locked and trying to acquire the reader again by calling `getReader` the second time results in an error - it is not expected behaviour. Because of this error, we do not get to execute the code added by us.

https://github.com/MattiasBuelens/web-streams-polyfill/blob/master/src/lib/readable-stream/default-reader.ts#L105-L116

```ts
constructor(stream: ReadableStream<R>) {
  assertRequiredArgument(stream, 1, 'ReadableStreamDefaultReader');
  assertReadableStream(stream, 'First parameter');

  if (IsReadableStreamLocked(stream)) { // <- this evaluates to true the second time getReader is called
    throw new TypeError('This stream has already been locked for exclusive reading by another reader');
  }

  ReadableStreamReaderGenericInitialize(this, stream);

  this._readRequests = new SimpleQueue();
}
```

Because of that, the need to handle this matter did not occur obvious to me, but it might be worth looking into if someone decided to use different polyfill, which would not cause an error to get throw, although by MDN specification, it should throw.

Given that, even if we do not handle this situation, this needs to be documented.

pFornagiel: Similar to situation above - this override handles the case where the request gets aborted because of stream being cancelled by the user or due to an error. When it happens - all the handlers of network methods get disconnected, thus we need a way to detect when the stream gets closed to inform the Network Inspector that request got finished in this situation.

```ts
// stream and streamController creation in fetchPolyfill code
// callback causes unsubscribing to all network event handlers
const { stream, streamController } = createStream(() => {
    this.__clearNetworkSubscriptions();
    Networking.abortRequest(this._requestId);
});
```

https://github.com/react-native-community/fetch/blob/master/src/Fetch.js#L142-L145

packages/vscode-extension/lib/babel_transformer.js
pFornagiel: Discussed this internally with @filip131311 and @kmagiera - during my testing, the `try ... catch ...` approach was working fine, but we have historically had problems with this approach and opted for resolution through `babel_transform.js` instead. The comment referencing this is here:
https://github.com/software-mansion/radon-ide/pull/1738#discussion_r2539648464

At first, I attempted to create a separate file, which would be transformed by babel based on whether the polyfill was installed or not, later changed this approach due to this comment:
https://github.com/software-mansion/radon-ide/pull/1738#discussion_r2557012623
From those two - this one does not require additional file dependencies, so seemed like a better solution.

We may try and resolve this using the `try ... catch ...` if we wish to reduce the complexity and see whether this works out. As said - did not cause issues on my end and so far, we've also been using this technique when trying to get the device orientation using external libraries if they are present in `lib/orientation/ExpoOrientationStrategy.js`. This was implemented back when I did not know about the issues connected with this form of module resolution and seemed to have not caused issues so far.

The approach is up to be discussed then.

jwajgelt: 1. If it's done with `try`/`catch` elsewhere, I fail to see how handling it in this single case helps. I think I'd rather not overcomplicate it here and implement this `require`-substitution mechanism in a separate PR, with a generic solution that can handle all of the cases where we `require` a potentially non-existent module.
2. If we have to do it this way, I'd prefer that we had some magic constant / string literal that is replaced by `require` call, rather than looking at a specific `const`-declaration in a specific file.
In the code, you'd have
```ts
const Fetch = "__RADON_REQUIRE_FETCH__";
```
And in the transformer you'd look for this string literal and replace it with `require()`/`undefined` as needed.

This approach also generalizes nicely to a solution where in the code you can write:
```ts
const Fetch = "__RADON_REQUIRE_(react-native-fetch-api/src/Fetch)"
```
And in the babel transformer, whenever you're transforming a file under the `/lib/` directory, you apply a regex matching 
```
"__RADON_REQUIRE_\(([correct_character_class]+)\)"
```
 extract the package names from matches, and replace each occurrence with the correct expression.

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
jwajgelt: Is it not possible to dispose of the reader (unlocking the stream), then call this method again?

jwajgelt: I could imagine code like
```ts
useEffect(() => {
  const reader = response.getReader();
  /* ...process stream... */
  return () => reader.releaseLock();
}, [...deps])
```

jwajgelt: Can we have a guard which ensures this CDPMessage is sent only once?
- extract `interceptorInstance.sendCDPMessage("Network.loadingFinished", ...)` to a helper method `sendLoadingFinished(...)`
- keep a flag `finished` in a field
- set it the first time `sendLoadingFinished` is called, and perhaps add a warning if it's called again (<- this is tricky in that we don't want to report this error when the user runs Radon, only when we test it. Perhaps it should send a message to Radon that we only log in DEBUG builds, perhaps we could handle it another way)

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
pFornagiel: Removed in d77829f7, thanks

packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
pFornagiel: Implemented proper `releaseLock` handling in d77829f7 with added.

Extracted sendLoadingFinished and sendLoadingFailed methods into private helper methods and added  `CompletedRequestTracker` which is checked before sending those messages to avoid duplicates.

Added `lockReleased` flag set by overriding `releaseLock` method of the reader, to properly tell apart cases when error occurred and when the reader was released, as both the release and internal errors cause the `_closedPromise` checked to be rejected.

pFornagiel: Implemented in d77829f7, described in the comment under the above `releaseLock` conversation.

packages/vscode-extension/lib/babel_transformer.js
pFornagiel: Discussed internally, changed to the approach with `"__RADON_REQUIRE_FETCH__"` in 2b9494eb.

`babel_transformer.js`:
```js
const fetchDeclaration = polyfillPath ? `require("${polyfillPath}")?.default` : "undefined";
src = src.replace(/"__RADON_REQUIRE_FETCH__"/g, fetchDeclaration);
```

Although the solution with `"__RADON_REQUIRE_(react-native-fetch-api/src/Fetch)"` is really great, I will leave it to be implemented in another PR, when we will need to properly handle more cases like this one or patch the existing `orientationStrategy` `try ... catch ...` approach.

PR #1740, opened 2025-11-14
tests: Adjust for local hosted runner

KeyJayY:
This PR improves and adjusts the test execution flow for a locally hosted runner. It also adds Android to the set of platforms included in the nightly test runs.

### How Has This Been Tested: 

The changes were tested by running the test suites on a locally hosted GitHub runner. Tests after update appear to work more reliably than before, although additional refinements may still be needed in the future.

No documentation changes needed.

### DIFF
--- a/packages/vscode-extension-tester/ui-tests/07-radonTools.test.js
+++ b/packages/vscode-extension-tester/ui-tests/07-radonTools.test.js
@@ -79,9 +79,9 @@ safeDescribe("7 - Radon tools tests", () => {
 
     const position = centerCoordinates(originalPosition);
 
-    const PhoneRect = await phoneScreen.getRect();
-    const phoneWidth = PhoneRect.width;
-    const phoneHeight = PhoneRect.height;
+    const phoneRect = await phoneScreen.getRect();
+    const phoneWidth = phoneRect.width;
+    const phoneHeight = phoneRect.height;
 
     const actions = driver.actions({ bridge: true });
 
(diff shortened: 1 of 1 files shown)


packages/vscode-extension-tester/ui-tests/07-radonTools.test.js
```diff
@@ -64,6 +64,58 @@ describe("7 - Radon tools tests", () => {
...
+    const phoneScreen = await elementHelperService.findAndWaitForElement(
+      By.css(`[data-testid="phone-screen"]`),
+      "Timed out waiting for phone-screen"
+    );
+
+    const position = centerCoordinates(originalPosition);
+
+    const PhoneRect = await phoneScreen.getRect();
+    const phoneWidth = PhoneRect.width;
+    const phoneHeight = PhoneRect.height;
```
jwajgelt:
```suggestion
    const phoneRect = await phoneScreen.getRect();
    const phoneWidth = phoneRect.width;
    const phoneHeight = phoneRect.height;
```

packages/vscode-extension-tester/ui-tests/setupTest.js
```diff
@@ -107,6 +92,26 @@ afterEach(async function () {
...
+      `${this.currentTest.title}-${Date.now()}.png`
+    );
+
+    fs.mkdirSync(screenshotDir, { recursive: true });
+    fs.writeFileSync(filePath, image, "base64");
+    console.log(`Saved screenshot: ${filePath}`);
+    failedTests.push(this.currentTest.fullTitle());
+  }
+
+  await cleanUpAfterTest();
```
jwajgelt, line 114: Is `safeDescribe` necessary if we already cleanup here, in `afterEach`?

KeyJayY: The tests are grouped into test sets called `describes`. Each `describe` block can define its own `before` function, which is executed before any test in that block runs. The problem is that if a `before` hook fails, the `afterEach` hook will not run because no test has actually started.
consider following situation:
- The `before` hook fails for some reason in test set number 6.
- The `afterEach` hook with `cleanUpAfterTest` does not run because no test was executed.
- As a result, all following tests fail because the environment was not cleaned up. 

 To solve this, safeDescribe is a wrapper that adds: 
```
    after(async () => {
      await cleanUpAfterTest();
    })
```
This `after` hook is executed after the entire describe block finishes, even if the `before` hook failed.
Since the test framework I'm using does not provide any `afterEachDescribe` or equivalent hook, I use `safeDescribe` to avoid repeating the same `after` logic in every `describe` block.

PR #1741, opened 2025-11-14
feat: open main activity defined in android manifest if product flavor is specified

filip131311:
This PR supports opening android applications with: 
1) applicationId's modified by build type or product flavor
2) non standard main activities 

The following modifications use a different method of launching such applications in order to prevent to possible issues:
- android not knowing which app with the same name to open 
- dev client navigating to some what random paths right after app is launched 

### How Has This Been Tested: 
- run a expo-dev-client application - it works 
- modify build type in app -> build.gradle and choose it in launch configuration 
- change it again ( you then test " android not knowing which app with the same name to open" ) 
- open bare android application 
- test it in close source set up where the "dev client navigating to some what random paths right after app is launched" was happening

auto config detection does not require user facing documentation

### DIFF
--- a/packages/vscode-extension/src/builders/buildAndroid.ts
+++ b/packages/vscode-extension/src/builders/buildAndroid.ts
@@ -1,3 +1,4 @@
+import assert from "assert";
 import fs from "fs";
 import path from "path";
 import semver from "semver";
@@ -54,15 +55,13 @@ function locateAapt() {
  * This function extracts the application Id that allows for identifying the application
  * while managing its runtime it might be defined by multiple configurations
  * https://developer.android.com/build/configure-app-module
- * @param artifactPath
- * @param cancelToken
- * @returns Promise<string>
  */
 async function extractAppId(artifactPath: string, cancelToken: CancelToken) {
   const aaptPath = locateAapt();
   const { stdout } = await cancelToken.adapt(exec(aaptPath, ["dump", "badging", artifactPath]));
   const packageLine = stdout.split("\n").find((line: string) => line.startsWith("package: name="));
-  const appId = packageLine!.split("'")[1];
+  assert(packageLine, "aapt file always contains the line specifying the package name");
+  const appId = packageLine.split("'")[1];
   return appId;
 }
 
@@ -73,11 +72,10 @@ async function getApplicationManifest(
     const filePath = await AndroidConfig.Paths.getAndroidManifestAsync(projectRoot);
     const androidManifest = await AndroidConfig.Manifest.readAndroidManifestAsync(filePath);
 
-    if (!androidManifest) {
-      return undefined;
+    if (androidManifest) {
+      return androidManifest;
     }
-    return androidManifest;
-  } catch (e) {
+  } catch {
     // ignore errors and just return undefined here
   }
   return undefined;
@@ -96,8 +94,9 @@ async function resolveAppIdFromNative(projectRoot: string): Promise<string | nul
     const androidManifest = await AndroidConfig.Manifest.readAndroidManifestAsync(filePath);
     // Assert MainActivity defined.
     await AndroidConfig.Manifest.getMainActivityOrThrow(androidManifest);
-    if (androidManifest.manifest?.$?.package) {
-      return androidManifest.manifest.$.package;
+    const appId = androidManifest.manifest?.$?.package;
+    if (appId) {
+      return appId;
     }
   } catch (error: any) {
     Logger.debug("Expected error resolving the package name from the AndroidManifest.xml:", error);


packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -46,12 +48,96 @@ function locateAapt() {
...
+      return undefined;
+    }
+    return androidManifest;
+  } catch (e) {
+    // ignore errors and just return undefined here
+  }
+  return undefined;
+}
+
+async function resolveAppIdFromNativeAsync(packageName: string): Promise<string | null> {
```
jwajgelt:
```suggestion
async function resolveAppIdFromNative(packageName: string): Promise<string | null> {
```

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -46,12 +48,96 @@ function locateAapt() {
...
+      return androidManifest.manifest.$.package;
+    }
+  } catch (error: any) {
+    Logger.debug("Expected error resolving the package name from the AndroidManifest.xml:", error);
+  }
+
+  return null;
+}
+
+async function getLaunchActivityAsync(
```
jwajgelt:
```suggestion
async function getLaunchActivity(
```

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -46,12 +48,96 @@ function locateAapt() {
...
+  // in production, so we restrain the usage of it only to the situations in which we observed
+  // a problem: when appId used by build application is different then the one defined by the
+  // applications manifest. It is quite common and happens when the productFlavor or buildType
+  // defines a special prefix/suffix to the appId. Most of the code is inspired by how expo CLI
+  // handles this case with the added bonus that radons solution does not require additional
+  // user configuration. You can explore the expo solution here:
+  // https://github.com/expo/expo/blob/645e63df903d28149ee9eda6682f6032b31601d7/packages/%40expo/cli/src/start/platforms/android/AndroidPlatformManager.ts#L93
+  if (packageName == appId) {
+    return undefined;
+  }
```
jwajgelt: This seems weird to me.
I'd prefer it if we instead changed the type of `AndroidBuildResult` to contain all of `appId`,`packageName` and `mainActivityName`, and have `AndroidDevice` decide how to launch the app instead of having this logic in `buildAndroid`.

filip131311: yeah i think that would make sense artefact of prototyping I'll move it around

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -46,12 +50,88 @@ function locateAapt() {
...
 }
 
-async function extractPackageName(artifactPath: string, cancelToken: CancelToken) {
+/**
+ * This function extracts the application Id that allows for identifying the application
+ * while managing its runtime it might be defined by multiple configurations
+ * https://developer.android.com/build/configure-app-module
+ * @param artifactPath
+ * @param cancelToken
+ * @returns Promise<string>
```
jwajgelt: useful

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -46,12 +50,88 @@ function locateAapt() {
...
+ */
+async function extractAppId(artifactPath: string, cancelToken: CancelToken) {
   const aaptPath = locateAapt();
   const { stdout } = await cancelToken.adapt(exec(aaptPath, ["dump", "badging", artifactPath]));
   const packageLine = stdout.split("\n").find((line: string) => line.startsWith("package: name="));
-  const packageName = packageLine!.split("'")[1];
-  return packageName;
+  const appId = packageLine!.split("'")[1];
```
jwajgelt: NIT:
```suggestion
  assert(packageLine, "aapt file always contains the line specifying the package name");
  const appId = packageLine.split("'")[1];
```
This will still fail on malformed files, but at least we will get a descriptive error instead of "cannot read property 'split' of undefined"

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -46,12 +50,88 @@ function locateAapt() {
...
+): Promise<AndroidConfig.Manifest.AndroidManifest | undefined> {
+  try {
+    const filePath = await AndroidConfig.Paths.getAndroidManifestAsync(projectRoot);
+    const androidManifest = await AndroidConfig.Manifest.readAndroidManifestAsync(filePath);
+
+    if (!androidManifest) {
+      return undefined;
+    }
+    return androidManifest;
+  } catch (e) {
```
jwajgelt: NIT:
```suggestion
  } catch {
```

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -46,12 +50,88 @@ function locateAapt() {
...
+  projectRoot: string
+): Promise<AndroidConfig.Manifest.AndroidManifest | undefined> {
+  try {
+    const filePath = await AndroidConfig.Paths.getAndroidManifestAsync(projectRoot);
+    const androidManifest = await AndroidConfig.Manifest.readAndroidManifestAsync(filePath);
+
+    if (!androidManifest) {
+      return undefined;
+    }
+    return androidManifest;
```
jwajgelt:
```suggestion
    if (androidManifest) {
      return androidManifest;
    }
```
and you simply step out of the try/catch into the final `return undefined`

packages/vscode-extension/src/builders/buildAndroid.ts
```diff
@@ -46,12 +50,88 @@ function locateAapt() {
...
+  }
+
+  try {
+    const filePath = await AndroidConfig.Paths.getAndroidManifestAsync(projectRoot);
+    const androidManifest = await AndroidConfig.Manifest.readAndroidManifestAsync(filePath);
+    // Assert MainActivity defined.
+    await AndroidConfig.Manifest.getMainActivityOrThrow(androidManifest);
+    if (androidManifest.manifest?.$?.package) {
+      return androidManifest.manifest.$.package;
+    }
```
jwajgelt: VERY NIT:
```suggestion
    const appId = androidManifest.manifest?.$?.package
    if (appId) {
      return appId
    }
```

PR #1750, opened 2025-11-19
fix: crashes when auto reload is triggered before Android device boot

jwajgelt:
Fixes sim-server failing to start when an auto-reload is triggered before an Android Emulator first boots.
The issue was caused by `makePreview` method being called by autoreload (in order to restart sim-server in case of crashes/disconnects/killswitch triggering) and accessing the `AndroidDevice`'s `serial` property before it's initialized by the Emulator boot procedure.
The fix checks the `serial` is actually set before accessing it, and throwing an error in that case (instead of passing `"undefined"` to the sim-server).

### How Has This Been Tested: 
- open Radon
- select an Android Emulator device
- trigger an autoreload immediately, before the device finishes booting
- the autoreload should succeed

Internal

### DIFF
--- a/packages/vscode-extension/src/devices/AndroidDevice.ts
+++ b/packages/vscode-extension/src/devices/AndroidDevice.ts
@@ -31,7 +31,18 @@ export const ADB_PATH = path.join(
 );
 
 export abstract class AndroidDevice extends DeviceBase implements Disposable {
-  protected serial: string | undefined;
+  private _serial: string | undefined;
+
+  protected get serial(): string {
+    if (this._serial === undefined) {
+      throw new Error("Device used before boot completed");
+    }
+    return this._serial;
+  }
+
+  protected set serial(value: string) {
+    this._serial = value;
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/devices/AndroidDevice.ts
```diff
@@ -31,7 +31,18 @@ export const ADB_PATH = path.join(
 );
 
 export abstract class AndroidDevice extends DeviceBase implements Disposable {
-  protected serial: string | undefined;
+  private _serial: string | undefined;
+
+  protected get serial(): string {
+    if (this._serial === undefined) {
+      throw new Error("Device used before boot completed");
```
filip131311, line 38: can we make it a custom Errror like "DeviceError" or something like that so the catch can filter it out?

jwajgelt: We can change it once we need it.

PR #1759, opened 2025-11-21
tests: Location tests for iOS and fixes for local CI

KeyJayY:
This pull request adds test coverage for location and localization features on iOS devices. It also includes several fixes for tests run on a local machine to improve overall stability.

### How Has This Been Tested:

The new tests have been executed on the local CI environment, and previously failing tests have been verified on a locally hosted machine to ensure they now work as expected.

To reproduce the results, the new tests should be run on the `react-native-81-additional-libs` test application, which provides the necessary tools for location tests.

Not documented yet.

### DIFF
--- a/packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
+++ b/packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
@@ -330,31 +330,36 @@ safeDescribe("8 - Device Settings", () => {
     }, 5000);
   });
 
-  itIf(
-    !getConfiguration().IS_ANDROID && IS_APP_WITH_ADDITIONAL_LIBS,
-    "change location",
-    async () => {
-      execSync(
-        "xcrun simctl --set ~/Library/Caches/com.swmansion.radon-ide/Devices/iOS privacy booted grant location org.reactjs.native.example.reactNative81AdditionalLibs"
-      );
+  const locationsToTest = [
+    { lattitude: 1, longitude: 1 },
+    { lattitude: -1, longitude: -1 },
+    { lattitude: 0, longitude: 0 },
+  ];
+  let permissionGranted = false;
 
-      await appManipulationService.restartDevice();
-      await appManipulationService.waitForAppToLoad();
+  for (const actualLocation of locationsToTest) {
+    itIf(
+      !getConfiguration().IS_ANDROID && IS_APP_WITH_ADDITIONAL_LIBS,
+      `Should change location to (${actualLocation.lattitude}, ${actualLocation.longitude})`,
+      async () => {
+        if (!permissionGranted) {
+          execSync(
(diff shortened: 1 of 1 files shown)


packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
```diff
@@ -324,4 +329,97 @@ safeDescribe("8 - Device Settings", () => {
...
+
+      await elementHelperService.findAndClickElementByTag("modal-close-button");
+      await driver.wait(async () => {
+        try {
+          const location =
+            await appManipulationService.sendMessageAndWaitForResponse(
+              appWebsocket,
+              "getLocation"
+            );
+          assert.approximately(location.value.latitude, 1, 0.1);
```
filip131311: can you duplicate this test and test 0 0 location as well? this is a special case that fails on some platform i don't know which but our fix is a bit hacky and I would like to have it covered

thank you <3

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
```diff
@@ -324,4 +329,97 @@ safeDescribe("8 - Device Settings", () => {
...
+      }, 5000);
+
+      await driver.wait(async () => {
+        try {
+          const localization =
+            await appManipulationService.sendMessageAndWaitForResponse(
+              appWebsocket,
+              "getLocalization"
+            );
+          assert.equal(localization.value.countryCode.toLowerCase(), "es");
```
filip131311, line 442: can we check for language to?

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
KeyJayY: I've added that test case, however changing location to that value appears to not work on iOS. I'm unable to set the coordinates to (0, 0), when I try, the longitude either does not change or it sets to our default value (19° 57′ 55″ E).

filip131311: That's "good" it is probably a bug thank you for adding a test case and I will look into it in the core

filip131311: does it work on android?

packages/vscode-extension-tester/ui-tests/08-deviceSettings.test.js
```diff
@@ -324,4 +329,119 @@ safeDescribe("8 - Device Settings", () => {
...
+      }, 5000);
+
+      await appManipulationService.hideExpoOverlay(appWebsocket);
+
+      const locationsToTest = [
+        { lattitude: 1, longitude: 1 },
+        { lattitude: 0, longitude: 0 },
+        { lattitude: -1, longitude: -1 },
+      ];
+      for (const actualLocation of locationsToTest) {
```
filip131311: If only one of does will fail will it be readable from the test output?

PR #1762, opened 2025-11-24
tests: Radon diagnostics test coverage 

KeyJayY:
This pull request adds test coverage for diagnostic functionality in Radon IDE extension. 

### How Has This Been Tested: 

The tests have been verified to pass on locally hosted GitHub CI runner:

Screenshot 2025-11-24 at 13 17 03

Screenshot 2025-11-24 at 13 12 21

to reproduce tests can be ran via Actions tab in GitHub.

No documentation yet.

### DIFF
--- a/.github/matrices/autorun_test_params.json
+++ b/.github/matrices/autorun_test_params.json
@@ -13,15 +13,6 @@
       "node": "22",
       "isRecording": "true",
       "testsToRun": "",
-      "isAndroid": "true",
-      "testsToExclude": "",
-      "codeVersion": "latest"
-    },
-    {
-      "testApp": "expo-53",
-      "node": "22",
-      "isRecording": "true",
-      "testsToRun": "",
       "testsToExclude": "",
       "codeVersion": "latest"
     },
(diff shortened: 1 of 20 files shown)


packages/vscode-extension-tester/ui-tests/17-diagnostics.test.js
```diff
@@ -0,0 +1,320 @@
...
+        execSync("brew unlink node");
+      },
+      async () => {
+        execSync("brew link node");
+      },
+      `Node.js was not found, or the version in the PATH does not satisfy minimum version requirements. You can find more information in our documentation.`
+    );
+  });
+
+  it("should show correct diagnostic for npm", async function () {
```
filip131311, line 124: maybe i don't get it but it seems to me that what this test is testing is "should show npm uninstalled when uninstalled" or I don't de how it tests the happy path? 

also is there a clean up mechanism for dose tests? that will make sure that the env changes made by the test block do not affect the other tests? maybe we should have a "build environment" job that rebuilds environment afterwords (it could also be used at the beginning of all the tests so we have fully replicable environment)

filip131311 (changes requested):
I left my worries inline

packages/vscode-extension-tester/ui-tests/17-diagnostics.test.js
KeyJayY: In our diagnostics panel we have field that checks if user has installed npm.

Image

This test verifies if this information actually appears if npm is not installed, and if it disappears if npm is installed again. Npm is part of brew node and `brew link node` fixes this after removing symlink using `rm /opt/homebrew/bin/npm`. 
Let me know if you think this should be tested different way.

PS
I know that this tests are dependent on how each tool is installed on testing machine, but I don't know better way to test it. The tests are compatible with locally hosted CI and tests pass there.

PR #1771, opened 2025-12-01
feat: user logs export

filip131311:
This PR adds a new option to export extension, build, etc. logs as a user to lower the friction of diagnostic process. Additionally we add initial ui for allowing to send logs with user feedback as well, but without actually sending anything as it will require addition infrastructure on customer portal side to store the logs in a secure way. 

Screenshot 2025-12-01 at 16 59 03
Screenshot 2025-12-01 at 16 59 09

### How Has This Been Tested: 

Logs export:
- run radon 
- open settings drop down 
- select which logs you would like to export 
- export them and go through them to make sure they are not corrupted 

New "include logs checkbox": 
- click it around make sure it does not look bad

will write a full docs once all the functionality is ready

### DIFF
--- a/packages/vscode-extension/src/webview/views/ExportLogsView.tsx
+++ b/packages/vscode-extension/src/webview/views/ExportLogsView.tsx
@@ -14,8 +14,12 @@ function ExportLogsView() {
   const [selected, setSelected] = useState<Record<string, boolean>>({});
 
   useEffect(() => {
+    let cancelled = false;
+
     async function fetchLogs() {
       const logs = await project.getLogFileNames();
+      if (cancelled) return;
+
       // Move "Radon IDE.log" to the front if it exists
       const radonLog = "Radon IDE.log";
       const hasRadonLog = logs.includes(radonLog);
@@ -25,7 +29,11 @@ function ExportLogsView() {
       setAllLogs(orderedLogs);
       setSelected(Object.fromEntries(orderedLogs.map((name) => [name, true])));
     }
+
     fetchLogs();
+    return () => {
+      cancelled = true;
+    };
   }, [project]);
 
   const selectedFiles = useMemo(
@@ -34,7 +42,12 @@ function ExportLogsView() {
   );
 
   function toggle(name: string) {
-    setSelected((prev) => ({ ...prev, [name]: !prev[name] }));
+    setSelected((prev) => {
+      if (!(name in prev)) {
+        return prev;
+      }
+      return { ...prev, [name]: !prev[name] };
+    });
   }
 
   async function onExport() {
--- a/packages/vscode-extension/src/webview/views/FeedbackView.tsx
+++ b/packages/vscode-extension/src/webview/views/FeedbackView.tsx
@@ -11,6 +11,11 @@ import { Sentiment } from "../../common/types";
 
 const CLOSE_MODAL_AFTER = 2400;
 
+// Note(Filip Kamiński): this feature is disabled for no as it
+// would require a lot of additional work on customer portal side,
+// but the ground work for it from extension side is already here
+const SHOW_INCLUDE_DIAGNOSTICS_CHECKBOX = false;
+
 type FeedbackViewProps = {
   initialSentiment: Sentiment | undefined;
 };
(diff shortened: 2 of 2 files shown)


packages/vscode-extension/src/project/EditorBindings.ts
```diff
@@ -21,6 +23,38 @@ type KeybindingType = {
...
+      window.showErrorMessage(`Failed to build diagnostics report: ${error}`);
+    }
+  }
+
+  public async getLogFileNames(): Promise<string[]> {
+    const logsUri = extensionContext.logUri;
+    const logsPath = logsUri.path;
+    const files = await fs.promises.readdir(logsPath);
+    return files;
+  }
```
jwajgelt, line 56: This feels like it should be in `State`, but I understand why it's not.

packages/vscode-extension/src/webview/views/ExportLogsView.tsx
```diff
@@ -0,0 +1,86 @@
...
+      // Move "Radon IDE.log" to the front if it exists
+      const radonLog = "Radon IDE.log";
+      const hasRadonLog = logs.includes(radonLog);
+      const orderedLogs = hasRadonLog
+        ? [radonLog, ...logs.filter((name) => name !== radonLog)]
+        : logs;
+      setAllLogs(orderedLogs);
+      setSelected(Object.fromEntries(orderedLogs.map((name) => [name, true])));
+    }
+    fetchLogs();
```
jwajgelt, line 33:
```suggestion
      if (cancelled) return;
      // Move "Radon IDE.log" to the front if it exists
      const radonLog = "Radon IDE.log";
      const hasRadonLog = logs.includes(radonLog);
      const orderedLogs = hasRadonLog
        ? [radonLog, ...logs.filter((name) => name !== radonLog)]
        : logs;
      setAllLogs(orderedLogs);
      setSelected(Object.fromEntries(orderedLogs.map((name) => [name, true])));
    }
    let cancelled = false;
    fetchLogs();
    return () => {cancelled = true};
```
This effect can rerun, causing concurrent updates.
I don't know if `project` is stable now (I hope it is), but at some point in the future someone will add another dependency here, and it will cause problems.

packages/vscode-extension/src/webview/views/ExportLogsView.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }, [project]);
+
+  const selectedFiles = useMemo(
+    () => allLogs.filter((name) => selected[name]),
+    [allLogs, selected]
+  );
+
+  function toggle(name: string) {
+    setSelected((prev) => ({ ...prev, [name]: !prev[name] }));
```
jwajgelt: NIT:
```suggestion
    setSelected((prev) => {
    if (!(name in prev)) {
      return prev;
    }
    return { ...prev, [name]: !prev[name] }
    });
```
for the sake of soundness.
Sure, you only _should_ call it on `name` that's already in `selected`, but:
1. that's an implicit invariant that results from code in two places different from this function
2. since setting state is asynchronous, you _could_ call it on a name which later disappears from `selected`, before this callback is called <- again, it shouldn't if you read the rest of this component, but it's not a local property of `toggle`, and it's easy to break it when poking around this component

packages/vscode-extension/src/webview/views/FeedbackView.tsx
```diff
@@ -70,8 +78,21 @@ function FeedbackView({ initialSentiment }: FeedbackViewProps) {
         placeholder="Tell us why (optional)"
         rows={5}
       />
+      {/* Note(Filip Kamiński): this feature is disabled for no as it  
+      would require a lot of additional work on customer portal side,
+      but the ground work for it from extension side is already here */}
+      {/* eslint-disable-next-line */}
+      {false && sentiment === "negative" && (
```
jwajgelt: I'd prefer if we had a module-level constant
```
/* Note(Filip Kamiński): this feature is disabled for now as it  
      would require a lot of additional work on customer portal side,
      but the ground work for it from extension side is already here */
const SHOW_INCLUDE_DIAGNOSTICS_CHECKBOX = false;
```
and checked that in the component
```suggestion
      {SHOW_INCLUDE_DIAGNOSTICS_CHECKBOX && sentiment === "negative" && (
```

packages/vscode-extension/src/project/EditorBindings.ts
filip131311: Yeah it probably would be nicer but would require a watchman to observe the logs directory and this seems like an unnecessary overhead

PR #1775, opened 2025-12-02
fix: cannot remove a broken iOS simulator

jwajgelt:
Fixes the "Remove Device" button in the "Manage Devices" dialog not working for "broken" devices.
The issue was caused by us disabling the whole Device Row for unavailable devices, preventing touch events from ever reaching the "Remove" button...

image

This PR removes the `disabled` attribute, and instead relies on a descriptive error dialog when the user selects the unavailable device.
(example for physical device)
image
The "Play" button in "Manage Devices" remains disabled, suggesting to the user that they won't be able to start the device before they take any action.

### How Has This Been Tested: 
- create an iOS device and close Radon / vscode completely
- remove the runtime for that device via xcode
- restart Radon verify you can remove the device through Radon

Internal / Bug Fix

### DIFF
--- a/packages/vscode-extension/src/webview/views/ManageDevicesView.tsx
+++ b/packages/vscode-extension/src/webview/views/ManageDevicesView.tsx
@@ -78,7 +78,6 @@ function DeviceRow({
   return (
     <button
       className="device-row"
-      disabled={disabled}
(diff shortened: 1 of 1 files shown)


filip131311 (approved):
thank you

PR #1776, opened 2025-12-02
fix: Radon does not pick up schemes for XCode workspaces

jwajgelt:
- fixes Radon not detecting XCode project schemes correctly for applications using XCode workspaces (which seems to include all of our test apps)
- changes the scheme select in the Launch Configurations dialog to a combobox, allowing setting custom schemes without editing `launch.json` manually

### How Has This Been Tested: 
- open reanimated's FabricExample in Radon
- the "Debug FabricExample" scheme should be detected

Bugfix

### DIFF
--- a/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
+++ b/packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
@@ -449,6 +449,13 @@ function StandardBuildConfiguration({
       return { value: xcodeScheme, label: xcodeScheme };
     });
 
+    const selectedScheme = config?.ios?.scheme;
+
+    // NOTE: this is necessary for a custom scheme to appear in the list when initially opening the modal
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/webview/views/LaunchConfigurationView.tsx
```diff
@@ -449,6 +449,13 @@ function StandardBuildConfiguration({
       return { value: xcodeScheme, label: xcodeScheme };
     });
 
+    const selectedScheme = config?.ios?.scheme;
+
+    // NOTE: this is necessary for a custom scheme to appear in the list when initially opening the modal
+    if (selectedScheme && !availableXcodeSchemes.find((s) => s.value === selectedScheme)) {
```
filip131311, line 455: there is probably some smart  and unnecessary optimisation to be made by performing this logic during: 
```
const availableXcodeSchemes = xcodeSchemes.map((xcodeScheme) => {
      return { value: xcodeScheme, label: xcodeScheme };
    });
```
but you can leave it

PR #1780, opened 2025-12-03
feat: add an option to disable features by administration 

filip131311:
This PR adds a requested ability to restrict radons functionality from customer portal side, by license administrator. 
The turned off functionality should be removed from the ui if possible if not it should be disabled. If This is still imposiblle because it is in form of a command or is "@radon" chat participant it should inform a user why it is not working. 

In critical fallback situations a user will be met with custom "Restricted by administration screen. 

### How Has This Been Tested: 

- run test ups with modified `BASE_CUSTOMER_PORTAL_URL` to hit the staging customer portal 
- disable radon AI and send files and check out if it does not work.
- check if chat responds with correct message.
- check if element inspector AI tool is removed from the ui. 

wip

Screenshot 2025-12-03 at 21 10 24

Screenshot 2025-12-03 at 17 00 59


packages/vscode-extension/src/ai/mcp/RadonMcpController.ts
```diff
@@ -113,31 +149,42 @@ class RadonMcpController implements Disposable {
...
+    setGlobalEnableAI(radonAiEnabled);
+
+    if (!shouldUseDirectRegistering()) {
+      if (radonAiEnabled) {
+        this.enableServer();
+      } else {
+        this.disableServer();
+      }
+    }
+  };
```
jwajgelt: 1. change this to something like `updateAvailabliity`
2. call this in the ctor as well

packages/vscode-extension/src/project/project.ts
```diff
@@ -2,6 +2,7 @@ import { EventEmitter } from "stream";
 import os from "os";
 import path from "path";
 import { env, Disposable, commands, workspace, window, Uri } from "vscode";
+import { env, Disposable, commands, workspace, window, Uri } from "vscode";
```
jwajgelt:
```suggestion
```

packages/vscode-extension/src/utilities/license.ts
```diff
@@ -212,6 +228,40 @@ export async function checkLicenseToken(token: string): Promise<SimServerLicense
...
+   * JWT payload structure returned by the Customer Portal (cp_* fields)
+   */
+  type LicenseJWTPayload = {
+    cp_sub: string;
+    cp_pri: string;
+    cp_fpr: string;
+    cp_ent: number;
+    cp_usr: string;
+    cp_plan: string;
+    cp_features: FeaturesAvailability;
```
jwajgelt: This comes from the aether, you cannot assert the type like this.
You should:
- parse this into an object (doing this together with the rest of the `JWTPayload` is fine)
- check that this _actually_ is a `Record<string, string>` (or `Record<string, unknown>`)
- ignore keys that are not `Feature`s
- and then verify all values are `FeatureAvailabilityStatus`, potentially with a fallback to `Paywalled` or otherwise an error-ey value

And you should probably parse the rest of the fields as well.

At some point, the customer portal and Radon versions might mismatch (we may at some point decide to change the shape of this payload for whatever reason, and not pay enough attention to make the change backwards-compatible), and in that case this will start behaving unexpectedly wherever you assume that the types are actually correct.

We already have `zod` in our dependencies, it's great for stuff like this.

packages/vscode-extension/src/webview/components/DeviceSelect.tsx
```diff
@@ -197,6 +205,14 @@ function DeviceSelect() {
     }
   };
 
+  const isAdminDisabledDeviceSections = ([label]: [
+    string,
+    IOSDeviceInfo[] | AndroidEmulatorInfo[] | AndroidPhysicalDeviceInfo[],
+  ]): boolean => {
+    const isRemoteAndroidAdminDisabled = useIsFeatureAdminDisabled(Feature.AndroidPhysicalDevice);
+    return label !== "Connected Android Devices" || !isRemoteAndroidAdminDisabled;
```
jwajgelt: NIT: magic strings bad
It was fine(-ish) before, since the value was only used for display purposes, but if you actually care about the value, we need to protect ourselves from accidentally changing it in one place and not the other.
```ts
    /* ... someplace else ... */
    enum DeviceSection {
      PhysicalAndroid,
      /* etc... */
    }

    const DeviceSectionLabels = {
      [DeviceSection.PhysicalAndroid]: "Connected Android Devices",
       /* etc... */
    } as const
```

```suggestion
    return section !== DeviceSection.PhysicalAndroid || !isRemoteAndroidAdminDisabled;
```

jwajgelt: I'd deMorgan the shit out of this

```suggestion
    return !(label === "Connected Android Devices" && isRemoteAndroidAdminDisabled);
```

packages/vscode-extension/src/common/License.ts
```diff
@@ -112,51 +78,45 @@ export enum Feature {
...
+  [Feature.AndroidPhysicalDevice]: FeatureAvailabilityStatus.PAYWALLED,
+  [Feature.AppSwitcherButton]: FeatureAvailabilityStatus.AVAILABLE,
+  [Feature.Biometrics]: FeatureAvailabilityStatus.PAYWALLED,
+  [Feature.ComponentPreview]: FeatureAvailabilityStatus.AVAILABLE,
+  [Feature.Debugger]: FeatureAvailabilityStatus.AVAILABLE,
+  [Feature.DeviceAppearanceSettings]: FeatureAvailabilityStatus.AVAILABLE,
+  [Feature.DeviceFontSizeSettings]: FeatureAvailabilityStatus.AVAILABLE,
+  [Feature.DeviceLocalizationSettings]: FeatureAvailabilityStatus.PAYWALLED,
+  // FRYTKI - this is changed only for testing
+  [Feature.DeviceRotation]: FeatureAvailabilityStatus.ADMIN_DISABLED,
```
jwajgelt: Unfrytkify this before merging

```suggestion
  [Feature.DeviceRotation]: FeatureAvailabilityStatus.PAYWALLED,
```

packages/vscode-extension/src/common/License.ts
```diff
@@ -112,51 +78,45 @@ export enum Feature {
...
+  [Feature.ReactProfiler]: FeatureAvailabilityStatus.AVAILABLE,
+  [Feature.ReactQueryDevTools]: FeatureAvailabilityStatus.AVAILABLE,
+  [Feature.RadonConnect]: FeatureAvailabilityStatus.AVAILABLE,
+  [Feature.RadonAI]: FeatureAvailabilityStatus.PAYWALLED,
+  [Feature.ReduxDevTools]: FeatureAvailabilityStatus.AVAILABLE,
+  [Feature.ScreenRecording]: FeatureAvailabilityStatus.PAYWALLED,
+  [Feature.ScreenReplay]: FeatureAvailabilityStatus.PAYWALLED,
+  [Feature.Screenshot]: FeatureAvailabilityStatus.PAYWALLED,
+  // Frytki
+  [Feature.SendFile]: FeatureAvailabilityStatus.ADMIN_DISABLED,
```
jwajgelt: Likewise here

```suggestion
  [Feature.SendFile]: FeatureAvailabilityStatus.PAYWALLED,
```

packages/vscode-extension/src/ai/mcp/RadonMcpController.ts
```diff
@@ -92,15 +99,44 @@ export function isAiEnabledInSettings() {
...
+  }
+
+  constructor(
+    context: ExtensionContext,
+    options: { isAiEnabledInSettings: boolean; radonAvailabilityStatus: FeatureAvailabilityStatus }
+  ) {
+    const isAiEnabled =
+      options.isAiEnabledInSettings ||
+      options.radonAvailabilityStatus !== FeatureAvailabilityStatus.AVAILABLE;
+    setGlobalEnableAI(isAiEnabled);
```
jwajgelt:
```suggestion
```
`this.updateAvailability` should take care of this.

packages/vscode-extension/src/ai/mcp/RadonMcpController.ts
```diff
@@ -112,32 +148,41 @@ class RadonMcpController implements Disposable {
...
+
+    this.disposables.push(
+      watchLicenseTokenChange(() => {
+        this.updateAvailability();
       })
     );
   }
 
+  // used in callbacks needs to be an arrow function
+  private updateAvailability = async () => {
```
jwajgelt: You could make this sync by accepting the `availabilityStatus`, and only update it in `watchLicenseTokenChange`.

packages/vscode-extension/src/common/License.ts
```diff
@@ -1,54 +1,20 @@
...
 
 // #endregion License Helpers
 
 // #region License State
 
+export type FeaturesAvailability = { [F in Feature]: FeatureAvailabilityStatus };
+
 export type LicenseState = {
   status: LicenseStatus;
+  featuresAvailability: FeaturesAvailability;
```
jwajgelt, line 17: I think that:
- This should be 
```suggestion
  featuresAvailability: Partial<FeaturesAvailability>;
```
- when you check availability and the feature is not there, you refetch the token

in order to avoid issues where after we add a feature but before the token is refreshed, a paying customer doesn't have access to the new features (as we fallback to `DEFAULT`)

packages/vscode-extension/src/utilities/license.ts
```diff
@@ -212,6 +230,83 @@ export async function checkLicenseToken(token: string): Promise<SimServerLicense
...
+} | null> {
+  // Zod schema for runtime validation of features object
+  const featureAvailabilityStatusSchema = z.enum([
+    FeatureAvailabilityStatus.AVAILABLE.toString(),
+    FeatureAvailabilityStatus.PAYWALLED.toString(),
+    FeatureAvailabilityStatus.ADMIN_DISABLED.toString(),
+  ]);
+
+  const featuresSchema = z.record(z.string(), z.unknown()).transform((obj) => {
+    const result: FeaturesAvailability = { ...DefaultFeaturesAvailability };
```
jwajgelt: Like I mentioned in the other comment, I think this function should return a `Partial<FeaturesAvailability>`

```suggestion
    const result: Partial<FeaturesAvailability> = {};
```

packages/vscode-extension/src/webview/components/DeviceSelect.tsx
```diff
@@ -239,15 +267,17 @@ function DeviceSelect() {
...
-                selectedDevice,
-                runningSessionIds,
-                handleDeviceStop
-              )
-            )}
+            {Object.entries(deviceSections)
+              .filter(isAdminDisabledDeviceSections)
+              .map(([section, sectionDevices]) =>
+                renderDevices(
+                  DeviceSectionLabels[section as keyof typeof DeviceSectionLabels],
```
jwajgelt: I think you should be able to type the `DeviceSectionLabels` as `Record<DeviceSection, string>` to avoid this cast.

packages/vscode-extension/src/common/License.ts
filip131311: but this is handled at jwt decoding level when we substitute any missing values with a default this would not be true that it is a partial

filip131311: We may consider refetching, tho this PR is quite big already i would preffer to do it in a follow up if at all

packages/vscode-extension/src/webview/components/DeviceSelect.tsx
filip131311: I made it better but DeviceSection is lost along the way unfortunately

packages/vscode-extension/src/utilities/license.ts
filip131311: I made changes here tho

packages/vscode-extension/src/utilities/license.ts
```diff
@@ -212,6 +258,81 @@ export async function checkLicenseToken(token: string): Promise<SimServerLicense
...
+  cp_ent?: number;
+  cp_usr?: string;
+  cp_plan?: string;
+  cp_features?: Partial<FeaturesAvailability>;
+} | null> {
+  // Zod schema for runtime validation of features object
+  const featureAvailabilityStatusSchema = z.enum([
+    FeatureAvailabilityStatus.AVAILABLE.toString(),
+    FeatureAvailabilityStatus.PAYWALLED.toString(),
+    FeatureAvailabilityStatus.ADMIN_DISABLED.toString(),
```
latekvo: `...Status.FeatureAvailabilityStatus.AVAILABLE` is already a string, is there a reason why we're doing `.toString()`?


```suggestion
    FeatureAvailabilityStatus.AVAILABLE,
    FeatureAvailabilityStatus.PAYWALLED,
    FeatureAvailabilityStatus.ADMIN_DISABLED,
```

packages/vscode-extension/src/project/project.ts
```diff
@@ -469,22 +469,34 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
...
+    if (validationResult.status === SimServerLicenseValidationStatus.Success) {
+      newState.status = validationResult.licensePlan;
+      newState.featuresAvailability = validationResult.featuresAvailability;
+    }
+
+    commands.executeCommand(
+      "setContext",
+      "RNIDE.licenseActive",
+      newState.status !== LicenseStatus.Inactive
+    );
```
latekvo: I think `RNIDE.licenseActive` was made redundant in this PR, as `RNIDE.AIEnabled` is already set to `false` when `this.radonAvailabilityStatus !== FeatureAvailabilityStatus.AVAILABLE`, and `RNIDE.licenseActive` is only ever used in pair with `RNIDE.AIEnabled`. 

Also please remember to remove all references of `RNIDE.licenseActive` from `package.json` if you decide to remove it.

latekvo (commented):
Thanks for this PR! I just looked through the MCP side of things, looks good just left a few inlines.

packages/vscode-extension/src/ai/mcp/RadonMcpController.ts
```diff
@@ -112,32 +147,46 @@ class RadonMcpController implements Disposable {
...
+      watchLicenseTokenChange(async () => {
+        const newRadonAIAvailabilityStatus = await radonAIAvailabilityStatus();
+        this.updateAvailability(newRadonAIAvailabilityStatus);
       })
     );
   }
 
+  // used in callbacks needs to be an arrow function
+  private updateAvailability = (newRadonAIAvailabilityStatus?: FeatureAvailabilityStatus) => {
+    if (newRadonAIAvailabilityStatus) {
```
jwajgelt:
```suggestion
    if (newRadonAIAvailabilityStatus !== undefined) {
```

packages/vscode-extension/src/webview/components/DeviceSelect.tsx
jwajgelt: Yeah.
The issue is this:
```ts
  Object.entries(deviceSections)
```
since TS has structural typing, you're not guaranteed `deviceSections` doesn't have extra keys that are not `DeviceSection`.
Instead, what you could do is: something along the lines of:
```ts
Object.values(DeviceSection)
              .filter(isAdminDisabledDeviceSection) // now takes just a single `DeviceSection` argument
              .map((section) =>
                renderDevices(
                  DeviceSectionLabels[section],
                  deviceSections[section],
                  selectedDevice,
                  runningSessionIds,
                  handleDeviceStop
                )
              )
```

packages/vscode-extension/src/common/License.ts
jwajgelt: Discussed internally.

jwajgelt (approved):

PR #1781, opened 2025-12-04
feat/fix: network inspector - fixes and improvements in network inspector

pFornagiel:
### Description

This PR introduces fixes and small features which were bugged / missing from network inspector:

Frontend:
- Fixed a bug, where a time value of `0` was incorrectly evaluated as falsy in a conditional check, causing incorrect behavior when the time was exactly zero.
- Fixed a CSS overflow issue in the network log details panel by changing `min-width` to `width`, preventing content from overflowing its container.

Application (lib folder):
- Removed the unused `data` field from the `Network.responseReceived` CDP message response object. This field was being sent but never consumed by any part of the codebase.
- Added handling for React Native's `FormData` objects in the `deserializeRequestData` function. FormData in React Native has a `_parts` array structure (e.g., `{_parts: [['key', 'value'], ...]}`), which is now properly serialized to a readable JSON format for display in the network inspector.
- Cleaned up network event messages by removing unnecessary fields
- Added proper `resourceType` handling for network messages (e.g., distinguishing between XHR, Script, Image, Media, etc.)
- Added `getResourceType` function to determine the resource type based on MIME type
- Unified the timestamp format across both XHR and Fetch polyfill interceptors:
  - Changed from `Date.now()` (milliseconds) to `Date.now() / 1000` (seconds) to match CDP protocol expectations and `DebuggerNetworkInspector` format
- Added handling for the XHR `progress` event to properly track incremental data transfer. This enables the network inspector to show real-time progress for large responses and properly calculate `encodedDataLength` during streaming responses.

### How Has This Been Tested: 

WIP.

Not applicable.

### DIFF
--- a/packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
+++ b/packages/vscode-extension/lib/network/interceptors/PolyfillFetchInterceptor.ts
@@ -330,7 +330,7 @@ class PolyfillFetchInterceptor {
     this.sendCDPMessage("Network.loadingFinished", {
       requestId,
       timestamp: timeStamp,
-      duration: timeStamp - this.startTime,
+      duration: Math.round((timeStamp - this.startTime) * 1000),
       response: {
         type: response.type,
         status: fetchInstance._responseStatus,


packages/vscode-extension/lib/network/networkRequestParsers.ts
```diff
@@ -360,6 +357,41 @@ export function mimeTypeFromResponseType(responseType: string): string | undefin
...
+ * We've been using the plain "XHR" type for all requests so far,
+ * but we can improve this behaviour by mapping MIME types to resource types.
+ * @see https://github.com/facebook/react-native/blob/main/packages/react-native/ReactCommon/jsinspector-modern/network/CdpNetwork.cpp#L144-L165
+ */
+//
+export function getResourceType(mimeType: string): ResourceType {
+  console.log(mimeType);
+  // Normalize casing to ensure robustness (optional, but recommended for MIME types)
+  const normalizedType = mimeType.toLowerCase().trim();
+  console.log(normalizedType === "application/json");
```
```suggestion
  // Normalize casing to ensure robustness (optional, but recommended for MIME types)
  const normalizedType = mimeType.toLowerCase().trim();
```
remember to get rid of _all_ of the logs under `lib` before merging (since they would be visible as application logs).

packages/vscode-extension/lib/network/networkRequestParsers.ts
```diff
@@ -420,10 +460,23 @@ function reconstructTypedArray(serializedData: SerializedTypedArray): Uint8Array
...
     return data;
   }
 
+  // Handle React Native FormData with _parts structure
+  if (isFormDataObject(data)) {
+    try {
+      const obj: Record<string, unknown> = {};
+      data._parts.forEach(([key, value]) => {
+        obj[key] = value;
+      });
```
jwajgelt: How is this different from `Object.fromEntries`?

packages/vscode-extension/src/network/hooks/useNetworkTracker.tsx
```diff
@@ -68,7 +68,8 @@ class RequestTimingTracker {
     if (timeStart === undefined || timeEnd === undefined) {
       return undefined;
     }
-    return timeEnd - timeStart;
+    console.log("Calculating time difference:", (timeEnd - timeStart) * 1000);
```
jwajgelt: I feel this is excessive.
```suggestion
```

pFornagiel: Leaving this was an oversight, thanks. Removed in c5b0938d

packages/vscode-extension/lib/network/networkRequestParsers.ts
pFornagiel: My bad, removed in c5b0938d

packages/vscode-extension/lib/network/networkRequestParsers.ts
pFornagiel: It isn't and `Object.fromEntries` is cleaner, thanks for pointing that out.

Changed in 625350a2

PR #1786, opened 2025-12-05
fix: maestro ios fails in prod builds

jwajgelt:
Fixes an issue where Maestro tests would fail in production builds, due to the required shims not being bundled with the extension.

### How Has This Been Tested: 
- build the vscode package
- open a test app with a maestro test
- run the test on an iOS device
- the test should run

Bugfix


filip131311 (approved):
🚀

PR #1791, opened 2025-12-05
feat: add remove license command

filip131311:
Now that there is a noticeable difference between experience of activated and un activated license we add a "remove license" command with two goals in mind: 
1) make it easier to test multiple setups 
2) if a user has any problem with corrupted license (which hopefully wont happen) we can point them to this command in order to replace it with a new one

### How Has This Been Tested: 

https://github.com/user-attachments/assets/d1f642a8-b3d6-4e01-a4d0-e71446f30c63

We have quite a large trouble shooting backlog to document we will handle it all at once in a future PR

### DIFF
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -90,7 +90,6 @@ export enum ActivateDeviceResult {
 
 export interface ProjectEventMap {
   projectStateChanged: ProjectState;
-  licenseActivationChanged: boolean;
 }
 
 export interface ProjectEventListener<T> {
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/common/Project.ts
```diff
@@ -90,7 +90,6 @@ export enum ActivateDeviceResult {
 
 export interface ProjectEventMap {
   projectStateChanged: ProjectState;
-  licenseActivationChanged: boolean;
```
latekvo, line 93: Is this change relevant to this PR?

latekvo (approved):
Thx!

packages/vscode-extension/src/common/Project.ts
filip131311: no but i noticed it working on it. It is a dead code from previous state management approach

PR #1798, opened 2025-12-08
feat: network inspector - add launch configuration for new inspector

pFornagiel:
### Description

This PR introduces additional Launch Configuration option `Use Native Network Inspector Implementation`, which controls whether the new network inspector is enabled.

The option can be explicitly set to `Yes` or `No` with caveat, that for react native version < 0.83.0 the inspector will not showcase any massages, which is provided in the description of option.

When set to `Detect Automatically` the extension will detect whether current version is >= 0.83.0 and set the flag accordingly.

Screenshot 2025-12-08 at 15 08 20

### How Has This Been Tested: 

Changed configuration and used debugger to determine which inspector instance is initialised in `network-plugin.ts` in the constructor, changed the `package.json` of react native module in `node_modules` to `0.83.0` to check whether automatic detection works correctly, verified that changing the option properly edits `launch.json`

Will have to be done asap, wip for now.

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -880,7 +880,7 @@
               },
               "useNativeNetworkInspector": {
                 "type": "boolean",
-                "description": "When set to true, Radon will use the Native (\"Fusebox\") Network Inspector implementation introduced in 0.83.0. Note, that this requires at least React Native 0.83.0 and it will not intercept network messages for earlier versions."
+                "description": "When set to true, Radon will use the Native (\"Fusebox\") Network Inspector implementation introduced in 0.83.0. Note that this is only supported on React Native 0.83.0 and newer, and might not intercept network messages on earlier versions."
               }
             }
           }


packages/vscode-extension/package.json
```diff
@@ -884,6 +877,10 @@
               "disableNativeBuildStaleChecks": {
                 "type": "boolean",
                 "description": "When set to true, Radon will not run fingerprint checks on reloads and after file changes that would detect whether the native code needs to be rebuilt."
+              },
+              "useNativeNetworkInspector": {
+                "type": "boolean",
+                "description": "When set to true, Radon will use the Native (\"Fusebox\") Network Inspector implementation introduced in 0.83.0. Note, that this requires at least React Native 0.83.0 and it will not intercept network messages for earlier versions."
```
jwajgelt:
```suggestion
                "description": "When set to true, Radon will use the Native (\"Fusebox\") Network Inspector implementation introduced in 0.83.0. Note that this is only supported on React Native 0.83.0 and newer, and might not intercept network messages on earlier versions."
```

pFornagiel: Done in e319865b

PR #1799, opened 2025-12-08
feat: disable maestro UI until it`s ready

jwajgelt:
Removes entry points for Maestro tests from the UI:
- CodeLens Provider for editor inline hints
- Tools drop-down entry
- Diagnostics entry
- VSCode commands

### How Has This Been Tested: 
- run Radon with maestro installed
- check that you can't start maestro tests inside Radon

meh


packages/vscode-extension/src/extension.ts
```diff
@@ -286,8 +266,6 @@ export async function activate(context: ExtensionContext) {
   context.subscriptions.push(
     commands.registerCommand("RNIDE.showInlinePreview", showInlinePreview)
   );
-  context.subscriptions.push(commands.registerCommand("RNIDE.startMaestroTest", startMaestroTest));
```
filip131311, line 289: this wasn't declared in package.json as well?

jwajgelt: nope, apparently.

PR #1804, opened 2025-12-10
fix: delay before starting maestro tests

jwajgelt:
Fixes a 30s delay after starting maestro tests and before the app is launched on the device.
The delay was caused by `maestro` calling `applesimutils` utility in order to set permissions on the device. The utility does not support passing simulator device sets (and even if it did, `maestro` would not set it anyway), so it would repeatedly try to set permissions on a simulator device which does not exist, and timeout after 30s, after which `maestro` would proceed with running the application to be tested.

The solution allows `applesimutils` to modify the permissions on Radon simulator by symlinking it to the `CoreSimulator` directory -- since `applesimutils` accesses the simulator's files directly (without checking the simulators database, like `xcodebuild`), a symlink is sufficient for it to function.

### How Has This Been Tested: 
- run maestro test on an iOS simulator
- the test should start soon after the Maestro TestRunner opens

Bugfix

### DIFF
--- a/packages/vscode-extension/src/project/MaestroTestRunner.ts
+++ b/packages/vscode-extension/src/project/MaestroTestRunner.ts
@@ -108,6 +108,10 @@ export class MaestroTestRunner implements Disposable {
       pty.writeLine(`Starting ${fileNames.length} Maestro flows: ${fileNames.join(", ")}`);
     }
 
+    // NOTE: maestro sets application permissions using the `applesimutils` utility,
+    // which does not support custom device sets.
+    // To work around this, we create a symlink to the target Radon device
+    // in the default simulator location for the duration of the test, and remove it afterwards.
     const cleanupSymlink = await this.setupSimulatorSymlink();
     try {
       await this.runMaestro(fileNames, pty);


packages/vscode-extension/src/project/MaestroTestRunner.ts
```diff
@@ -145,54 +205,53 @@ export class MaestroTestRunner implements Disposable {
...
-    this.pty?.writeLine("");
-    this.pty?.writeLine(`\x1b[33mAborting Maestro test...\x1b[0m`);
-
-    const proc = this.maestroProcess;
-    try {
-      proc.kill();
-    } catch (e) {}
+    const id = this.device.id;
+    const devicePath = path.join(getOrCreateDeviceSet(id), id);
+    const coreSimulatorPath = path.join(
```
filip131311, line 219: Is it always the same even for users with beta versions of the Xcode? shouldn't we really on `xcode-select` or something like that?

packages/vscode-extension/src/project/MaestroTestRunner.ts
```diff
@@ -108,6 +108,65 @@ export class MaestroTestRunner implements Disposable {
...
+      if (exitCode !== null && exitCode !== 143) {
+        pty.writeLine(`\x1b[31mMaestro test failed with exit code ${exitCode}\x1b[0m`);
+      }
+    } finally {
+      await cleanupSymlink();
+    }
+    this.maestroProcess = undefined;
+  }
+
+  public async stopMaestroTest(): Promise<void> {
```
filip131311, line 132: why does it not cleanup simlinks?

filip131311 (commented):
I have questions inline and another here: 

We observed a problem with the permissions script only at the beginning of the test suite run. Is it because of our example tests or is it something normal? If it is only an issue with the setup could we detect that this part of the process is done and clean up the sim links then? this way we could minimize the risk of leaving symlinks behind

packages/vscode-extension/src/project/MaestroTestRunner.ts
jwajgelt: It's not stored within XCode, it's in your `Library` path.
This path was taken from [`applesimutils` implementation](https://github.com/wix/AppleSimulatorUtils/blob/753749d51ccdb709fddb4f24dd04436bed3bb6b4/applesimutils/applesimutils/SimUtils.m#L80) so even if the system decided to place the sims somewhere else, for `applesimutils` to work we need to link them here.

packages/vscode-extension/src/project/MaestroTestRunner.ts
jwajgelt: Why would it?
Stopping the process will make the `maestroProcess` "promise" reject, triggering the `finally` clause set up in `start`.
You should cleanup wherever you initialize the resource, where possible.

PR #1806, opened 2025-12-10
feat: collapse unused section of launch configuration dropdown 

filip131311:
This PR collapses  the "detected" or "custom" section of launch configuration depending on which one is currently selected
this should improve a user experience for users that have a lot of custom configurations and the "detected" section is cluttering their already large selection. 

Moreover for the first setup we select the custom configuration if exists: 
```js
// Otherwise, return the first launch config or a default one
if (this._launchConfigurations.length > 0) {
  return this._launchConfigurations[0];
}
```

so for enterprise projects with already existing configurations the user should never have a need to expand the "detected" section 

### How Has This Been Tested: 

https://github.com/user-attachments/assets/0d3a21b0-b821-4e83-aedd-ccb58e97e2d6

nah

### DIFF
--- a/packages/vscode-extension/vite.config.js
+++ b/packages/vscode-extension/vite.config.js
@@ -71,7 +71,7 @@ export default defineConfig({
     port: 2137,
     strictPort: true,
     hmr: {
-      host: "localhost",
+      host: "127.0.0.1",
     },
   },
 });


packages/vscode-extension/src/webview/components/AppRootSelect.tsx
```diff
@@ -145,10 +161,17 @@ function AppRootSelect() {
     customLaunchConfigurations: customConfigurations,
   } = projectState;
 
+  console.log("Frytki", selectedConfiguration);
+
```
filip131311:
```suggestion
```

jwajgelt: 🕸️👨 👉  👈 👨🕸️

packages/vscode-extension/src/webview/components/AppRootSelect.tsx
```diff
@@ -46,6 +46,8 @@ function renderLaunchConfigurations(
   prefix: string,
   customLaunchConfigurations: LaunchConfiguration[],
   selectedValue: string | undefined,
+  isExpanded: boolean,
+  onToggleExpand: () => void,
```
jwajgelt, line 50: I think this function should be made into a full-blown component, and these should be the internal state of that component, rather than arguments that are passed through three separate functions.

filip131311: sure Ill cook it in a minute

jwajgelt (approved):

PR #1816, opened 2025-12-12
fix: network inspector - disable native implementation on android 0.83.0 due to issues

pFornagiel:
### Description

This PR introduces additional check for React Native `0.83.0` running on an Android device, which helps to disable native network inspector in this case.

When using native network inspector implementation with Android device on version `0.83.0`, the `Size` of request made to an `https` endpoint is always `-1`. This behaviour is caused by React Native DevTools bug and is not our fault (it is in fact also apparent in the devtools window itself), but given that our implementation works in this case, we should switch to it to provide proper user experience when we can.

image

This change required moving the compatibility check from `ApplicationContext` to `ApplicationSession`, as we do not have information about used device in the LaunchConfig (it is tailored for both platforms, the platform information itself comes at later initialisation stage).

### How Has This Been Tested: 

Verified that, on version `0.83.0` native network inspector is used on iOS and disabled on android, switched to other version (`0.82.0`), then back to `0.83.0` to check whether the version updates correctly.

Changed `package.json` of react-native to state that version is `0.83.2` and checked that in this case, android properly uses the native implementation.

Checked that explicitly toggling the option `useNativeInspectorImplementation` in launch configuration works as expected.

Not Applicable.

### DIFF
--- a/packages/vscode-extension/src/project/applicationSession.ts
+++ b/packages/vscode-extension/src/project/applicationSession.ts
@@ -239,7 +239,8 @@ export class ApplicationSession implements Disposable {
       this.applicationContext,
       devtoolsInspectorBridge,
       this.networkBridge,
-      this.metro.port
+      this.metro.port,
+      this.shouldEnableNativeNetworkInspector()
     );
 
     this.disposables.push(this.toolsManager);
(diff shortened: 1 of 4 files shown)


pFornagiel:
@filip131311 
If accepted, also candidate for `1.14.1`

packages/vscode-extension/src/project/applicationSession.ts
```diff
@@ -316,6 +317,33 @@ export class ApplicationSession implements Disposable {
...
+    const isAndroid = this.device.platform === DevicePlatform.Android;
+    const isAndroid0830 = reactNativeVersion?.compare("0.83.0") === 0 && isAndroid;
+
+    // RN 0.83.0 on Android has known issues with native network inspector
+    if (isAndroid0830) {
+      return false;
+    }
+
+    return meetsMinimumVersion;
+  }
```
jwajgelt, line 345: I don't like this, but I don't have a better idea.

jwajgelt (approved):

PR #1818, opened 2025-12-12
fix: don`t refetch Android phone models list on fail

jwajgelt:
Changes the behaviour of `DeviceNameCache` so that if the model list fetch fails, the extension won't attempt to refetch it.
This should prevent making unnecessary request and spamming the IDE logs with errors when e.g. the machine doesn't have Internet access.

### How Has This Been Tested: 
- open Radon in Radon
- connect a physical Android device
- disconnect from the Internet
- remove the `~/Library/Application Support/Code/User/globalStorage/swmansion.react-native-ide/RNIDE_device_models.json` file
- verify "Error fetching device models: fetch failed" isn't logged more than once

Internal


filip131311 (approved):
🚢 🇮🇹

PR #1820, opened 2025-12-12
feat: add seasonal visual improvements

pFornagiel:
### Description

Add optional seasonal visuals, toggleable between 19th December and 20th of January.

### How Has This Been Tested: 

Verified that toggle does not display in incorrect time periods, made sure the visual changes look good, made sure the visuals are not obstructive.

Not applicable.

### DIFF
--- a/packages/vscode-extension/src/webview/components/christmas/ChristmasModeToggle.tsx
+++ b/packages/vscode-extension/src/webview/components/christmas/ChristmasModeToggle.tsx
@@ -1,33 +1,9 @@
 import * as Switch from "@radix-ui/react-switch";
 import { use$ } from "@legendapp/state/react";
 import { useStore } from "../../providers/storeProvider";
-
-function isChristmasSeason(): boolean {
-  const now = new Date();
-  const month = now.getMonth(); // 0-based: 0 = January, 11 = December
-  const day = now.getDate();
-
-  // December 19th to December 31st
-  if (month === 11 && day >= 19) {
-    return true;
-  }
-
-  // January 1st to January 20th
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/src/webview/components/christmas/ChristmasModeToggle.tsx
```diff
@@ -0,0 +1,46 @@
+import * as Switch from "@radix-ui/react-switch";
+import { use$ } from "@legendapp/state/react";
+import { useStore } from "../../providers/storeProvider";
+
+function isChristmasSeason(): boolean {
```
kmagiera: If it's behind a switch let just make it available without any date based limitations. We will remove this code in first update in 2026 anyway

Also, the setting is persistent so once we hide the switch in Jan people who had it on, won't be able to turn it off

pFornagiel: This change was not made persistent (between extension sessions), exactly for the reason stated above, so that we would not accidentally leave people in this mode without an option to turn it off.

If we plan to  explicitly clean up the changes after, making them persistent across extension openings would require minimal extra work.

kmagiera: I think it'd be nice to keep it saved if someone enabled it. We will cleanup the code later so with some future update it will be switched off entirely.

pFornagiel: Made persistent and removed the date limitations in 87da9712.

PR #1825, opened 2025-12-16
chore: Remove deprecated `@radon` chat participant

latekvo:
This PR removes the deprecated `@radon` chat participant.
That feature is already marked as deprecated, not sure ***when*** we want to remove it, but it's okay for this PR to wait for the right moment.

Documented by #1794


packages/vscode-extension/src/ai/mcp/api.ts
```diff
@@ -11,6 +8,8 @@ const BACKEND_URL = "https://radon-ai-backend.swmansion.com/";
 
 export const AI_API_PLACEHOLDER_ID = "3241"; // This placeholder is needed by the API, but the value doesn't matter
 
+export const AI_LOG = "[AI]";
```
filip131311: seems like a left over

latekvo:
moved internally

PR #1827, opened 2025-12-17
fix: maestro tested app fails to connect to metro on iOS

jwajgelt:
Fixes the application failing to connect to Metro when running Maestro tests with `clearState` command.

### How Has This Been Tested: 
- open an app in Radon on iOS
- try to run a maestro test for the app with `clearState` option passed to the `launchApp` command
```
launchApp:
  clearState: true
```
- the app should connect to metro

Bugfix


filip131311 (approved):
thank you

PR #1833, opened 2025-12-22
tests: Extend network inspector test coverage

KeyJayY:
This pull requests introduces extended test coverage for network inspector tool. These changes are connected with this [PR](https://github.com/software-mansion-labs/radon-ide-test-apps/pull/40) in test apps repository.

### How Has This Been Tested: 

Tests have been checked to work correct both on iOS and Android devices. Also it has been verified if tests pass on Locally hosted CI.

No documentation needed

### DIFF
--- a/packages/vscode-extension-tester/ui-tests/04-appInteractions.test.js
+++ b/packages/vscode-extension-tester/ui-tests/04-appInteractions.test.js
@@ -57,6 +57,8 @@ safeDescribe("4 - App interaction tests", () => {
   });
 
   afterEach(async function () {
+    await driver.switchTo().defaultContent();
+    await new EditorView().closeAllEditors();
     await driver.sleep(TIMEOUTS.SHORT);
     await vscodeHelperService.openCommandLineAndExecute(
       "Remove All Breakpoints"


packages/vscode-extension-tester/server/webSocketServer.js
```diff
@@ -1,33 +1,230 @@
...
+    deletedId: id,
+  });
+});
+
+app.post("/api/multipart", upload.single("multipart_data"), (req, res) => {
+  if (!req.file) {
+    return res.status(400).send("No file uploaded.");
   }
+  const description = req.body.description;
+  console.log("MULTIPART RECEIVED");
```
pFornagiel: Are logs like these needed during the tests or are these left accidentally?

packages/vscode-extension-tester/server/webSocketServer.js
```diff
@@ -1,33 +1,230 @@
...
 let wss = null;
 let appWebsocket = null;
 
-export function initServer(port = 8080) {
-  if (wss) {
-    console.warn("WebSocket server already initialized");
-    return;
+app.use(express.static("."));
+app.use(express.json());
+app.use(express.urlencoded({ extended: true }));
```
pFornagiel: NIT: For readability, it would be better to split the express and websocket server implementations into different files, as they serve different purposes. We could, for example, create `RequestTestServer.js` and `WebSocketSever.js`, where the coresponding functionalities would be defined and then something like `server.js`, where the `initServers`, `closeServers` and other functions which govern the orchestration of the servers would be.

packages/vscode-extension-tester/server/webSocketServer.js
```diff
@@ -1,33 +1,230 @@
...
+
+app.post("/api/query-and-body", (req, res) => {
+  const { type, source } = req.query;
+  const { payload, extra } = req.body;
+  res.json({
+    received_query: { type, source },
+    received_body: { payload, extra },
+    message: "Query params and body received successfully",
+  });
+});
```
pFornagiel: The endpoint is defined here, but not referenced in `data_for_network_tests.json` or anywhere in the codebose

packages/vscode-extension-tester/files_for_tests/data_for_network_tests.json
```diff
@@ -0,0 +1,421 @@
...
+  "api/get": {
+    "request": {
+      "url": "http://localhost:8080/api/get?page=2&sort=desc",
+      "method": "GET",
+      "headers": {},
+      "query": {
+        "page": 2,
+        "sort": "desc"
+      }
+    },
```
pFornagiel, line 11: There are query parameters in the request, but the corresponding server endpoint does not parse them - instead it returns hardcoded value in `meta` field. Is this expected?

```js
app.get("/api/get", (req, res) => {
  res.setHeader("Content-Type", "application/json");
  res.json({
    meta: {
      page: "2",
      sort: "desc",
      total: 2,
    },
    data: [
      { id: 1, name: "John Doe", email: "john@example.com" },
      { id: 2, name: "Jane Doe", email: "jane@example.com" },
    ],
  });
});
```

packages/vscode-extension-tester/files_for_tests/data_for_network_tests.json
```diff
@@ -0,0 +1,421 @@
...
+      "headers": {
+        "connection": "keep-alive",
+        "content-length": "5242880",
+        "content-type": "text/html; charset=utf-8",
+        "keep-alive": "timeout=5",
+        "x-powered-by": "Express"
+      },
+      "body": "Response too large, showing truncated data"
+    }
+  },
```
pFornagiel, line 349: Should we not also check, that the rest of the body contains the truncated resposne or at least that resposne is properly formatted (check the presence of trailing `...` for example)?

packages/vscode-extension-tester/files_for_tests/data_for_network_tests.json
```diff
@@ -0,0 +1,421 @@
...
+        "connection": "keep-alive",
+        "content-length": "54181",
+        "content-type": "image/jpeg",
+        "keep-alive": "timeout=5",
+        "x-powered-by": "Express"
+      },
+      "body": "<image>"
+    }
+  },
+  "api/stream-xhr": {
```
pFornagiel, line 370: Shouldn't the response body of the image, displayed in the `Preview`tab, be the url-encoding of this image? Where exactly are we getting `"<image>"` string from, given that the server sends us the `jpg` file?

```js
app.get("/api/image", (req, res) => {
  res.sendFile(path.join(__dirname, "img", "img.jpg"));
});
```

packages/vscode-extension-tester/files_for_tests/data_for_network_tests.json
pFornagiel: Same goes for `/api/error/truncated`, `/api/error/hang`, `/api/large-image`. If we do not plan to use those endpoints, we may get rid of them.

packages/vscode-extension-tester/server/webSocketServer.js
```diff
@@ -1,33 +1,263 @@
...
+  res.sendFile(path.join(__dirname, "img", "large_img.jpg"));
+});
+
+app.use((req, res) => {
+  res.status(404).json({ error: "Not Found", endpoint: req.originalUrl });
+});
+
+app.get("/", (req, res) => {
+  res.sendFile(path.join(__dirname, "index.html"));
+});
```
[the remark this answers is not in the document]
pFornagiel: I don't think the `index.html` exists and that this endpoints should exist.

packages/vscode-extension-tester/server/webSocketServer.js
```diff
@@ -1,33 +1,230 @@
...
+      page: "2",
+      sort: "desc",
+      total: 2,
+    },
+    data: [
+      { id: 1, name: "John Doe", email: "john@example.com" },
+      { id: 2, name: "Jane Doe", email: "jane@example.com" },
+    ],
+  });
+});
```
pFornagiel: Proposal, altough not neccessity: Would this be an overkill if we read the hardcoded response messages from the `data_for_network_tests.json`?, in order to make sure we get consistent results?

KeyJayY: I'm not sure if that's what you mean, but I do check if fields shown in Radon's network inspector are the same as in `data_for_network _tests.json`, both for response and request. It happens in `testDetails` function in `ui-tests/05-networkPanel.test.js`, this function comperes expected values (from `data_for_network _tests.json`) with actual data in network inspector

packages/vscode-extension-tester/files_for_tests/data_for_network_tests.json
KeyJayY: `<image>` is not the actual data received from the server, the real data would be a string of image bytes. I believe that in this test case we shouldn't compare bytes, but just check if the network inspector shows the Preview tab and image in this tab. The `testDetails` function defines how to test endpoints with the `<image>` tag in the body.
```js
    if (expectedString !== "<image>") {
      assert.include(
        responseText.replace(/\s/g, ""),
        expectedString.replace(/\s/g, "")
      );
    } else {
      await elementHelperService.findAndClickElementByTag(
        `network-panel-tab-header-preview`
      );
      await elementHelperService.findAndWaitForElementByTag(
        `network-panel-preview-image`
      );
    }
 ```

packages/vscode-extension-tester/files_for_tests/data_for_network_tests.json
KeyJayY: You are right I haven't checked that before, however I think it's worth testing if this data appears in `payload` tab, I have added such a check in tests (before only request body was checked).

packages/vscode-extension-tester/server/webSocketServer.js
KeyJayY: I have added it to tests, I think it's worth testing.

packages/vscode-extension-tester/server/webSocketServer.js
KeyJayY: It's separated into two different files now.

packages/vscode-extension-tester/files_for_tests/data_for_network_tests.json
pFornagiel: Discussed internally, we should still perform a check, where we test whether there is any content in the `Response` tab and the content is not `No response body` message, which would indicate something is not right.

KeyJayY: It has been added

pFornagiel:
There is one potential issue connected to this PR - in our test-apps repository `shared/automatedTests.tsx` file, the handler of `"fetchData"` event does not explicitly handle the `XHR` network request, which we test in the `/api/stream-xhr` endpoint. 

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_XMLHttpRequest

packages/vscode-extension-tester/server/webSocketServer.js
pFornagiel: Discussed internaly, has been implemented by @KeyJayY

PR #1834, opened 2025-12-23
fix: loading metro config with new (newer then 54.0.20) Expo CLI

filip131311:
Tis PR solves the a problem with loading metro configuration in Radon, that was introduced with [this change](https://github.com/expo/expo/pull/41142) in expo CLI. 

Previously Expo CLI has relied on `loadConfig`, that we have overridden to adapt the config. After the change mentioned above expo decided to use `resolveConfig` instead, so we temporally override it as well. 

To avoid overriding even more expos internal functionality. We check the expo CLI version to determine if `resolveConfig` should be overridden. An alternative solution would be to override both `resolveConfig`  and `getDefaultConfig` and modify the logic of `loadConfig` to merge the changes correctly after being given already modified `default`, but this option would potentially affect how the old setups behave, which we want to avoid.

Most of the changes made in this PR should be removed once expo implements more stable solution for providing custom `metro-configs`. 

### How Has This Been Tested: 

- run expo app with expo version 54.0.30 (this matches the CLI version 54.0.20) - radon should work after this fix 
- run expo app with expo version 54.0.29 (this matches the CLI version 54.0.19) - radon should work after this fix and before
- run expo app with expo version 53 - radon should work after this fix and before 

NOTE: If you test multiple versions by reinstalling expo in the same project you need to use "restart metro" option from radons reload menu 

internal


packages/vscode-extension/lib/expo/expo_start.js
```diff
@@ -4,15 +4,65 @@ const {
...
+  }
+}
+
+function createRadonResolveConfig(original) {
+  return async (expoConfig, projectRoot, ...rest) => {
+    const originalResolvedConfig = await original(expoConfig, projectRoot, ...rest);
+    const getDefaultConfig = requireGetDefaultConfig();
+    if (getDefaultConfig === undefined) {
+      return originalResolvedConfig;
+    }
```
kmagiera: I wonder if we could just use empty config  here instead of loading the default one. I think that the resolved config is later merged with the default anyway, so maybe we can simplify the logic here and also avoid loading the default twice

PR #1837, opened 2025-12-23
feat: improve paywalled features signalling

pFornagiel:
### Description

The PR introduces changes to how paywalled and admin-blocked features are signaled to the user.

Changes mainly apply to the `Device Settings` menu and `Icon Buttons` in the top bar of the extension. Instead of silently disabling items, paywalled are now greyed-out, show a **Pro** badge on hover and  allow interaction to open the paywall modal.

https://github.com/user-attachments/assets/71394b4f-c2a0-4cae-a594-af59d9085cfa

When features are disabled due to administrative restrictions (`ADMIN_DISABLED` state), the options are also graded out but trigger appropriate modal and do not showcase the **Pro** badge.

The `DeviceSettingsDropdown` has been refactored, made a little more readable and edited so that the paywalled functionality checks are hidden within the `PaywolledDropdownMenu` components, where proper checks take place and showcase the paywall accordingly. The name of the feature (via `feature` prop accepting `Feature` enum value) can be provided to the component to make proper checks. 

Note: The Appearance menu does not currently use the above API. If there is a potential need to disable or gate it in the future, the necessary changes can be discussed and implemented.

Similar changes have been made to the `IconButton` which now also accepts `feature` prop and makes necessary checks when needed.

The paywall modal now also accepts a `feature` prop, which helps to display the information which of the features caused it to be activated.

### How Has This Been Tested: 

Performed tests with: lack of license, free license pro license - verified that lack of license and free license cause proper behaviour and visual look.

Made features disabled via setting `ADMIN_DISABLED` in `DefaultFeaturesAvailability` and commented the line, which would override it via token - checked whether the behaviour is correct with pro license.

Not applicable.

### DIFF
--- a/packages/vscode-extension/src/project/project.ts
+++ b/packages/vscode-extension/src/project/project.ts
@@ -284,7 +284,7 @@ export class Project implements Disposable, ProjectInterface, DeviceSessionsMana
       [Feature.DeviceFontSizeSettings]: "contentSize",
       [Feature.DeviceLocalizationSettings]: "locale",
       [Feature.LocationSimulation]: "location",
-    };
+    } as const;
 
     const configState = this.workspaceStateManager.getState();
     const deviceSettings = configState.deviceSettings;


packages/vscode-extension/src/webview/components/shared/IconButton.tsx
```diff
@@ -41,15 +46,21 @@ const IconButton = React.forwardRef<HTMLButtonElement, PropsWithDataTest<IconBut
...
+      feature,
+      paywallCallbackDependencies = [],
       ...rest
     } = props;
 
-    const shouldPing = usePing(counter ?? 0, counterMode);
+    const isPaywalled = useIsFeaturePaywalled(feature);
+    const isProFeature = feature !== undefined && isPaywalled;
+
+    const handleOnClick = usePaywalledCallback(onClick, feature, paywallCallbackDependencies);
```
jwajgelt: This doesn't feel right.
I think I'd prefer that we simply had something like:

```suggestion
    const handleOnClick = usePaywalledCallback(onClick, feature, [onClick]);
```

And if the component that passes `onClick` here wants the callback to be memoized, it can simply call `useCallback` itself and pass that here.

packages/vscode-extension/src/webview/components/shared/PaywallDropdownMenu.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+interface PaywallSwitchItemProps extends DropdownMenuComponents.SwitchItemProps {
+  feature: Feature;
+}
+
+export function SwitchItem({ children, className, feature, ...props }: PaywallSwitchItemProps) {
+  const isPaywalled = useIsFeaturePaywalled(feature);
+  const isAvailable = useIsFeatureAvailable(feature);
+
+  const handleOnClick = usePaywalledCallback(() => {}, feature, []);
```
jwajgelt, line 107: Why is it always a (paywalled) noop?

packages/vscode-extension/src/webview/components/shared/PaywallDropdownMenu.tsx
```diff
@@ -0,0 +1,135 @@
...
+      onSelect={handleSelect}
+      className={classnames(className, !isAvailable && "unavailable")}
+      {...props}>
+      {children}
+      {isPaywalled && <ProBadge />}
+    </DropdownMenu.Item>
+  );
+}
+
+interface PaywallSub extends DropdownMenu.DropdownMenuSubProps {
```
jwajgelt:
```suggestion
interface PaywallSubProps extends DropdownMenu.DropdownMenuSubProps {
```

packages/vscode-extension/src/webview/hooks/useFeatureAvailabilityCheck.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export function useIsFeaturePaywalled(feature: Feature | undefined): boolean {
+  const featuresAvailability = useFeaturesAvailability();
+  return !!feature && featuresAvailability[feature] === FeatureAvailabilityStatus.PAYWALLED;
+}
+
+export function useIsFeatureAvailable(feature: Feature | undefined): boolean {
+  const featuresAvailability = useFeaturesAvailability();
+  return !!feature && featuresAvailability[feature] === FeatureAvailabilityStatus.AVAILABLE;
+}
```
jwajgelt: NIT: Maybe just

```suggestion
export function useFeatureAvailability(feature: Feature | undefined): FeatureAvailabilityStatus | undefined {
  const featuresAvailability = useFeaturesAvailability();
  return feature && featuresAvailability[feature];
}
```

packages/vscode-extension/src/webview/hooks/usePaywalledCallback.ts
```diff
@@ -11,19 +11,21 @@ import { useAdminBlock } from "./useAdminBlock";
 
 function withPaywallGuard<F extends (...args: any[]) => Promise<void> | void>(
   fn: F,
-  feature: Feature,
+  feature: Feature | undefined,
```
jwajgelt: Why is `undefined` needed?

jwajgelt (dismissed):

packages/vscode-extension/src/webview/components/shared/DropdownMenuComponents.tsx
```diff
@@ -0,0 +1,115 @@
...
+}
+
+export function DropdownMenuRoot({
+  children,
+  open: openExternal,
+  onOpenChange,
+}: DropdownMenuRootProps) {
+  const [openInternal, setOpenInternal] = useState(false);
+
+  const open = openExternal !== undefined ? openExternal : openInternal;
```
kacperkapusciak:
```suggestion
  const open = openExternal ?? openInternal;
```

i think it's a great place to use [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)

packages/vscode-extension/src/webview/components/shared/DropdownMenuComponents.tsx
```diff
@@ -0,0 +1,115 @@
...
+  icon: string;
+  disabled?: boolean;
+}>) {
+  return (
+    <DropdownMenu.Item
+      className="dropdown-menu-item"
+      onSelect={onSelect}
+      disabled={disabled}
+      data-testid={dataTest}>
+      <span className="dropdown-menu-item-wraper">
```
kacperkapusciak:
```suggestion
      <span className="dropdown-menu-item-wrapper">
```
wrap**p**er typo

packages/vscode-extension/src/webview/components/shared/PaywallDropdownMenu.tsx
```diff
@@ -0,0 +1,135 @@
...
+  feature: Feature;
+  paywallCallbackDependencies?: unknown[];
+  children: React.ReactNode;
+}
+
+export function Sub({
```
kacperkapusciak: i don't know what Sub means

edit: oh it's a submenu, we should call it maybe?

packages/vscode-extension/src/webview/hooks/useFeatureAvailabilityCheck.ts
kacperkapusciak: yes +1

packages/vscode-extension/src/webview/hooks/usePaywalledCallback.ts
```diff
@@ -11,19 +11,21 @@ import { useAdminBlock } from "./useAdminBlock";
...
-    const featureAvailability = featuresAvailability[feature];
+    const featureAvailability = feature
+      ? featuresAvailability[feature]
+      : FeatureAvailabilityStatus.AVAILABLE;
     switch (featureAvailability) {
       case FeatureAvailabilityStatus.AVAILABLE:
         break;
       case FeatureAvailabilityStatus.PAYWALLED:
-        openPaywall();
+        openPaywall(undefined, feature);
```
kacperkapusciak:
```suggestion
        openPaywall({ feature });
```
this is a bit confusing. maybe passing an object as argument would be cleaner here

packages/vscode-extension/src/webview/components/shared/PaywallDropdownMenu.tsx
```diff
@@ -0,0 +1,135 @@
...
+  feature,
+  paywallCallbackDependencies = [],
+  children,
+  open,
+  ...props
+}: PaywallSub) {
+  const isPaywalled = useIsFeaturePaywalled(feature);
+  const isAvailable = useIsFeatureAvailable(feature);
+
+  const isOpen = isAvailable ? open : false;
```
kacperkapusciak:
```suggestion
  const isOpen = isAvailable && open;
```

packages/vscode-extension/src/webview/components/shared/IconButton.tsx
pFornagiel: Thanks for the suggestion, it is much cleaner.

Implemented in 2d880fda, also changing implementation to this behaviour in `PaywallDropdownMenu.tsx` -> `Item` and deleted `paywallCallbackDependencies` from `Sub` component.

packages/vscode-extension/src/webview/components/shared/PaywallDropdownMenu.tsx
pFornagiel: The passed handler is used inside `SwitchItem`'s`div` wrapper, which was never assigned any `onClick` behaviour as the item is normally meant to react only to the wrapped `Switch.Root` component clicks (`onCheckedChange` events to be precise).

```tsx
export function SwitchItem({
  icon,
  checked,
  onMenuItemClick,
  onCheckedChange,
  dataTestId,
  id,
  children,
  className,
  ...props
}: SwitchItemProps) {
  return (
    <div className={classNames("dropdown-menu-item", className)} onClick={onMenuItemClick}>
      {icon}
      {children}
      <Switch.Root
        {...props}
        className="switch-root small-switch"
        data-testid={dataTestId}
        id={id}
        onCheckedChange={onCheckedChange}
        defaultChecked={checked}
        style={{ marginLeft: "auto" }}>
        <Switch.Thumb className="switch-thumb" />
      </Switch.Root>
    </div>
  );
}
```

The `onMenuItemClick` prop could be added, tough, if we wish to support such API option, similarly to the implementation mentioned in the above comment.

As for why `usePaywalledCallback` is used with empty arrow function - the hook handles opening both the paywall modal and admin-block modal with necessary checks. Tried to make an implementation using `usePaywall` and `useAdminBlock`, but the approaches ended up pretty much identical to the `usePaywalledCallback` implementation. Thus, decided that code redundancy is not worth it. If needed, we can define another function, which would not be a wrapper and just opens the appropriate modal with checks performed on click.

packages/vscode-extension/src/webview/components/shared/PaywallDropdownMenu.tsx
pFornagiel: An oversight, thanks. Changed in f3d80dad

packages/vscode-extension/src/webview/hooks/usePaywalledCallback.ts
pFornagiel: Undefined was needed because of how I generalised the `IconButton` to support the paywall checks.

In a hindsight, I believe this was not the best design choice, as everywhere, where providing `feature` parameter was needed, the parameter itself had to be checked for being `undefined.

Decided to not generalise it that much and introduced `PaywallIconButton` in 8a7f0ad2 as an alternative, which allows us to require the feature to be provided where needed (thus eliminating the `undefined` checks). 

Because of the above, changed to `feature: Feature` in ee53ba9b

packages/vscode-extension/src/webview/hooks/useFeatureAvailabilityCheck.ts
pFornagiel: Made suggested change in 16aedf2e

packages/vscode-extension/src/webview/components/shared/DropdownMenuComponents.tsx
pFornagiel: Made suggested change in 630a67de

packages/vscode-extension/src/webview/hooks/usePaywalledCallback.ts
pFornagiel: Made suggested change in 630a67de and adjusted the usages accordingly.

```tsx
interface OpenPaywallOptions {
  title?: string;
  feature?: Feature;
}

export function usePaywall() {
  const { project } = useProject();

  const openPaywall = ({ title, feature }: OpenPaywallOptions = {}) => {
    project.sendTelemetry("paywall:open");
    openModal(<PaywallView title={title} feature={feature} />, { fullscreen: true });
  };

  return { openPaywall };
}

```

packages/vscode-extension/src/webview/components/shared/PaywallDropdownMenu.tsx
pFornagiel: Made suggested change in https://github.com/software-mansion/radon-ide/commit/630a67de29ad38ff1d9da603d68b1c8dcc3d27ff

packages/vscode-extension/src/webview/components/shared/DropdownMenuComponents.tsx
pFornagiel: Fixed the typo in 2e691278

packages/vscode-extension/src/webview/components/shared/PaywallDropdownMenu.tsx
jwajgelt: I get it now, it's ok as is.

packages/vscode-extension/src/webview/components/shared/PaywallDropdownMenu.tsx
jwajgelt: The idea is to replicate the `radix-ui` interface, `DropdownMenu.Sub` -> `PaywallDropdownMenu.Sub`.
I think it's reasonable.

PR #1847, opened 2026-01-08
fix: react-query plugin crash with react-compiler

### LINKED ISSUE #1839: App crashes with JS errors in RadonIDE if reactCompiler is turned on in app.config.js
**Description**

When running an Expo Router project with `experiments.reactCompiler: true` enabled in `app.config.js`, Radon IDE fails to detect route files correctly. All routes show "missing the required default export" warnings and throw `TypeError: Cannot read property 'prototype' of undefined` errors.

The issue occurs because:
1. `experiments.reactCompiler: true` enables React Compiler transformation at the Metro bundler level
2. Metro transforms route files before Expo Router can perform static analysis
3. Expo Router relies on static analysis to detect default exports in route files
4. The transformation breaks route detection, causing all routes to fail

**Important Note:** This issue only occurs when running the app through Radon IDE. Running the app normally (via `expo start` or `expo run:ios`) works correctly, suggesting Radon IDE handles the React Compiler transformation differently than standard Metro bundler.

**Error Messages**

```
Route "./app/profile/index.tsx" is missing the required default export. Ensure a React component is exported as default.
TypeError: Cannot read property 'prototype' of undefined, js engine: hermes
```

This error repeats for every route file in the `app/` directory.

**Environment**

- **Radon IDE Version:** 1.14.4
- **Expo SDK:** ~53.0.24
- **Expo Router:** ~5.1.8
- **React Native:** 0.79.6
- **React:** 19.0.0
- **React Compiler Runtime:** 19.0.0-beta-af1b7da-20250417
- **Metro:** ^0.82.0
- **Metro Config:** ^0.82.0
- **Node.js:** v22.19.0
- **OS:** macOS (darwin 25.2.0)
- **Platform:** iOS Simulator

**Configuration**

`app.config.js`:
```javascript
experiments: {
  reactCompiler: true,
}
```

**Steps to Reproduce**

1. Create an Expo Router project with `experiments.reactCompiler: true` in `app.config.js`
2. Ensure route files have proper default exports (e.g., `app/profile/index.tsx`)
3. Run the app through Radon IDE
4. Observe that all routes fail with "missing default export" errors

**Expected Behavior**

Routes should be detected correctly, and the app should run without route detection errors, similar to running via standard `expo start` or `expo run:ios`.

**Actual Behavior**

All routes fail to be detected, showing "missing default export" warnings and `TypeError: Cannot read property 'prototype' of undefined` errors.

**Workaround**

Disable `experiments.reactCompiler` in `app.config.js`:

```javascript
experiments: {
  reactCompiler: false,
}
```

This disables React Compiler transformation at the Metro bundler level, allowing Expo Router to perform static analysis on route files correctly.

**Additional Context**

- The issue appears to be specific to how Radon IDE handles React Compiler transformation
- Standard Metro bundler runs work correctly with `experiments.reactCompiler: true`
- This suggests a compatibility issue between Radon IDE's Metro configuration and Expo Router's static route analysis when React Compiler is enabled

Image

Image

jwajgelt:
Fixes crashes on Expo <=53 when using `react-query` with `react-compiler` enabled.
The crash is caused by react-compiler reordering the `plugins/react-query-devtools.js` import injected into `@tanstack/react-query` _before_ the `QueryClient` reexport from `query-core`:
```ts
  /* inside the transformed `@tanstack/react-query/src/index.ts`... */
  require(_dependencyMap[18], "__RNIDE_lib__/plugins/react-query-devtools.js");
  var _tanstackQueryCore = require(_dependencyMap[19], "@tanstack/query-core");
  for (var _key in _tanstackQueryCore) {
    exports[_key] = _tanstackQueryCore[_key];
  }
  /* etc... */
```
Causing `QueryClient` to be `undefined` in the plugin.
This PR instead import the `QueryClient` from `query-core`, resolving the issue.
It also wraps the code overriding the `QueryClient.prototype` in a `try/catch` block so that even if the plugin fails to apply itself, it won't bring the app down with it.

Fixes #1839 

### How Has This Been Tested: 
1. 
- enable react-compiler in `expo-52-prebuild-with-plugins` (https://docs.expo.dev/guides/react-compiler/)
- run the app
- ???
- profit
2. 
- check on a pnpm workspace, since indirect dependency imports break there

BUGFIX


filip131311 (changes requested):
Please add pnpm workspace to test plan because I think this solution will break it

kacperkapusciak (approved):
didn't test

PR #1852, opened 2026-01-09
tests: Improve tests stability

KeyJayY:
This pull request improves stability and fixes issues for tests running on locally hosted CI. While some flakiness may persist, it has been reduced. A subsequent PR will address the remaining instability by implementing a percentage based passing threshold.

Additionally, this PR redefines the apps on which we run tests to prioritize the latest versions of expo and react native.

### How Has This Been Tested: 

Tests have been ran via GitHub Actions and verified to work correctly.

No documentation needed


packages/vscode-extension-tester/services/radonViewsService.js
```diff
@@ -192,13 +192,20 @@ export default class RadonViewsService {
...
+    // in case it does not save on enter
+    try {
+      const quickInputButton =
+        await this.elementHelperService.findAndWaitForElement(
+          By.css(".quick-input-action"),
+          "Timed out waiting for quick input button"
+        );
+
+      await quickInputButton.click();
+    } catch {}
```
jwajgelt, line 207: Won't this cause false positives if the form never appears? I guess at that point the previous 
```
this.elementHelperService.findAndWaitForElement(
          By.css(".quick-input-widget .input")
```
would fail?

I guess I don't understand under what circumstances the previous test failed. Why would the element we expect here (`By.css(".quick-input-action")`, which I guess is the confirmation for the form) not be available here, even though the flow actually succeeds?

packages/vscode-extension-tester/ui-tests/04-appInteractions.test.js
```diff
@@ -66,7 +67,16 @@ safeDescribe("4 - App interaction tests", () => {
     await vscodeHelperService.openFileInEditor(
       "/data/react-native-app/shared/automatedTests.tsx"
     );
-    const editor = new TextEditor();
+    let editor = new TextEditor();
```
jwajgelt: https://github.com/redhat-developer/vscode-extension-tester/wiki/TextEditor#lookup
In the docs, they mention you can retrieve an open editor by the file name. Could we use that instead?

packages/vscode-extension-tester/ui-tests/04-appInteractions.test.js
```diff
@@ -66,7 +67,16 @@ safeDescribe("4 - App interaction tests", () => {
...
+
+    // Reopen the file in case first attempt didn't work
+    if ((await editor.getTitle()) !== "automatedTests.tsx") {
+      await vscodeHelperService.openFileInEditor(
+        "/data/react-native-app/shared/automatedTests.tsx"
+      );
+      editor = new TextEditor();
```
jwajgelt: I feel like this deserves it's own helper function. Maybe `vscodeHelperService.openFileInEditor` could return the `TextEditor` object?

packages/vscode-extension-tester/ui-tests/10-expo.test.js
```diff
@@ -86,6 +87,8 @@ describeIf(IS_EXPO, "10 - Expo router tests", () => {
     await urlInput.click();
     await urlInput.sendKeys("/explore", Key.ENTER);
 
+    await driver.sleep(TIMEOUTS.SHORT);
+
```
jwajgelt, line 91: Why does this extra `driver.sleep` help here? Shouldn't the `driver.wait` below already resolve any timing-related issues?

jwajgelt: Or is the problem that the button in the application changes its position, and we need to wait a while for it to settle in place?

KeyJayY: The problem before was that there's an animation in app on view change (change tab, open modal etc.), if communication happens to fast the app returns coordinates of moving button. The wait below doesn't resolve this it waits for any response from app.

packages/vscode-extension-tester/services/radonViewsService.js
KeyJayY: The lines above actively use this form, like finding the input box and typing into it. I believe there is no risk of a false positive, since if the form hadn't appeared, the test would have failed earlier and wouldn't even reach these lines.

The problem was that in some cases the button was clicked (it changed its appearance and didn't cause any error), but the file did not actually save and the form stayed open. There are two methods of confirming this form I know: pressing `Enter` and clicking the `OK` button. The current implementation is a retry mechanism, if the form stays open after pressing `Enter`, we try to confirm this by clicking the `OK` button.

jwajgelt: Could we wait until the input box disappears in the test, and fail if it doesn't? To ensure that if the test passes the whole flow actually completed.

packages/vscode-extension-tester/ui-tests/04-appInteractions.test.js
KeyJayY: I have moved this line to `vscodeHelperService.openFileInEditor` function and used 
`await new EditorView().openEditor('package.json');` instead of `new TextEditor();`

packages/vscode-extension-tester/services/radonViewsService.js
KeyJayY: We could do that, but I'm not sure if it's the correct approach. The VS Code instance used by the test framework uses a different save file form than the standard VS Code. Normally, the editor opens the native macOS save dialog, but in the test environment, the built-in VS Code dialog (image below) is used instead. I really doubt this fail is radon's fault, so test failing at this point wouldn't provide any useful informations.

Image

jwajgelt: If it doesn't break the further test cases (because we expected this pop-up to be closed there, but it remains open), then it's fine.

packages/vscode-extension-tester/services/helperServices.js
```diff
@@ -114,6 +115,19 @@ export class VSCodeHelperService {
...
+
+    await this.driver.sleep(TIMEOUTS.SHORT);
+    const fileName = path.basename(filePath);
+
+    const editorView = new EditorView();
+    if (
+      (await editorView.getOpenEditorTitles(0)).includes(
+        path.basename(fileName)
+      )
+    ) {
```
jwajgelt: NIT: I'd extract the condition to a separate variable for readability.

```suggestion
    const isFileOpen = (await editorView.getOpenEditorTitles(0)).includes(path.basename(fileName))
    if (isFileOpen) {
```

packages/vscode-extension-tester/services/helperServices.js
```diff
@@ -114,6 +115,19 @@ export class VSCodeHelperService {
...
+      const editor = await editorView.openEditor(fileName);
+      return editor;
```
jwajgelt, line 128: I think this function should throw if we don't hit this `return`.
In the test that calls this, we use the `editor` as if it were non-null(ish), but now this function can return `undefined` in that case, which will result in errors that may be difficult to debug.

packages/vscode-extension-tester/ui-tests/04-appInteractions.test.js
```diff
@@ -63,10 +64,17 @@ safeDescribe("4 - App interaction tests", () => {
...
       "/data/react-native-app/shared/automatedTests.tsx"
     );
-    const editor = new TextEditor();
+
+    // Reopen the file in case first attempt didn't work
+    if ((await editor.getTitle()) !== "automatedTests.tsx") {
+      editor = await vscodeHelperService.openFileInEditor(
+        "/data/react-native-app/shared/automatedTests.tsx"
+      );
+    }
```
jwajgelt, line 76: Can this check occur now that we use the API to get `editor` for this specific file?

PR #1861, opened 2026-01-13
feat: support storybook with custom configuration path

filip131311:
note: this is only a first PR in a series of necessary fixes for storybook integration to work as expected

In this PR we introduce a way of detecting custom configurationPath for storybook. To achieve that we load metro config with overloaded `wiehStorybook` utility that receives the information (if provided by the user) and then pass it to our metroConfig adapter to be used by utilities running in application runtime. 

### Still to do in feature PRs:

- [ ] additional launch configuration that will allow for full storybook customization 
- [ ] fix problems with ` const { toId, storyNameFromExport } = require("__APPDIR__/node_modules/storybook/dist/csf");` import in new versions 
- [ ] fix problems with reading the StoriesStore if the stories have a title 

### How Has This Been Tested: 

- use "show story" functionality in a project that has storybook configured in non standard location
- we will still receive an error  in debug console of the application but because of the following line:
` const { toId, storyNameFromExport } = require("__APPDIR__/node_modules/storybook/dist/csf");` 
that is after the line that was failing before 

We will review the storybook documentation once it's working

### DIFF
--- a/packages/vscode-extension/lib/storybook/storybook_config_reader.js
+++ b/packages/vscode-extension/lib/storybook/storybook_config_reader.js
@@ -2,7 +2,7 @@ const { requireFromAppDependency, requireFromAppDir, overrideModuleFromAppDir } 
 
 const { loadConfig } = requireFromAppDependency("react-native", "metro-config");
 
-const newWithStorybook = (_config, options) => {
+const overrideWithStorybook = (_config, options) => {
   if (!options?.configPath) {
     process.exit(1);
   }
@@ -17,17 +17,17 @@ const newWithStorybook = (_config, options) => {
   process.exit(0);
 }
 
-const oldWithStorybookModule = requireFromAppDir("@storybook/react-native/metro/withStorybook");
-
-let newWithStorybookModule = newWithStorybook;
+const originalWithStorybookModule = requireFromAppDir("@storybook/react-native/metro/withStorybook");
 
 // since Storybook 10 the module export has changed so we need to override it correctly 
 // https://github.com/storybookjs/react-native/pull/786
-if (oldWithStorybookModule?.withStorybook) {
-  newWithStorybookModule = { withStorybook: newWithStorybook }
-}
+const usesDefaultExport = originalWithStorybookModule?.withStorybook === undefined;
 
-overrideModuleFromAppDir("@storybook/react-native/metro/withStorybook", newWithStorybookModule);
+if (usesDefaultExport) {
+  overrideModuleFromAppDir("@storybook/react-native/metro/withStorybook", overrideWithStorybook);
+} else {
+  overrideModuleFromAppDir("@storybook/react-native/metro/withStorybook", { withStorybook: overrideWithStorybook });
+}
 
 async function main() {
   const customMetroConfigPath = process.env.RN_IDE_METRO_CONFIG_PATH;


packages/vscode-extension/lib/storybook/storybook_config_reader.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  if (options?.enabled === false) {
+    process.exit(1);
+  }
+
+  const { configPath } = options;
+
+  process.stdout.write(`RADON_STORYBOOK_CONFIG_PATH:${configPath}\n`);
+  process.exit(0);
```
jwajgelt, line 17: INB4 someone call `withStorybook` twice in their config...

packages/vscode-extension/lib/storybook/storybook_config_reader.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const oldWithStorybookModule = requireFromAppDir("@storybook/react-native/metro/withStorybook");
+
+let newWithStorybookModule = newWithStorybook;
+
+// since Storybook 10 the module export has changed so we need to override it correctly 
+// https://github.com/storybookjs/react-native/pull/786
```
jwajgelt: When reading this, I thought `old` and `new` referred to the Storybook version (since you mention it here).
Could we rename these to `originalWithStorybookModule` and `overrideWithStorybookModule` (or `stubbed-`, or whatever sound good to you)?

packages/vscode-extension/lib/storybook/storybook_config_reader.js
```diff
@@ -0,0 +1,42 @@
...
+if (oldWithStorybookModule?.withStorybook) {
+  newWithStorybookModule = { withStorybook: newWithStorybook }
+}
+
+overrideModuleFromAppDir("@storybook/react-native/metro/withStorybook", newWithStorybookModule);
```
jwajgelt: VERY NIT:
I think it'd read more naturally if you had something like

```suggestion
const usesDefaultExport = oldWithStorybookModule?.withStorybook === undefined;

if (usesDefaultExport) {
  overrideModuleFromAppDir("@storybook/react-native/metro/withStorybook", newWithStorybook);
} else {
  overrideModuleFromAppDir("@storybook/react-native/metro/withStorybook", { withStorybook: newWithStorybook });
}
```

packages/vscode-extension/src/utilities/applicationDependencies.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (dependencyVersion === null) {
+    return false;
+  }
+
+  if (!minVersion || semver.gte(dependencyVersion, new SemVer(minVersion))) {
+    return true;
+  }
+
+  return false;
+}
```
jwajgelt, line 66: IMO: 
I'd move this function back to the `ApplicationDependencyManager` and make it return `"installed"`/`"notInstalled"`, and handle the version directly in `utilities/storybook.ts` (since it needs to read the version anyway).

packages/vscode-extension/src/utilities/storybook.ts
```diff
@@ -0,0 +1,96 @@
...
+
+type StorybookConfig =
+  | {
+      installed: false;
+    }
+  | {
+      installed: true;
+      version: SemVer;
+      configPath: string;
+    };
```
jwajgelt, line 20: I think I'd change this to

```suggestion
type StorybookConfig = {
      version: SemVer;
      configPath: string;
    };
```

and simply return `null` from `getStorybookConfiguration` if a supported version is not installed.

packages/vscode-extension/lib/storybook/storybook_config_reader.js
filip131311: probably would break your storybook by itself

packages/vscode-extension/src/utilities/storybook.ts
filip131311: I do not agree. This utility is designed to generate a config that describes the Storybook State in users environment, so I can just use it to generate the whole project configuration in the future. Explicit "installed" information is very useful in such configurations. 

On the side note I prefer `installed: false;` pattern over nulling the whole objects because then I don't have to guess (or look up) that null means it is clear from the code.

packages/vscode-extension/src/utilities/applicationDependencies.ts
filip131311: I think having a utility that checks if a package is installed is useful

jwajgelt: That utility exists, and is called `readApplicationDependencyVersion` -- since it returns `null` if it's not.
The comparison seems overly specific for a util -- the only place where you really want to just check `semver.gte(version, MIN_VERSION)` is in the `ApplicationDependencyManager`, usually you will want to inspect the version (like in the storybook case, where you want to differentiate between not-installed/9.x/10.x).

packages/vscode-extension/src/utilities/storybook.ts
jwajgelt: I disagree.
You have a function that returns the storybook configuration for the project.
If storybook is not configured (that is, "installed"), `null` is what I'd expect.
Having an extra boolean field that signifies whether storybook is configured feels off to me, and has some ergonomics issues later down the line:
- JS provides some QoL features for handling null-ish and falsy values that you can't use with this approach
- each and every time you deal with a `StorybookConfig` value, you _have_ to inspect the `installed` variable -- while with the other approach you can always be sure that the code path only runs if `storybook` is present in the project

packages/vscode-extension/lib/storybook/storybook_config_reader.js
jwajgelt: Yeah, I wouldn't worry about it.
Just planting the flag.

packages/vscode-extension/src/utilities/storybook.ts
filip131311: I was convinced will change that

PR #1863, opened 2026-01-14
feat: copy screenshot to clipboard

pFornagiel:
### Description

This PR adds compatibility with introduced copy-to-clipboard functionality https://github.com/software-mansion-labs/simulator-server/pull/249.

Upon successfully taking a screenshot, the taken screenshot now gets saved to user's clipboard automatically, with popup message below the button informing about it. The screenshot does not get copied when the AI agent uses the `getScreenshot` function. Technical details about screenshot-taking are listed in the PR linked above.

### How Has This Been Tested: 

Built the simulator server from branch https://github.com/software-mansion-labs/simulator-server/pull/249, checked that upon saving a screenshot it gets properly copied to clipboard and the popup appears.

WIP.

### DIFF
--- a/packages/vscode-extension/src/project/deviceSession.ts
+++ b/packages/vscode-extension/src/project/deviceSession.ts
@@ -33,6 +33,7 @@ import {
   RecursivePartial,
   REMOVE,
   StartupMessage,
+  CaptureResult,
 } from "../common/State";
 import { ReloadAction } from "./DeviceSessionsManager";
 import { StateManager } from "./StateManager";
@@ -904,12 +905,12 @@ export class DeviceSession implements Disposable {
   }
 
   public async captureScreenshot() {
-    const wasSaved = await this.screenCapture.captureScreenshot();
-    if (wasSaved) {
+    const captureResult = await this.screenCapture.captureScreenshot();
+    if (captureResult === CaptureResult.Saved) {
       const rotation = this.applicationContext.workspaceConfiguration.deviceSettings.deviceRotation;
       this.device.copyLastScreenshotToClipboard(rotation);
     }
-    return wasSaved;
+    return captureResult;
   }
 
   public async getScreenshot() {
--- a/packages/vscode-extension/src/common/Project.ts
+++ b/packages/vscode-extension/src/common/Project.ts
@@ -4,6 +4,7 @@ import { LaunchConfiguration } from "./LaunchConfig";
 import { Output } from "./OutputChannel";
 import {
   AndroidSystemImageInfo,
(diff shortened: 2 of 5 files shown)


packages/vscode-extension/src/project/deviceSession.ts
```diff
@@ -904,7 +904,12 @@ export class DeviceSession implements Disposable {
   }
 
   public async captureScreenshot() {
-    this.screenCapture.captureScreenshot();
+    const wasSaved = await this.screenCapture.captureScreenshot();
+    if (wasSaved) {
+      const rotation = this.applicationContext.workspaceConfiguration.deviceSettings.deviceRotation;
+      this.device.copyLastScreenshotToClipboard(rotation);
+    }
+    return wasSaved;
```
jwajgelt: I'd prefer if `ScreenCapture.captureScreenshot()` threw an error when the screenshot cannot be taken, instead of returning a boolean with the status of the operation.
In that case, you can rewrite this function as:
```suggestion
    await this.screenCapture.captureScreenshot();
    const rotation = this.applicationContext.workspaceConfiguration.deviceSettings.deviceRotation;
    await this.device.copyLastScreenshotToClipboard(rotation);
```
and you get error handling "for free" without having to bubble the return value all the way to the frontend.

pFornagiel: The problem here is similar to what I brought up in https://github.com/software-mansion-labs/simulator-server/pull/249#discussion_r2695135746. After taking the screenshot  the user is prompted to decide where they want to save it (`PeviewView` -> `project.captureScreenshot` -> `deviceSession.captureScreenshot` -> `screenCapture.captureScreenshot` -> `saveMultimedia`, where `saveMultimedia` uses `saveFile` API prompting user to choose save directory)  . If the user cancels saving the file, the `saveMultimedia` used returns `undefined`. 

In this situation, I don't think the canceling of file saving should throw an error conceptually, as it is rather an expected action that user may take, but if it makes the code cleaner in your opinion I am be fine with that.

jwajgelt: Yeah, I didn't think about that when I wrote this comment.
The return value makes sense in that case.

jwajgelt: I'd change it to return an
```ts
enum CaptureResult {
  Saved,
  Aborted
}
```
but that's a nitpick.

pFornagiel: Done in 9a8e902e

PR #1885, opened 2026-02-02
docs: make the Trusted by on Hero dynamically fetch data

kacperkapusciak:
## How to test:

See Vercel preview

### DIFF
--- a/packages/docs/src/hooks/useInstallCount.ts
+++ b/packages/docs/src/hooks/useInstallCount.ts
@@ -33,30 +33,14 @@ async function fetchInstallCount(): Promise<number | null> {
   }
 }
 
-function formatInstallCount(count: number, roundToThousands = false): string {
-  if (roundToThousands) {
-    const thousands = Math.floor(count / 1000);
-    return `${thousands}K+`;
-  }
-  return count.toLocaleString("en-US");
-}
-
-interface UseInstallCountOptions {
-  defaultValue: string;
-  roundToThousands?: boolean;
-}
-
 interface UseInstallCountResult {
-  data: string;
+  data: number | null;
   error: Error | null;
   isLoading: boolean;
 }
 
-export default function useInstallCount({
-  defaultValue,
-  roundToThousands = false,
-}: UseInstallCountOptions): UseInstallCountResult {
-  const [data, setData] = useState(defaultValue);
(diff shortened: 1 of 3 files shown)


packages/docs/src/hooks/useInstallCount.ts
```diff
@@ -0,0 +1,80 @@
...
+    return null;
+  } catch (err) {
+    console.error(err);
+    return null;
+  }
+}
+
+function formatInstallCount(count: number, roundToThousands = false): string {
+  if (roundToThousands) {
+    const thousands = Math.round(count / 1000);
```
jwajgelt:
```suggestion
    const thousands = Math.floor(count / 1000);
```
I know it's not meant to be precise anyway, but saying we have e.g. "2K+" when the actual number is "1500" seems weird.

packages/docs/src/hooks/useInstallCount.ts
```diff
@@ -0,0 +1,80 @@
...
+          setData(formatInstallCount(count, roundToThousands));
+        }
+      } catch (err) {
+        setError(err instanceof Error ? err : new Error("Failed to fetch install count"));
+      } finally {
+        setIsLoading(false);
+      }
+    }
+    fetchData();
+  }, [roundToThousands]);
```
jwajgelt: I'd change this so that:
- `data` state is the actual number
- return the result of `formatInstallCount(data, roundToThousands)` from this hook
This way, this effect doesn't have to depend on `roundToThousands` and refetch if that argument changes for whatever reason.

jwajgelt: I would even argue the formatting should probably be done in a component, not in this hook -- but I'll leave that up to you.

packages/docs/src/hooks/useInstallCount.ts
kacperkapusciak: you're right, that would be deceptive

packages/docs/src/hooks/useInstallCount.ts
kacperkapusciak: i agree, fixed in [250b018](https://github.com/software-mansion/radon-ide/pull/1885/commits/250b01824be399dd43d29aacd1ee1471540994bc)

PR #1897, opened 2026-02-13
chore: early access to new features

kacperkapusciak:
This PR reverts changes introduced in https://github.com/software-mansion/radon-ide/pull/1895 related to "Early access to new features". 

Explanation: Radon publishes nightly releases that allow to use features before they are released in stable channel.


filip131311 (approved):
I have no idea whats going on

PR #1926, opened 2026-04-28
fix: fix missing testimonial avatars

mateuszaliyev:
This pull request fixes missing testimonial avatars, by using URLs archived by [Wayback Machine](https://web.archive.org/). This should prevent the URLs from ever expiring.

### How has this been tested


jwajgelt (approved):
Thank you 🙏

PR #1931, opened 2026-05-06
fix: add `scroll-padding-top`

mateuszaliyev:
This pull request adds `scroll-padding-top` to the `<html>` element so anchor-link jumps land below the sticky navbar instead of hiding the section heading underneath it and enables `scroll-behavior: smooth`.

### How has this been tested


kacperkapusciak (approved):
damn, this feels soo good

PR #1945, opened 2026-07-09
docs: update element inspector chat context section

mateuszaliyev:
This pull request updates element inspector chat context section.

### How has this been tested


jwajgelt (approved):
:+1:

